PR #3, opened 2023-01-17
Rename trait types to include trait name

cemozerr:


preston-evans98 (approved):
LGTM. We should add a note somewhere explaining the convention that we follow (use *Trait IFF there's an associated type that would otherwise have the same name), but that can be done as a fast follow.

PR #4, opened 2023-01-17
Add readme and a note on trait naming convention

cemozerr:


preston-evans98 (commented):
Looks good overall. 

I'd maybe consider moving the Note on trait naming somewhere, but that is low priority.

PR #7, opened 2023-01-26
Simplify STF interface

preston-evans98:
- Remove Address, Error and Header types
- Combine begin_batch, deliver_tx, and end_batch methods
- Simplify return type of apply_batch
- Simplify Consensus set management


cemozerr (approved):
LGTM.

PR #8, opened 2023-01-26
Update README.md

cemozerr:


preston-evans98 (approved):
Would be good to add an even more obvious warning, but lgtm otherwise

PR #10, opened 2023-01-31
Update serialisation traits

bkolad:
Changes in this PR:
1. The `Serialize & Deser` traits are renamed to `Encode & Decode`.
2. A new associated type Error is defined in the `Decode` trait. This way every de-serializable type can have a custom error type.


preston-evans98 (approved):
Looks great overall, I like adding an associated error type for `Decode`. 

The only thing we should think about is whether to rename the module from `serial` to `encoding` or something, but that's not blocking.

PR #12, opened 2023-02-01
Impl Encode/Decode for smart pointers

preston-evans98:
- Implement Encode for all type that impl `Deref<T> where T: Encode`
- Implement Decode for `Rc<T> where T: Decode`
- Implement Decode for `Arc<T> where T: Decode` if feature `sync` is enabled
- Add `IoError` variant to `CodecError`


bkolad (approved):
LGTM

PR #16, opened 2023-02-01
Remove jmt (fixed); Add EventKey and EventValue types

preston-evans98:
The previous commit failed to remove penumbra's `jmt` from the `Cargo.toml` file. This PR fixes that issue, and adds new types for EventKey and EventValue.

### DIFF
--- a/src/state_machine/serial.rs
+++ b/src/state_machine/serial.rs
@@ -35,11 +35,10 @@ pub trait Decode: Sized {
     fn decode(target: &mut &[u8]) -> Result<Self, Self::Error>;
 }
 
-// Automatically implement encode for smart pointers
-impl<T, U> Encode for T
+#[cfg(feature = "sync")]
+impl<T> Encode for Arc<T>
 where
-    T: Deref<Target = U>,
-    U: Encode,
+    T: Encode,
 {
     fn encode(&self, target: &mut impl std::io::Write) {
         self.deref().encode(target)
@@ -58,6 +57,15 @@ where
     }
 }
 
+impl<T> Encode for Rc<T>
+where
+    T: Encode,
+{
+    fn encode(&self, target: &mut impl std::io::Write) {
+        self.deref().encode(target)
+    }
+}
+
 impl<T, E> Decode for Rc<T>
 where
     T: Decode<Error = E>,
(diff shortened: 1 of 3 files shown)


bkolad (approved):
LGTM

src/state_machine/serial.rs
```diff
@@ -68,3 +76,65 @@ where
...
+            .expect("Serialization should not fail");
+        for item in self.iter() {
+            item.encode(target)
+        }
+    }
+}
+
+impl<T: Decode<Error = E>, E> Decode for Vec<T>
+where
+    DeserializationError: From<E>,
```
bkolad, line 93: I think we shouldn't assume an `Error` type for any type T.

What about:
1. Moving the length-checking logic to a separate method and checking it before decoding.
2. Defining a distinct type, `PrefixedVec', then we could check the length in a constructor, so the error is emitted as soon as we get the vec not when we try to deserialize it.

src/state_machine/serial.rs
```diff
@@ -68,3 +76,65 @@ where
...
+        self.0.encode(target);
+        self.1.encode(target)
+    }
+}
+
+impl<T, U, E1, E2> Decode for (T, U)
+where
+    T: Decode<Error = E1>,
+    U: Decode<Error = E2>,
+    DeserializationError: From<E1>,
```
bkolad, line 132: Here T, U are generic parameters, so we shouldn't make assumptions about their deserialization error type since we might not know it in advance.

What about:

pub enum CombinedError<E1, E2> {
    E1(E1),
    E2(E2),
}

impl<T, U, E1, E2> Decode for (T, U)
where
    T: Decode<Error = E1>,
    U: Decode<Error = E2>,
    CombinedError<E1, E2>: From<E1>,
    CombinedError<E1, E2>: From<E2>,
{
    type Error = CombinedError<E1, E2>;

    fn decode(target: &mut &[u8]) -> Result<Self, Self::Error> {
        Ok((T::decode(target)?, U::decode(target)?))
    }
}

src/state_machine/serial.rs
preston-evans98: Hmm, it would be nice to allow decoding of `Vec<T>` directly, but let me think about the mechanism. 

I guess one thing we could do would be to swap DeserializationError for a more generic error type like `CodecError`, which implements From: std::error::Error.

src/state_machine/serial.rs
preston-evans98: Synced with @bkolad offline. We're going to make significant revisions to encode/decode in #18, so I'll remove this piece of the PR

src/state_machine/serial.rs
preston-evans98: Canceling this in favor of #18

PR #17, opened 2023-02-02
Loosen trait bound on BobTransactionTrait::Data from AsRef<[u8]> to Buf

preston-evans98:
This relaxation makes it easier to implement efficient share decoding inside Jupiter, without changing the intent of the trait.

### DIFF
--- a/src/state_machine/core/run.rs
+++ b/src/state_machine/core/run.rs
@@ -1,3 +1,5 @@
+use bytes::Buf;
+
 use crate::{
     core::traits::{BlockheaderTrait, CanonicalHash},
     da::{BlobTransactionTrait, DaLayerTrait},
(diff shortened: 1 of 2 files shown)


preston-evans98:
This PR should wait and rebase on top of  #16. Marked as draft for now

src/state_machine/serial.rs
```diff
@@ -68,3 +76,65 @@ where
...
+        self.0.encode(target);
+        self.1.encode(target)
+    }
+}
+
+impl<T, U, E1, E2> Decode for (T, U)
+where
+    T: Decode<Error = E1>,
+    U: Decode<Error = E2>,
+    DeserializationError: From<E1>,
```
bkolad, line 132: Here `T, U` are generic parameters so we shouldn't assume their deserialization error type since might not know it in advance.  

What about:
```
pub enum CombinedError<E1, E2> {
    E1(E1),
    E2(E2),
}

impl<T, U, E1, E2> Decode for (T, U)
where
    T: Decode<Error = E1>,
    U: Decode<Error = E2>,
    CombinedError<E1, E2>: From<E1>,
    CombinedError<E1, E2>: From<E2>,
{
    type Error = CombinedError<E1, E2>;

    fn decode(target: &mut &[u8]) -> Result<Self, Self::Error> {
        Ok((T::decode(target)?, U::decode(target)?))
    }
}
```

bkolad (approved):
lgtm

PR #19, opened 2023-02-03
Split Decode trait to allow zero-copy deserialization

preston-evans98:
- Create DecodeBorrowed trait, which ties the liftime of its output to the lifetime of its input. This allows for borrow-based deserialization (aka zero-copy).
- Create Decode trait, whose output is not tied to its input lifetime.
- Implement DecodeBorrowed + Decode in terms of Borsch

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2021"
 [dependencies]
 bytes = "1.2.1"
 sha2 = "0.10.6"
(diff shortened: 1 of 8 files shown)


preston-evans98:
This PR still doesn't build yet, since we need to swap all of the manual Encode impls to Borsch derives. Sharing early in case it's useful.

bkolad (approved):
lgtm

PR #26, opened 2023-02-06
Add sov-modules to monorepo

bkolad:


.github/workflows/rust.yml
```diff
@@ -11,12 +11,22 @@ env:
...
-      run: cargo +stable build --verbose
-    - name: Run tests
-      run: cargo +stable test --verbose
+      - uses: actions/checkout@v2
+      - name: Install latest nightly
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: nightly
+            override: true
+            components: rustc
```
preston-evans98, line 22: We should open an issue to get back on stable if we haven't already

preston-evans98 (approved):

.github/workflows/rust.yml
bkolad: We have, but I need to bring over all the issues from the other repos.

bkolad: #29

PR #31, opened 2023-02-07
Change TreeReader::Error bound from std to anyhow

preston-evans98:
Allow TreeReader implementers to use anyhow::Errors instead of std::error::Errors. This is necessary to allow any usage of anyhow, since anyhow::Error does not (and cannot) implement the std error trait.

This PR should wait for #30 to merge


bkolad (approved):
lgtm

PR #32, opened 2023-02-07
Draft: Add a new "PhysicalTreeReader" trait for databases

preston-evans98:
Add a new trait `PhysicalTreeReader` which is suitable for use at the database layer. It is unaware of the hash function used by the JMT, but stores all of they physical bits of each node.


bkolad (approved):
Lgtm

PR #48, opened 2023-02-08
Basic integration tests for `sov-modules-impl`

### LINKED ISSUE #35: sov-modules: Decide how to combine a module prefix and a state key.
Each module has a globally unique prefix that has to be combined with a state key before db insertion. This way, the same state key in different modules will end up in separate db slots. 

```
pub struct Prefix {
    storage_name: String,
    module_name: String
    module_path: String,
}
```

How do we combine the state key and the prefix? There are two ways:
1. Hash the prefix and concatenate it with the state key: This is a bad idea because it could destroy db caching (we will access many state variables in the same module).
2. Concatenate all the prefix fields and the state key.

TODO: What about `prefix||state_key` padding?

bkolad:
1. Basic integration tests in `sov-modules-impl`.
2. Logic for `pferix` and `state_key` concatentation.
3. Code stub for `AlignedVec `
4. Implementations for `get/set in StateMap`

close #35

### DIFF
--- a/sov-modules/sov-modules-api/src/lib.rs
+++ b/sov-modules/sov-modules-api/src/lib.rs
@@ -10,6 +10,9 @@ use sovereign_sdk::{
 };
 use std::{convert::Infallible, io::Read};
 
+// separator == "/"
+const DOMAIN_SEPARATOR: [u8; 1] = [47];
+
 // A unique identifier for each state variable in a module.
 #[derive(Debug, PartialEq, Eq)]
 pub struct Prefix {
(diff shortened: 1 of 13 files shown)


sov-modules/sov-modules-api/src/lib.rs
```diff
@@ -31,12 +31,19 @@ impl Prefix {
 impl From<Prefix> for sov_state::Prefix {
     fn from(prefix: Prefix) -> Self {
         let mut combined_prefix = Vec::with_capacity(
-            prefix.module_path.len() + prefix.module_name.len() + prefix.storage_name.len(),
+            prefix.module_path.len() + prefix.module_name.len() + prefix.storage_name.len() + 3,
```
preston-evans98, line 34: nit: magic number. Maybe add a `const DOMAIN_SEPARATOR: [u8] = [47]` and then say "+ DOMAIN_SEPARATOR.len() * 3`?

sov-modules/sov-state/src/storage.rs
```diff
@@ -1,25 +1,70 @@
...
+    pub fn new<K: Encode>(prefix: &Prefix, key: K) -> Self {
+        let mut encoded_key = Vec::default();
+        key.encode(&mut encoded_key);
+
+        let encoded_key = AlignedVec::new(encoded_key);
+
+        let full_key = Vec::<u8>::with_capacity(prefix.len() + encoded_key.len());
+        let mut full_key = AlignedVec::new(full_key);
+        full_key.extend(prefix.as_aligned_vec());
+        full_key.extend(&encoded_key);
```
preston-evans98, line 36: (non-blocking) It seems like we should be able to do this without encoding the key separately and then copying. Maybe we should add a `size_hint` method to the key type?

sov-modules/sov-modules-api/src/lib.rs
bkolad: done

preston-evans98 (approved):

sov-modules/sov-state/src/storage.rs
bkolad: Resolved offline, we are skipping it for now.

PR #49, opened 2023-02-08
api:Prefix references `&'static str` instead of owning String

### LINKED ISSUE #46: sov-modules: `api:Prefix` should reference `&'static str` instead of owning `String`
The `Prefix` struct should reference `&'static str` instead of owning `String`. The names are known during the compilation phase, and we will avoid unnecessary calls to `.to_string()`
```
pub struct Prefix {
    module_path: String,
    ...
}
```
Should be changed to:
```
pub struct Prefix {
    module_path: &'static str,
    ...
}
```
I made it a standalone issue because it requires minor changes to the `ModuleInfo` macro.

bkolad:
close #46


preston-evans98 (approved):
Looks great!

PR #50, opened 2023-02-09
Add StateValue type to storage

### LINKED ISSUE #34: Add `StateItem` type to storage
Currently the module system supports maps only. We should add `StateItem` type that stores a single state value.

bkolad:
1. `StateValue` struct: a container for a single value.
2. Updates tests to use the new `StateValue`.
3. Removes `MockStorage`. We can use `JMT` for tests as we can store data inside the cache for now. 

close #34

### DIFF
--- a/sov-modules/sov-state/src/backend.rs
+++ b/sov-modules/sov-state/src/backend.rs
@@ -9,7 +9,7 @@ use std::marker::PhantomData;
 pub(crate) struct Backend<K, V, S> {
     _phantom: (PhantomData<K>, PhantomData<V>),
     storage: S,
-    // Every instance of the `StateMap` contains a unique prefix.
+    // Every instance of the `Backend` contains a unique prefix.
     // The prefix is prepended to each key before insertion and retrieval from the storage.
     prefix: Prefix,
 }
--- a/sov-modules/sov-state/src/backend.rs
+++ b/sov-modules/sov-state/src/backend.rs
@@ -9,8 +9,8 @@ use std::marker::PhantomData;
 pub(crate) struct Backend<K, V, S> {
     _phantom: (PhantomData<K>, PhantomData<V>),
     storage: S,
-    // Every instance of the `Backend` contains a unique prefix.
-    // The prefix is prepended to each key before insertion and retrieval from the storage.
+    /// Every instance of the `Backend` contains a unique prefix.
+    /// The prefix is prepended to each key before insertion and retrieval from the storage.
     prefix: Prefix,
 }
 
--- a/sov-modules/sov-state/src/backend.rs
+++ b/sov-modules/sov-state/src/backend.rs
@@ -1,0 +1,45 @@
+use crate::{
+    storage::{StorageKey, StorageValue},
+    Prefix, Storage,
+};
+use sovereign_sdk::serial::{Decode, Encode};
+use std::marker::PhantomData;
+
+#[derive(Debug)]
+pub(crate) struct Backend<K, V, S> {
+    _phantom: (PhantomData<K>, PhantomData<V>),
+    storage: S,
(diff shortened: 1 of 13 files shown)


sov-modules/sov-state/src/value.rs
```diff
@@ -0,0 +1,50 @@
...
+/// We provide a custom efficient Encode implementation for SingletonKey while Encode for `()`
+/// is likely already implemented by an external library (like borsh), which is outside of our control.
+#[derive(Debug)]
+pub struct SingletonKey;
+
+impl Encode for SingletonKey {
+    fn encode(&self, _: &mut impl std::io::Write) {
+        // Do nothing.
+    }
+}
```
preston-evans98, line 15: Note, non-blocking: Since we manually implement encode, any types containing SingletonKey won't be able to use the derive macros

sov-modules/sov-state/src/value.rs
```diff
@@ -0,0 +1,50 @@
...
+
+/// Container for a single value.
+#[derive(Debug)]
+pub struct StateValue<V, S> {
+    _phantom: PhantomData<V>,
+    // StateValue is equivalent to a StateMap with a single key.
+    map: StateMap<SingletonKey, V, S>,
+}
```
preston-evans98, line 23: It's a little unintuitive that `StateValue` is a wrapper around `StateMap`. I'd consider either keeping them completely separate, or defining a third type which contains the shared logic and having them both contain that type.

sov-modules/sov-state/src/jmt_storage.rs
```diff
@@ -1,24 +1,50 @@
...
+            // TODO If the value does not exist in the cache, then fetch it from the JMT.
+            cache::ExistsInCache::No => todo!(),
+        }
     }
 
-    fn set(&mut self, _key: StorageKey, _value: StorageValue) {
-        todo!()
+    fn set(&mut self, key: StorageKey, value: StorageValue) {
+        let cache_key = key.into();
+        let cache_value = value.into();
```
preston-evans98, line 40: non-blocking: It's not clear at a glance how costly these conversions are. Have we considered
1. Switching to `as_`, which is zero-cost by convention or
2. Replacing the `CacheKey` and `CacheValue` types with `StorageKey` and `StorageValue`

sov-modules/sov-state/src/value.rs
bkolad: I made the type private to the module so it won't leak (before I needed it in tests in another crate, but I found a walk around).  Here we won't mix it with other types.

sov-modules/sov-state/src/value.rs
bkolad: I carted a new type `Backend` as suggested.

sov-modules/sov-state/src/jmt_storage.rs
bkolad: Added `as_` methods.

sov-modules/sov-state/src/backend.rs
```diff
@@ -0,0 +1,45 @@
...
+    Prefix, Storage,
+};
+use sovereign_sdk::serial::{Decode, Encode};
+use std::marker::PhantomData;
+
+#[derive(Debug)]
+pub(crate) struct Backend<K, V, S> {
+    _phantom: (PhantomData<K>, PhantomData<V>),
+    storage: S,
+    // Every instance of the `Backend` contains a unique prefix.
```
preston-evans98, line 12: (nit, non-blocking): It would be nice to make this a doc comment unless there's a reason not to

preston-evans98 (approved):

sov-modules/sov-state/src/backend.rs
bkolad: done

PR #51, opened 2023-02-13
`ZkStorage` implementation

### LINKED ISSUE #36: sob-modules: Basic mocks for `ZkVm` environment
The logic of a module will be run in two environments: `Prover` and `ZkVm`.
The `Prover` has access to  functionality not avaliaible in  the `ZkVm` (for example a database) and has to produce  (among other things) a witness for the `ZkVm`. 

We want to be able to mock  `ZkVm` environment in our unit tests without running a full-blown `ZkVm` (for example risc0). 
In the first iteration, we will only check if all the state updates are visible in the `ZkVm` and if they are correctly cached. 

In the subsequent iterations, we will add MerkleTree verification, etc..

bkolad:
**Overview:** 

The first time a value is accessed from a module, it has to be loaded from an external source.
If a rollup runs in `native` context the external source will be a database. In `zk` context the first accessed values will appear "out of thin air" (for example, via the `env::read()` mechanism in Risc0)

1. `ValueReader` trait & `FirstReads` struct: encapsulate the code representing "external source".
2. Implementation for `ZkStorage` & `JmtStorage` using the `ValueReader ` trait.
3. Test update to reflect the above changes.

close #36

### DIFF
--- a/first-read-last-write-cache/src/cache.rs
+++ b/first-read-last-write-cache/src/cache.rs
@@ -31,12 +31,17 @@ pub struct CacheLog {
     log: HashMap<CacheKey, Access>,
 }
 
-#[derive(Default, Clone)]
+#[derive(Default, Clone, Debug)]
 pub struct FirstReads {
     reads: Arc<HashMap<CacheKey, CacheValue>>,
 }
 
 impl FirstReads {
+    pub fn new(reads: HashMap<CacheKey, CacheValue>) -> Self {
+        Self {
+            reads: Arc::new(reads),
+        }
+    }
     pub fn read(&self, key: &CacheKey) -> ExistsInCache {
         match self.reads.get(key) {
             Some(read) => ExistsInCache::Yes(read.clone()),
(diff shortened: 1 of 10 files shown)


first-read-last-write-cache/src/cache.rs
```diff
@@ -30,6 +31,28 @@ pub struct CacheLog {
     log: HashMap<CacheKey, Access>,
 }
 
+/// Represents all reads form a CacheLog.
```
preston-evans98, line 34: nit: form -> from

sov-modules/sov-state/src/zk_storage.rs
```diff
@@ -0,0 +1,53 @@
...
+        self.internal_cache.get_or_fetch(key, &self.first_reads)
+    }
+
+    fn set(&mut self, key: StorageKey, value: StorageValue) {
+        self.internal_cache.set(key, value)
+    }
+
+    fn delete(&mut self, key: StorageKey) {
+        self.internal_cache.delete(key)
+    }
```
preston-evans98, line 52: It looks like this logic is shared with JmtStorage. Can we make a generic storage struct and then just have JmtStorage and ZkStorage be type aliases to the generic version with a particular value for the `ValueReader`?

bkolad: Yes, I was thinking about it. I kept it this way in case there are some further differences between the two in the future, but nothing comes to my mind  (maybe we want logging in the JmtStorage?). I will go ahead and make a generic struct.

preston-evans98 (approved):

PR #53, opened 2023-02-14
Adds example of a basic module.

bkolad:
1. Simplifies the `Module` trait in `sov-modules-api:`.
2. Refactors `sov-modules-api:lib.rs`: splits the code into several files.
3. Introduces "simple_module", an example of a very basic module + tests.


sov-modules/sov-modules-api/src/error.rs
```diff
@@ -0,0 +1,47 @@
...
+#[derive(Debug)]
+pub struct DecodingError {}
+
+/// General error type in the module system.
+pub enum Error {
+    /// Custom error thrown by a module.
+    Module(ModuleError),
+}
+
+// We derive `Debug` by hand, because `ModuleError` doesn't implement `Debug` trait.
```
bkolad, line 14: All the errors in this file can be converted to `this error`

sov-modules/sov-modules-impl/src/example_simple_module/query.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl<C: sov_modules_api::Context> ValueAdderModule<C> {
+    /// Queries the state of the module.
+    pub fn query_value(&self) -> QueryResponse {
+        let response = Response {
+            value: self.value.get(),
+        };
+
+        let response = serde_json::to_vec(&response).unwrap();
```
preston-evans98, line 22: On second thought, maybe serialization should happen outside of the module system. If responses are always serialized, it will be hard to implement queries in higher-level modules. 

```
// token_module.rs
pub fn query_balance(&self, msg: address) -> QueryResponse { 
   let acct: Account = serde_json::deserialize(self.accounts_module.query(address)).unwrap();
   QueryResponse { balance: acct.balance }
 }

Comment (bkolad) [reply sov-modules/sov-modules-impl/src/example_simple_module/query.rs:22]: Good point, I think methods that return `QueryResponse ` should be callable only from the outside. So in the example above `accounts_module` would have 2 methods:
- `get_balance->Balance` 
- `query-> QueryResponse ` which calls `get_balance` internally + adds serialization

The `token_module.rs` would call `get_balance` instead of `query`.
WDT?

Comment (preston-evans98) [reply sov-modules/sov-modules-impl/src/example_simple_module/query.rs:22]: Yeah, that seems reasonable - but it would be nice to use the type system to show module designers what the correct pattern is as well. Let me think about this a bit more :)

File: sov-modules/sov-modules-api/src/error.rs [hunk 3]
```diff
@@ -0,0 +1,47 @@
...
+impl Debug for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Module(e) => f.debug_tuple("Module").field(&e.err).finish(),
+        }
+    }
+}
+
+/// Custom error thrown by a module.
+// We can't derive `Debug` because it conflicts with blanket `From` implementation.
```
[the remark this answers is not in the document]
bkolad: Yes if we won't go with ThisError I will add `Display` implementation.

sov-modules/sov-modules-impl/src/example_simple_module/query.rs
preston-evans98: Ok, after thinking a bit more I don't know that using the type system to show the design pattern is worth all of the complexity. Let's stick with this for now.

preston-evans98 (approved):

PR #69, opened 2023-02-17
fuzz testing for merge fn in cache.rs

irfanshaik11:
https://github.com/Sovereign-Labs/sovereign/issues/28

### DIFF
--- a/first-read-last-write-cache/src/cache.rs
+++ b/first-read-last-write-cache/src/cache.rs
@@ -381,8 +381,9 @@ mod tests {
     proptest! {
         #[test]
         fn test_merge_fuzz(s: u8) {
-            let mut testvec = vec![0; 15];
-            for i in 0..15 {
+            let num_cases = 15;
+            let mut testvec = vec![0; num_cases];
+            for i in 0..num_cases {
                 testvec[i] = s.wrapping_add(i as u8);
             }
 


first-read-last-write-cache/src/cache.rs
```diff
@@ -3,6 +3,7 @@ use crate::{CacheKey, CacheValue};
 use std::collections::{hash_map::Entry, HashMap};
 use std::sync::Arc;
 use thiserror::Error;
+use proptest::prelude::*;
```
bkolad, line 6: This import should go to `mod tests`.

first-read-last-write-cache/src/cache.rs
```diff
@@ -427,4 +428,42 @@ mod tests {
...
+                TestCase {
+                    left: None,
+                    right: Some(ReadWrite::Write(new_cache_entry(a, b))),
+                },
+                TestCase {
+                    left: Some(ReadWrite::Write(new_cache_entry(a, b))),
+                    right: Some(ReadWrite::Read(new_cache_entry(a, b))),
+                },
+            ];
+
```
bkolad, line 466: This only generates test cases but doesn't test the `merge` method. You can check the `test_merge_fail` test for details. I suggest extracting the code:
```
for TestCase { left, right } in test_cases.clone() {
            match (left, right) {
                (None, None) => {}
                (None, Some(rw)) => right_cache.add_to_cache(rw).unwrap(),
                (Some(rw), None) => left_cache.add_to_cache(rw).unwrap(),
                (Some(left_rw), Some(right_rw)) => {
                    left_cache.add_to_cache(left_rw).unwrap();
                    right_cache.add_to_cache(right_rw).unwrap();
                }
            }
        }

        let merged = left_cache.merge(right_cache).unwrap();
```
into a function and using it in all the `test_merge` tests.

first-read-last-write-cache/src/cache.rs
```diff
@@ -427,4 +428,42 @@ mod tests {
             }
         }
     }
+
+    proptest! {
+        #[test]
+        fn merge_doesnt_crash(a: u8, b: u8) {
```
bkolad, line 434: let's keep the convention and add `test_` prefix to the function name.

bkolad (changes requested):
Thanks for the PR! 
I added a couple of suggestions. There is also a merge conflict that has to be resolved.

first-read-last-write-cache/Cargo.toml
```diff
@@ -6,3 +6,4 @@ edition = "2021"
 
 [dependencies]
 thiserror = { workspace = true }
+proptest = "1.1.0"
```
bkolad, line 9: This should go to `[dev-dependencies]` and let's use the one declared in the `workspace`.

first-read-last-write-cache/src/cache.rs
bkolad: Thanks, I can't see this change in `Files changed` on github, have you pushed the commit?

first-read-last-write-cache/src/cache.rs
```diff
@@ -427,4 +393,88 @@ mod tests {
             }
         }
     }
+
+    proptest! {
+        #[test]
+        fn test_merge_fuzz(s: u8) {
+            let mut testvec = vec![0; 15];
+            for i in 0..15 {
+                testvec[i] = ((s as u16 + i as u16) % 255) as u8;
```
bkolad, line 402: Two small nits:
1. You can make `s: usize` and remove `s as u16` cast
2. Let's extract 15 a variable in this function.

first-read-last-write-cache/src/cache.rs
```diff
@@ -427,4 +393,88 @@ mod tests {
...
+
+            test_merge_ok_helper(test_cases);
+        }
+    }
+
+    fn test_merge_ok_helper(test_cases: Vec<TestCase>) {
+        let mut left_cache = CacheLog::default();
+        let mut right_cache = CacheLog::default();
+
+        for TestCase { left, right } in test_cases.clone() {
```
bkolad, line 444: We can extract this logic as well, and use it here and in `test_merge_fail`

first-read-last-write-cache/src/cache.rs
bkolad, line 402: let's replace it with `testvec[i] = s.wrapping_add(i as u8)`

bkolad: Ohh, I just noticed another thing. Let's extract 15 to a variable in this function.

bkolad (approved):

PR #70, opened 2023-02-17
test error cases

irfanshaik11:
https://github.com/Sovereign-Labs/sovereign/issues/56

### DIFF
--- a/sov-modules/sov-modules-impl/src/example_simple_module/tests.rs
+++ b/sov-modules/sov-modules-impl/src/example_simple_module/tests.rs
@@ -76,6 +76,14 @@ fn test_err_on_sender_is_not_admin() {
 
         test_err_on_sender_is_not_admin_helper(context, storage.clone());
     }
+
+    // Test Zk-Context
+    {
+        let zk_context = ZkMockContext { sender };
+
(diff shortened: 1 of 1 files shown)


sov-modules/sov-modules-impl/src/example_simple_module/tests.rs
```diff
@@ -61,3 +61,28 @@ fn test_module<C: Context>(context: C, storage: C::Storage) {
...
+    let sender = MockPublicKey::try_from("not_admin").unwrap();
+    let storage = JmtStorage::default();
+
+    // Test Native-Context
+    {
+        let context = MockContext {
+            sender: sender.clone(),
+        };
+
+        test_err_on_sender_is_not_admin_helper(context, storage.clone());
```
bkolad, line 77: We can also add test for test the `ZkContext` see `test_simple_module`

bkolad (changes requested):
This looks good overall, but the structure of the repo has changed. Now `example_simple_module` is in a separate crate. While separating it, I made a mistake which will be fixed in #68. 

Here is the plan on how to proceed:
1. We merge the #68 fix (will be done today).
2. You update this PR to the latest code.
3. We merge this PR!

PR #72, opened 2023-02-17
modules: get on state variables returns `Result` instead of `Option`

bkolad:
The `get` method for state variables returns  `Option`, but the module methods return `Result`. This means that we have to use `ok_or` to get a value from the storage. 
```
let value = self.some_varable.get().ok_or("anyhow!("some error message")")?;
```
This is not very ergonomic. The idea is to cheng the signature of `get` method to return  `Result`. 
This way, developers can write:

```
let value = self.some_varable.get()?;
```

### DIFF
--- a/sov-modules/sov-modules-impl/example-election/Cargo.toml
+++ b/sov-modules/sov-modules-impl/example-election/Cargo.toml
@@ -13,9 +13,10 @@ sov-modules-macros = { workspace = true }
 sov-state = { workspace = true }
 sovereign-sdk = { workspace = true }
 serde = { workspace = true, optional = true }
-serde_json = { workspace = true, optional = true}
+serde_json = { workspace = true, optional = true }
 thiserror = { workspace = true }
 borsh = { workspace = true, features = ["rc"]}
+hex = { workspace = true }
 
 [features]
 default = ["native"]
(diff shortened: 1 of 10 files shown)


preston-evans98:
I really like these ergonomics, but the semantics are a bit unusual. Intuitively, it doesn't feel like "this value doesn't exist" is a recoverable error. Either the value really is `option`al, or the failure is unrecoverable. (Reading an uninitialized non-optional variable is a bug, and if the value was initialized then the backing store must have gotten corrupted.)

In the JMT, we handle this situation by having two separate methods: a `get_option` method, which returns a `Result<Option<_>>`, and a `get` method which is just:

```rust
/// Gets a value which the caller expects to exist - returning an error if it was absent
fn get(&self, key: Key) -> Result<Value> { 
  Ok(self.get_option(key)?.ok_or(anyhow::error!("Value must exist")))
}
```

bkolad:
> I really like these ergonomics, but the semantics are a bit unusual. Intuitively, it doesn't feel like "this value doesn't exist" is a recoverable error. Either the value really is `option`al, or the failure is unrecoverable. (Reading an uninitialized non-optional variable is a bug, and if the value was initialized then the backing store must have gotten corrupted.)
> 
> In the JMT, we handle this situation by having two separate methods: a `get_option` method, which returns a `Result<Option<_>>`, and a `get` method which is just:
> 
> ```rust
> /// Gets a value which the caller expects to exist - returning an error if it was absent
> fn get(&self, key: Key) -> Result<Value> { 
>   Ok(self.get_option(key)?.ok_or(anyhow::error!("Value must exist")))
> }
> ```

Added second method `get_or_err`

preston-evans98 (approved):
Looks great!

PR #78, opened 2023-02-21
Design of the tx dispatch mechanism

bkolad:
1. `demo` module in the integration tests: proposed design of how the `dispatch` marcors will work
2. renames `tests.rs` to `nested_modules_tests.rs`


sov-modules/sov-modules-impl/integration-tests/tests/demo/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    mocks::{MockContext, MockPublicKey},
+    CallResponse, Context, Error, Module, ModuleInfo, QueryResponse,
+};
+use sov_state::{CacheLog, ValueReader};
+use sovereign_sdk::serial::{Decode, Encode};
+use std::{io::Cursor, marker::PhantomData};
+
+/// dispatch_tx is a high level interface used by the sdk.
+/// Transaction signature must be checked outside of this function.
+fn dispatch_tx<C: Context, VR: ValueReader>(
```
preston-evans98, line 14: It would be great to encode the constraint "Transaction signature must be checked outside of this function" into the type system. 

One potential idea - maybe we should make separate structs for `Context` and `Spec` - the `Spec` struct would contain type information (what signature scheme is being used, etc.), while the `Context` contained concrete tx data. 

So the API might look something like this:

```rust
pub struct Transaction<S: Spec> { 
   context: Context<S>,
   msg: Vec<u8>
}

impl<S: Spec> Transaction<S>{ 
  pub fn from_raw_tx(data: &mut &[u8]) -> Result<Self, anyhow::Error> { 
    let signature = S::Signature::decode(data)?;
    let (sender, msg) = signature.verify()?;
    // ... 
    let context = Context { 
       sender,
       // ...
    };
    Ok(Self { 
      context,
      msg
    })
  }
}
```

sov-modules/sov-modules-impl/integration-tests/tests/demo/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    fn query_value_adder(&self, data: <ValueAdderModule<C> as Module>::QueryMessage) -> Vec<u8> {
+        let query = RuntimeQuery::<C>::ValueAdder(data);
+        let mut data = Vec::default();
+        query.encode(&mut data);
+
+        data
+    }
+}
```
preston-evans98, line 157: nit: all of these functions can use the `encode_to_vec` method from our `Encode` trait

bkolad: done

preston-evans98 (approved):

sov-modules/sov-modules-impl/integration-tests/tests/demo/mod.rs
preston-evans98: Sure, it makes sense to keep that change separate if it will be invasive

PR #95, opened 2023-02-23
Implement temporary StateDB feature

preston-evans98:
Allow creation of StateDBs in temporary directories

### DIFF
--- a/db/sovereign-db/src/rocks_db_config.rs
+++ b/db/sovereign-db/src/rocks_db_config.rs
@@ -3,7 +3,8 @@
 
 use rocksdb::Options;
 
-/// Port selected RocksDB options for tuning underlying rocksdb instance of AptosDB.
+/// Port selected RocksDB options for tuning underlying rocksdb instance of our state db.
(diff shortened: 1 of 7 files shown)


db/sovereign-db/src/rocks_db_config.rs
```diff
@@ -0,0 +1,51 @@
+// Adapted from Aptos-Core.
+// Modified to remove serde dependency
+
+use rocksdb::Options;
+
+/// Port selected RocksDB options for tuning underlying rocksdb instance of AptosDB.
```
bkolad, line 6: Should we call it `SovDb` (last word of the above sentence.)

bkolad (approved):

db/sovereign-db/src/rocks_db_config.rs
preston-evans98: Good catch!

preston-evans98: Fixed!

PR #96, opened 2023-02-23
Use RocksDB for JMT Storage in tests

preston-evans98:
This commit revises the `Storage` trait with a configurable `new` function instead of relying on Default. It updates the `Genesis` trait and macro to use this new functionality, and it replaces the stubbed JmtStorage type with a real implementation backed by RocksDB.


sov-modules/sov-modules-impl/integration-tests/tests/demo/mod.rs
```diff
@@ -40,13 +40,30 @@ trait DispatchCall {
...
 // #[derive(Genesis, DispatchCall, DispatchQuery, Client)]
-#[derive(Genesis)]
+// #[derive(Genesis)]
 struct Runtime<C: Context> {
     election: Election<C>,
     value_adder: ValueSetter<C>,
     //..
 }
 
+impl<C: Context> Genesis for Runtime<C> {
```
bkolad, line 50: We can remove this and enable #[derive(Genesis)].

sov-modules/sov-state/src/jmt_storage.rs
```diff
@@ -34,3 +44,30 @@ impl JmtStorage {
         cache.get_first_reads()
     }
 }
+
+impl Storage for JmtStorage {
+    /// Instead of a config object, we just pass in the
+    /// db instance directly for now.
+    //  TODO: decide whether to use an actual config, or
+    //  rename this type
+    type Config = JmtDb;
```
bkolad, line 53: I think the config should propagate via `Spec` trait, but we can do it later.

sov-modules/sov-state/src/storage.rs
```diff
@@ -105,7 +108,13 @@ pub struct GenericStorage<VR: ValueReader> {
     pub(crate) internal_cache: StorageInternalCache,
 }
 
-impl<VR: ValueReader> Storage for GenericStorage<VR> {
+impl<VR: ValueReader + Default> Storage for GenericStorage<VR> {
```
bkolad, line 111: Let's remove the `GenericStorage` type and implement `ZkStorage` directly, the same way as `JmtStorage`.

bkolad (approved):

sov-modules/sov-modules-impl/integration-tests/tests/demo/mod.rs
preston-evans98: Done!

bkolad (approved):

PR #102, opened 2023-02-27
Add support for `#[derive(DispatchQuery))`

### LINKED ISSUE #84: sov-modules: Create `#[derive(DispatchQuery)]` macro
Create a mechanism for dispatching a query to the correct module.

bkolad:
1. Implements derive for `sov-modules-api::DispatchQuery` trait and adds relevant tests.
2. Extracts common code (`StructDef`) for `DispatchQuery, DispatchCall` to common.rs

close #84


sov-modules/sov-modules-api/src/dispatch.rs
```diff
@@ -23,3 +23,14 @@ pub trait DispatchCall {
         context: &Self::Context,
     ) -> Result<CallResponse, Error>;
 }
+
+/// A trait that needs to be implemented for any query message.
+pub trait DispatchQuery {
+    type Context: Context;
+
+    /// Dispatches a query message to the appropriate module.
+    fn dispatch(
```
preston-evans98, line 32: Should we name this `dispatch_query` instead of `dispatch` so that we don't need to qualify it with the trait name everywhere?

bkolad: Done!

preston-evans98 (approved):

PR #105, opened 2023-03-01
Basic runtime demo

bkolad:
1. Adds a high level doc to `Runtime`
2. Imoproves the tests for `Runtime`
3. Moves `demo.rs` to `basic_runtime.rs`


sov-modules/sov-modules-impl/integration-tests/tests/demo/basic_runtime.rs
```diff
@@ -0,0 +1,199 @@
...
+use sovereign_db::state_db::StateDB;
+
+/// dispatch_tx is a high level interface used by the sdk.
+/// Transaction signature must be checked outside of this function.
+fn dispatch_tx<C: Context, VR: ValueReader>(
+    _tx_data: Vec<u8>,
+    _context: C,
+    _value_reader: VR,
+) -> Result<(CallResponse, CacheLog), Error> {
+    // 1. Create Storage (with fresh Cache)
```
preston-evans98, line 21: (non-blocking): In the zk enviornment, we might want to pass in the storage from the context. That way, if the prover knows that a transaction won't revert they can run it against the main storage and save all the overhead of creating and merging an extra cache.

But that's an optimization, and it only applies in zk - so definitely not a blocker. Maybe not worth doing at all.

sov-modules/sov-modules-impl/integration-tests/tests/demo/basic_runtime.rs
```diff
@@ -0,0 +1,199 @@
...
+/// 3. Queries:
+///    The `Module` interface defines a `query` method, which allows querying the state of the module.
+///     Queries are read only i.e they don't change the state of the rollup.
+///     
+/// `#[derive(MessageCodec)` adds deserialization capabilities to the `Runtime` (implements `decode_call` method).
+/// `Runtime::decode_call` accepts serialized call message and returns a type that implements the `DispatchCall` trait.
+///  The `DispatchCall` implementation (derived by a macro) forwards the message to the appropriate module and executes its `call` method.
+///
+/// Similar mechanism works for queries with the difference that queries are submitted by users directly to the rollup node
+/// instead of going trough the DA layer.
```
preston-evans98, line 62: nit: "trough" -> "through"

sov-modules/sov-modules-impl/integration-tests/tests/demo/basic_runtime.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Definition of the first module in the rollup (must implement the sov_modules_api::Module trait).
+    election: example_election::Election<C>,
+    /// Definition of the second module in the rollup (must implement the sov_modules_api::Module trait).
+    value_setter: example_value_setter::ValueSetter<C>,
+}
+
+fn run_example() {
+    type C = MockContext;
+    let sender = MockPublicKey::try_from("admin").unwrap();
+    let admin_context = MockContext::new(sender);
```
preston-evans98, line 74: nit: since we have the type alias `C`, it would nice to use it everywhere. That way, we can swap out the MockContext with just a change to the alias.

preston-evans98 (approved):

sov-modules/sov-modules-impl/integration-tests/tests/demo/basic_runtime.rs
bkolad: This is a good point. Let's discuss it tmr on the daily.

sov-modules/sov-modules-impl/integration-tests/tests/demo/basic_runtime.rs
bkolad: fixed

PR #109, opened 2023-03-02
Implementation of ValueReader for StateDb

bkolad:
1. Adds `get_value_option_by_key ` method in `StateDB`
2. Implements `ValueReader` for `StateDB`
3. Updates the `Genesis` trait, now the `genesis` method accepts `Storage` instead of `Config`. 
This change decoupled  storage creation  from the module system and enabled point 4. and 5. 
5. Removes `Config` associated type from the `Storage` trait.
6. Introduces `GenericStorage` struct which is shared by `JmtStorage` and `ZkStorage` (we used to have it)

### DIFF
--- a/sov-modules/sov-modules-api/src/dispatch.rs
+++ b/sov-modules/sov-modules-api/src/dispatch.rs
@@ -3,12 +3,9 @@ use crate::{CallResponse, Context, Error, QueryResponse, Spec};
 /// Methods from this trait should be called only once during the rollup deployment.
 pub trait Genesis {
     type Context: Context;
-    type Config;
 
     /// Initializes the state of the rollup.
-    fn genesis(
-        config: Self::Config,
-    ) -> Result<<<Self as Genesis>::Context as Spec>::Storage, Error>;
+    fn genesis(storage: <<Self as Genesis>::Context as Spec>::Storage) -> Result<(), Error>;
 }
 
 /// A trait that needs to be implemented for any call message.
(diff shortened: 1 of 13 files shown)


sov-modules/sov-modules-api/src/dispatch.rs
```diff
@@ -3,12 +3,9 @@ use crate::{CallResponse, Context, Error, QueryResponse, Spec};
 /// Methods from this trait should be called only once during the rollup deployment.
 pub trait Genesis {
     type Context: Context;
-    type Config;
 
     /// Initializes the state of the rollup.
-    fn genesis(
-        config: Self::Config,
-    ) -> Result<<<Self as Genesis>::Context as Spec>::Storage, Error>;
+    fn genesis(storage: <<Self as Genesis>::Context as Spec>::Storage) -> Result<(), Error>;
```
preston-evans98, line 8: Is there any situation where a user would want to initialize a module with something other than a `Storage` instance? For example, maybe you want to pass in a reference to the `account` or `token` module as well?

bkolad: This will work out of the box. Modules can reference each other with `#[module]` tag; they share the same `Storage` (but manage their own `Prefixes`). Here is a unit test for this:
https://github.com/Sovereign-Labs/sovereign/blob/research/sov-modules/sov-modules-macros/tests/module_info/mod_and_state.rs#L31 

I will add more tests for this.

preston-evans98: Oh, great! Nevermind then!

preston-evans98 (approved):

PR #120, opened 2023-03-07
JMT version handling in the Storage

### LINKED ISSUE #114: sov-modules: Update JMT version when batch is written to the db
1. We should read the latest JMT version from the DB on rollup startup.
2. On every `Storage::finalize` we should bump the version by one and save it in the DB.

bkolad:
close #114

### DIFF
--- a/db/sovereign-db/src/state_db.rs
+++ b/db/sovereign-db/src/state_db.rs
@@ -20,7 +20,7 @@ use crate::{
 #[derive(Clone)]
 pub struct StateDB {
     db: Arc<DB>,
-    version: Arc<Mutex<Version>>,
+    next_version: Arc<Mutex<Version>>,
 }
 
 impl StateDB {
@@ -32,13 +32,13 @@ impl StateDB {
             &gen_rocksdb_options(&Default::default(), false),
         )?;
 
-        let version = Self::last_version_written(&inner)?
+        let next_version = Self::last_version_written(&inner)?
             .map(|v| v + 1)
             .unwrap_or_default();
 
         Ok(Self {
             db: Arc::new(inner),
-            version: Arc::new(Mutex::new(version)),
+            next_version: Arc::new(Mutex::new(next_version)),
         })
     }
 
@@ -76,13 +76,13 @@ impl StateDB {
         }
     }
 
-    pub fn inc_version(&mut self) {
-        let mut version = self.version.lock().unwrap();
+    pub fn inc_next_version(&mut self) {
+        let mut version = self.next_version.lock().unwrap();
         *version += 1;
     }
 
-    pub fn get_version(&self) -> Version {
-        let version = self.version.lock().unwrap();
+    pub fn get_next_version(&self) -> Version {
+        let version = self.next_version.lock().unwrap();
         *version
     }
 
--- a/db/sovereign-db/src/state_db.rs
+++ b/db/sovereign-db/src/state_db.rs
@@ -7,6 +7,7 @@ use jmt::{
     storage::{TreeReader, TreeWriter},
     KeyHash, Version,
 };
+
 use schemadb::DB;
 
 use crate::{
@@ -76,7 +77,19 @@ impl StateDB {
         }
     }
 
-    fn inc_next_version(&self) {
+    pub fn update_db(
+        &self,
+        key: StateKey,
+        key_hash: KeyHash,
+        value: Option<Vec<u8>>,
+        next_version: Version,
+    ) -> anyhow::Result<()> {
+        self.put_preimage(key_hash, &key)?;
+        self.db.put::<JmtValues>(&(key, next_version), &value)?;
+        Ok(())
+    }
+
+    pub fn inc_next_version(&self) {
         let mut version = self.next_version.lock().unwrap();
         *version += 1;
     }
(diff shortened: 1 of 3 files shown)


sov-modules/sov-state/src/jmt_storage.rs
```diff
@@ -86,23 +114,81 @@ impl Storage for JmtStorage {
...
-            data.push(((0, key_hash), value));
+            let value =
+                cache_value.map(|v| Arc::try_unwrap(v.value).unwrap_or_else(|arc| (*arc).clone()));
+
+            data.push(((self.db.version, key_hash), value));
         }
 
-        batch.extend(vec![], data);
-        self.db.write_node_batch(&batch).unwrap();
+        if !data.is_empty() {
```
bkolad, line 125: Question: we don't bump the version if there are no writes to save.  Is it ok?

sov-modules/sov-state/src/jmt_storage.rs
```diff
@@ -8,14 +8,41 @@ use crate::{
...
 
-impl ValueReader for StateDB {
+type VersionAndKeyHash = (Version, KeyHash);
+
+#[derive(Clone)]
+struct StateDBAndVersion {
+    db: StateDB,
+    version: Version,
+}
+
```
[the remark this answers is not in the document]
bkolad: Done

db/sovereign-db/src/state_db.rs
```diff
@@ -17,6 +17,7 @@ use crate::{
 #[derive(Clone)]
 pub struct StateDB {
     db: Arc<DB>,
+    version: Version,
```
preston-evans98, line 20: It think this needs to be an `Arc<Mutex<Version>>` since we clone the StateDB everywhere

sov-modules/sov-state/src/jmt_storage.rs
bkolad: Ok, I have changed it.

db/sovereign-db/src/state_db.rs
```diff
@@ -27,8 +28,14 @@ impl StateDB {
             STATE_TABLES.iter().copied(),
             &gen_rocksdb_options(&Default::default(), false),
         )?;
+
+        let version = Self::last_version(&inner)?
+            .map(|v| v + 1)
+            .unwrap_or_default();
+
```
preston-evans98, line 35: Thinking about this more, I'm afraid that the semantics might be confusing. Is `version` the next version to write, or the last version written? 

Personally, I think `last_version_written` is probably preferable, since there might be circumstances where people write to the same version multiple times (for example, when recovering from a crash).

db/sovereign-db/src/state_db.rs
bkolad: This is a good catch! We only increment the version in the `Storage::finalize,` and only the top-level version of the storage calls it, that's why the tests didn't catch it. Updated it to use Mutex so we are safe in the future.

db/sovereign-db/src/state_db.rs
bkolad: I have changed it to `last_version_written`

db/sovereign-db/src/state_db.rs
```diff
@@ -17,6 +20,7 @@ use crate::{
 #[derive(Clone)]
 pub struct StateDB {
     db: Arc<DB>,
+    version: Arc<Mutex<Version>>,
```
preston-evans98, line 23: Can we change this to `last_version_written` as well?

bkolad: But this is `last_version_written +1`, so maybe `next_version`?

preston-evans98: We can do `next_version`!

preston-evans98 (commented):
Looks great overall! One more thing - I guess we should update `write_node_batch` to automatically update the `next_version` if any of the nodes have a version greater than or equal to `next_version`

bkolad:
> Looks great overall! One more thing - I guess we should update `write_node_batch` to automatically update the `next_version` if any of the nodes have a version greater than or equal to `next_version`

I updated the `write_node_batch`:
1. we assert that `nodes` version is eq to next_version
2. we inc `next_version` at the end of `write_node_batch` even if there are no data in the `node_batch`

db/sovereign-db/src/state_db.rs
```diff
@@ -97,10 +128,12 @@ impl TreeReader for StateDB {
 impl TreeWriter for StateDB {
     fn write_node_batch(&self, node_batch: &jmt::storage::NodeBatch) -> anyhow::Result<()> {
         for (node_key, node) in node_batch.nodes() {
+            assert_eq!(node_key.version(), self.get_next_version());
             self.db.put::<JmtNodes>(node_key, node)?;
         }
 
         for ((version, key_hash), value) in node_batch.values() {
+            assert_eq!(version, &self.get_next_version());
```
preston-evans98, line 136: Should we pull the `get_next_version()` call out into a variable so that we're not locking and unlocking the mutex in a loop? Use something like `let next_version = self.get_next_version();` outside of the loop, and then just `assert_eq!(version, next_version);` in each iteration.

If we do, I guess we should also handle the failure case since the local copy of `next_version` could be stale.
```rust
if version != next_version { 
  next_version = self.get_next_version;
  assert_eq!(version, next_version);
}
```

preston-evans98 (approved):

db/sovereign-db/src/state_db.rs
preston-evans98: Synced with @bkolad offline. We removed version checks from the TreeWriter entirely in favor of the new `update_db` method. Resolving this conversation.

preston-evans98:
This version looks great! Thanks for the persistence @bkolad!

PR #121, opened 2023-03-07
`basic_runtime` refactoring

bkolad:
1. Separates message creation from execution (see the `simulate_da()` function)
2. Introduces the `execute` function, which executes a single message 

These changes simplify the demo and structure the code closer to how it's going to be used with the `sovereign-sdk`

### DIFF
--- a/sov-modules/sov-modules-impl/integration-tests/tests/demo/basic_runtime.rs
+++ b/sov-modules/sov-modules-impl/integration-tests/tests/demo/basic_runtime.rs
@@ -2,14 +2,13 @@
 //! This `Runtime` serves as a basic example of how to wire up module system and trigger the rollup logic.
 
 #![allow(dead_code)]
-
-use example_election::Candidate;
 use sov_modules_api::{
     mocks::{MockContext, MockPublicKey},
     CallResponse, Context, DispatchCall, DispatchQuery, Error, Genesis, Module,
 };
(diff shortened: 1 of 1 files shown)


sov-modules/sov-modules-impl/integration-tests/tests/demo/basic_runtime.rs
```diff
@@ -68,186 +67,189 @@ pub struct Runtime<C: Context> {
...
     let query_response = module.dispatch_query(storage);
-    let response: example_value_setter::query::Response =
-        serde_json::from_slice(&query_response.response).unwrap();
 
-    response
-        == example_value_setter::query::Response {
-            value: Some(new_value),
-        }
+    let response = str::from_utf8(&query_response.response).unwrap();
+    assert_eq!(response, expected_response)
```
preston-evans98, line 88: (non-blocking): I think it would be preferable to decode the json response here instead of checking string equality. That way, we'll be robust to changes in the default json serialization behavior

preston-evans98 (approved):

sov-modules/sov-modules-impl/integration-tests/tests/demo/basic_runtime.rs
bkolad: The problem is that every module defines its own response type, so if we want to keep the function generic, we have to work with strings or add generics. Since it is not blocking, I will merge it, and we can catch up tmr.

PR #122, opened 2023-03-08
Create Demo-app package. Implement core of STF trait for Demo App. 

preston-evans98:
This PR Creates the Demo package and begins implementing the STF interface for it. Support for the `apply_proof` and `end_slot` methods will be added in a follow up.


demo-app/src/main.rs
```diff
@@ -0,0 +1,253 @@
+use std::marker::PhantomData;
+
```
bkolad, line 2: Let's move the all the data structures `DemoTransaction`, `Demo` etc.. and trait impls to a separate file `types.rs` or `impl.rs`

demo-app/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+    },
+    jmt,
+    serial::Decode,
+    stf::{ConsensusSetUpdate, OpaqueAddress, StateTransitionFunction},
+};
+
+// we should re-export anyhow from the modules_api
+
+#[derive(Debug, PartialEq, Eq, Clone, BorshDeserialize, BorshSerialize)]
+pub struct DemoTransaction<C: Context> {
```
bkolad, line 26: non-blocking: in `basic_runtime.rs` we have the `Message` type. I think we should unify DemoTransaction and Message let's add TODO for it.

demo-app/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+    mock_signature: MockSignature,
+    msg: Vec<u8>,
+    phantom_context: PhantomData<C>,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, BorshDeserialize, BorshSerialize)]
+pub struct MockSignature {
+    pub_key: MockPublicKey,
+    msg_hash: [u8; 32],
+}
```
bkolad, line 36: non-blocking: 
1. We should use `sov-modules-api::mocks::MockSignature` instead, let's add TODO
2. Should we keep the public key outside of signature object?

demo-app/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+        Vec<Vec<sovereign_sdk::stf::Event>>,
+        sovereign_sdk::stf::ConsensusSetUpdate<OpaqueAddress>,
+    > {
+        let mut storage = self.current_storage.clone();
+        let mut events = Vec::new();
+        for tx in batch.take_transactions() {
+            // Do mock signature verification
+            // We just check that the signature hash matches the tx hash
+            let expected_hash = tx.mock_signature.msg_hash;
+            let found_hash = <MockContext as Spec>::Hasher::hash(&tx.msg);
```
bkolad, line 78: non-blocking:
This logic is common for the native and zk logic so we should pass Context from the outside instead of hardcoding MockContext (we can parametric Demo struct by C)

demo-app/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+impl<C: Context> CanonicalHash for DemoTransaction<C> {
+    type Output = [u8; 32];
+
+    fn hash(&self) -> Self::Output {
+        self.mock_signature.msg_hash.clone()
+    }
+}
+
+fn main() {
+    type RT = Runtime<MockContext>;
```
bkolad, line 165: Let's move the content of the `basic_runtime ` here and delete `sov-modules/sov-modules-impl/integration-tests/tests/demo` So the integration test becomes the `demo app`.  

`basic_runtime`  defines the `execute` function which we can keep for this PR but the intention is to replace it with `apply_batch` and the code there is structured to make this change easy.

sov-modules/sov-modules-macros/src/dispatch/common.rs
```diff
@@ -110,8 +110,8 @@ impl<'a> StructDef<'a> {
             #[allow(non_camel_case_types)]
             // TODO we should not hardcode the serialization format inside the macro:
             // https://github.com/Sovereign-Labs/sovereign/issues/97
-            #[derive(borsh::BorshDeserialize, borsh::BorshSerialize)]
-            enum #enum_ident #impl_generics #where_clause {
+            #[derive(borsh::BorshDeserialize, borsh::BorshSerialize, ::core::fmt::Debug, PartialEq)]
+            pub enum #enum_ident #impl_generics #where_clause {
```
bkolad, line 114: Let's keep it private and use `Runtime::<C>::decode_call` for decoding

demo-app/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+            if let Ok(msg) = RuntimeCall::<MockContext>::decode(&mut std::io::Cursor::new(tx.msg)) {
+                let ctx = MockContext::new(tx.mock_signature.pub_key);
+                let tx_result = msg.dispatch_call(storage.clone(), &ctx);
+                match tx_result {
+                    Ok(resp) => {
+                        events.push(resp.events);
+                        storage.merge();
+                    }
+                    Err(_) => {
+                        storage.merge_reads_and_discard_writes();
```
bkolad, line 93: Question: why do we need to update storage on faulty tx?

bkolad (approved):
This looks great.
Added a couple of nits which we can do in separate PRs (marked as non-blocking). 
I think we should copy `basic_runtime.rs` to the demo app instead of using only the `election` module.

demo-app/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+                    }
+                }
+            } else {
+                // If the serialization is invalid, the sequencer is malicious. Slash them
+                return Err(ConsensusSetUpdate::slashing(sequencer));
+            }
+        }
```
bkolad, line 100: Should we save db here:`storage.finalize()`?

demo-app/src/main.rs
preston-evans98: 1. TODO added!

2. For a real signature scheme, the `pub_key` is either part of the signature or recovered from it using EcRecover, right? 
Maybe we should plan to use EcRecover and introduce separate types for unchecked and verified signatures. Maybe something like :

```rust 

struct RawSignature(Vec<u8>);
struct VerifiedSignature { 
   pub_key: PublicKey,
   msg_hash: [u8;32],
}
```

demo-app/src/main.rs
preston-evans98: Hmm, yeah this is a good point. Can we get away with using a concrete Spec instead of generic context here?

demo-app/src/main.rs
bkolad: But then we are limited to schemas that support pubkey recovery. Can we just send (PubKey, Sig) pair and verify the sig against the key?

demo-app/src/main.rs
bkolad: If the `ZkHasher` can have a different implementation than `NativeHasher` then it looks like we have to be generic...

demo-app/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+            // Do mock signature verification
+            // We just check that the signature hash matches the tx hash
+            let expected_hash = tx.mock_signature.msg_hash;
+            let found_hash = <MockContext as Spec>::Hasher::hash(&tx.msg);
+            // If the (mock) signature is invalid, the sequencer is malicious. Slash them.
+            if expected_hash != found_hash {
+                return Err(ConsensusSetUpdate::slashing(sequencer));
+            }
+
+            if let Ok(msg) = RuntimeCall::<MockContext>::decode(&mut std::io::Cursor::new(tx.msg)) {
```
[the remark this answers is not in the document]
preston-evans98: RuntimeCall has to be public if we're going to use it in the DemoTransaction struct. We could (a type like) `DemoTransaction` into `sov-modules` and auto-generate it as well, but I haven't thought through the tradeoffs

demo-app/src/main.rs
preston-evans98: Removed the existing `main` body.

demo-app/src/main.rs
preston-evans98: I think we want to finalize during `end_slot` for better throughput

demo-app/src/main.rs
preston-evans98: Oh, sure! I guess what I meant was that (once a signature is verified) we always have access to the pubkey. So we probably want that to be in the `VerifiedSignature` struct. It's fine if the pubkey is also included in the RawSignature type! 

The suggestion was really just adding a type boundary between Raw and Verified signatures

demo-app/src/main.rs
preston-evans98: Yep, that makes sense. I think I'd lean towards abstracting the difference between zk and native hashing inside the `Hasher` using conditional compilation. What do you think?

sov-modules/sov-modules-macros/src/dispatch/common.rs
preston-evans98: Yep, makes sense!

demo-app/src/main.rs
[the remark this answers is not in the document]
preston-evans98: Disregard the previous comment! We want RuntimeCall to be private, and we'll just use a Vec<u8> inside of `DemoTransaction`

demo-app/src/main.rs
bkolad: We can do this or  do the conditional compilation one level higher (here):
Context = MockContext or Context=ZkContext and hardcode Context everywhere. Let's merge it as it is and decide later.

PR #125, opened 2023-03-09
Demo App runs with mock data

bkolad:
1. The demo app executes mock transactions. 
2. Deletes the demo integration tests (now redundant)
3. Moves some of the data types to `sov-modules-api::mock.rs`. 
4. Moves `slot_cache` (previously `batch_cache`) and `tx_cache` to the same structure.


sov-modules/sov-state/src/jmt_storage.rs
```diff
@@ -26,8 +28,7 @@ impl ValueReader for StateDB {
 
 #[derive(Clone)]
 pub struct JmtStorage {
-    batch_cache: StorageInternalCache,
-    tx_cache: StorageInternalCache,
+    cache: Rc<RefCell<StorageInternalCache>>,
```
preston-evans98, line 31: (Non-blocking) If/when we support parallel execution, I think this will panic when we try to merge the `tx_cache`s in parallel.

In the long term, it might make sense to introduce a new abstraction of something like a `tx_group_cache` and remove the `tx_cache` and `slot_cache`. (In this setup a `tx_group` is a set of transactions that have to be executed sequentially due to data dependencies). We would also need to add a shared caching layer (with an RW Lock) to the DB. 

Then we could execute each tx_group in parallel and merge the resulting `tx_group_cache`s in parallel to create the `slot_cache`

preston-evans98 (approved):

sov-modules/sov-state/src/jmt_storage.rs
bkolad: Yep we will have to change the structure a bit but in general the logic is parallelizable.

PR #126, opened 2023-03-09
Allow Zk storage to finalize

preston-evans98:
This PR is marked as a draft because of some open design questions. Currently, the APIs are brittle and very sensitive to the ordering of different calls. For example, each of these three calls has a differen result!

```rust
// This version is correct
jmt_storage.get_first_reads(); 
jmt_storage.finalize();
jmt_storage.take_treedb_log();
```

```rust
// This version returns an empty FirstReads
jmt_storage.finalize();
jmt_storage.get_first_reads();
jmt_storage.take_treedb_log();
```

```rust
// This version panics
jmt_storage.take_treedb_log();
jmt_storage.finalize();
jmt_storage.get_first_reads();
```

I've opened this as a draft anyway for a few reasons: 
- We can use it temporarily to set up a demo for tomorrow, even though we probably won't merge the current design
- I hope that it sparks a useful discussion about how to set up host->guest interaction

### DIFF
--- a/sov-modules/sov-state/src/jmt_storage.rs
+++ b/sov-modules/sov-state/src/jmt_storage.rs
@@ -9,6 +9,7 @@ use std::{
 use crate::{
     internal_cache::{StorageInternalCache, ValueReader},
     storage::{StorageKey, StorageValue},
+    tree_db::TreeReadLogger,
     Storage,
 };
 use first_read_last_write_cache::cache::FirstReads;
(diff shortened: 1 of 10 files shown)


bkolad:
Some thoughts, maybe we could use `type state patter` to ensure the correct ordering of operations:
https://rustype.github.io/notes/notes/rust-typestate-series/rust-typestate-part-1.html

sov-modules/sov-state/src/jmt_storage.rs
```diff
@@ -67,9 +73,14 @@ impl JmtStorage {
...
     }
+
+    /// Take the log
+    pub fn take_treedb_log(&self) -> Option<TreeReadLogger> {
+        self.read_logger.borrow_mut().take()
+    }
 }
 
-impl Storage for JmtStorage {
+impl<H: SimpleHasher> Storage for JmtStorage<H> {
```
bkolad, line 83: What about parameterizing only the `finalize` fn by SimpleHasher instead of the whole struct?  This way we don't have to add generic parameter everywhere. (Not for this draft, but maybe we could consider it when we bring this change over to `research`)

PR #131, opened 2023-03-15
Adds `remove` method to `StateValue` and `StateMap`

bkolad:
1. Adds `remove` method to `StateValue` and `StateMap`
2. Introduces  tests for the new `remove` method with different interweavings of `Storage::merge and Storage::finalize`


preston-evans98 (approved):
Change looks good to me! The tests will get broken by the big storage refactor later today though :(. But I think we can adapt them without too much trouble, so go ahead and merge

PR #133, opened 2023-03-16
Add `Accounts` module

bkolad:
1. `Accounts` module, which keeps track of users `PublicKeys, Addresses, and Nonces`
2. Tests for the accounts module.

Missing:
1. Integration with the `DemoApp`
2. `Nonce` update (will be done with 1)
3. Documentation will be done once we have the `MultiSig` (it may introduce some changes to the `Accounts`)

### DIFF
--- a/sov-modules/sov-modules-impl/accounts/src/call.rs
+++ b/sov-modules/sov-modules-impl/accounts/src/call.rs
@@ -9,7 +9,10 @@ pub const UPDATE_ACCOUNT_MSG: [u8; 32] = [0; 32];
 
 #[derive(BorshDeserialize, BorshSerialize, Debug, PartialEq)]
 pub enum CallMessage<C: sov_modules_api::Context> {
+    // Creates a new account.
     CreateAccount,
+    // Updates a PublicKey for the corresponding Account.
+    // The sender must be in possession of the new PublicKey.
     UpdatePublicKey(C::PublicKey, C::Signature),
 }
 
@@ -42,12 +45,18 @@ impl<C: sov_modules_api::Context> Accounts<C> {
         self.exit_if_account_exist(&new_pub_key)?;
 
         let account = self.accounts.remove_or_err(context.sender())?;
-        self.public_keys.remove_or_err(&account.addr)?;
 
-        // Proof that the sender is in possession of the `new_pub_key`
+        // Sanity check.
+        anyhow::ensure!(
+            // This is guaranteed to be true.
+            self.public_keys.get(&account.addr).is_some(),
+            "Missing PublicKey"
+        );
+
+        // Proof that the sender is in possession of the `new_pub_key`.
         signature.verify(&new_pub_key, UPDATE_ACCOUNT_MSG)?;
 
-        // Only the public key is updated, but account data remains the same.
+        // Update the public key (account data remains the same).
         self.accounts.set(&new_pub_key, account);
         self.public_keys.set(&account.addr, new_pub_key);
         Ok(CallResponse::default())
@@ -56,7 +65,7 @@ impl<C: sov_modules_api::Context> Accounts<C> {
     fn exit_if_account_exist(&self, new_pub_key: &C::PublicKey) -> Result<()> {
         anyhow::ensure!(
             self.accounts.get(new_pub_key).is_none(),
-            "New Public Key already exists"
+            "New PublicKey already exists"
         );
         Ok(())
     }
--- a/sov-modules/sov-modules-impl/accounts/src/call.rs
+++ b/sov-modules/sov-modules-impl/accounts/src/call.rs
@@ -1,10 +1,8 @@
-use crate::Account;
-
 use super::Accounts;
-
+use crate::Account;
 use anyhow::Result;
 use borsh::{BorshDeserialize, BorshSerialize};
-use sov_modules_api::{Address, CallResponse};
+use sov_modules_api::{Address, CallResponse, PublicKey};
 
 #[derive(BorshDeserialize, BorshSerialize, Debug, PartialEq)]
 pub enum CallMessage<C: sov_modules_api::Context> {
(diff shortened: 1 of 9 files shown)


sov-modules/sov-modules-impl/accounts/src/call.rs
```diff
@@ -0,0 +1,80 @@
+use crate::Account;
+use crate::Accounts;
+use anyhow::Result;
+use borsh::{BorshDeserialize, BorshSerialize};
+use sov_modules_api::Signature;
+use sov_modules_api::{Address, CallResponse, PublicKey};
+
+pub const UPDATE_ACCOUNT_MSG: [u8; 32] = [0; 32];
```
preston-evans98, line 8: (Question) Can setting the message to all zeros interacts poorly with signature schemes since they interpret messages as numbers?

sov-modules/sov-modules-impl/accounts/src/call.rs
```diff
@@ -0,0 +1,80 @@
...
+    UpdatePublicKey(C::PublicKey, C::Signature),
+}
+
+impl<C: sov_modules_api::Context> Accounts<C> {
+    pub(crate) fn create_account(&mut self, context: &C) -> Result<CallResponse> {
+        self.exit_if_account_exist(context.sender())?;
+
+        let default_address = context.sender().to_address();
+        self.exit_if_address_exist(&default_address)?;
+
```
preston-evans98, line 25: nit: `exit_if_address_exist` -> `exit_if_address_exists`

sov-modules/sov-modules-impl/accounts/src/call.rs
```diff
@@ -0,0 +1,80 @@
...
+    // Updates a PublicKey for the corresponding Account.
+    // The sender must be in possession of the new PublicKey.
+    UpdatePublicKey(C::PublicKey, C::Signature),
+}
+
+impl<C: sov_modules_api::Context> Accounts<C> {
+    pub(crate) fn create_account(&mut self, context: &C) -> Result<CallResponse> {
+        self.exit_if_account_exist(context.sender())?;
+
+        let default_address = context.sender().to_address();
```
preston-evans98, line 23: non-blocking: Should we change `context.sender()` to be an address instead of a pubkey?

sov-modules/sov-modules-impl/accounts/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub nonce: u64,
+}
+
+#[derive(ModuleInfo)]
+pub struct Accounts<C: sov_modules_api::Context> {
+    #[state]
+    pub(crate) public_keys: sov_state::StateMap<Address, C::PublicKey, C::Storage>,
+
+    #[state]
+    pub(crate) accounts: sov_state::StateMap<C::PublicKey, Account, C::Storage>,
```
preston-evans98, line 23: Question: Do we need this mapping to be stored in state? Or could we get away with just having `accounts: sov_state::StateMap<C::Address, Account, C::Storage>,` and store pub_key directly in the account? 

If so, I think we could just use `accounts.get(address).unwrap_or_default()` instead of having a separate `create_account` function

sov-modules/sov-modules-impl/accounts/src/query.rs
```diff
@@ -0,0 +1,26 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(BorshDeserialize, BorshSerialize, Debug, PartialEq)]
+pub enum QueryMessage<C: sov_modules_api::Context> {
+    GetAccount(C::PublicKey),
+}
+
+#[derive(Deserialize, Serialize, Debug, Eq, PartialEq)]
+pub enum Response {
+    AccountExist { addr: [u8; 32], nonce: u64 },
```
preston-evans98, line 12: nit: `AccountExist` -> `AccountExists`

sov-modules/sov-modules-impl/accounts/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+                Ok(self.update_public_key(new_pub_key, sig, context)?)
+            }
+        }
+    }
+
+    #[cfg(feature = "native")]
+    fn query(&self, msg: Self::QueryMessage) -> sov_modules_api::QueryResponse {
+        match msg {
+            query::QueryMessage::GetAccount(pub_key) => {
+                let response = serde_json::to_vec(&self.get_account(pub_key)).unwrap();
```
preston-evans98, line 54: non-blocking: We should use an encoding with a checksum to display the address to humans. Maybe bech32?

sov-modules/sov-modules-impl/accounts/src/call.rs
bkolad: This has to be handled by signature schema itself. 
1. Messages are handled as numbers but [0;32] message does not necessarily maps to `0`  point. It can be that `[27;32]` means 0 for a particular schema so we can't really defend agains it. 
2. BLS for example has this property that `0` private key can produce signatures valid for any message and this is handled on the verification part.
3. We can (very unlikely but still) can generate a valid message that hashes to 0 and the signature schema should work.

But now, I think maybe we want our `burn` address to be `[0;32]` so it would be good to use a different number here (just to avoid confusion). I will change it to [1;32]

sov-modules/sov-modules-impl/accounts/src/call.rs
bkolad: I think we should have both, `context.sender_pub_key()` and `context.sender_addr()` but I will be able to populate the address only after we integrate `Accounts` module with the `DemoApp` so I will add it then (in 1 or 2 PRs).

sov-modules/sov-modules-impl/accounts/src/lib.rs
bkolad: Discussed offline and we will leave it as it is for now

sov-modules/sov-modules-impl/accounts/src/lib.rs
bkolad: Added an issue: https://github.com/Sovereign-Labs/sovereign/issues/134

sov-modules/sov-modules-impl/accounts/src/call.rs
```diff
@@ -0,0 +1,80 @@
...
+    // Creates a new account.
+    CreateAccount,
+    // Updates a PublicKey for the corresponding Account.
+    // The sender must be in possession of the new PublicKey.
+    UpdatePublicKey(C::PublicKey, C::Signature),
+}
+
+impl<C: sov_modules_api::Context> Accounts<C> {
+    pub(crate) fn create_account(&mut self, context: &C) -> Result<CallResponse> {
+        self.exit_if_account_exist(context.sender())?;
```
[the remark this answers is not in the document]
bkolad: done

PR #135, opened 2023-03-17
Add `address` method to `ModuleInfo` trait

bkolad:
1. Adds `address()` method to `ModuleInfo` trait
2. Adds `module_address` to `DispatchCall` trait
3. Converts `ModuleInfo` trait to use associated types instead of generics.


sov-modules/sov-modules-macros/src/dispatch/dispatch_query.rs
```diff
@@ -28,7 +28,7 @@ impl<'a> StructDef<'a> {
 
             quote::quote!(
                 #enum_ident::#name(message)=>{
-                    let #name = <#ty as sov_modules_api::ModuleInfo::#type_generics>::new(working_set.clone());
+                    let #name = <#ty as sov_modules_api::ModuleInfo>::new(working_set.clone());
```
preston-evans98, line 31: (Question): Why are we removing the generics here?

sov-modules/sov-modules-macros/src/module_info.rs
```diff
@@ -199,6 +204,20 @@ fn make_prefix_func(
...
 
+fn make_fn_address(module_ident: &proc_macro2::Ident) -> proc_macro2::TokenStream {
+    quote::quote! {
+        fn address() -> sov_modules_api::Address {
+            use sov_modules_api::Hasher;
+            let module_path = module_path!();
+
+            let mut hasher = <<Self::Context as sov_modules_api::Spec>::Hasher>::new();
+            hasher.update(module_path.as_bytes());
+            hasher.update(stringify!(#module_ident).as_bytes());
```
preston-evans98, line 215: Is there any way that a malicious module can cause a collision, or is there  a natural domain separator at the end of module path?

bkolad: Fixed, now we calculate it as a hash of the module the prefix (which has a domain separator).

sov-modules/sov-modules-macros/src/dispatch/dispatch_query.rs
bkolad: `ModuleInfo` used to have generic type `ModuleInfo<C: Context>` but I changed it to the associated type, and now we don't need the generic parameter anymore.

preston-evans98 (approved):

PR #138, opened 2023-03-20
`DemoApp` and `Accounts` integration

bkolad:
1. Integrates `Accounts` module with `DemoApp`
2. Changes `Context::sender` type from `PublicKey` to `Address` + updates all relevant unit tests
3. Introduces `TxHooks` trait. `TxHooks` allows injecting custom logic into a transaction processing pipeline. For example,
now we have a way to update a nonce after a tx is processed

### DIFF
--- a/demo-app/src/stf.rs
+++ b/demo-app/src/stf.rs
@@ -14,28 +14,25 @@ use sovereign_sdk::{
     stf::{ConsensusSetUpdate, OpaqueAddress, StateTransitionFunction},
 };
 
-pub(crate) struct Demo<C: Context, V: TxVerifier, H: TxHooks> {
+pub(crate) struct Demo<C: Context, V: TxVerifier> {
     pub current_storage: C::Storage,
     pub verifier: V,
-    pub tx_hooks: H,
     pub working_set: RefCell<Option<WorkingSet<C::Storage>>>,
 }
 
-impl<C: Context> Demo<C, DemoAppTxVerifier<C>, DemoAppTxHooks<C>> {
+impl<C: Context> Demo<C, DemoAppTxVerifier<C>> {
     pub fn new(storage: C::Storage) -> Self {
         Self {
             current_storage: storage,
             verifier: DemoAppTxVerifier::new(),
-            tx_hooks: DemoAppTxHooks::new(),
             working_set: RefCell::new(None),
         }
     }
 }
 
-impl<C: Context, V, H> StateTransitionFunction for Demo<C, V, H>
+impl<C: Context, V> StateTransitionFunction for Demo<C, V>
 where
     V: TxVerifier<Context = C>,
-    H: TxHooks<Context = C>,
 {
     type StateRoot = jmt::RootHash;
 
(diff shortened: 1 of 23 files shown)


demo-app/src/tx_hooks.rs
```diff
@@ -0,0 +1,63 @@
...
+        anyhow::ensure!(
+            acc_nonce == tx_nonce,
+            "Tx bad nonce, expected: {acc_nonce}, but found: {tx_nonce}",
+        );
+
+        Ok(VerifiedTx {
+            pub_key: tx.pub_key,
+            sender: acc.addr,
+            runtime_msg: tx.runtime_msg,
+        })
```
preston-evans98, line 53: non-blocking: I think this logic is going to be used in basically every `hooks` for a chain that uses the accounts module. Maybe we can abstract it into a function?

demo-app/src/tx_verifier.rs
```diff
@@ -48,13 +46,6 @@ pub(crate) trait TxVerifier {
 
         Ok(txs)
     }
-
-    /// Runs stateful checks against a Transaction. This method can modify the storage.
-    fn verify_tx_stateful(
```
preston-evans98, line 53: We will definitely want `check_tx` abstraction for the mempool, and it's probably useful to keep the stateful vs. stateless distinction (since stateful checks need to be re-run during block building, but the stateless checks can be skipped). Let's think about how that interacts with the new hooks

demo-app/src/stf.rs
```diff
@@ -74,18 +75,21 @@ where
...
-                .verify_tx_stateful(tx, batch_workspace.clone())
+            let verified_tx = tx_hooks
+                .pre_dispatch_tx_hook(tx)
                 .or(Err(ConsensusSetUpdate::slashing(sequencer)))?;
 
             if let Ok(msg) = Runtime::<C>::decode_call(&verified_tx.runtime_msg) {
                 let ctx = C::new(verified_tx.sender);
                 let tx_result = msg.dispatch_call(batch_workspace.clone(), &ctx);
 
+                tx_hooks.post_dispatch_tx_hook(verified_tx);
```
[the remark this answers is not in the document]
bkolad: I think we will need it at some point, but then, we should add two methods, `post_dispatch_tx_hook_sucess` and `post_dispatch_tx_hook_failure`, and insert them at lines 95 and 99.  I prefer to postpone adding these methods until we have a need for them (so we understand better what the API should look like).

preston-evans98 (approved):

demo-app/src/tx_hooks.rs
```diff
@@ -61,3 +52,23 @@ impl<C: Context> TxHooks for DemoAppTxHooks<C> {
             .unwrap_or_else(|e| panic!("Inconsistent nonce {e}"));
     }
 }
+
+impl<C: Context> DemoAppTxHooks<C> {
+    fn accounts_hook(
```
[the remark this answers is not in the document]
bkolad: done

demo-app/src/tx_verifier.rs
bkolad: Discussed offline

PR #140, opened 2023-03-20
Update sov modules to work with `&mut WorkingSet`.

bkolad:
Add `&mut WorkingSet` to all methods in `sov-modules`

See `sov-modules/sov-modules-impl/accounts/src/call.rs` for how it impacts the API.

### DIFF
--- a/sov-modules/sov-state/src/backend.rs
+++ b/sov-modules/sov-state/src/backend.rs
@@ -27,13 +27,22 @@ impl<K: Encode, V: Encode + Decode, S: Storage> Backend<K, V, S> {
         &self.prefix
     }
 
-    pub(crate) fn set_value(&mut self, storage_key: StorageKey, value: V) {
+    pub(crate) fn set_value(
+        &mut self,
+        storage_key: StorageKey,
+        value: V,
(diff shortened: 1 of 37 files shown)


sov-modules/sov-state/src/backend.rs
```diff
@@ -42,9 +50,13 @@ impl<K: Encode, V: Encode + Decode, S: Storage> Backend<K, V, S> {
         )
     }
 
-    pub(crate) fn remove_value(&mut self, storage_key: StorageKey) -> Option<V> {
-        let storage_value = self.get_value(storage_key.clone())?;
-        self.storage.delete(storage_key);
+    pub(crate) fn remove_value(
```
preston-evans98, line 53: (non-blocking, out-of-scope): Should we add an additional `delete` method which doesn't return the old value? That way, we can avoid a read in the MPT if clients don't need it

preston-evans98 (approved):
Looks great!

sov-modules/sov-state/src/backend.rs
bkolad: Created an issue: https://github.com/Sovereign-Labs/sovereign/issues/141

PR #146, opened 2023-03-22
Remove backend from sov-state

bkolad:
After removing the `ValueReader` trait, we don't share that much code between `StateValue` and `StateMap`, so the `Backend` is not very useful anymore. 
1. Moves the few common methods to `WorkingSet` and removes `Backend`
2. `StorageKey & StorageValue` are created in the same place (`WorkingSet`), before the logic was spread across two files.


sov-modules/sov-state/src/map.rs
```diff
@@ -1,11 +1,14 @@
-use crate::{backend::Backend, storage::StorageKey, Prefix, Storage, WorkingSet};
+use std::marker::PhantomData;
+
+use crate::{storage::StorageKey, Prefix, Storage, WorkingSet};
 use sovereign_sdk::serial::{Decode, Encode};
 use thiserror::Error;
 
 /// A container that maps keys to values.
 #[derive(Debug)]
 pub struct StateMap<K, V, S: Storage> {
```
preston-evans98, line 9: Now that we're not keeping a reference to storage, should we remove the generic and make `StateMap<K, V>`? That way, users wouldn't have to think about what kind of storage they wanted when they instantiated a state map.

If we move the method to the accessor methods, the compiler will infer it from the WorkingSet so users should never have to think about it except when they're creating a WorkingSet

```rust
    /// Inserts a key-value pair into the map.
    pub fn set<S: Storage>(&mut self, key: &K, value: V, working_set: &mut WorkingSet<S>) {
        working_set.set_value(self.prefix(), key, value)
    }

    /// Returns the value corresponding to the key or None if key is absent in the StateMap.
    pub fn get<S: Storage>(&mut self, key: &K,  working_set: &mut WorkingSet<S>)-> Option<V> {
        working_set.get_value(self.prefix(), key)
    }
    
    ...
```

sov-modules/sov-state/src/value.rs
```diff
@@ -17,8 +19,8 @@ impl Encode for SingletonKey {
 /// Container for a single value.
 #[derive(Debug)]
 pub struct StateValue<V, S: Storage> {
```
preston-evans98, line 21: If we remove the generic `S` from `StateMap`, we should do it here too

sov-modules/sov-state/src/map.rs
bkolad: This is a bit unfortunate, but I had to keep it this way because a struct implementing a `sov_modules_api::Module` needs to contain `C`: 

```
impl<C: sov_modules_api::Context> sov_modules_api::Module for MyModule<C> {
    type Context = C;
   ...
}
```
Without MyModule<**C**>  rust complains about unconstrained C, we could add `PhantomData` to `MyModule`, but then we would need to change the macros to understand `PhantomData` This is not a big deal, but I before doing this I would like to check if there is a more elegant solution.

PR #150, opened 2023-03-23
Extract `Runtime` from `Demo`

bkolad:
1. Hardcoded `Runtime` and `TxHooks` are extracted from the `Demo`. Now `Demo` is "runtime independent" (depends only on `DispatchCall+Genesis` traits)
2.  In order to improve ergonomics and make 1 possible, `DispatchCall/Query` traits are extended with `decode_call/query` methods. Before, these methods were implemented (macro-generated) directly on the `Runtime` struct.


preston-evans98 (approved):
Looks great!

PR #151, opened 2023-03-23
Add ledger_db

preston-evans98:
This PR adds the ability to persist ledger data (DA layer "slot" information, as well as batches, transactions, and events).

### DIFF
--- a/db/sovereign-db/src/ledger_db.rs
+++ b/db/sovereign-db/src/ledger_db.rs
@@ -1,0 +1,251 @@
+use std::{
+    collections::HashMap,
+    path::Path,
+    sync::{Arc, Mutex},
+};
+
+use anyhow::ensure;
+use schemadb::{Schema, DB};
+use sovereign_sdk::{db::SlotStore, services::da::SlotData, stf::Event};
+
+use crate::{
+    rocks_db_config::gen_rocksdb_options,
+    schema::{
+        tables::{
+            BatchByHash, BatchByNumber, EventByKey, EventByNumber, SlotByHash, SlotByNumber,
+            TxByHash, TxByNumber, LEDGER_TABLES,
+        },
+        types::{
+            BatchNumber, EventNumber, SlotNumber, StoredBatch, StoredSlot, StoredTransaction,
+            TxNumber,
+        },
+    },
+};
+
+const LEDGER_DB_PATH_SUFFIX: &'static str = "ledger";
+
+#[derive(Clone)]
+/// A database which stores the ledger history (slots, transactions, events, etc).
+/// Ledger data is first ingested into an in-memory map before being fed to the state-transition function.
+/// Once the state-transition function has been executed and finalzied, the results are committed to the final db
+pub struct LedgerDB<S> {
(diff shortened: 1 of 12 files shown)


db/sovereign-db/src/ledger_db.rs
```diff
@@ -0,0 +1,245 @@
...
+        // Insert transactions and events
+        for (idx, (tx, event_group)) in data_to_commit
+            .txs
+            .into_iter()
+            .zip(data_to_commit.events.into_iter())
+            .enumerate()
+        {
+            let tx_number = TxNumber(item_numbers.tx_number + idx as u64);
+            for event in event_group.into_iter() {
+                self.put_event(&event, &EventNumber(event_number), tx_number)?;
```
bkolad, line 196: Nit: it looks like here we have full ownership of `EventNumber`, so we could pass it by value instead of by reference and avoid cloning inside `put_value`: https://github.com/Sovereign-Labs/sovereign/pull/151/files#diff-a2bb8d643ebc5191d070ef5ad4243cb92e2821b839b7b64f37f85bb1ad486b75R152 (unless the cloning is cheap but then we could derive `Copy`)

db/sovereign-db/src/ledger_db.rs
```diff
@@ -0,0 +1,245 @@
...
+        tx_number: TxNumber,
+    ) -> Result<(), anyhow::Error> {
+        self.db.put::<EventByNumber>(event_number, event)?;
+        self.db
+            .put::<EventByKey>(&(event.key.clone(), tx_number, event_number.clone()), &())
+    }
+
+    pub fn commit_slot(&self, data_to_commit: SlotCommit) -> Result<(), anyhow::Error> {
+        let item_numbers = {
+            let mut next_item_numbers = self.next_item_numbers.lock().unwrap();
```
bkolad, line 157: Here we can use a method `get_next_items_numbers`

db/sovereign-db/src/ledger_db.rs
```diff
@@ -0,0 +1,245 @@
...
+        event_number: &EventNumber,
+        tx_number: TxNumber,
+    ) -> Result<(), anyhow::Error> {
+        self.db.put::<EventByNumber>(event_number, event)?;
+        self.db
+            .put::<EventByKey>(&(event.key.clone(), tx_number, event_number.clone()), &())
+    }
+
+    pub fn commit_slot(&self, data_to_commit: SlotCommit) -> Result<(), anyhow::Error> {
+        let item_numbers = {
```
bkolad, line 156: Nit: Can we make it a function or add a comment that the scope releases the lock? I am afraid someone might remove the brackets in the future without noticing the lock is released at the end of the scope.

bkolad (approved):

db/sovereign-db/src/ledger_db.rs
preston-evans98: It's already `Copy`!  I'll just change the `clone` into a dereference

db/sovereign-db/src/ledger_db.rs
preston-evans98: Done!

PR #156, opened 2023-03-29
Initial design of the `Bank` module

bkolad:


sov-modules/sov-modules-impl/bank/src/call.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use crate::{Amount, Bank, Coins};
+
+#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq)]
+pub enum CallMessage<C: sov_modules_api::Context> {
+    CreateToken {
+        token_name: String,
+        initial_balance: Amount,
+
+        minter_address: C::Address,
```
bkolad, line 13: Q: should we allow only the sender of the tx to be the minter?

sov-modules/sov-modules-impl/bank/src/lib.rs
```diff
@@ -0,0 +1,64 @@
+mod call;
+mod genesis;
+mod query;
+use sov_modules_api::Error;
+use sov_modules_macros::ModuleInfo;
+use sov_state::WorkingSet;
+
+// Q: is u64 good enough?
+type Amount = u64;
```
bkolad, line 9: Q: Is u64 good enough?

preston-evans98: Yeah, I think so

sov-modules/sov-modules-impl/bank/src/call.rs
```diff
@@ -0,0 +1,56 @@
+use anyhow::Result;
+use sov_modules_api::CallResponse;
+use sov_state::WorkingSet;
+
+use crate::{Amount, Bank, Coins};
+
+#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq)]
+pub enum CallMessage<C: sov_modules_api::Context> {
+    CreateToken {
```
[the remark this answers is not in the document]
bkolad: I would prefer to add it when we need it (maybe we can improve upon ERC20)

sov-modules/sov-modules-impl/bank/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+    balances: sov_state::StateMap<Address, Amount>,
+}
+
+#[derive(ModuleInfo)]
+pub struct Bank<C: sov_modules_api::Context> {
+    #[address]
+    pub address: C::Address,
+
+    #[state]
+    pub names: sov_state::StateMap<String, C::Address>,
```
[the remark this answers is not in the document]
bkolad: We decided  (offline) to support multiple tokens with the same name. I will remove the mapping. And because of that, `name squatting` is less of an issue.

preston-evans98 (approved):

PR #158, opened 2023-04-01
Introduce `sov-app-template` crate

bkolad:
1. Removes `stf.rs` and associated traits from the `demo-app`
2. Introduces a new `sov-app-template` crate, the crate contains template code for an `App`  (all application-specific logic is injected via the `new` constructor)
3. Implements the `demo-app` using the newly created `sov-app-template`


demo-app/src/tx_hooks_impl.rs
```diff
@@ -1,24 +1,25 @@
...
+    pub(crate) runtime_msg: Vec<u8>,
+}
 
-    /// post_dispatch_tx_hook runs after the tx is dispatched to an appropriate module.
-    fn post_dispatch_tx_hook(
-        &mut self,
-        tx: VerifiedTx<Self::Context>,
-        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
-    );
+impl<C: Context> VerifiedTx for AppVerifiedTx<C> {
```
preston-evans98, line 13: (non-blocking): This trait looks very similar to the `BlobWithSender` trait. Maybe in the future we can unify the two of them?

preston-evans98 (approved):

demo-app/src/tx_hooks_impl.rs
bkolad: I will have a look!

PR #159, opened 2023-04-01
Implement begin_slot in app template

bkolad:


preston-evans98 (approved):
Wow, whoever thought of this was really good 😂

PR #160, opened 2023-04-01
Update Address to use `[u8;32]` internally 

bkolad:


sov-modules/sov-modules-api/src/lib.rs
```diff
@@ -34,22 +34,25 @@ impl AsRef<[u8]> for Address {
...
-            addr: addr.to_vec(),
-        })
-    }
-}
-
 impl AddressTrait for Address {}
 
 /// Default implementation of AddressTrait for the module system
 #[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq, Clone, Eq)]
 pub struct Address {
```
preston-evans98, line 41: Looks great. I think we can remove the custom address from the demo-rollup once this lands too

preston-evans98 (approved):

PR #162, opened 2023-04-07
Bank module implementation

bkolad:
1. Implements `Bank` module functionality
2. Adds `From<[u8;32]>` implementation for `Address` (Our hash is always 32 bytes, and it makes sense to be able to derive an Address from hash)

Issue for better test coverage: https://github.com/Sovereign-Labs/sovereign/issues/163

### DIFF
--- a/sov-modules/sov-modules-impl/bank/src/call.rs
+++ b/sov-modules/sov-modules-impl/bank/src/call.rs
@@ -1,8 +1,8 @@
+use crate::{Amount, Bank, Coins, Token};
 use anyhow::{bail, Result};
+use sov_modules_api::Hasher;
 use sov_modules_api::{CallResponse, Spec};
 use sov_state::WorkingSet;
-
-use crate::{Amount, Bank, Coins, Token};
 
 #[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq)]
 pub enum CallMessage<C: sov_modules_api::Context> {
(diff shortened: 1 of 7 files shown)


sov-modules/sov-modules-impl/bank/src/call.rs
```diff
@@ -1,56 +1,102 @@
...
 use sov_state::WorkingSet;
 
-use crate::{Amount, Bank, Coins};
-
+/// This enumeration represents the available call messages for interacting with the bank module.
 #[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq)]
 pub enum CallMessage<C: sov_modules_api::Context> {
+    /// Creates a new token with the specified name and initial balance.
     CreateToken {
+        /// salt: a random value use to create a unique token address.
```
preston-evans98, line 11: (question, non-blocking): Is there an advantage to starting each comment with the name of the field?

sov-modules/sov-modules-impl/bank/src/call.rs
```diff
@@ -1,56 +1,102 @@
...
+        initial_balance: Amount,
+        minter_address: C::Address,
+        context: &C,
+        working_set: &mut WorkingSet<C::Storage>,
     ) -> Result<CallResponse> {
-        // This function will create a unique address for `token_name` and insert, the new `Token` to self.tokens
-        todo!()
+        let token_address = super::create_token_address::<C>(&token_name, context.sender(), salt);
+
+        match self.tokens.get(&token_address, working_set) {
```
preston-evans98, line 48: nit: It think it would be slightly more idiomatic not to `match` here. 

```
if self.tokens.get(&token_address, working_set).is_some() {
  bail!("Token address already exists");
}
...
```

preston-evans98 (approved):

sov-modules/sov-modules-impl/bank/src/call.rs
bkolad: Not really I can remove it

sov-modules/sov-modules-impl/bank/src/call.rs
bkolad: done

PR #164, opened 2023-04-11
Genesis configuration 

bkolad:
1. Adds `Config` type  to `Module` trait. This way every module specifies its own config type.
2. Adds `Config` type to `Genesis` trait. `Config` from the `Genesis` is a container for module-specific Configs, something like:
```
struct GenesisConfig{
  mod_1: <Mod_1 as Module>::Config,
  mod_2: <Mod_2 as Module>::Config,
  ...
}
```
3. Updates `Genesis` macro to handle the additional `Config` type and code to autogenerate `GenesisConfig` from `Runtime` definition.
4. Fixes a bug in `Genesis` macro: The, letter `C` for  `Context` was hardcoded inside the macro, so `K: Context` was not compiling. 

All relevant tests will be updated in the next PR.


sov-modules/sov-modules-api/src/dispatch.rs
```diff
@@ -4,10 +4,12 @@ use sov_state::WorkingSet;
 /// Methods from this trait should be called only once during the rollup deployment.
 pub trait Genesis {
     type Context: Context;
+    type Config;
 
     /// Initializes the state of the rollup.
     fn genesis(
         &self,
+        config: &Self::Config,
```
preston-evans98, line 12: (Question, non-blocking) - why do we take a reference instead of an owned Config here?

sov-modules/sov-modules-macros/tests/dispatch/modules.rs
```diff
@@ -54,20 +59,25 @@ pub mod second_test_module {
     pub struct TestType {}
 
     #[derive(ModuleInfo)]
-    pub(crate) struct SecondTestStruct<C: Context> {
+    pub struct SecondTestStruct<K: Context> {
```
preston-evans98, line 62: Nit: Using `K` for `Context` is a little bit confusing. Could we use `Ctx` instead?

preston-evans98 (approved):

sov-modules/sov-modules-api/src/dispatch.rs
bkolad: `init_chain` method does not take `Config` as a parameter (which is fine), so we pass it as `AppTemplate` field. The consequence is that we can't move it (unless we do the trick where we use `Option<Config>` and then we empty it, I can  do it in the next PR).

preston-evans98: Hmm - maybe we should just add it to `init_chain`

sov-modules/sov-modules-macros/tests/dispatch/modules.rs
bkolad: fixed

sov-modules/sov-modules-macros/tests/module_info/derive_on_enum_not_supported.rs
```diff
@@ -2,12 +2,12 @@ use sov_modules_macros::ModuleInfo;
...
 
 #[derive(ModuleInfo)]
-enum TestStruct {
+enum TestStruct<C: sov_modules_api::Context> {
     #[state]
     TestState1(StateMap<String, String>),
 
     #[state]
-    TestState2(StateMap<String, String>),
+    TestState2(C),
```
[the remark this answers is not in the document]
preston-evans98: Makes sense, thanks!

sov-modules/sov-modules-api/src/dispatch.rs
bkolad: Added an issue: https://github.com/Sovereign-Labs/sovereign/issues/165

PR #166, opened 2023-04-12
Add bech32 serialization 

### LINKED ISSUE #134: sov-modules: Add a way to display `Address` in human readable way
See: https://github.com/Sovereign-Labs/sovereign/pull/133#discussion_r1138738615

LukaszRozmej:
Closes #134 
Add bech32 serialization for [u8], make prefix parameterizable, add serialization for address

### DIFF
--- a/sov-modules/sov-modules-api/src/bech32.rs
+++ b/sov-modules/sov-modules-api/src/bech32.rs
@@ -1,0 +1,106 @@
+use std::{str::FromStr, fmt::{Display, self}};
+use bech32::{ToBase32, FromBase32, Error};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use crate::Address;
+
+pub fn vec_to_bech32(vec: &[u8], hrp: &str) -> Result<String, Error> {        
+    let data = vec.to_base32();
+    let bech32_addr = bech32::encode(hrp, data, bech32::Variant::Bech32)?;
+    Ok(bech32_addr.to_string())
+}
+
+pub fn bech32_to_vec(bech32_addr: &str) -> Result<(String, Vec<u8>), Error> {
+    let (hrp, data, _) = bech32::decode(bech32_addr)?;
+    let vec = Vec::<u8>::from_base32(&data)?;
+    Ok((hrp, vec))
+}
+
+const HRP: &str = "sov";
+
+#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq, Clone, Eq)]
+pub struct AddressBech32{    
+    value: String
+}
+
+impl TryFrom<&[u8]> for AddressBech32 {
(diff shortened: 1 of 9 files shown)


sov-modules/sov-modules-impl/accounts/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![feature(adt_const_params)]
```
LukaszRozmej, line 1: This is an unstable feature. I used it to pass HRP to the serialization/deserialization code `const HRP: &'static str`. It works here fine.

If not like this we can:
1. Hardcode the HRP making the serialization/deserialization only work for addresses (we can invoke general methods from it)
2. In response have a custom struct instead of just `Vec<u8>` that would contain the HRP as well as the data
3. Create a custom macro (overkill here)

sov-modules/sov-modules-impl/accounts/src/query.rs
```diff
@@ -29,3 +35,42 @@ impl<C: sov_modules_api::Context> Accounts<C> {
         }
     }
 }
+
+// TODO: Move to separate module?
```
LukaszRozmej, line 39: As this is general code we could move it somewhere else

sov-modules/sov-modules-api/src/bech32.rs
```diff
@@ -0,0 +1,106 @@
...
+            return Err(bech32::Error::InvalidLength);
+        }
+        let string = vec_to_bech32(addr, HRP)?;
+        Ok(AddressBech32{ value: string })
+    }    
+}
+
+impl From<&Address> for AddressBech32 {
+    fn from(addr: &Address) -> Self {
+        let string = vec_to_bech32(&addr.addr, HRP).unwrap();
```
bkolad, line 40: If this can fail then let's implement `TryFrom ` instead.

sov-modules/sov-modules-api/src/bech32.rs
```diff
@@ -0,0 +1,106 @@
...
+        AddressBech32{ value: string }
+    }    
+}
+
+#[derive(Debug)]
+pub enum Bech32ParseError {
```
bkolad, line 46: Let's use `thiserror` crate for this, it automatically derives some basic traits for an error. The manual impl of `From/Display`  won't be needed anymore. See example in `serial.rs` the crate is also well documented.

sov-modules/sov-modules-api/src/bech32.rs
LukaszRozmej: Yes was thinking about it, but this is guaranteed to be 32 bytes already and prefix is fixed. I think this just can't fail. At least I don't see how it could.

bkolad: ok

bkolad (approved):

PR #171, opened 2023-04-13
Add QueryResponse associated type to DispatchQuery

preston-evans98:
This commit implements four changes related to exposing typed data from the query apis (rather than just JSON strings, as in the current implementation):
1. Add DispatchQuery::QueryResponse type
2. Adjust the derive macros to handle the new associated type
3. Change the return type of the Bank::query to a new enum using the associated type
4. Adjust the derive macros to generate enums with proper casing

### DIFF
--- a/sov-modules/sov-modules-macros/src/dispatch/dispatch_query.rs
+++ b/sov-modules/sov-modules-macros/src/dispatch/dispatch_query.rs
@@ -1,12 +1,37 @@
-use super::common::{parse_generic_params, StructDef, StructFieldExtractor, QUERY, get_serialization_attrs};
+use super::common::{
+    get_serialization_attrs, parse_generic_params, StructDef, StructFieldExtractor, QUERY,
+    QUERY_RESPONSE,
+};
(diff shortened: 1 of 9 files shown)


sov-modules/sov-modules-macros/src/dispatch/dispatch_query.rs
```diff
@@ -1,12 +1,37 @@
...
+use super::common::{
+    get_serialization_attrs, parse_generic_params, StructDef, StructFieldExtractor, QUERY,
+    QUERY_RESPONSE,
+};
+use proc_macro2::Ident;
+use quote::format_ident;
 use syn::DeriveInput;
 
+/// A handy function that gpt4 generated to convert snake-case identifiers to camel-case
+pub(crate) fn convert_snake_case_to_upper_camel_case(ident: &Ident) -> Ident {
```
bkolad, line 10: let's move this function to `/dispatch/common.rs`

bkolad (approved):

PR #172, opened 2023-04-14
Design of the sequencer module

bkolad:
Draft of the `Sequencer` design. We can go over it on the next daily call. 
The `demo-app` tests will be updated in the next PR: https://github.com/Sovereign-Labs/sovereign/issues/174


sov-modules/sov-modules-impl/sequencer/src/hooks.rs
```diff
@@ -0,0 +1,37 @@
...
+        // Transfer `self.coins_to_lock` form `self.seq_rollup_address` to `self.address`
+        todo!()
+    }
+
+    pub fn next_sequencer(&self, working_set: &mut WorkingSet<C::Storage>) -> Result<Vec<u8>> {
+        Ok(self.inner.da_address.get_or_err(working_set)?)
+    }
+
+    pub fn slash(&self, working_set: &mut WorkingSet<C::Storage>) -> Result<()> {
+        todo!()
```
bkolad, line 28: Should we just burn `self.inner.slash` amount of coins?

sov-modules/sov-modules-impl/sequencer/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    fn genesis(
+        &self,
+        config: &Self::Config,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<(), Error> {
+        Ok(self.init_module(config, working_set)?)
+    }
+
+    // Questions:
```
bkolad, line 50: TODO: Resolve it before merging this PR.

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -146,6 +198,12 @@ where
         Self::StateRoot,
         Vec<sovereign_sdk::stf::ConsensusSetUpdate<OpaqueAddress>>,
     ) {
+        // TODO (discussion) possible alternative design for slashing sequencer:
```
bkolad, line 201: TODO: Resolve it before merging this PR.

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -84,24 +105,46 @@ where
         _misbehavior_hint: Option<Self::MisbehaviorProof>,
     ) -> Result<
         Vec<Vec<sovereign_sdk::stf::Event>>,
+        // Question: Should we return an enum error here? (see TODOs bellow)
         sovereign_sdk::stf::ConsensusSetUpdate<OpaqueAddress>,
     > {
+        let mut batch_workspace = WorkingSet::new(self.current_storage.clone());
+        batch_workspace = batch_workspace.to_revertable();
+
+        // TODO: Error handling.
```
bkolad, line 114: TODO: Resolve all todos in `apply_batch` before merging this PR.

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -84,24 +105,46 @@ where
...
             batch_workspace = batch_workspace.to_revertable();
             // Run the stateful verification, possibly modifies the state.
-            let verified_tx = self
-                .tx_hooks
-                .pre_dispatch_tx_hook(tx, &mut batch_workspace)
-                .or(Err(ConsensusSetUpdate::slashing(sequencer)))?;
+            let verified_tx = match self.tx_hooks.pre_dispatch_tx_hook(tx, &mut batch_workspace) {
+                Ok(verified_tx) => verified_tx,
+                Err(_) => {
+                    // TODO: should we revert here? // pre_dispatch_tx_hook changes nonces
```
preston-evans98, line 143: Yeah, I think revert is the right thing here - a malicious bundle should have no effect except for slashing the sequencer (so that we can short-circuit by jumping to the invalid transaction in some cases)

sov-modules/sov-modules-impl/sequencer/src/lib.rs
preston-evans98: (non-blocking) I think it would be nice to have a query via the sequencer module that returned the address and balance. That way, people can still query even if the bank module is private for some reason. I don't think we need any external calls.

demo-app/src/tx_hooks_impl.rs
```diff
@@ -64,11 +66,46 @@ impl<C: Context> TxHooks for DemoAppTxHooks<C> {
...
+        &self,
+        sequencer: &[u8],
+        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
+    ) -> Result<()> {
+        // TODO: Handle errors
+        match self.sequencer_hooks.next_sequencer(working_set) {
+            Ok(next_sequencer) => {
+                if next_sequencer != sequencer {
+                    // TODO: Return an error, should we slash in this case?
+                    todo!()
```
preston-evans98, line 80: I don't think we can slash here, because the invalid sequencer almost definitely won't happen to be bonded. 

Plus, this mechanism could potentially get abused to get other sequencers slashed (by holding onto a stale bundle that they signed and then posting it later. So I think we just want to skip execution

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -41,6 +46,22 @@ impl<C: Context, V, RT, H, GenesisConfig> AppTemplate<C, V, RT, H, GenesisConfig
             working_set: None,
         }
     }
+
+    fn revert_and_slash(&mut self, batch_workspace: WorkingSet<C::Storage>) {
+        // Revert all the changes (the sequencer funds are no longer locked)
+        let batch_workspace = batch_workspace.revert();
+
+        // Create a new batch_workspace
+        let mut batch_workspace = batch_workspace.to_revertable();
```
preston-evans98, line 55: Does this one need to be revertable? I don't think we ever revert a slashing, do we?

preston-evans98 (commented):
Looks good overall! For the centralized sequencer, I think it's probably fine if we don't even implement slashing - but it probably doesn't hurt to have it

demo-app/src/tx_hooks_impl.rs
bkolad: Done: Only error is returned.

sov-modules/sov-app-template/src/lib.rs
bkolad: Fixed

preston-evans98 (approved):

PR #178, opened 2023-04-19
Add sequencer tests in demo-app

bkolad:
1. close https://github.com/Sovereign-Labs/sovereign/issues/173
2. close https://github.com/Sovereign-Labs/sovereign/issues/174
4. changes the return type of `apply_batch` (now the errors are managed with `anyhow`)
5. adds error handling inside `apply_batch`
6. extends tests in `demo-app`
7. TODO: https://github.com/Sovereign-Labs/sovereign/issues/179

### DIFF
--- a/sdk/src/state_machine/core/run.rs
+++ b/sdk/src/state_machine/core/run.rs
@@ -114,7 +114,7 @@ impl<DaLayer: DaLayerTrait, App: StateTransitionFunction> Rollup<DaLayer, App> {
                         match self.app.apply_batch(batch, tx.sender().as_ref(), None) {
                             // TODO: handle events
                             Ok(_events) => {}
-                            Err(slashing) => current_sequencers.process_update(slashing),
+                            Err(_slashing) => todo!(), //current_sequencers.process_update(slashing),
                         };
                     }
                 }
--- a/sov-modules/sov-app-template/src/lib.rs
+++ b/sov-modules/sov-app-template/src/lib.rs
@@ -110,7 +110,10 @@ where
             .tx_hooks
             .enter_apply_batch(sequencer, &mut batch_workspace)
         {
-            anyhow::bail!("Error: enter_apply_batch. {}", e)
+            anyhow::bail!(
+                "Error: The transaction was rejected by the 'enter_apply_batch' hook. {}",
+                e
+            )
         }
 
         let mut events = Vec::new();
@@ -123,7 +126,7 @@ where
             Ok(txs) => txs,
             Err(e) => {
                 self.revert_and_slash(batch_workspace);
-                anyhow::bail!("Stateless verification error: {}", e);
+                anyhow::bail!("Stateless verification error - the sequencer included a transaction which was known to be invalid. {}", e);
             }
         };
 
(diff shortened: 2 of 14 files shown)


sdk/src/state_machine/core/run.rs
```diff
@@ -114,7 +114,7 @@ impl<DaLayer: DaLayerTrait, App: StateTransitionFunction> Rollup<DaLayer, App> {
                         match self.app.apply_batch(batch, tx.sender().as_ref(), None) {
                             // TODO: handle events
                             Ok(_events) => {}
-                            Err(slashing) => current_sequencers.process_update(slashing),
+                            Err(_slashing) => todo!(), //current_sequencers.process_update(slashing),
```
bkolad, line 117: @preston-evans98 are we going to maintain the `run.rs`? My refactoring broke the logic here but I remember we said in the past that `run.rs` is not up to date anymore.

preston-evans98: Now that we have the demo, I think it's fine to deprecated `run.rs`

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -100,21 +102,15 @@ where
...
 
-        if self
+        if let Err(e) = self
             .tx_hooks
             .enter_apply_batch(sequencer, &mut batch_workspace)
-            .is_err()
         {
-            // TODO: Enable it after: https://github.com/Sovereign-Labs/sovereign/issues/174
-            //return Ok(Vec::default());
+            anyhow::bail!("Error: enter_apply_batch. {}", e)
```
preston-evans98, line 113: nit: Could we expand the error message here a little. Maybe something like `"Error: The transaction was rejected by the 'enter_apply_batch' hook"`.

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -155,27 +151,21 @@ where
                         batch_workspace = batch_workspace.commit();
                     }
                     Err(e) => {
-                        // TODO: all the previous "successful" txs will be reverted, is that ok?
                         self.revert_and_slash(batch_workspace);
                         panic!("Demo app txs must succeed but failed with err: {}", e)
```
preston-evans98, line 155: Now that we have the option of returning an error, do we still want to panic here?

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -155,27 +151,21 @@ where
...
                     }
                 }
             } else {
                 // If the serialization is invalid, the sequencer is malicious. Slash them.
-                // TODO all the previous "successful" txs will be reverted, is that ok?
                 self.revert_and_slash(batch_workspace);
-                return Err(ConsensusSetUpdate::slashing(sequencer));
+                anyhow::bail!("Tx decoding error");
```
preston-evans98, line 161: nit: Let's add an explanation if we can.

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -155,27 +151,21 @@ where
...
         // TODO: calculate the amount based of gas and fees
-        if self
-            .tx_hooks
+        self.tx_hooks
             .exit_apply_batch(0, &mut batch_workspace)
-            .is_err()
-        {
-            // TODO handle error: https://github.com/Sovereign-Labs/sovereign/issues/174
-        }
+            .unwrap();
```
preston-evans98, line 168: Are we sure we want to `panic` on failure here? If so, let's add a comment explaining or use `expect`

preston-evans98 (commented):
Good change overall! Looks like there were a bunch of formatting changes from Lukasz's PR - we should take a note to get him set up with `format_on_save: true` in his editor

sdk/src/state_machine/core/run.rs
bkolad: Ok

sov-modules/sov-app-template/src/lib.rs
bkolad: Fixed

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -136,9 +132,9 @@ where
             // Run the stateful verification, possibly modifies the state.
             let verified_tx = match self.tx_hooks.pre_dispatch_tx_hook(tx, &mut batch_workspace) {
                 Ok(verified_tx) => verified_tx,
-                Err(_) => {
+                Err(e) => {
                     self.revert_and_slash(batch_workspace);
-                    return Err(ConsensusSetUpdate::slashing(sequencer));
+                    anyhow::bail!("Stateful verification error: {}", e);
```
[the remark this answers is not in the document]
bkolad: So the exact reason is inside `e`. Now it can be only bad nonce but in the future it might be not enough balance to cover fees etc (but the correct message will be in e )

sov-modules/sov-app-template/src/lib.rs
bkolad: Let's leave it for now, I want to write a specific test for this (and then I will remove the panic), but first we need to fix the other bug (#179)

sov-modules/sov-app-template/src/lib.rs
bkolad: I added `?` so the error will be handled like all the other error.

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -134,38 +137,39 @@ where
                 Ok(verified_tx) => verified_tx,
                 Err(e) => {
                     self.revert_and_slash(batch_workspace);
```
preston-evans98, line 139: This isn't a new change, but I'm not sure that we want to slash here. I think there are potentially non-malicious (and unpreventable) events where the same transaction is included twice in the same slot by different sequencers - and the second one will fail stateful verification

bkolad: Removed the slashing and added TODO to check one more time in the future.

preston-evans98 (approved):

PR #180, opened 2023-04-19
Implement module-level rpc macros

preston-evans98:
Derive a `jsonrpsee` implementation for the underlying module, just by adding a couple of annotations.

 Syntax is identical to `jsonrpsee`'s `#[rpc]` execept that:
 1. `#[rpc]` is renamed to `#[rpc_gen]` to avoid confusion with `jsonrpsee`'s `#[rpc]`
 2. `#[rpc_gen]` is applied to an `impl` block instead of a trait
 3. `#[method]` is renamed to with `#[rpc_method]` to avoid import confusion and clarify the purpose of the annotation

 ## Example
  ```rust,ignore
  struct MyModule {};

 #[rpc_gen(client, server, namespace ="myNamespace")]
 impl MyModule {
    #[rpc_method(name = "myMethod")]
     fn my_method(&self, param: u32) -> u32 {
          1
     }
 }
 ```

 This is exactly equivalent to hand-writing
 ```rust,ignore
 struct MyModule<C: Context> {
 ...
 };

 impl MyModule {
     fn my_method(&self, working_set: &mut WorkingSet<C::Storage>, param: u32) -> u32 {
         1
     }  
 }

 #[jsonrpsee::rpc(client, server, namespace ="myNamespace")]
 pub trait MyModuleRpc {
    #[jsonrpsee::method(name = "myMethod")]
    fn my_method(&self, param: u32) -> Result<u32, jsonrpsee::Error>;
    #[method(name = "health")]
    fn health() -> Result<(), jsonrpsee::Error> {
        Ok(())
    }
 }
 ```
 
 
This proc macro also generates an implementation trait intended to be used by a Runtime struct. This trait
is named `MyModuleRpcImpl`, which allows a Runtime to be converted into a functional RPC server 
by simply implementing the two required methods - `get_backing_impl(&self) -> MyModule` and `get_working_set(&self) -> ::sov_modules_api::WorkingSet<C>`
 
 ```rust,ignore
 pub trait MyModuleRpcImpl<C: sov_modules_api::Context> {
     fn get_backing_impl(&self) -> &TestStruct<C>;
     fn get_working_set(&self) -> ::sov_modules_api::WorkingSet<C>;
     fn my_method(&self, param: u32) -> u32 {
         Self::get_backing_impl(self).my_method(self, &mut Self::get_working_set(self), param)
     }
 }
 ```


sov-modules/sov-modules-macros/src/dispatch/common.rs
```diff
@@ -173,29 +176,28 @@ pub fn get_serialization_attrs(item: &syn::DeriveInput) -> Result<Vec<TokenStrea
     let mut has_deserialize = false;
     let mut has_other = false;
 
-    let attributes: Vec<String> = 
-        serialization_attrs.iter().map(|t| t.to_string()).collect();        
+    let attributes: Vec<String> = serialization_attrs.iter().map(|t| t.to_string()).collect();
 
-    for attr in &attributes {        
+    for attr in &attributes {
```
citizen-stig, line 181: I guess it is not critical for compile time, but we can check if all 3 booleans are true and then break.

sov-modules/sov-modules-macros/src/dispatch/derive_rpc.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub(crate) method_name: Ident,
+    pub(crate) method_signature: Signature,
+    pub(crate) idx_of_working_set_arg: Option<usize>,
+}
+
+
+impl RpcImplBlock {
+
+
+    /// Builds the trait `_RpcImpl` That will be implemented by the runtim
```
citizen-stig, line 84:
```suggestion
    /// Builds the trait `_RpcImpl` That will be implemented by the runtime
```

sov-modules/sov-modules-macros/src/dispatch/derive_rpc.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        #impl_rpc_trait_impl
+
+        #rpc_attribute
+        pub trait #intermediate_trait_name  #generics {
+
+            #(#intermediate_trait_items)*
+
+            #[method(name = "health")]
+            fn health(&self) -> ::jsonrpsee::core::RpcResult<()> {
```
citizen-stig, line 260: What do you think about checking connection to the database in health call?

sov-modules/sov-modules-macros/src/dispatch/derive_rpc.rs
```diff
@@ -0,0 +1,285 @@
...
+            }
+        }
+    }
+    None
+}
+
+fn intermediate_trait_name(ident: &Ident) -> Ident {
+    let mut ident_str = ident.to_string();
+    ident_str.push_str("Rpc");
+    format_ident!("{}", ident_str)
```
citizen-stig, line 25: I think I am missing something, but why is it not possible to do:

```
format_ident!({}Rpc", ident)
```
So there's less allocations

sov-modules/sov-modules-macros/src/dispatch/derive_rpc.rs
```diff
@@ -0,0 +1,285 @@
...
+}
+
+fn find_working_set_argument(sig: &Signature) -> Option<usize> {
+    for (idx, input) in sig.inputs.iter().enumerate(){
+        if let FnArg::Typed(PatType { ty, .. }) = input {
+            if let syn::Type::Reference(syn::TypeReference { elem, .. }) = *ty.clone() {
+                if let syn::Type::Path(syn::TypePath { path, .. }) = elem.as_ref() {
+                    if let Some(segment) = path.segments.last() {
+                        // TODO: enforce that the working set has exactly one angle bracketed argument
+                        if segment.ident == "WorkingSet" && !segment.arguments.is_empty() {
```
bkolad, line 55: If someone uses `WorkingSet as SomeOtherType` this will break, I don't have a good solution for this, we just need to remember to write a good compile time error message

sov-modules/sov-modules-macros/src/dispatch/derive_rpc.rs
preston-evans98: You don't need this at all! Copilot generated it for me :). Removed the function in a later commit

sov-modules/sov-modules-macros/src/dispatch/derive_rpc.rs
preston-evans98: Fixed!

sov-modules/sov-modules-macros/src/dispatch/derive_rpc.rs
preston-evans98: Since we use RocksDB, there isn't really the same notion of a connection - since it's embedded, you can't disconnect without having massive disk corruption on the machine running the RPC server

citizen-stig: Oh, that makes sense.

bkolad (approved):

sov-modules/sov-modules-macros/src/dispatch/common.rs
preston-evans98: Going to save this for future cleanup, since all that changed in this PR was formatting

PR #182, opened 2023-04-20
Fix for #179

### LINKED ISSUE #179: Bug in `AppTemplate::init_chain`
The `init_chain` method runs `genesis(..)` method for each module defined in `Runtime`. 
The `Bank::genesis` contains the following code:
```
...
if self.tokens.get(&token_address, working_set).is_some() {
     bail!("Token address already exists");
}
...
``` 
And it causes the following error: 
```
---- test::test_demo_values_in_db stdout ----
thread 'test::test_demo_values_in_db' panicked at 'Storage update must succeed: Missing state root node at version 0, probably pruned.', /Users/blaze/programming/sovereign-labs/sovereign/sov-modules/sov-app-template/src/lib.rs:93:14
```
After commenting out the above cod `init_chain` runs without problem.

bkolad:
close #179

### DIFF
--- a/sov-modules/sov-state/src/prover_storage.rs
+++ b/sov-modules/sov-state/src/prover_storage.rs
@@ -75,6 +75,21 @@ impl<S: StorageSpec> Storage for ProverStorage<S> {
         let read_logger = TreeReadLogger::with_db_and_witness(self.db.clone(), witness);
         let untracked_jmt = JellyfishMerkleTree::<_, S::Hasher>::new(&self.db);
 
+        // Handle empty untracked_jmt
+        if untracked_jmt
+            .get_root_hash_option(latest_version)?
+            .is_none()
+        {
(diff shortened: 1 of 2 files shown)


sov-modules/sov-state/src/prover_storage.rs
```diff
@@ -75,6 +75,21 @@ impl<S: StorageSpec> Storage for ProverStorage<S> {
         let read_logger = TreeReadLogger::with_db_and_witness(self.db.clone(), witness);
         let untracked_jmt = JellyfishMerkleTree::<_, S::Hasher>::new(&self.db);
 
+        // Handle empty untracked_jmt
+        if untracked_jmt
+            .get_root_hash_option(latest_version)?
+            .is_none()
```
preston-evans98, line 81: Looks good  - but maybe we want to enforce that the version is 0?

PR #183, opened 2023-04-20
Adding more tests for banking module

### LINKED ISSUE #163: sov-modules: Add more tests for the `Bank` module
We should add more tests for `Bank` module, especially for all the error cases (insufficient funds, etc..)

citizen-stig:
# Scope

* Existing test cases are moved changed to be 'integration' tests, so it only works with public methods, so we can catch misshaped API, for example non public `call` enum.
* Common test code has changed, for configurability. 

# Bugs found
* if sender equals to receiver it will mint coins for sender of the message. Fixed by doing early return.
* If prefix is not valid `utf-8` string, `Display` trait implementation would panic. This can happen as it is described in #185 . Fixed by hex encoding raw bytes if it is not `utf-8` string.  

# Open questions

* All tokens in current implementation are fixed supply, there's no way to mint new tokens after token is deployed. Is it intended?
* Burning amount does not changes total supply, because it transfers tokens to `0x0` address. Is it inteded behavoiur?

Resolves #163


sov-modules/sov-modules-impl/bank/tests/transfer_test.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    let salt = 10;
+    let token_name = "Token1".to_owned();
+    let initial_balance = 1000;
+    let token_address = create_token_address::<C>(&token_name, sender_address.as_ref(), salt);
+
+    assert_ne!(sender_address, receiver_address);
+
+    // Preparation
+    let query_user_balance =
```
bkolad, line 223: Let's refactor `query_user_balance & query_total_supply` as helpers. It looks like they are used in several places.

bkolad (approved):

sov-modules/sov-modules-impl/bank/tests/transfer_test.rs
citizen-stig: Although it does repeat few times, actual value of this function is to capture token address inside closure, so it can be skipped on follow calls. Extracting this into its own function won't bring that much benefit in my opinion.

PR #186, opened 2023-04-20
Expose bank_balanceOf RPC method

preston-evans98:

PR #187, opened 2023-04-20
Implement remaining methods on `LedgerRpcProvider` trait. Add `get_head` method

preston-evans98:
N.B. - this commit is based on top of #186

### DIFF
--- a/sov-modules/sov-modules-api/src/lib.rs
+++ b/sov-modules/sov-modules-api/src/lib.rs
@@ -42,6 +42,7 @@ impl AddressTrait for Address {}
 
 /// Default implementation of AddressTrait for the module system
 // TODO: decide if we want to feature gate the serde implementations
+// See https://github.com/Sovereign-Labs/sovereign/issues/175
 #[derive(
     borsh::BorshDeserialize,
     borsh::BorshSerialize,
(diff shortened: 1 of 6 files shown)


sov-modules/sov-modules-impl/bank/src/query.rs
```diff
@@ -23,7 +24,9 @@ pub struct TotalSupplyResponse {
     pub amount: Option<Amount>,
 }
 
+#[rpc_gen(client, server, namespace = "bank")]
 impl<C: sov_modules_api::Context> Bank<C> {
+    #[rpc_method(name = "balanceOf")]
```
bkolad, line 29: Is it possible to unit test the rpc call?

sov-modules/sov-modules-api/src/lib.rs
```diff
@@ -40,7 +41,17 @@ impl AsRef<[u8]> for Address {
 impl AddressTrait for Address {}
 
 /// Default implementation of AddressTrait for the module system
-#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq, Clone, Eq)]
+// TODO: decide if we want to feature gate the serde implementations
```
bkolad, line 44: lets add issue number to all this TODOS: https://github.com/Sovereign-Labs/sovereign/issues/175

bkolad (approved):

sov-modules/sov-modules-impl/bank/src/query.rs
preston-evans98: Yes, it should be. I'll add a test

preston-evans98: This actually proved to be a bit tricky because there's no way to discover the address of a created token (#189). Opened #190 instead

sov-modules/sov-modules-api/src/lib.rs
preston-evans98: Done!

PR #202, opened 2023-04-25
Simplify Da Layer interface

### LINKED ISSUE #170: Deduplicate DaService and DaTrait interfaces
Currently, the workflow for processing Da layer block data has two steps. First, data is fetched from the RPC via the `DaService` trait, which is generic over a Future (to enable async) and a `FilteredBlock`, which is corresponds to the Da layer block with the irrelevant data filtered out. In a typical workflow, the returned future is `await`ed immediately and the result is stored in the database. 

Second, the filtered block is retrived from the Database using the `get_relevant_txs_with_proof` from the `DaTrait`, which (in practice) is a relatively thin wrapper around a db lookup. This seems redundant. Since `get_relevant_txs_with_proof` is never called within the state machine (only `verify_relevant_tx_list` is), it seems like all of its work should be pushed into the DaService trait. This would leave us with two traits - DaService, which is responsible for fetching blocks via RPC *and* performing ETL on the returned data, and `DaTrait` which is part of the rollups provable execution and simply retrieves data from the host.

preston-evans98:
Simplify the interfaces to be implemented by DA Layers, and improve documentation. The following changes are made:
- Rename `DaLayerTrait` to `DaVerifier`
- Move all ETL from `DaLayerTrait` to `DaService`. As a side effect, this removes the need to store SlotData in the db before it can be processed. Resolves #170. 
- Simplify the SlotData trait, since slots no longer have to be stored in the db in their raw form. 
- Remove the generic used for handling raw slot data from LedgerDB
- Remove the (now unused) RollupSpec trait
- Create a new `da::Spec` trait which defines the types shared between `DaVerifier` and `DaService`

Finally, to minimize the headache from this breaking change, remove the long-unused `run.rs` file. Remove types from `core/types.rs` which were only used in `run.rs`.


sdk/src/state_machine/mod.rs
```diff
@@ -1,12 +1,11 @@
-mod env {
-    pub fn read_unchecked<T>() -> T {
-        todo!()
-    }
-}
+// mod env {
```
bkolad, line 1: let's remove the unused code?

bkolad (approved):

sdk/src/state_machine/mod.rs
preston-evans98: Good catch - I had meant to delete completely. Fixed!

Issue #208, opened 2023-04-26
Add `send_transaction` method DaService trait


citizen-stig:
Method signature:
 
 ```rust
 fn send_transaction(tx: Self::Transaction, sender: Self::Address)
 ```

citizen-stig:
Reopening until https://github.com/Sovereign-Labs/Jupiter/pull/22 is merged

PR #212, opened 2023-04-27
Adding send_transaction method to trait

xBalbinus:
Adds a `send_transaction` method according to the description in #208

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,12 +12,13 @@ env:
 jobs:
   check:
     runs-on: ubuntu-latest
+    name: ubuntu / check
     env:
       RUSTFLAGS: -D warnings
       SCCACHE_GHA_ENABLED: "true"
       RUSTC_WRAPPER: "sccache"
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Run sccache-cache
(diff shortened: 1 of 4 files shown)


sdk/src/node/services/da.rs
```diff
@@ -52,8 +52,8 @@ pub trait DaService {
         <Self::Spec as DaSpec>::CompletenessProof,
     );
 
-    // TODO: add a send_transaction method
-    // fn send_transaction(tx: Self::Transaction, sender: Self::Address)
+    /// A method to send a transaction. Takes in some arbitrary transaction payload and the address of the sender
+    fn send_transaction(&self, tx: <Self::Spec as DaSpec>::BlobTransaction, sender: <Self::Spec as DaSpec>::Address);
```
preston-evans98, line 56: Ah sorry, this comment is a bit misleading. The API for sending transactions is still somewhat unclear, because someone has to *sign* the transaction before it can be broadcast - which also means that we probably can't specify what address we want it to be signed with. 

My current feeling is that this method should probably just take a single `&[u8]` argument, and forward it to some external service which handles signing and broadcasting - but we might want to handle some more complicated orchestration instead (for example - maybe we'd add a separate method for "sign transaction" which invokes a wallet API, and then modify `send_transaction` to accept the pre-signed transaction as its argument).

PR #216, opened 2023-04-28
Use `From` instead of `TryFrom` for MockPublicKey

bkolad:


sov-modules/sov-modules-api/src/mocks.rs
```diff
@@ -26,21 +25,17 @@ impl MockPublicKey {
...
-    type Error = Infallible;
-
-    fn try_from(key: String) -> Result<Self, Self::Error> {
+impl From<String> for MockPublicKey {
+    fn from(key: String) -> Self {
         let key = key.as_bytes().to_vec();
-        Ok(Self { pub_key: key })
+        Self { pub_key: key }
     }
 }
```
preston-evans98, line 40: non-blocking: It seems like we should be able to deduplicate these by using `AsRef<str>`. Is that right?

preston-evans98 (approved):
Good change! Left one nit on de-duplicating the From, implementations but feel free to merge as is

sov-modules/sov-modules-api/src/mocks.rs
bkolad: Good point, done

PR #220, opened 2023-05-02
Feature gating sov-module-api and cleanup sov-app-template

citizen-stig:


Cargo.toml
```diff
@@ -40,8 +40,8 @@ sequencer = { path = "sov-modules/sov-modules-impl/sequencer" }
 bank = { path = "sov-modules/sov-modules-impl/bank" }
 schemadb = { path = "db/schemadb" }
 sovereign-db = { path = "db/sovereign-db" }
-sov-state = { path = "sov-modules/sov-state" }
-sov-modules-api = { path = "sov-modules/sov-modules-api" }
+sov-state = { path = "sov-modules/sov-state", default-features = false }
+sov-modules-api = { path = "sov-modules/sov-modules-api", default-features = false }
```
citizen-stig, line 44: This is breaking change, so I kept it to progress on risc0 integration

sov-modules/sov-app-template/Cargo.toml
```diff
@@ -8,7 +8,6 @@ edition = "2021"
 [dependencies]
 anyhow = { workspace = true }
 sovereign-sdk = { workspace = true }
-sov-state = { workspace = true, features = ["temp"] }
-sovereign-db = { workspace = true }
```
citizen-stig, line 12: Apparently it does not use it

sov-modules/sov-modules-api/src/mocks.rs
```diff
@@ -46,8 +49,10 @@ impl TryFrom<String> for MockPublicKey {
 
 impl PublicKey for MockPublicKey {
     fn to_address<A: AddressTrait>(&self) -> A {
-        let pub_key_hash = <MockContext as Spec>::Hasher::hash(&self.pub_key);
-        A::try_from(&pub_key_hash).expect("todo")
+        let mut hasher = sha2::Sha256::new();
```
citizen-stig, line 52: Replaced with concrete hasher, so it does not depend on MockContext as Spec, because it brings db dependency from storage type

preston-evans98: This isn't ideal - we want to be able to configure the hash function (for efficiency in a zkvm), and it will be surprising to future developers when changing the `Hash` in `Context` has no effect. Can we feature-gate the db dependency instead?

citizen-stig: Done!

PR #223, opened 2023-05-03
WIP: Cleanup for Alpha

preston-evans98:
This PR begins two important pieces of cleanup. 

## Add Initialization to Core Traits
Prior to this PR, none of the core traits from the SDK do not contained initializers. This made it impossible to write generic `start` code in the node implementation. This PR adds initialization for the `DaVerifier`, `DaService`, and `STF` traits.

## Allow Replacing Custom  STF in Demo Rollup with Demo APP

Prior to this PR, the demo-app was only usable for internal testing. This PR makes several changes required to use the demo-app as a library.

- Make DemoApp type public.
- Remove generic GenesisConfig from AppTemplate. Replace with the existing InitialState associated type (renamed from ChainParams).
- Create new DemoAppRunner implementing StateTransitionRunner.

### DIFF
--- a/sdk/src/node/services/da.rs
+++ b/sdk/src/node/services/da.rs
@@ -1,5 +1,7 @@
 use std::future::Future;
 
+use serde::de::DeserializeOwned;
+
 use crate::{
     da::DaSpec,
     serial::{Decode, Encode},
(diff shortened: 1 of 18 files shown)


citizen-stig (approved):

sdk/src/node/services/da.rs
```diff
@@ -11,6 +13,9 @@ use crate::{
 /// The DaService has two responsibilities - fetching data from the DA layer, transforming the
 /// data into a representation that can be efficiently verified in circuit.
 pub trait DaService {
+    /// A handle to the types used by the DA layer.
```
dubbelosix, line 16: wrong description

dubbelosix (approved):

PR #228, opened 2023-05-05
RPC: Use range queries to minimize db access in ledger RPC WIP

xBalbinus:
Hey folks, another PR here to follow up with the issue that @preston-evans98 brought up last time as a good suggestion for moving forward.

I've yet to implement the same for the other methods, because I'm currently stuck on some functions that the DB implements (or rather, doesn't). As soon as I'm sure that the implementation for slots is correct, expanding to the others will be very fast.

Do we know the way to do `db.get`, but `db.get_range` instead?

I appreciate the help in advance.

### DIFF
--- a/db/sovereign-db/src/ledger_db/rpc.rs
+++ b/db/sovereign-db/src/ledger_db/rpc.rs
@@ -1,3 +1,4 @@
+use anyhow::Result;
 use borsh::{BorshDeserialize, BorshSerialize};
 use serde::{Deserialize, Serialize};
 use sovereign_sdk::{
@@ -78,19 +79,51 @@ impl LedgerRpcProvider for LedgerDB {
     ) -> Result<Vec<Option<Self::SlotResponse>>, anyhow::Error> {
         // TODO: https://github.com/Sovereign-Labs/sovereign/issues/191 Sort the input
         //      and use an iterator instead of querying for each slot individually
-        let mut out = Vec::with_capacity(slot_ids.len());
-        for slot_id in slot_ids {
-            let slot_num = self.resolve_slot_identifier(slot_id)?;
-            out.push(match slot_num {
-                Some(num) => {
-                    if let Some(stored_slot) = self.db.get::<SlotByNumber>(&num)? {
-                        Some(self.populate_slot_response(num.into(), stored_slot, query_mode)?)
-                    } else {
-                        None
+
+        // First resolve the slot identifiers and store the slot numbers
+        let mut slot_nums: Vec<Option<SlotNumber>> = slot_ids.iter()
+            .map(|slot_id| self.resolve_slot_identifier(slot_id))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        // Then sort the slot numbers
+        slot_nums.sort_unstable();
+
+        // Then identify sequential ranges
+        let mut slot_ranges: Vec<(SlotNumber, SlotNumber)> = Vec::new();
+        let mut current_range: Option<(SlotNumber, SlotNumber)> = None;
+
+        for slot_num in slot_nums {
+            if let Some(num) = slot_num {
+                match current_range {
+                    Some((start, end)) if SlotNumber(Into::<u64>::into(end) + 1) == num => {
+                        current_range = Some((start, num));
+                    }
+                    _ => {
+                        if let Some(range) = current_range {
+                            slot_ranges.push(range);
+                        }
+                        current_range = Some((num, num));
                     }
                 }
-                None => None,
-            })
+            }
+        }
+
+        if let Some(range) = current_range {
+            slot_ranges.push(range);
+        }
+
+        // For each identified range, make a single query to fetch tx's
+        // From the corresponding blocks, then store into result vector.
+        let mut out = Vec::with_capacity(slot_ids.len());
+        for (start, end) in slot_ranges {
+            let stored_slots = self.db.get_range::<SlotByNumber>(start, end)?;
+            for num in start..=end {
+                if let Some(stored_slot) = stored_slots.get(&num) {
+                    out.push(Some(self.populate_slot_response(num.into(), stored_slot.clone(), query_mode)?));
+                } else {
+                    out.push(None);
+                }
+            }
         }
         Ok(out)
     }
@@ -102,6 +135,7 @@ impl LedgerRpcProvider for LedgerDB {
     ) -> Result<Vec<Option<Self::BatchResponse>>, anyhow::Error> {
         // TODO: https://github.com/Sovereign-Labs/sovereign/issues/191 Sort the input
         //      and use an iterator instead of querying for each slot individually
+        
         let mut out = Vec::with_capacity(batch_ids.len());
         for batch_id in batch_ids {
             let batch_num = self.resolve_batch_identifier(batch_id)?;


preston-evans98:
Thanks for the PR @xBalbinus! I would recommend using the [`get_x_range()`](https://github.com/Sovereign-Labs/sovereign/blob/fbf8b628ecf8ddcf3e6da631d3f22ab66280a703/db/sovereign-db/src/ledger_db/mod.rs#L123) functions on the outer `LedgerDB`. They provide a convenient wrapper around the iterator interface exposed by the underlying `SchemaDB `instance.

db/sovereign-db/src/ledger_db/rpc.rs
```diff
@@ -78,19 +79,51 @@ impl LedgerRpcProvider for LedgerDB {
...
-                    } else {
-                        None
+
+        // First resolve the slot identifiers and store the slot numbers
+        let mut slot_nums: Vec<Option<SlotNumber>> = slot_ids.iter()
+            .map(|slot_id| self.resolve_slot_identifier(slot_id))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        // Then sort the slot numbers
+        slot_nums.sort_unstable();
```
preston-evans98, line 89: Currently, this logic will return results in sorted order rather than the requested order. What if we keep track of the original indices and return the results in that order too. Something like this:

```rust
     let mut slot_nums: Vec<Option<SlotNumber>> = slot_ids.iter().enumerate().
            .map(|(idx, slot_id)| (idx, self.resolve_slot_identifier(slot_id)))
            .collect::<Result<Vec<_>, _>>()?;

     slot_nums.sort_unstable_by(|(_, slot_id_a), (_, slot_id_b)| slot_id_a.cmp(slot_id_b));

    // ...
    for (orig_idx, (start, end)) in slot_ranges {
            let stored_slots = self.db.get_range::<SlotByNumber>(start, end)?;
            for num in start..=end {
                if let Some(stored_slot) = stored_slots.get(&num) {
                    out[orig_idx] = Some(self.populate_slot_response(num.into(), stored_slot.clone(), query_mode)?);
                } else {
                    out[orig_idx]  = None;
                }
            }
        }

```

db/sovereign-db/src/ledger_db/rpc.rs
```diff
@@ -78,19 +79,51 @@ impl LedgerRpcProvider for LedgerDB {
...
+        }
+
+        // For each identified range, make a single query to fetch tx's
+        // From the corresponding blocks, then store into result vector.
+        let mut out = Vec::with_capacity(slot_ids.len());
+        for (start, end) in slot_ranges {
+            let stored_slots = self.db.get_range::<SlotByNumber>(start, end)?;
+            for num in start..=end {
+                if let Some(stored_slot) = stored_slots.get(&num) {
+                    out.push(Some(self.populate_slot_response(num.into(), stored_slot.clone(), query_mode)?));
```
preston-evans98, line 122: Currently, we're doing one range-query per *slot* to populate the sub-fields inside of `populate_slot_response`. But with the new logic, we should be able to reduce that to one range-query per *range*. So I think we'll also want to add a batched version of `populate_slot_response` too. 

That doesn't have to happen in this PR though - just flagging it for later!

db/sovereign-db/src/ledger_db/rpc.rs
```diff
@@ -78,19 +79,51 @@ impl LedgerRpcProvider for LedgerDB {
...
+
+        if let Some(range) = current_range {
+            slot_ranges.push(range);
+        }
+
+        // For each identified range, make a single query to fetch tx's
+        // From the corresponding blocks, then store into result vector.
+        let mut out = Vec::with_capacity(slot_ids.len());
+        for (start, end) in slot_ranges {
+            let stored_slots = self.db.get_range::<SlotByNumber>(start, end)?;
```
preston-evans98, line 119: You can use `self.get_slot_range` here: https://github.com/Sovereign-Labs/sovereign/blob/fbf8b628ecf8ddcf3e6da631d3f22ab66280a703/db/sovereign-db/src/ledger_db/mod.rs#L123

xBalbinus: Got you, thanks! Fixed.

db/sovereign-db/src/ledger_db/rpc.rs
```diff
@@ -78,19 +79,67 @@ impl LedgerRpcProvider for LedgerDB {
...
+        // From the corresponding blocks, then store into result vector.
+        let mut out = Vec::with_capacity(slot_ids.len());
+        for (start, end) in slot_ranges {
+            let range = std::ops::Range {
+                start: start.into(),
+                end: end.into(),
+            };
+            let stored_slots = self._get_slot_range(&range)?;
+            for (stored_slot, slot_num) in stored_slots.iter().zip(range.iter()) {
+                out.push(Some(self.populate_slot_response(
```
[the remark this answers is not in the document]
xBalbinus: Makes a lot of sense - though, I think I'm still struggling on one last thing regarding the out[original_idx] = Some(...) - the iterator uses references to each member of the `stored_slots` array, but StoredSlot doesn't implement the `Copy` trait; would love if you have any recommendations for the fix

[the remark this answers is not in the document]
xBalbinus: Thank you for your guidance! I've fixed the common before that

db/sovereign-db/src/ledger_db/rpc.rs
```diff
@@ -78,19 +84,67 @@ impl LedgerRpcProvider for LedgerDB {
...
+        }
+
+        // For each identified range, make a single query to fetch tx's
+        // From the corresponding blocks, then store into result vector.
+        let mut out = Vec::with_capacity(slot_ids.len());
+        for slot_range in slot_ranges {
+            let stored_slots = self.get_slot_range(&slot_range.inner)?;
+
+            for ((queried_slot, original_idx), slot_num) in stored_slots
+                .iter()
```
preston-evans98, line 141: The build is failing because of this borrow. Looks like it needs to change to `into_iter()`. 

On a related note - looks like CI has failed a couple of times in a row. Is your repo out of sync with the upstream? If not, you might want to try using the Makefile we've included! From the root directory (`sovereign/`), just run `make check` - it should catch pretty much everything that might cause CI to fail.

preston-evans98 (changes requested):

db/sovereign-db/src/ledger_db/rpc.rs
```diff
@@ -78,19 +84,67 @@ impl LedgerRpcProvider for LedgerDB {
...
+                .zip(slot_range.original_indices)
+                .zip(slot_range.inner.start.into()..slot_range.inner.end.into())
+            {
+                out[original_idx] =
```
preston-evans98, line 145: This looks great, thanks!

db/sovereign-db/src/ledger_db/rpc.rs
xBalbinus: Was able to run this after rebasing - I made a slight error in forgetting one of the steps that made things fail, sorry about that. Most of the checks are passing now.

preston-evans98 (approved):
This looks great! The last blocker for merging is adding test coverage on this method. @xBalbinus do you have any interest in handling that? If not, lmk and we'll have someone from the team pick it up

xBalbinus:
Hey @preston-evans98 ! Sure thing, happy to get that done as well - I also should expand this similar change to the other queries (e.g. get_batches) too right?

Also apologies for the slowness in my reply - feel free to let me know if I'm blocking.

preston-evans98:
> Sure thing, happy to get that done as well - I also should expand this similar change to the other queries (e.g. get_batches) too right?

Both of those would be amazing, thanks! 

> Also apologies for the slowness in my reply - feel free to let me know if I'm blocking.

Not at all, we're super grateful for the contributions. We're pretty focused on other areas atm, so we're happy to have you take your time.

preston-evans98:
Hey @xBalbinus, we'd love to get this over the line - we're happy to allocate some points in the next week or two, but if you'd prefer to finish it yourself let us know!

PR #229, opened 2023-05-05
Prover demo

citizen-stig:


preston-evans98:
@bkolad, We made some pretty significant changes to the SDK while you were gone. This PR is big, and will probably need a few iterations - let's plan hop on a call and go through  it all as a team on Monday. Before then, here's a summary of what happened to help get you up to speed.

Broadly speaking, these changes fall into three camps - updates to feature-gating / dependencies (necessary to allow building under Risc0), API changes, and functional changes. The API changes are made with the goal of making the state transition function a black box to the full node implementation. 

There are three major API changes, which I recommend you review separately. 
- [Combine the `apply_batch`/`apply_proof` API](https://github.com/Sovereign-Labs/sovereign/pull/229/commits/0d6bb5d600771753f627c9fd7ee85211e64e4dbf). Previously, the apply_batch and apply_proof methods required their arguments to be typed - which meant that the full node implementation needed to run a deserializer. This gave the full node discretion over what should happen when the serialization was invalid, which we don't want - the STF should decide for itself what to do in that case. This commit updates the API to work on serialized inputs - which meant that `apply_batch`/`apply_proof` now have identical signatures - so the commit merged them into a single `apply_blob` method.
- [Have `apply_blob` return a receipt](https://github.com/Sovereign-Labs/sovereign/pull/229/commits/0d6bb5d600771753f627c9fd7ee85211e64e4dbf). Rather than returning a set of events, `apply_blob` now returns a structured receipt - which allows it to communicate to the full node exactly what it wants to have stored and served via RPC in a generic way.
- [Pass a `witness` into `begin_slot()` and return it from `end_slot`.](https://github.com/Sovereign-Labs/sovereign/pull/229/commits/0d8e50a6a49f63428ca627a92546a94ac3fe376f) This is necessary to support zk verification.

There is also one significant functional change. `RevertableDelta` has been simplified so that all writes are immediately applied to the underlying `Delta`, only writes are cached. This change spans the four commits from "[Simplify revertable delta. Bug fixes in apply_blob](https://github.com/Sovereign-Labs/sovereign/pull/229/commits/9454cec9448982016863b6ad3d810a097d9305d6)" to "[Cleanup. Introduce OrderedReadsAndWrites type.](https://github.com/Sovereign-Labs/sovereign/pull/229/commits/93bc397a51ed14c1953b570421f299c9a7f76294)". This change was introduced to fix a bug which caused the prover and the verifier to disagree about what the contents of the witness should be. At this point, I'm not certain of the exact cause of that bug because the caching logic was too complex to debug properly. The new logic fixes the bug and is easier to reason about.

bkolad (approved):

sdk/src/state_machine/stf.rs
```diff
@@ -56,18 +83,11 @@ pub trait StateTransitionFunction {
     fn begin_slot(&mut self);
 
     /// Apply a batch of transactions to the rollup, slashing the sequencer who proposed the batch on failure
```
bkolad, line 91: Let's change `batch` to `blob` in the doc.

sdk/src/state_machine/stf.rs
```diff
@@ -56,18 +83,11 @@ pub trait StateTransitionFunction {
...
-    fn apply_batch(
+    fn apply_blob(
         &mut self,
-        batch: Self::Batch,
-        sequencer: &[u8],
+        blob: impl BlobTransactionTrait,
```
bkolad, line 94: Why do we pass here `impl BlobTransactionTrait` why the other types are defined as `StateTransitionFunction`  type parameters?

sov-modules/sov-state/src/prover_storage.rs
```diff
@@ -59,6 +59,10 @@ impl<S: StorageSpec> Storage for ProverStorage<S> {
     type Witness = S::Witness;
     type RuntimeConfig = &'static str;
 
+    fn with_config(config: Self::RuntimeConfig) -> Result<Self, anyhow::Error> {
```
bkolad, line 62: Not sure when was the method introduced, but do we need it? It looks like it is not used anywhere.

demo-app/src/data_generation/mod.rs
```diff
@@ -31,6 +31,8 @@ pub fn simulate_da_with_bad_sig() -> Vec<RawTx> {
     election.create_raw_txs()
 }
 
+// TODO: Remove once we fix test with bad nonce
```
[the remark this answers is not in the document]
citizen-stig: https://github.com/Sovereign-Labs/sovereign/issues/235

sdk/src/state_machine/stf.rs
bkolad: Let's add a comment why we use `impl BlobTransactionTrait`

sov-modules/sov-app-template/src/lib.rs
```diff
@@ -71,33 +96,61 @@ where
...
+        let sequencer = sequencer.as_ref();
 
         if let Err(e) = self
             .tx_hooks
             .enter_apply_batch(sequencer, &mut batch_workspace)
         {
-            anyhow::bail!(
-                "Error: The transaction was rejected by the 'enter_apply_batch' hook. {}",
+            error!(
+                "Error: The transaction was rejected by the 'enter_apply_batch' hook. Skipping batch without slashing the sequencer {}",
```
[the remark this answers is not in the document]
preston-evans98: Fixed by @citizen-stig

demo-app/src/tests.rs
```diff
@@ -12,21 +13,54 @@ mod test {
...
-            demo.begin_slot();
+            demo.begin_slot(Default::default());
 
             let txs = simulate_da();
 
-            demo.apply_batch(Batch { txs }, &SEQUENCER_DA_ADDRESS, None)
-                .expect("Batch is valid");
-
-            demo.end_slot();
+            if let SequencerOutcome::Rewarded = demo
```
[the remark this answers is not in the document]
preston-evans98: Done!

sdk/src/state_machine/stf.rs
preston-evans98: Fixed by @citizen-stig

sdk/src/state_machine/stf.rs
preston-evans98: Done!

PR #254, opened 2023-05-09
Remove mock prefix in `MockContext` 

### LINKED ISSUE #231: Move from `Mock(ZK)Context` to ``(ZK)Context``
After choosing a suitable signature scheme we should be able to replace `Mock(ZK)Context` with `(ZK)Context` that will be used in production

bkolad:
1. Renames `Mock[PublicKey, Signature, Context]` to `Default[PublicKey, Signature, Context]`
2. Removes `mock` feature from `sov-modules-api
3. Adds `From[u8;32]` bound to `AddressTrait`

close #231

### DIFF
--- a/sov-modules/sov-modules-api/src/default_context.rs
+++ b/sov-modules/sov-modules-api/src/default_context.rs
@@ -2,9 +2,11 @@ use crate::default_signature::{DefaultPublicKey, DefaultSignature};
 use crate::{Address, AddressTrait, Context, PublicKey, Spec};
 
 use jmt::SimpleHasher;
+
 #[cfg(feature = "native")]
 use serde::{Deserialize, Serialize};
 use sov_state::DefaultStorageSpec;
+
 #[cfg(feature = "native")]
 use sov_state::ProverStorage;
 use sov_state::ZkStorage;


sov-modules/sov-modules-api/src/default_context.rs
```diff
@@ -0,0 +1,71 @@
...
+#[cfg(feature = "native")]
+use serde::{Deserialize, Serialize};
+use sov_state::DefaultStorageSpec;
+#[cfg(feature = "native")]
+use sov_state::ProverStorage;
+use sov_state::ZkStorage;
+use sovereign_sdk::core::types::ArrayWitness;
+
+// TODO: consider feature gating the serde implementations, since they are only needed for RPC
+// https://github.com/Sovereign-Labs/sovereign/issues/175
```
citizen-stig, line 14: I believe this comment is not relevant here, because whole `DefaultContext` is not available in zk (non-native) feature

citizen-stig (approved):

sov-modules/sov-modules-api/Cargo.toml
```diff
@@ -16,6 +16,9 @@ sha2 = { workspace = true }
 bech32 = { workspace = true }
 derive_more = { workspace = true }
 
+ed25519-dalek = { version = "1.0.1" }
+rand = "0.7"
```
preston-evans98, line 20: We should only need randomness for key generation, right? If so, let's feature gate that behind the "native" flag.

If we need randomness for signature *verification*, then we'll probably need to use a different signature scheme - since the host can provide malicious "randomness"

sov-modules/sov-modules-api/src/default_signature.rs
```diff
@@ -1,50 +1,92 @@
...
+use rand::{CryptoRng, RngCore};
+
+pub struct DefaultPrivateKey {
+    key_pair: Keypair,
 }
 
-impl DefaultPublicKey {
-    pub fn new(pub_key: Vec<u8>) -> Self {
-        Self { pub_key }
+impl DefaultPrivateKey {
```
preston-evans98, line 13: We won't ever sign in zk, so let's feature gate this.

preston-evans98 (changes requested):

sov-modules/sov-modules-api/Cargo.toml
bkolad: We decided on daily that the signature scheme will be implemented in the following PR.

sov-modules/sov-modules-api/src/default_signature.rs
bkolad: We decided on daily that the signature scheme will be implemented in the following PR.

sov-modules/sov-modules-api/src/default_signature.rs
```diff
@@ -0,0 +1,50 @@
...
+        let key = key.as_ref().as_bytes().to_vec();
+        Self { pub_key: key }
+    }
+}
+
+#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, PartialEq, Eq, Debug, Clone, Default)]
+pub struct DefaultSignature {
+    pub msg_sig: Vec<u8>,
+    pub should_fail: bool,
+}
```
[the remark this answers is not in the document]
bkolad: We decided on daily that the signature scheme will be implemented in the following PR.

preston-evans98 (approved):

PR #256, opened 2023-05-09
Feature gate serde derive in `sov-modules-api`

### LINKED ISSUE #175: Hide `bech32, derive_more & serde` behind `native` compilation flag.
We added bunch of crates to `sov-module-api`. The crates are used only in the `native` context, when the `AppTemplate` runs in `zk` context we pay for the code size and we don't need these libraries (we don't answear queries) so we should exclude them.

bkolad:
close #175

### DIFF
--- a/sov-modules/sov-modules-api/src/lib.rs
+++ b/sov-modules/sov-modules-api/src/lib.rs
@@ -25,8 +25,8 @@ use sovereign_sdk::{
     core::traits::Witness,
     serial::{Decode, Encode},
 };
-use std::fmt;
 
+use core::fmt::{self, Debug, Display};
 use thiserror::Error;
 
 impl AsRef<[u8]> for Address {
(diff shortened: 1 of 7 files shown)


sov-modules/sov-modules-api/src/lib.rs
```diff
@@ -129,7 +115,7 @@ pub trait Spec {
         + Eq
         + TryFrom<&'static str>
         + Clone
-        + Debug
+        + fmt::Debug
```
citizen-stig, line 118: Why?

citizen-stig (approved):

sov-modules/sov-modules-api/src/lib.rs
bkolad: It saved one `#[cfg(feature = "native")]` in the imports as `fmt::Display` is use only with `native` feature. So I can use `use std::fmt` for both `native` and `not(native)`.

bkolad: fixed

PR #265, opened 2023-05-10
Add request limits to ledger rpc

### LINKED ISSUE #230: Limit size of RPC range requests
Currently, the ledger RPC allows users to request ranges of items (slots, batches, etc.), but it doesn't enforce any limits on the size of the range. This is a trivial DOS vector. Set reasonable default limits.

preston-evans98:
Closes #230

### DIFF
--- a/db/sovereign-db/src/ledger_db/rpc.rs
+++ b/db/sovereign-db/src/ledger_db/rpc.rs
@@ -18,6 +18,15 @@ use crate::schema::{
     },
 };
 
+/// The maximum number of slots that can be requested in a single RPC range query
+const MAX_SLOTS_PER_REQUEST: u64 = 10;
+/// The maximum number of batches that can be requested in a single RPC range query
+const MAX_BATCHES_PER_REQUEST: u64 = 20;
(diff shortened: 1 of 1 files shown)


db/sovereign-db/src/ledger_db/rpc.rs
```diff
@@ -76,6 +85,11 @@ impl LedgerRpcProvider for LedgerDB {
         slot_ids: &[sovereign_sdk::rpc::SlotIdentifier],
         query_mode: QueryMode,
     ) -> Result<Vec<Option<Self::SlotResponse>>, anyhow::Error> {
+        anyhow::ensure!(
+            slot_ids.len() <= MAX_SLOTS_PER_REQUEST as usize,
+            "requested too many slots. Max: {}",
```
citizen-stig, line 90: This is minor, but I believe it is nice to have actual length in the error message:

```suggestion
            "requested too many slots: {} . Max: {}", slot_ids.len(),
```

preston-evans98: Updated!

citizen-stig (approved):

PR #270, opened 2023-05-11
Add module template

### LINKED ISSUE #169: Create module template
# Background

Currently, there's no guidance for implementers who want to add a new module. A simple template would be very helpful for preventing issues such as accidentally making the *query* module private (which prevents anyone from querying the module!), and would reduce our development time as we add modules.

preston-evans98:
@bkolad I wanted the module template to start working on the proof-logic refactor, so I created a very simple on based on the value-setter. Please feel free to take over and make any changes that would be helpful!

Fixes #169


bkolad (approved):
LGTM, we can merge and we can add more changes via PRs.

PR #284, opened 2023-05-15
Implement risc0 adapter

preston-evans98:


risc0-adapter/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "risc0-adapter"
```
bkolad, line 2: Doe is make sense to create `adapters/risc0` , this way there will be a natural place to add other adapters?

risc0-adapter/src/host.rs
```diff
@@ -0,0 +1,64 @@
...
+impl<'prover> Zkvm for Risc0Host<'prover> {
+    type CodeCommitment = Risc0MethodId;
+
+    type Error = anyhow::Error;
+
+    fn verify<'a>(
+        serialized_proof: &'a [u8],
+        code_commitment: &Self::CodeCommitment,
+    ) -> Result<&'a [u8], Self::Error> {
+        // let journal:  = bincode::deserialize(serialized_proof)?;
```
citizen-stig, line 46: Let's remove commented code.

risc0-adapter/src/host.rs
```diff
@@ -0,0 +1,64 @@
...
+            ),
+            &code_commitment.0,
+            &receipt.seal,
+            receipt.journal,
+        )?;
+        Ok(receipt.journal)
+    }
+}
+
+#[derive(Serialize, Deserialize)]
```
citizen-stig, line 60: According to #285 we prefere fully qualified serde derive, so you don't need to feature gate import.

risc0-adapter/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+pub mod guest;
+#[cfg(feature = "native")]
+pub mod host;
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Risc0MethodId([u32; 8]);
+
+impl Matches<Self> for Risc0MethodId {
+    fn matches(&self, other: &Self) -> bool {
+        self.0 == other.0
```
citizen-stig, line 14: Why not implement automatic implementation for `PartialEq` types?

```rust
impl<T: PartialEq> Matches<T> for T {
    fn matches(&self, other: &T) -> bool {
        self == other
    }
}
```

risc0-adapter/Cargo.toml
preston-evans98: Good idea. Moved!

risc0-adapter/src/host.rs
preston-evans98: Oops, good catch. Removed!

risc0-adapter/src/host.rs
preston-evans98: This package can't feature gate the import, so we don't need to qualify the paths to make it easier. I've qualified it just to make it uniform with the other packages though

risc0-adapter/src/lib.rs
preston-evans98: `Matches` is a looser concept than `PartialEq` - so a type where `a != b`  might still return `true` for `a.matches(b)`.

bkolad (approved):

PR #290, opened 2023-05-16
Update ledger_db to store and serve generic BatchReceipts and TransactionReceipts

preston-evans98:


sov-modules/sov-modules-api/src/default_signature.rs
```diff
@@ -37,6 +41,10 @@ pub mod private_key {
                 pub_key: self.key_pair.public,
             }
         }
+
+        pub fn default_address<H: SimpleHasher>(&self) -> Address {
+            H::hash(self.pub_key().pub_key.as_bytes()).into()
```
bkolad, line 46: We have `self.pub_key().to_address()`

bkolad (approved):

sov-modules/sov-modules-api/src/default_signature.rs
preston-evans98: Nice, updating. Thanks!

sdk/src/node/rpc/mod.rs
```diff
@@ -56,83 +58,110 @@ pub enum QueryMode {
...
-    fn get_head(&self) -> Result<Option<Self::SlotResponse>, anyhow::Error>;
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum ItemOrHash<T> {
+    Hash([u8; 32]),
+    Full(T),
+}
 
-    fn get_slots(
+pub trait LedgerRpcProvider {
```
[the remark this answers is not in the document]
preston-evans98: This is a good question. I think (surprisingly), that adding the generics to the trait actually makes it more difficult to use! 

This is because the implementer of this trait (LedgerDB) doesn't have generics of its own - so it has to have a blanket impl for all choices of `B` and `T`. Which means that every time you want to use the trait, you still have to provide the generics, and you have to do it using the ugly fully-qualified syntax `< LedgerRpcProvider<MyBatchType, MyTxType>>::get_head(ledger_db)`.

PR #292, opened 2023-05-17
rpc impl traits

dubbelosix:
current model for generating RPC impl traits is

rpc_gen -> annotates the module struct and generates <module_struct>RpcImpl trait along with the blanket implementation of <module_struct>RpcServer which provides jsonrpsee for any type that implements the  __RpcImpl trait (as well as send + sync)

expose_rpc(module1<context>,module2<context>) -> implements the <module_struct>RpcImpl  trait for the AppRunner (a type that implements `StateTransitionRunner`.

an example would be

```
#[derive(ModuleInfo)]
pub struct TestStruct<C: Context> {
        #[address]
        pub(crate) address: C::Address,
}

#[rpc_gen(client, server, namespace="test")]
impl<C: Context> TestStruct<C> {
    #[rpc_method(name = "firstMethod")]
    pub fn first_method(&self, _working_set: &mut WorkingSet<C::Storage>) -> u32 {
        11
    }
}
```
the above macros generate
```
    pub trait TestStructRpcImpl<C: Context> {
        fn get_working_set(&self) -> WorkingSet<C::Storage>;

        fn first_method(&self) -> u32 {
            <TestStruct<C> as ::sov_modules_api::ModuleInfo>::new()
                .first_method(&mut Self::get_working_set(self))
        }
    }
```
this is the key part of the code that ties in the AppRunner / higher layers to the module. since modules are stateless, we can create a new module directly and call the necessary method on it. 
```
 <TestStruct<C> as ::sov_modules_api::ModuleInfo>::new()
                .first_method(&mut Self::get_working_set(self))
```

expose_rpc is used with parameters specifying which concrete modules need the RpcImpl trait implementations.

```
#[expose_rpc((TestStruct<DefaultContext>,))]
impl<Vm: Zkvm> StateTransitionRunner<ProverConfig, Vm> for TestAppRunner<DefaultContext, Vm> {
```

The core logic in this macro is to provide the working set.


dubbelosix:
The logic has been modified to include a new struct
`::sov_modules_api::RpcStorage`
This is the struct that get the module_nameRpcImpl trait implementation and thus the blanket RPC server implementation

PR #293, opened 2023-05-17
Read runtime configuration from files

citizen-stig:


demo-app/src/config.rs
```diff
@@ -0,0 +1,90 @@
+use serde::de::DeserializeOwned;
+pub use sov_state::config::Config as StorageConfig;
+use std::fs::File;
+use std::io::Read;
+use std::path::Path;
+
+pub fn from_toml_path<P: AsRef<Path>, R: DeserializeOwned>(path: P) -> anyhow::Result<R> {
```
preston-evans98, line 7: Non-blocking, nit: I think it's slightly easier for callers if we use the syntax `impl trait` syntax:

`pub fn from_toml_path<R: DeserializeOwned>(path: impl AsRef<Path>) -> anyhow::Result<R>`

This way, there's only one generic for them to specify, so they can say `from_toml_path::<MyConfig>("/path");` instead of `from_toml_path::<_, MyConfig>("/path");`

preston-evans98:
This change looks fantastic! It would be awesome to provide the `from_toml_path` as a library method (maybe via a trait?) so that we can use it with other apps as well.

PR #296, opened 2023-05-18
Calculate batch and transaction hashes in sov-app-template

### LINKED ISSUE #250: Calculate batch and transaction hashes in sov-app-template::apply_blob

bkolad:
close #250

### DIFF
--- a/module-system/sov-app-template/src/lib.rs
+++ b/module-system/sov-app-template/src/lib.rs
@@ -23,6 +23,7 @@ use sovereign_core::{
     traits::BatchTrait,
 };
 use std::io::Read;
+
 pub struct AppTemplate<C: Context, V, RT, H, Vm> {
     pub current_storage: C::Storage,
     pub runtime: RT,
@@ -216,6 +217,7 @@ struct BatchDataAndHash {
     hash: [u8; 32],
     data: Vec<u8>,
 }
+
 impl BatchDataAndHash {
     fn new<C: Context>(batch: impl Buf) -> BatchDataAndHash {
         let mut reader = batch.reader();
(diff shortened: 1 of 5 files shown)


module-system/sov-app-template/src/lib.rs
```diff
@@ -202,13 +206,33 @@ where
...
+struct BatchDataAndHash {
+    hash: [u8; 32],
+    data: Vec<u8>,
+}
+
+impl BatchDataAndHash {
+    fn new<C: Context>(batch: impl Buf) -> BatchDataAndHash {
+        let mut reader = batch.reader();
+        let mut batch_data = Vec::new();
+        reader
```
bkolad, line 225: The only way I found to calculate the batch hash was by converting it to a slice of bytes.

module-system/sov-app-template/src/lib.rs
```diff
@@ -120,8 +122,10 @@ where
         let mut tx_receipts = Vec::with_capacity(txs.len());
 
         // Process transactions in a loop, commit changes after every step of the loop.
-        for tx in txs {
+        for (tx, raw_tx_hash) in txs {
             batch_workspace = batch_workspace.to_revertable();
+
+            //let tx_hash = tx.hash();
```
preston-evans98, line 128: nit: let's remove the commented code

preston-evans98 (approved):
This looks great overall. Since all that we need these hashes for is indexing receipts (which is currently not part of the rollup state), we can skip this hashing in the zk context. No need to do that in this PR, but let's open an issue so we don't forget

module-system/sov-app-template/src/lib.rs
bkolad: done

PR #297, opened 2023-05-18
WIP: Starting to add demo nft module

citizen-stig:


examples/demo-nft-module/README.md
```diff
@@ -0,0 +1,374 @@
...
+    pub(crate) owners: sov_state::StateMap<u64, C::Address>,
+}
+```
+
+Where it has:
+
+1. Address. Each module has to have address. TBD: Why, and how it is defined and used.
+2. Several state attributes. In this case it is address of admin, and map of token id to owner address. Token id is u64 for simplicity.
+
+TBD: What is context, what does it hold?
```
citizen-stig, line 43: This question

preston-evans98: Context is a trait which allows the runtime to pass verified data to modules during execution. Currently the only required method in Context `sender()` which returns the address of the person who initiated the transaction (the signer).

Context also inherits the `Spec` trait, which defines the concrete types used by the rollup for `Hash`ing, persistant data `Storage`, digital `Signature`s, and `Addresses`. The `Spec` trait allows rollups to easily tailor themselves to different Zkvms. By being generic over a `Spec`, a rollup can ensure that any potentially SNARK-unfriendly cryptography can be easily swapped out.

examples/demo-nft-module/README.md
```diff
@@ -0,0 +1,374 @@
...
+    pub(crate) admin: sov_state::StateValue<C::Address>,
+
+    #[state]
+    pub(crate) owners: sov_state::StateMap<u64, C::Address>,
+}
+```
+
+Where it has:
+
+1. Address. Each module has to have address. TBD: Why, and how it is defined and used.
```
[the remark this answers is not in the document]
preston-evans98: Having addresses allows modules to interact with each other (in the same way that smart contracts can interact with each other in Ethereum). For example, you can use the `bank.transfer` function to transfer tokens to a module.

IIRC, the address is calculated deterministically based on the module name. CC @bkolad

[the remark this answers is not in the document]
bkolad: A module address is similar to a `smart contract` address in Ethereum. So, for example, a module can receive money. The address is calculated from a hash of the module prefix. This means that:
- The module address is unique 
- No one knows private key that produces this address

examples/demo-nft-module/README.md
```diff
@@ -0,0 +1,436 @@
...
+1. Add new dependencies: serde, borsh, sov-state
+2. Define 'native' feature flag to separate logic that is not needed in ZK context
+3. Define Call and Query messages
+4. Define Config
+
+## Preparation
+
+
+
+1. Define `native` feature in `Cargo.toml`:
```
bkolad, line 57: Let's specify what `native` means: the logic runs outside of zkvm.

bkolad: Ok it is explained below, closing the comment:)

citizen-stig:
I will be dumping more questions in comments, @bkolad @preston-evans98 @dubbelosix 

Can module be upgraded? How upgrade logic is handled?

bkolad (approved):

bkolad:
> I will be dumping more questions in comments, @bkolad @preston-evans98 @dubbelosix
> 
> Can module be upgraded? How upgrade logic is handled?

We can't change the logic of a single module without affecting the whole rollup, but we should be able to upgrade the whole rollup (but this will change the proving and and verification keys as well). @preston-evans98 any thoughts

preston-evans98:
> > I will be dumping more questions in comments, @bkolad @preston-evans98 @dubbelosix
> > Can module be upgraded? How upgrade logic is handled?
> 
> We can't change the logic of a single module without affecting the whole rollup, but we should be able to upgrade the whole rollup (but this will change the proving and and verification keys as well). @preston-evans98 any thoughts

Yep, @bkolad is spot-on here

examples/demo-nft-module/README.md
```diff
@@ -0,0 +1,481 @@
...
+```toml
+[dependencies]
+anyhow = { anyhow = "1.0.62" }
+sov-modules-api = { git = "https://github.com/Sovereign-Labs/sovereign.git", branch = "main", default-features = false }
+sov-modules-macros = { git = "https://github.com/Sovereign-Labs/sovereign.git", branch = "main" }
+```
+
+## Module definition
+
+NFT module is defined as following:
```
preston-evans98, line 19: This feels a little surprising - we haven't told readers _that_ we're building an NFT module yet, but we jump right into telling them how to do it. 

We also haven't explained where this code should live, or what the directory structure should look like. Maybe we should start by telling them that we'll follow the structure of the module template (with a link to the template that they can clone), and then we can add a note saying that this code replaces the `ExampleStruct` in `lib.rs`

examples/demo-nft-module/README.md
```diff
@@ -0,0 +1,481 @@
...
+}
+```
+
+Where it has:
+
+1. Address. Each module has to have an address. A module address is similar to a smart contract address in Ethereum. For example, a module can receive tokens. The address is calculated from a hash of the module prefix. This means that:
+    - The module address is unique
+    - No one knows the private key that produces this address
+
+2. Several state attributes. In this case it is an address of admin, and map of token id to owner address. Token id is u64 for simplicity.
```
preston-evans98, line 44: I think we need to give users some context about what `state` attributes are. It doesn't have to happen in this exact part of the tutorial, but we should at least let them know that we'll explain more about state values later. Broadly, things we want users to know about `state` values are that they are stored in the JMT, they can be read and written to, that the state values aren't part of a module (so the module can be cloned cheaply), that they're populated in a particular `WorkingSet`, etc.)

examples/demo-nft-module/README.md
```diff
@@ -0,0 +1,481 @@
...
+
```
preston-evans98, line 45: It would be good to mention modules can also use `#[module]` to store a reference to another module, even though we don't use that feature in this example.

examples/demo-nft-module/README.md
```diff
@@ -0,0 +1,481 @@
...
+sov-modules-api = { git = "https://github.com/Sovereign-Labs/sovereign.git", branch = "main", default-features = false }
+sov-modules-macros = { git = "https://github.com/Sovereign-Labs/sovereign.git", branch = "main" }
+sov-state = { git = "https://github.com/Sovereign-Labs/sovereign.git", branch = "main", default-features = false }
+
+
+[features]
+default = ["native"]
+serde = ["dep:serde", "dep:serde_json"]
+native = ["serde", "sov-state/native", "sov-modules-api/native"]
```
cemozerr, line 83: When reading I was a bit unclear how "defining" the native flag look like. 

After the "Define 'native' feature flag to separate logic that is not needed in ZK context" would it make sense to explain why in this instance the native flag definition includes "serde", and "sov-state/native" etc.?

examples/demo-nft-module/README.md
```diff
@@ -0,0 +1,481 @@
...
+```
+
+It is needed for optimizing module for execution in ZK context, where none of RPC related logic is needed.
```
cemozerr, line 86:
```suggestion
The native flag is needed for optimizing module for execution in ZK context, where none of RPC related logic is needed.
```

examples/demo-nft-module/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+
+/// This enumeration responsible for querying the nft module.
+#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq)]
+pub enum QueryMessage {
+    GetOwner { token_id: u64 },
+}
+```
+
+Query message is only used in native mode, because it does not change state in any way, since it is not used in state transition function.
```
cemozerr, line 120: For clarity, how about? 


```suggestion
The query message is exclusively used in native mode as it does not alter the state. This is because it is not utilized in the state transition function.
```

cemozerr: BTW, just curious, how do different modules read the state of other modules? Do they not use the query message?

examples/demo-nft-module/README.md
citizen-stig: Done

examples/demo-nft-module/README.md
citizen-stig: Addressed this

examples/demo-nft-module/README.md
citizen-stig: Module should expose public methods, using query messages adds overhead on serialization/deserialization.

preston-evans98 (approved):

cemozerr (approved):
LGTM!

PR #301, opened 2023-05-18
Use rpc macros for bank and election modules

dubbelosix:
Hardcoded the usage of the new `rpc_gen` and `expose_rpc` macros.

Moved `RpcStorage` struct back to macro because of orphan rules complaining.


examples/demo-rollup/Cargo.toml
```diff
@@ -20,12 +20,21 @@ hex = { workspace = true }
 tendermint = "0.32"
 tokio = { version = "1", features = ["full"] }
 tracing-subscriber = "0.3.16"
+bytes = "1.1.0"
```
preston-evans98, line 23: The sovereign SDK re-exports the bytes crate to avoid version conflicts. Let's use that one instead!

examples/demo-rollup/src/config.rs
```diff
@@ -51,6 +61,10 @@ mod tests {
                     path: PathBuf::from("/tmp"),
                 },
             },
+            rpc_config: RpcConfig {
+                bind_host: "127.0.0.1".to_string(),
```
preston-evans98, line 65: Do we need them to specify the host? Or can we default to "127.0.0.1"?

examples/demo-rollup/src/main.rs
```diff
@@ -24,9 +26,20 @@ pub fn initialize_ledger() -> LedgerDB {
     ledger_db
 }
 
+async fn rpc(module: RpcModule<()>, address: SocketAddr) {
```
preston-evans98, line 32: nit: should we rename this to `start_rpc_server` or something a little more descriptive?

examples/demo-rollup/src/main.rs
```diff
@@ -40,6 +53,14 @@ async fn main() -> Result<(), anyhow::Error> {
 
     // Initialize the Celestia service
     let mut demo_runner = NativeAppRunner::<Risc0Host>::new(rollup_config.runner.clone());
+
+    let storj = demo_runner.inner().current_storage.clone();
```
preston-evans98, line 76: This looks like it depends on an implementation detail of `demo_runner` (that `inner()` returns an AppTemplate which happens to have a field called `current_storage`. Ideally, all of this should be implemented purely in terms of trait methods. Can we add a trait method to `StateTransitionRunner` which returns a handle to storage?

examples/demo-rollup/src/main.rs
```diff
@@ -57,13 +78,14 @@ async fn main() -> Result<(), anyhow::Error> {
...
-            117, 100, 54, 103, 57, 116, 52, 107, 122, 109, 115, 108, 112, 102, 50, 52, 121, 115,
-            97, 120, 113, 102, 110, 122, 101, 101, 53, 119, 57,
+            99, 101, 108, 101, 115, 116, 105, 97, 49, 113, 112, 48, 57, 121, 115, 121, 103, 99,
+            120, 54, 110, 112, 116, 101, 100, 53, 121, 99, 48, 97, 117, 54, 107, 57, 108, 110, 101,
+            114, 48, 53, 121, 118, 115, 57, 50, 48, 56,
         ],
         &sequencer_private_key,
         &sequencer_private_key,
     );
+    println!("priv: {}", sequencer_private_key.as_hex());
```
preston-evans98: nit: Let's remove the `print`

examples/demo-rollup/src/main.rs
```diff
@@ -97,15 +119,22 @@ async fn main() -> Result<(), anyhow::Error> {
...
         }
         let (next_state_root, _witness, _) = demo.end_slot();
         ledger_db.commit_slot(data_to_commit)?;
         prev_state_root = next_state_root.0;
     }
 
+    // uncomment this if we need to process a smaller range of blocks
+    // and we want to use the rpc to query the storage
+    // println!("waiting on RPC");
+    // futures::future::pending::<()>().await;
```
preston-evans98, line 156: nit: remove commented code

module-system/sov-modules-api/src/lib.rs
```diff
@@ -200,9 +200,9 @@ pub trait ModuleInfo {
...
-#[derive(Clone)]
-pub struct RpcStorage<C: Context> {
-    pub storage: C::Storage,
-}
+// /// this is the struct that users need to instantiate
+// /// with a storage object to convert into an RPC
+// #[derive(Clone)]
+// pub struct RpcStorage<C: Context> {
+//     pub storage: C::Storage,
+// }
```
preston-evans98, line 229: nit: Let's remove commented code

examples/demo-rollup/src/config.rs
dubbelosix: oh, the rollup_config.toml does have it set to 127.0.0.1 so it should default to that
the "127.0.0.1" here is for the test case.
i think sometimes users like binding to 0.0.0.0 or even a specific network interface

examples/demo-rollup/rollup_config.toml
```diff
@@ -13,3 +13,9 @@ max_celestia_response_body_size = 104_857_600
 [runner.storage]
 # The path to the rollup's data directory. Paths that do not begin with `/` are interpreted as relative paths.
 path = "demo_data"
+
+[rpc_config]
+# the host and port to bind the rpc server for
+bind_host = "127.0.0.1"
+bind_port = 12345
```
cemozerr, line 20: Is this the port Celestia currently exposes by default? Would be nice if users didn't need to change it.

examples/demo-rollup/src/main.rs
```diff
@@ -24,15 +27,24 @@ pub fn initialize_ledger(path: impl AsRef<std::path::Path>) -> LedgerDB {
...
 pub fn get_genesis_config() -> GenesisConfig<DefaultContext> {
     let sequencer_private_key = DefaultPrivateKey::generate();
     create_demo_genesis_config(
         100000000,
         sequencer_private_key.default_address(),
         vec![
-            99, 101, 108, 101, 115, 116, 105, 97, 49, 122, 102, 118, 114, 114, 102, 97, 113, 57,
-            117, 100, 54, 103, 57, 116, 52, 107, 122, 109, 115, 108, 112, 102, 50, 52, 121, 115,
-            97, 120, 113, 102, 110, 122, 101, 101, 53, 119, 57,
+            99, 101, 108, 101, 115, 116, 105, 97, 49, 113, 112, 48, 57, 121, 115, 121, 103, 99,
```
cemozerr, line 45: I know this was just an update, but it would be nice to define this magic vector somewhere and then use its variable for readability, so that first time readers would know what it means.

module-system/sov-modules-macros/src/dispatch/derive_rpc.rs
```diff
@@ -371,17 +371,32 @@ pub(crate) fn rpc_outer_impls(
...
+
     let original_impl = quote! {
         #(#attrs)*
         #input
+        #rpc_storage_struct
     };
+
     output_tokens.extend(original_impl);
 
+    // will be replaced in the below loop
```
cemozerr, line 396: Shall we add a TODO? or are we planning to fix this hack before merging?

module-system/sov-modules-macros/tests/dispatch/derive_rpc.rs
```diff
@@ -37,8 +37,16 @@ pub struct TestRuntime<C: sov_modules_api::Context> {
     test_struct: TestStruct<C>,
 }
 
+// this is generated by a macro
```
cemozerr, line 40: nit:

```suggestion
// This is generated by a macro annotating the state transition runner, 
// but we do not have that in scope here so generating the struct manually.
```

examples/demo-rollup/rollup_config.toml
dubbelosix: oh this is the configuration for the RPC that that full node would run, so the user can set whichever port and bind to whichever network interface they want

module-system/sov-modules-macros/src/dispatch/derive_rpc.rs
dubbelosix: yeah, i can add a todo for this - it's a bit complex because we let each module specify its own context, but the runner has only a single context and storage. we might need to reconcile that later

examples/demo-rollup/src/main.rs
dubbelosix: just checked with preston - since its consensus critical /  part of the genesis config, its better to have it in the codebase as opposed to runtime configuration. moved it back into the code for now

examples/demo-rollup/src/config.rs
preston-evans98: Hmm, yeah good point. We might want to add a serde default to 127.0.0.1 so that users can just not specify an interface if they want that, but that's out of scope - and you're right that it's nice to have the option to configure something else

examples/demo-rollup/src/main.rs
preston-evans98: Good call on the magic value though - we should add a doc explaining how and why to pick this value (and the magic 1000000).

module-system/sov-modules-api/src/lib.rs
```diff
@@ -221,9 +223,26 @@ pub trait ModuleInfo {
...
+    /// Return a mutable reference to the inner STF implementation
+    fn inner_mut(&mut self) -> &mut Self::Inner;
+
+    /// Return a mutable reference to the inner STF implementation
+    fn get_storage(&self) -> <Self::Context as Spec>::Storage;
+
+    // /// Report if the state transition function has been initialized.
+    // /// If not, node implementations should respond by running `init_chain`
+    // fn has_been_initialized(&self) -> bool;
+}
```
preston-evans98: Hmm, thinking about this change in more depth,  I think it becomes impossible to run the full node without taking a dependency on sov-modules-api (you need the state-transition runner to get started, right?). 

If so, I guess we need to go the route of splitting this into two traits: the old StateTransitionRunner (which would live in the core/rollup-interface crate) and a new trait like RpcRunner which lives in sov-modules-api and has the `Context` type and `get_storage` method. That way, users can still run the default full-node implementation even if they don't use sov-modules, but they get an easy way to add RPC by just importing the trait and running an initializer if they do use sov-modules

preston-evans98 (commented):
Looks good overall. Just one thing to get cleared up before merging - if we move StateTransitionRunner to sov-modules, users who don't rely on the module system won't be able to initialize their rollups.

So I guess we'll have to move STR back to its old location and add a new trait for the extra items.

module-system/sov-modules-api/src/lib.rs
dubbelosix: yeah that makes sense. was thinking of a way to avoid another trait implementation that they needed to provide, but being explicit is better than tying them together

preston-evans98 (approved):

examples/demo-rollup/src/main.rs
dubbelosix: added sequencer_da_address as a constant

PR #308, opened 2023-05-22
`sov-state` readme

bkolad:


module-system/sov-state/README.md
```diff
@@ -0,0 +1,82 @@
...
+state.value.set(some_value, working_set);
+let maybe_value = state.value.get(working_set);
+
+```
+
+## Low level explanation:
+It's important to note that this section is not necessarily required for efficient usage of the `sov-state`.
+
+### `Native` & `Zkp` execution: 
+During `Native` execution, the data is stored in a `key-value` store, which is accessed through the `WorkingSet`. It's worth mentioning that the actual storage mechanism, such as `RocksDB`, is only accessible during this phase when the full node executes the transaction and updates the state.
```
preston-evans98, line 48: nit: Can we add a quick explanation that working sets *read* data from the backing store but write to an in-memory map which can be reverted or committed?

module-system/sov-state/README.md
```diff
@@ -0,0 +1,82 @@
...
+    /// returning the new state root after applying all writes
+    fn validate_and_commit(
+        &self,
+        state_accesses: OrderedReadsAndWrites,
+        witness: &Self::Witness,
+    ) -> Result<[u8; 32], anyhow::Error>;
+}
+```
+The `sov-state` crate offers two implementations of the `Storage` trait, namely `ZkStorage` and `ProverStorage`, which handle the storage and retrieval of data in the context of the `Zkp` and `Prover` execution modes, respectively. These implementations encapsulate the necessary logic and interactions with the storage system, allowing module developers to work with a consistent interface regardless of the execution mode.
+
```
preston-evans98, line 76: Nit: Can we add a quick explanation that "ProverStorage" can be used in regular execution too, but that we'll add an optimized implementation which doesn't generate witnesses eventually

preston-evans98 (approved):

module-system/sov-state/README.md
```diff
@@ -0,0 +1,82 @@
+# sov-state
+This crate provides abstractions specifically designed for storing and retrieving data from a permanent storage, tailored to be used within the Sovereign `module-system`.
```
cemozerr, line 2: just a nit to align with the latest component renamings:

```suggestion
This crate provides abstractions specifically designed for storing and retrieving data from a permanent storage, tailored to be used within the `module-system`.
```

module-system/sov-state/README.md
bkolad: Added  relevant info to the `WorkingSet` section

module-system/sov-state/README.md
bkolad: Done

Issue #248, opened 2023-05-08
Cut v0.1.0 release


preston-evans98:
Once all other tasks for the Alpha release are finished, we should make our code available on `crates.io`. Packages to be released are

- [x] sov-core (done but renamed to rollup-interface)
- [x] sov-state
- [x] sov-modules-api
- [x] sov-value-setter
- [x] sov-election
- [x] sov-accounts
- [x] sov-app-template (sov-default-stf)
- [x] sov-bank
- [x] sov-sequencer-registry
- [x] sov-prover-incentives
- [x] sov-modules-macros
- [x] sovereign-db (Rename to sov-db)
- [x] schemadb (Rename to sov-schema-db)
- [x] create Slack thread to confirm names
- [x] Set `publish = false` for examples crates and name them like other, #337 

Update: Rethinking this task, I think we want to exclude `election` and `value-setter`.  Open question - do we want to publish jupiter and the risc0 adapter?

PR #336, opened 2023-05-25
prefix module names with `sov-`

bkolad:
# Description
Following modules name have been updated:
- sov-value-setter
- sov-election
- sov-accounts
- sov-bank
- sov-sequencer-registry
- sov-prover-incentives

## Linked Issues
- Related to #248

## Testing
Tests update with new imports.

## Docs
Docs updates with new module names.


citizen-stig:
Do we need to rename folders to match crate name? 
Like, can we keep Bank module in `bank` folder, but name crate `sov-bank` ?

bkolad:
> Do we need to rename folders to match crate name?
> Like, can we keep Bank module in `bank` folder, but name crate `sov-bank` ?
I think it is easier to navigate if folders match crate names so if this is not a big deal I would prefer to keep it this way.

citizen-stig:
> I think it is easier to navigate if folders match crate names so if this is not a big deal I would prefer to keep it this way.

I think it pollutes overview a little bit, but on a big deal. The only reason why we prefix everything with `sov-` is because we want to publish those.

citizen-stig (approved):

PR #347, opened 2023-05-28
mint and freeze tokens

dubbelosix:
add mint and freeze functionality to bank module.
changes
- create: 
  - store a vector of authorized minters. currently hardcoded to a single element "sender" who created the token
  - boolean flag freeze
- freeze: sets freeze to true if not set, errors otherwise. only authorized minter allowed
- mint: mints coins to specified address, only if not frozen and if sender is in authorized minters

wip, tests in progress

### DIFF
--- a/module-system/module-implementations/sov-bank/src/call.rs
+++ b/module-system/module-implementations/sov-bank/src/call.rs
@@ -43,7 +43,7 @@ pub enum CallMessage<C: sov_modules_api::Context> {
         /// The amount of tokens to mint.
         coins: Coins<C>,
         /// Address to mint tokens to
-        minter_address: C::Address
+        minter_address: C::Address,
     },
 
     /// Freeze a token so that the supply is frozen
@@ -116,7 +116,8 @@ impl<C: sov_modules_api::Context> Bank<C> {
         working_set: &mut WorkingSet<C::Storage>,
     ) -> Result<CallResponse> {
         let mut token = self.tokens.get_or_err(&coins.token_address, working_set)?;
-        let mint_response = token.mint(context.sender(), &minter_address, coins.amount, working_set)?;
+        let mint_response =
+            token.mint(context.sender(), &minter_address, coins.amount, working_set)?;
         token.total_supply += coins.amount;
         self.tokens.set(&coins.token_address, token, working_set);
 
--- a/module-system/module-implementations/sov-bank/src/call.rs
+++ b/module-system/module-implementations/sov-bank/src/call.rs
@@ -105,11 +105,11 @@ impl<C: sov_modules_api::Context> Bank<C> {
         working_set: &mut WorkingSet<C::Storage>,
     ) -> Result<CallResponse> {
         let mut token = self.tokens.get_or_err(&coins.token_address, working_set)?;
-        let burn_response = token.burn(context.sender(), coins.amount, working_set)?;
+        token.burn(context.sender(), coins.amount, working_set)?;
         token.total_supply -= coins.amount;
         self.tokens.set(&coins.token_address, token, working_set);
 
-        Ok(burn_response)
+        Ok(CallResponse::default())
     }
 
     pub(crate) fn mint(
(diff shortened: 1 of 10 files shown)


module-system/module-implementations/sov-bank/src/call.rs
```diff
@@ -93,6 +107,34 @@ impl<C: sov_modules_api::Context> Bank<C> {
...
+    pub(crate) fn mint(
+        &self,
+        coins: Coins<C>,
+        minter_address: C::Address,
+        context: &C,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<CallResponse> {
+        let mut token = self.tokens.get_or_err(&coins.token_address, working_set)?;
+        let mint_response = token.mint(context.sender(), &minter_address, coins.amount, working_set)?;
+        token.total_supply += coins.amount;
```
bkolad, line 120: Let's make the `mint` function to change the `token.total_supply ` this way all the minting logic  is in one place.

dubbelosix: sounds good !

module-system/module-implementations/sov-bank/src/token.rs
```diff
@@ -26,6 +26,10 @@ pub(crate) struct Token<C: sov_modules_api::Context> {
     pub(crate) total_supply: u64,
     /// Mapping from user address to user balance.
     pub(crate) balances: sov_state::StateMap<C::Address, Amount>,
+    /// Flag indicating if the supply is frozen.
+    pub(crate) frozen: bool,
+    /// Flag indicating if the supply is frozen.
```
bkolad, line 31: The comment is copy pasted.

module-system/module-implementations/sov-bank/src/token.rs
```diff
@@ -62,6 +66,43 @@ impl<C: sov_modules_api::Context> Token<C> {
         Ok(CallResponse::default())
     }
 
+    pub(crate) fn freeze(&mut self, sender: &C::Address) -> Result<CallResponse> {
+        self.is_authorized_minter(sender)?;
+        if self.frozen {
+            bail!("Token is already frozen")
+        }
+        self.frozen = true;
+        Ok(CallResponse::default())
```
bkolad, line 75: Nit: Can this function just return `()`,  It looks like this function don't have to be aware of this type.

module-system/module-implementations/sov-bank/src/token.rs
```diff
@@ -62,6 +66,43 @@ impl<C: sov_modules_api::Context> Token<C> {
...
+            bail!("Attempt to mint frozen token")
+        }
+        let to_balance = self
+            .balances
+            .get(minter_address, working_set)
+            .unwrap_or_default()
+            + amount;
+        self.balances.set(minter_address, to_balance, working_set);
+        self.total_supply += amount;
+        Ok(CallResponse::default())
```
bkolad, line 96: Same

module-system/module-implementations/sov-bank/src/token.rs
```diff
@@ -107,6 +149,8 @@ impl<C: sov_modules_api::Context> Token<C> {
             name: token_name.to_owned(),
             total_supply,
             balances,
+            frozen,
+            authorized_minters: authorized_minters.unwrap_or(vec![C::Address::try_from(sender)?]),
```
bkolad, line 153: Here we put the sender of the tx as a minter only when the vec is empty. I think it would be more consistent to always insert it always (or never).

dubbelosix: ah ok! so the sender/creator is always part of the authorized minter. yeah, that was one of the questions i had. I can make that change

module-system/module-implementations/sov-bank/src/token.rs
```diff
@@ -103,10 +150,20 @@ impl<C: sov_modules_api::Context> Token<C> {
             None => bail!("Total supply overflow"),
         };
 
+        let mut indices = HashSet::new();
+        let mut auth_minter_list = Vec::new();
+
+        for (i, item) in authorized_minters.iter().enumerate() {
```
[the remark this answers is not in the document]
dubbelosix: hmm, so I was using the loop because I'm also deduping the source vector in case a user passes in the same address multiple times in the authorized_vector list

bkolad (approved):

PR #348, opened 2023-05-29
Move `events` to WorkingSet

### LINKED ISSUE #233: Proposal: Change event handling in `sov-modules`
Currently, events thrown by a module are returned in the `response` object while the state is modified via mutable reference to `working_set`, this brakes the uniformity of the API. See the code below: 

Example:
```rust
    pub(crate) fn method_form_parent_module(
        &self,
        new_value: u32,
        context: &C,
        working_set: &mut WorkingSet<C::Storage>,
    ) -> Result<sov_modules_api::CallResponse> {
        let mut response = CallResponse::default();

        let other_events = self.some_other_module.method(working_set);
        response.add_events(other_events);     <----- ADD EVENTS

        ...
        self.value.set(new_value, working_set);
        response.add_event("set", &format!("value_set: {new_value:?}"));

        Ok(response)
    }
```

The parent module has to add `other_events` to the response "manually" while `working_set` is updated transparently. 
We propose the following API changes:

1. New structure:
```rust
struct Env<S:Storage>{
  working_set: WorkingSet<S>.
  events: Vec<Event>,
}
```

And the example code would change to:

```rust
    pub(crate) fn method_form_parent_module(
        &self,
        new_value: u32,
        context: &C,
        env: &mut Env<C::Storage>,
    ) -> Result<sov_modules_api::CallResponse> {
       let mut response = CallResponse::default();

       self.some_other_module.method(env);
  
       // ...
       self.value.set(new_value, env.working_set);
       env.add_event("set", &format!("value_set: {new_value:?}"));

       Ok(response)
    }
```
We no longer call `response.add_events(other_events)`. The events and state are modified the same way via mutable reference to `Env`.

Drawbacks:
1. We will have to update `sov-macros` to make it work.
2. Before, we had :
`self.value.set(new_value, working_set);` but now we have to get `working_set` from env 
`self.value.set(new_value, env.working_set);` for every `set` and `get` 

Things to consider:
Should we put `Context` inside `Env` as well?

bkolad:
# Description
1. The `events` field is moved from the `CallResponse` to the `WorkingSet`.
2. The `WorkingSet` is refactored with `events` handling.
3. The `election, value-setter & nft` modules are updated to include additional `events`.
4. The `sov-default-stf`  is refactored, and `TransactionReceipt` gets `events` from `WorkingSet`.

## Linked Issues
- Fixes #233
- Related to #349, #350

## Testing
1. A check for correct `events` propagation was added in the `nested-module-test`.
2. Updated tests in `election, value-setter & nft` modules to verify the correct propagation of events. 
3. Updated `demo-stf` tests to verify the correct propagation of events.
 
## Docs
Updated docs for the `nft` module.

### DIFF
--- a/module-system/sov-default-stf/src/lib.rs
+++ b/module-system/sov-default-stf/src/lib.rs
@@ -80,6 +80,9 @@ where
             };
         }
 
+        // TODO
+        let _ = batch_workspace.take_events();
+
         // Commit `enter_apply_batch` changes.
         batch_workspace = batch_workspace.commit().to_revertable();
 
@@ -128,25 +131,16 @@ where
             let verified_tx = match self.tx_hooks.pre_dispatch_tx_hook(tx, &mut batch_workspace) {
                 Ok(verified_tx) => verified_tx,
                 Err(e) => {
-                    // // Revert the batch.
-                    // batch_workspace = batch_workspace.revert();
-
-                    // // We reward sequencer funds inside `exit_apply_batch`.
-                    // self.tx_hooks
-                    //     .exit_apply_batch(0, &mut batch_workspace)
-                    //     .expect("Impossible happened: error in exit_apply_batch");
-
-                    // self.working_set = Some(batch_workspace);
-
                     // Don't revert any state changes made by the pre_dispatch_hook even if it rejects
                     error!("Stateful verification error - the sequencer included an invalid transaction: {}", e);
                     batch_workspace = batch_workspace.revert();
                     let receipt = TransactionReceipt {
                         tx_hash: raw_tx_hash,
                         body_to_save: None,
-                        events: vec![],
+                        events: batch_workspace.take_events(),
                         receipt: TxEffect::Reverted,
                     };
+
                     tx_receipts.push(receipt);
                     continue;
                 }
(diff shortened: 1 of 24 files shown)


module-system/sov-state/src/scratchpad.rs
```diff
@@ -40,73 +41,114 @@ impl<S: Storage> Debug for RevertableDelta<S> {
...
-            s @ WorkingSet::Standard(_) => s,
-            WorkingSet::Revertable(revertable) => WorkingSet::Standard(revertable.revert()),
+        let read_write_set = match self.read_write_set {
+            s @ ReadWriteSet::Standard(_) => s,
+            ReadWriteSet::Revertable(revertable) => ReadWriteSet::Standard(revertable.revert()),
+        };
+        Self {
+            read_write_set,
+            // The `revert` removes all events associated with the transaction
+            events: Vec::default(),
```
preston-evans98, line 104: I don't think this logic matches the state revert logic. To be consistent, don't we want to remove the events that have happened since the last call to `to_revertable()`?

module-system/sov-default-stf/src/lib.rs
```diff
@@ -128,25 +132,16 @@ where
...
-                    // self.working_set = Some(batch_workspace);
-
                     // Don't revert any state changes made by the pre_dispatch_hook even if it rejects
                     error!("Stateful verification error - the sequencer included an invalid transaction: {}", e);
                     batch_workspace = batch_workspace.revert();
                     let receipt = TransactionReceipt {
                         tx_hash: raw_tx_hash,
                         body_to_save: None,
-                        events: vec![],
+                        events: batch_workspace.take_events(),
```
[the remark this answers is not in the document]
bkolad: +1

module-system/sov-state/src/scratchpad.rs
preston-evans98: Synced with @bkolad offline. This divergence is intentional, since `events` are a per-transaction concept while state changes are accumulated per-batch.

preston-evans98 (approved):

PR #360, opened 2023-05-30
Fix demo-rollup restart issue

### LINKED ISSUE #340: Demo rollup and prover fails after restart
Original issue: https://github.com/Sovereign-Labs/demo-rollup/issues/9

two storages: ledger, state. demo_app is currently using ledger slot number to determine if storage has been initialized or not

```
let mut item_numbers = ledger_db.get_next_items_numbers();
    if item_numbers.slot_number == 1 {
        print!("No history detected. Initializing chain...");
        demo.init_chain(());
        println!("Done.");
    } else {
        println!("Chain is already initialized. Skipping initialization.");
    }
```

problem with this is, if for some reason after initializing the ledger, sync fails (eg: 
* due to incorrect authentication token, 
* or because the light client is not running, has crashed etc), 
slot number will remain at 1 and when the demo rollup tries to start again, it ends up attempting to re-initialize the db and panics.

```
No history detected. Initializing chain...initializing runtime
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ModuleError(Token address already exists)', 
sov-modules/sov-app-template/src/lib.rs:88:14
```

I think using `slot_number == 1` is not a good indication of `is chain initialized / do we need to run init_chain`. 
(a trivial case is to run `init_chain` and check slot number - and it's still 1!)

we need a better check to know that the state is "initialized" because there can be any number of failures once chain is initialized and before slot number advances. or we can make init_chain fail gracefully if its called multiple times, but this still requires some notion of knowing the chain has been "initialized"

in the shorter term, I can also just catch_unwind and detect the condition, or it can be as simple as adding "nuke the db" to the instructions, since that ends up working. a better solution is to have something queryable that tells us if both ledger and state have been "initialized". could also just be a boolean flag, but might be overkill?

@preston-evans98 lmk wyt 

since this is just a demo, we can also just add instructions

citizen-stig:
# Description
Adds `is_empty` method to `Storage` trait so it can be checked during initialization.

## Linked Issues
- Fixes #340 

## Testing
Added unit tests with restart lifecycle

## Docs
Trait documentation has been updated

### DIFF
--- a/module-system/sov-state/src/zk_storage.rs
+++ b/module-system/sov-state/src/zk_storage.rs
@@ -91,4 +91,9 @@ impl<S: MerkleProofSpec> Storage for ZkStorage<S> {
 
         Ok(new_root.0)
     }
+
+    // ZK Storage is never empty
+    fn is_empty(&self) -> bool {
+        false
+    }
 }
--- a/examples/demo-rollup/src/main.rs
+++ b/examples/demo-rollup/src/main.rs
@@ -27,6 +27,7 @@ use tracing::{debug, info};
 // RPC related imports
 use demo_stf::app::get_rpc_methods;
 use sov_modules_api::RpcRunner;
+use sov_state::Storage;
 
 // The rollup stores its data in the namespace b"sov-test" on Celestia
 // You can change this constant to point your rollup at a different namespace
@@ -93,6 +94,7 @@ async fn main() -> Result<(), anyhow::Error> {
 
     // Our state transition also implements the RpcRunner interface, so we use that to initialize the RPC server.
     let storj = demo_runner.get_storage();
+    let is_storage_empty = storj.is_empty();
     let mut methods = get_rpc_methods(storj);
     let ledger_rpc_module =
         ledger_rpc::get_ledger_rpc::<DemoBatchReceipt, DemoTxReceipt>(ledger_db.clone());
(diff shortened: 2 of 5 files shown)


module-system/sov-state/src/zk_storage.rs
```diff
@@ -91,4 +91,9 @@ impl<S: MerkleProofSpec> Storage for ZkStorage<S> {
 
         Ok(new_root.0)
     }
+
+    // ZK Storage is never empty
+    fn is_empty(&self) -> bool {
+        false
```
preston-evans98, line 97: This could cause a divergence between native and zk storage, which could lead to very unintuitive bugs. 
 I think the best thing to do is to check if its state_root is equal to the root of an empty JMT, but we can also leave it `unimplemented!()` for now.

preston-evans98 (changes requested):

module-system/sov-state/src/zk_storage.rs
```diff
@@ -91,4 +91,8 @@ impl<S: MerkleProofSpec> Storage for ZkStorage<S> {
 
         Ok(new_root.0)
     }
+
+    fn is_empty(&self) -> bool {
+        unimplemented!("Needs simplification in JellyfishMerkleTree")
```
preston-evans98, line 96: Do we have an open issue for this?

preston-evans98 (approved):

module-system/sov-state/src/zk_storage.rs
citizen-stig: Now we do: https://github.com/Sovereign-Labs/sovereign-sdk/issues/362

Issue #320, opened 2023-05-23
Auto-generate hooks based on Runtime


preston-evans98:
Currently, users have to manually define their TxHooks implementation based on the modules they use. In principle, though, it seems like it should be possible (and much less error-prone) to automatically generate the hooks based on the runtime. 

cc @bkolad

bkolad:
What about this:
1. We introduce `ModuleHook` trait with `pre_dispatch_tx_hook, post_dispatch_tx_hook` etc methods
2. Every module that introduces hooks needs to implement this trait. 
3.  Runtime gets an additional macro and `[hooks]` attribute:

```rust
#[derive(Genesis, DispatchCall, MessageCodec, Hooks)]
pub struct Runtime<C: Context> {
   #[hook]
    pub sequencer: sequencer::Sequencer<C>,

    pub bank: bank::Bank<C>,

    pub election: election::Election<C>,

    pub value_setter: value_setter::ValueSetter<C>,

   #[hook]
    pub accounts: accounts::Accounts<C>,
}
```
And the implementation of the `Hook` trait is automatically derived.

preston-evans98:
This looks great overall. Do we have a motivating a use case where users would want a module without its hooks though? If not, can we just derive a no-op `Hook` trait implementation for all Modules and then get rid of the `#[hook]` annotation here? I'm worried that forgetting to add the annotation will be a very subtle way to completely break security guarantees

bkolad:
I think/hope most of the modules won't expose hooks. Hopefully, our abstractions are strong enough that users can accomplish what they want just by simply implementing the `Module` trait, and only if a module has to do something very special, it exposes hooks. But I think it is ok to automatically derive a default `Hooks` implementation and do noop.

bkolad:
After incorporating #361, the hooks implementation has a well-defined structure, allowing us to implement the macro. However, there is a small remaining issue.

```Rust
impl<C: Context> TxHooks for Runtime<C> {
    type Context = C;

    fn pre_dispatch_tx_hook(
        &self,
        tx: Transaction<Self::Context>,
        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
    ) -> anyhow::Result<<Self::Context as Spec>::Address> {
        self.accounts.pre_dispatch_tx_hook(tx, working_set)
    }

    fn post_dispatch_tx_hook(
        &self,
        tx: &Transaction<Self::Context>,
[...]

PR #361, opened 2023-05-30
Hooks simplification 

### LINKED ISSUE #319: Create Default Tx Hooks and Default Tx Verifier
Using hooks is an advanced feature for modules. While a few special modules like accounts and sequencer need access to hooks, most modules will not. So, we should consider making moving the current DemoTxHooks and DemoTxVerifier into another package and exporting them as default values.

cc @bkolad, curious for your thoughts here

bkolad:
# Description
1. The `TxHooks` has been split into two separate hooks: `ApplyBlobTxHooks` and `ApplyBlobSequencerHooks`. These hooks now belong to the `sov-modules-api` and are implemented in `sov-accounts` and `sov-sequencer-registry`. Consequently, the `DempAppHook` structure has been removed.
2. The `TxVerifier` has been removed, and the stateless verification has been implemented in `sov-default-stf`. This verification only performs a signature check, and it makes sense to share it.

These two changes have simplified the `AppTemplate` since it no longer requires the aforementioned structures as parameters.

3. The `AppTemplate` requires that `Runtime` implements `ApplyBlobTxHooks` and `ApplyBlobSequencerHooks` (the implementation is trivial).

Now `ApplyBlobTxHooks` implementation for `Runtime`  has a well-defined structure and could be implemented as a macro, but it should be done in a separate PR.  

## Linked Issues
- Fixes #319
- Related to  #320 

## Testing
All relevant tests are updated.

## Docs
Relevant readmes updated (the biggest change in `demo-stf`)


examples/demo-stf/README.md
```diff
@@ -33,90 +33,90 @@ several parameters that specify its exact behavior. In order, these four generic
 1. `Context`: a per-transaction struct containing the message's sender. This also provides specs for storage access, so we use different `Context`
    implementations for Native and ZK execution. In ZK, we read values non-deterministically from hints and check them against a merkle tree, while in
    native mode we just read values straight from disk.
-2. `TxVerifier`: a struct that verifies the signatures on transactions and deserializes them into messages
-3. `Runtime`: a collection of modules which make up the rollup's public interface
-4. `TxHooks`: a set of functions which are invoked at various points in the transaction lifecycle
+2. `Runtime`: a collection of modules which make up the rollup's public interface
+3. `TxHooks & ApplyBlobHooks` derived from the `Runtime`: a set of functions which are invoked at various points in the transaction lifecycle.
```
preston-evans98, line 37: Since these generics don't show up on the AppTemplate anymore, I think we can drop them from the documentation.  We should also change the line above this list from "In order, these four generics are:" to "In order, these generics are:"

examples/demo-stf/README.md
```diff
@@ -33,90 +33,90 @@ several parameters that specify its exact behavior. In order, these four generic
...
-    sequencer_hooks: sov_sequencer_registry::hooks::Hooks<C>,
+    fn reward_sequencer(
+        &self,
+        amount: u64,
+        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
+    ) -> anyhow::Result<()> {
+        self.sequencer.reward_sequencer(amount, working_set)
+    }
 }
 ```
```
preston-evans98, line 119: nit: Can we make this section a sub-section of `Implementing Runtime`?

module-system/sov-default-stf/README.md
```diff
@@ -30,8 +29,7 @@ where
 
 1. The `DispatchCall` trait is responsible for decoding serialized messages and forwarding them to the appropriate module.
 1. The `Genesis` trait handles the initialization process of the rollup. It sets up the initial state upon the rollup deployment.
-1. The `TxVerifier` trait is responsible for validating transactions within the rollup. It ensures that incoming transactions meet the necessary criteria and are valid for execution.
-1. The `TxHooks` trait allows for the injection of custom logic into the transaction processing pipeline. It provides a mechanism to execute additional actions or perform specific operations during the transaction processing phase.
+1. The `TxHooks & ApplyBlobHooks` traits that allows for the injection of custom logic into the transaction processing pipeline. It provides a mechanism to execute additional actions or perform specific operations during the transaction processing phase.
```
preston-evans98, line 32: nit: Since there are two traits, we need to make this plural: 

`The `TxHooks & ApplyBlobHooks` traits that allow for the injection of custom logic into the transaction processing pipeline. They provide a mechanism to execute additional actions or perform specific operations during the transaction processing phase.`

module-system/sov-modules-api/src/transaction.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub fn verify(&self) -> anyhow::Result<()> {
+        // We check signature against runtime_msg and nonce.
+        let mut hasher = C::Hasher::new();
+        hasher.update(self.runtime_msg());
+        hasher.update(&self.nonce().to_le_bytes());
+        let msg_hash = hasher.finalize();
+        self.signature().verify(self.pub_key(), msg_hash)?;
+
+        Ok(())
+    }
```
preston-evans98, line 58: non-blocking: Is there any way that we can run this check in the constructor so that invalid transactions are unrepresentable. That way, we can't have any bugs from forgetting to call `verify`

examples/demo-stf/README.md
```diff
@@ -33,90 +33,90 @@ several parameters that specify its exact behavior. In order, these four generic
...
-    type Transaction;
-    /// Runs stateless checks against a single RawTx.
-    fn verify_tx_stateless(&self, raw_tx: RawTx) -> anyhow::Result<Self::Transaction>;
-```
+There are two kind of hooks:
 
-The semantics of the `TxVerifier` are pretty straightforward - it takes a `RawTx` (a slice of bytes) as an argument, and does
-some work to transform it into some output `Transaction` type _without looking at the current rollup state_. This output transaction
-type will eventually be fed to the `TxHooks` for _stateful_ verification.
+`ApplyBlobTxHooks` with the following methods:
```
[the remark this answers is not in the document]
bkolad: Fixed

examples/demo-stf/README.md
bkolad: Fixed

examples/demo-stf/README.md
bkolad: Done

module-system/sov-default-stf/README.md
bkolad: Done

module-system/sov-modules-api/src/transaction.rs
bkolad: We can, but I would prefer to do it in a separate PR. We have some unit tests using `Transaction::new()` with bad signatures so it will require additional refactoring.

preston-evans98 (approved):

PR #364, opened 2023-05-30
Theo/add macro runtime

### LINKED ISSUE #309: Macro for generating `Runtime::new()`
Maybe we should have `Default` trait and follow back to it?

theochap:
# Description
A proposal to fix the issue #309 , by adding a macro and an associated trait to define default configuration for the module runtime. I followed the suggestion in #309 to add a new trait associated with the macro, but maybe it would be more elegant (less imports) to add the implementation of new directly to the structure.

## Linked Issues
- Fixes #309 

## Testing

## Docs

### DIFF
--- a/examples/demo-stf/src/app.rs
+++ b/examples/demo-stf/src/app.rs
@@ -13,6 +13,7 @@ pub use sov_modules_api::default_context::ZkDefaultContext;
 #[cfg(feature = "native")]
 pub use sov_modules_api::default_signature::private_key::DefaultPrivateKey;
 use sov_modules_api::Context;
+use sov_modules_api::DefaultConfig;
 #[cfg(feature = "native")]
 use sov_modules_api::RpcRunner;
 #[cfg(feature = "native")]
--- a/examples/demo-stf/src/runtime.rs
+++ b/examples/demo-stf/src/runtime.rs
@@ -1,5 +1,5 @@
 use sov_modules_api::Context;
-use sov_modules_macros::{DispatchCall, Genesis, MessageCodec};
+use sov_modules_macros::{DefaultConfig, DispatchCall, Genesis, MessageCodec};
 
 /// The Rollup entrypoint.
 ///
@@ -36,7 +36,7 @@ use sov_modules_macros::{DispatchCall, Genesis, MessageCodec};
 /// Similar mechanism works for queries with the difference that queries are submitted by users directly to the rollup node
 /// instead of going through the DA layer.
 
-#[derive(Genesis, DispatchCall, MessageCodec)]
+#[derive(Genesis, DispatchCall, MessageCodec, DefaultConfig)]
 #[serialization(borsh::BorshDeserialize, borsh::BorshSerialize)]
 pub struct Runtime<C: Context> {
     pub sequencer: sov_sequencer_registry::Sequencer<C>,
@@ -48,19 +48,4 @@ pub struct Runtime<C: Context> {
     pub value_setter: sov_value_setter::ValueSetter<C>,
 
     pub accounts: sov_accounts::Accounts<C>,
-}
-
-// TODO add macro to generate the following code:
-//      https://github.com/Sovereign-Labs/sovereign/issues/309
-impl<C: Context> Runtime<C> {
-    pub(crate) fn new() -> Self {
-        use sov_modules_api::ModuleInfo;
-        Self {
-            sequencer: sov_sequencer_registry::Sequencer::new(),
-            bank: sov_bank::Bank::new(),
-            election: sov_election::Election::new(),
-            value_setter: sov_value_setter::ValueSetter::new(),
-            accounts: sov_accounts::Accounts::new(),
-        }
-    }
 }
--- a/examples/demo-stf/src/tests/mod.rs
+++ b/examples/demo-stf/src/tests/mod.rs
@@ -2,6 +2,7 @@ use borsh::BorshSerialize;
 use sov_default_stf::{AppTemplate, Batch, SequencerOutcome, TxEffect};
 use sov_modules_api::{
     default_context::DefaultContext, default_signature::private_key::DefaultPrivateKey, Address,
+    DefaultConfig,
 };
 use sov_rollup_interface::stf::BatchReceipt;
 use sov_state::ProverStorage;
--- a/examples/demo-stf/src/tests/stf_tests.rs
+++ b/examples/demo-stf/src/tests/stf_tests.rs
@@ -11,6 +11,7 @@ pub mod test {
     use sov_default_stf::{Batch, SequencerOutcome};
     use sov_modules_api::{
         default_context::DefaultContext, default_signature::private_key::DefaultPrivateKey,
+        DefaultConfig,
     };
     use sov_rollup_interface::{mocks::MockZkvm, stf::StateTransitionFunction};
     use sov_state::{ProverStorage, WorkingSet};
--- a/examples/demo-stf/src/tests/tx_revert_tests.rs
+++ b/examples/demo-stf/src/tests/tx_revert_tests.rs
@@ -6,6 +6,7 @@ use crate::{
 use sov_default_stf::{Batch, SequencerOutcome, SlashingReason};
 use sov_modules_api::{
     default_context::DefaultContext, default_signature::private_key::DefaultPrivateKey,
+    DefaultConfig,
 };
 use sov_rollup_interface::{mocks::MockZkvm, stf::StateTransitionFunction};
 use sov_state::{ProverStorage, WorkingSet};
--- a/module-system/sov-modules-api/src/lib.rs
+++ b/module-system/sov-modules-api/src/lib.rs
@@ -164,6 +164,10 @@ pub trait Context: Spec + Clone + Debug + PartialEq {
     fn new(sender: Self::Address) -> Self;
 }
 
+pub trait DefaultConfig<C: Context> {
+    fn new() -> Self;
+}
+
 impl<T> Genesis for T
 where
     T: Module,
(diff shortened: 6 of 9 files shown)


preston-evans98:
Looks super useful - but can we get away with using the `Default` trait instead of a custom trait? I think we can blanket implement `Default` for `T where T: ModuleInfo`.

cc @bkolad

bkolad:
> Looks super useful - but can we get away with using the `Default` trait instead of a custom trait? I think we can blanket implement `Default` for `T where T: ModuleInfo`.
> 
> cc @bkolad

+1 from me for using `Default`.
Can you also add a unit tests using  the `trybuild` crate. You can check how other macros are tested in `sov-modules-macros/tests`

theochap:
The default trait does not have a new method so I don't really see a way to avoid using a custom trait (which I can name Default, maybe that's actually better). I could also use the default method from the default trait but it doesn't seem to be a good solution (at least to me)

theochap:
So I fixed the tests and used the default method of the default trait from the std lib. Feel free to review the changes @bkolad

preston-evans98:
I think we can do this without a custom macro by blanket implementing `Default` in terms of `ModuleInfo` and then just deriving `Default` on the runtime.

```rust
impl<T> Default for T 
   where: ModuleInfo { 
  fn default() -> Self {
    <Self as ModuleInfo>::new()
  } 
}

```

theochap:
@preston-evans98  so it seems that is not possible if you use an externally defined trait:
https://stackoverflow.com/questions/58417250/implementing-a-foreign-trait-for-a-local-generic-type

bkolad:
So if we do this:
1. Remove the `new` method from `ModuleInfo`
2. Change `ModuleInfo` macro to derive `Default` trait instead of implementing the `new` method.
3. We do `derive(Default)` on Runtime without any other custom. macro.

module-system/sov-modules-macros/src/dispatch/default.rs
```diff
@@ -0,0 +1,62 @@
...
+            generics,
+            ..
+        } = input;
+
+        let (impl_generics, type_generics, where_clause) = generics.split_for_impl();
+
+        let fields = self.field_extractor.get_fields_from_struct(&data)?;
+        let runtime_fn_body = Self::make_default_fn_body(&fields);
+
+        // Implements the Default Config trait
```
bkolad, line 32: This comment is out of date.

module-system/sov-modules-macros/src/dispatch/default.rs
```diff
@@ -0,0 +1,62 @@
...
+        Ok(quote::quote! {
+        impl #impl_generics Default for #ident #type_generics #where_clause {
```
bkolad, line 34: We need a fully path here `std::default::Default`.

bkolad:
We spent some time with @theochap on this together, and we did the following changes:
1. Added a trait bound `ModuleInfo: Default`
2. Removed `new` from `ModuleInfo`
3. `derive(ModuleInfo)` derives both the `ModuleInfo` and `Default` traits.

We hoped that these changes would allow us to automatically derive `Default` for `Runtime`. Unfortunately, the `Default` macro from the standard library adds additional constraints to `C: Context`

```Rust
#[derive(Default)
pub struct Runtime<C: Context> {
}

creates:

impl <C: Context + Default> Default for Runtime<C> {

}
```

This `Context + Default` prohibits us from using it since our `C: Context` does not implement `Default`. This limitation is artificial since, in our case we can implement `Default` even if the `Context` does not implement it. This forced us to crate our custom `Default` macro which does the same thing as the standard one but without requiring that generics in the struct also have to immpl `Default.`

examples/demo-rollup/rollup_config.toml
```diff
@@ -4,7 +4,7 @@ start_height = 671431
 
 [da]
 # The JWT used to authenticate with the celestia light client. Instructions for generating this token can be found in the README
-celestia_rpc_auth_token = "MY.SECRET.TOKEN"
+celestia_rpc_auth_token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJwdWJsaWMiLCJyZWFkIiwid3JpdGUiLCJhZG1pbiJdfQ._gj_vbeV9QYWhZZpaHHyhbuUwucBiM0ZQK1S6eNpdjI"
```
preston-evans98, line 7: Looks like this got committed by mistake
```suggestion
celestia_rpc_auth_token = "MY.SECRET.TOKEN"
```

preston-evans98 (changes requested):

module-system/sov-modules-macros/src/dispatch/derive_rpc.rs
```diff
@@ -149,7 +149,7 @@ impl RpcImplBlock {
                     .filter(|arg| arg.to_string() != quote! { self }.to_string());
                 quote! {
                     #signature {
-                        <#type_name <#(#generics_params)*,> as ::sov_modules_api::ModuleInfo>::new().#method_name(#(#arg_values),*)
+                        <#type_name <#(#generics_params)*,> as std::default::Default>::default().#method_name(#(#arg_values),*)
```
citizen-stig, line 152: Shouldn't it be `::std::default::Default` so it won't collide if someone has module `std` ?

@bkolad I am not sure about this one.

module-system/sov-modules-macros/src/dispatch/default_runtime.rs
```diff
@@ -0,0 +1,62 @@
...
+        } = input;
+
+        let (impl_generics, type_generics, where_clause) = generics.split_for_impl();
+
+        let fields = self.field_extractor.get_fields_from_struct(&data)?;
+        let runtime_fn_body = Self::make_default_runtime_fn_body(&fields);
+
+        // Implements the Default Runtime Config trait
+        Ok(quote::quote! {
+        impl #impl_generics Default for #ident #type_generics #where_clause {
```
citizen-stig, line 34: Same not about fully qualified name, shouldn't it be
```suggestion
        impl #impl_generics ::std::default::Default for #ident #type_generics #where_clause {
```

module-system/sov-modules-macros/src/dispatch/default_runtime.rs
```diff
@@ -0,0 +1,62 @@
...
+            fn default() -> Self {
+                use sov_modules_api::ModuleInfo;
```
citizen-stig, line 36: And same here:
```suggestion
                use ::sov_modules_api::ModuleInfo;
```

module-system/sov-modules-macros/src/dispatch/default.rs
citizen-stig: I belive it should be even `::std...`

module-system/sov-modules-macros/src/module_info.rs
```diff
@@ -306,8 +313,8 @@ fn make_init_address(
             use sov_modules_api::Hasher;
             let module_path = module_path!();
             let prefix = sov_modules_api::Prefix::new_module(module_path, stringify!(#struct_ident));
-            let #field_ident =
-                <Self::Context as sov_modules_api::Spec>::Address::try_from(&prefix.hash:: <Self::Context> ())
+            let #field_ident : <#generic_param as sov_modules_api::Spec>::Address =
+                <#generic_param as sov_modules_api::Spec>::Address::try_from(&prefix.hash::<#generic_param>())
```
citizen-stig, line 317:
```suggestion
                <#generic_param as ::sov_modules_api::Spec>::Address::try_from(&prefix.hash::<#generic_param>())
```

module-system/sov-modules-macros/src/module_info.rs
```diff
@@ -283,14 +289,15 @@ fn make_init_module(field: &StructNamedField) -> Result<proc_macro2::TokenStream
     let ty = &field.ty;
 
     Ok(quote::quote! {
-        let #field_ident = <#ty as sov_modules_api::ModuleInfo>::new();
+        let #field_ident = <#ty as std::default::Default>::default();
```
citizen-stig, line 292:
```suggestion
        let #field_ident = <#ty as ::std::default::Default>::default();
```

module-system/sov-modules-macros/src/module_info.rs
```diff
@@ -125,18 +130,19 @@ impl<'a> StructDef<'a> {
 
         Ok(quote::quote! {
             use sov_modules_api::AddressTrait;
+            impl #impl_generics Default for #ident #type_generics #where_clause{
```
citizen-stig, line 133:
```suggestion
            impl #impl_generics ::std::default::Default for #ident #type_generics #where_clause{
```

module-system/sov-modules-macros/src/dispatch/derive_rpc.rs
citizen-stig, line 152:
```suggestion
                        <#type_name <#(#generics_params)*,> as ::std::default::Default>::default().#method_name(#(#arg_values),*)
```

module-system/sov-modules-macros/src/module_info.rs
```diff
@@ -125,18 +130,19 @@ impl<'a> StructDef<'a> {
 
         Ok(quote::quote! {
             use sov_modules_api::AddressTrait;
```
citizen-stig, line 132: It wasn't touched in this PR, but still:

```suggestion
            use ::sov_modules_api::AddressTrait;
```

module-system/sov-modules-macros/src/dispatch/default.rs
bkolad: I caught up with @citizen-stig. This is a good suggestion. I was not aware of this trick in macros before.

citizen-stig (approved):

preston-evans98 (approved):

Issue #358, opened 2023-05-30
Extend Tx Generator to be fully generic


preston-evans98:
Currently, we can generate simple transfers for our bank module using the `bank-cmd` utility. Since generating transactions is one of the most common things that developers will want to do, we should extend this tool to handle modules generically.

cc @bkolad, IIRC you had thoughts on how to do this

bkolad:
The `bank-cmd` is quite close to being generic, but we need to extend it.
I propose the following solution: 
1. Develop a macro that operates over the `Runtime` and parses all its fields. This macro would return a tuple `(ModuleName, Module::CallMessage)` for each module.
2. The macro should generate clap CLI options with the following functionality:
    - The command would be structured as `cmd module-name serialize-call path_to_priv_key.json path_to_call_message.json`
    - The tool will read the `Module::CallMessage` from JSON. This is possible because all call messages in all modules will have serde derive.
3. With the strongly typed `CallMessage` obtained, it can be serialized to bytes using `Runtime::<C>::encode_*module_name*_call(call_msg)`
4. Subsequently, the upstream code for all modules would be identical: save the bytes to a file.

Once the above functionality is implemented, we can enhance the tool further:
1. Currently, the `.dat` files are saved in the same directory as the `.json` files. However, we should allow the path to be specified as a parameter.
2. The files should be saved in a hex-encoded format, eliminating the need for an additional step before submitting to Celestia.

PR #371, opened 2023-05-31
draft commit

dubbelosix:
https://github.com/Sovereign-Labs/sovereign-sdk/issues/358

### DIFF
--- a/examples/demo-stf/src/runtime.rs
+++ b/examples/demo-stf/src/runtime.rs
@@ -1,5 +1,7 @@
 use sov_modules_api::Context;
-use sov_modules_macros::{DispatchCall, Genesis, MessageCodec};
+use sov_modules_macros::{DispatchCall, Genesis, MessageCodec, cmd};
+#[cfg(feature = "native")]
+pub use sov_modules_api::default_context::DefaultContext;
 
 /// The Rollup entrypoint.
 ///
@@ -36,6 +38,7 @@ use sov_modules_macros::{DispatchCall, Genesis, MessageCodec};
 /// Similar mechanism works for queries with the difference that queries are submitted by users directly to the rollup node
 /// instead of going through the DA layer.
 
+#[cfg_attr(feature = "native", cmd(DefaultContext))]
 #[derive(Genesis, DispatchCall, MessageCodec)]
 #[serialization(borsh::BorshDeserialize, borsh::BorshSerialize)]
 pub struct Runtime<C: Context> {
--- a/module-system/module-implementations/examples/sov-election/src/call.rs
+++ b/module-system/module-implementations/examples/sov-election/src/call.rs
@@ -8,7 +8,12 @@ use sov_modules_api::{CallResponse, Context};
 use sov_state::WorkingSet;
 
 /// Call actions supported byte the module.
-#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq)]
+#[cfg_attr(
+feature = "native",
+derive(serde::Serialize),
+derive(serde::Deserialize)
+)]
+#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq, Clone)]
 pub enum CallMessage<C: Context> {
     SetCandidates { names: Vec<String> },
     AddVoter(C::Address),
--- a/module-system/sov-modules-macros/src/lib.rs
+++ b/module-system/sov-modules-macros/src/lib.rs
@@ -5,7 +5,7 @@ use dispatch::{
     dispatch_call::DispatchCallMacro, genesis::GenesisMacro, message_codec::MessageCodec,
 };
 use proc_macro::TokenStream;
-use syn::parse_macro_input;
+use syn::{parse_macro_input, Type};
 
 /// Derives the `sov-modules-api::ModuleInfo` implementation for the underlying type.
 ///
@@ -152,5 +152,14 @@ pub fn expose_rpc(attr: TokenStream, item: TokenStream) -> TokenStream {
     let input = parse_macro_input!(item as syn::ItemImpl);
     handle_macro_error(
         dispatch::derive_rpc::rpc_outer_impls(attr.into(), input).map(|ok| ok.into()),
+    )
+}
+
+#[proc_macro_attribute]
+pub fn cmd(attr: TokenStream, input: TokenStream) -> TokenStream {
+    let context_type = parse_macro_input!(attr);
+    let input = parse_macro_input!(input);
+    handle_macro_error(
+        dispatch::cmd::build_cmd_parser(input, context_type).map(|ok| ok.into()),
     )
 }
--- a/module-system/module-implementations/examples/sov-value-setter/src/call.rs
+++ b/module-system/module-implementations/examples/sov-value-setter/src/call.rs
@@ -7,7 +7,12 @@ use thiserror::Error;
 use super::ValueSetter;
 
 /// This enumeration represents the available call messages for interacting with the `sov-value-setter` module.
-#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq)]
+#[cfg_attr(
+feature = "native",
+derive(serde::Serialize),
+derive(serde::Deserialize)
+)]
+#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq, Clone)]
 pub enum CallMessage {
     SetValue(u32),
 }
--- a/module-system/module-implementations/sov-accounts/src/call.rs
+++ b/module-system/module-implementations/sov-accounts/src/call.rs
@@ -6,7 +6,7 @@ use sov_state::WorkingSet;
 
 pub const UPDATE_ACCOUNT_MSG: [u8; 32] = [1; 32];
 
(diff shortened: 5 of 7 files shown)


module-system/sov-modules-macros/src/dispatch/cmd.rs
```diff
@@ -0,0 +1,102 @@
...
+        generics,
+        data,
+    } = input.clone(); // make a copy to use later
+
+    let data = match data {
+        Data::Struct(data) => data,
+        _ => return Err(syn::Error::new_spanned(ident, "expected a struct")),
+    };
+
+    let fields = match data.fields {
```
bkolad, line 21: We already have logic for field extraction in `common.rs` can we reuse it here?

examples/demo-stf/src/sov-cli/main.rs
```diff
@@ -0,0 +1,200 @@
...
+
+#[derive(Args)]
+struct UtilArgs {
+    #[command(subcommand)]
+    command: UtilCommands,
+}
+
+#[derive(Subcommand)]
+enum UtilCommands {
+    DeriveTokenAddress {
```
bkolad, line 47: The same for `/// comments` here. Let's add some more info here and check if they are available with `--help`

examples/demo-stf/src/sov-cli/main.rs
```diff
@@ -0,0 +1,200 @@
+use anyhow::Context;
```
bkolad, line 1: We need a `README.md` for this tool. We had one for `bank-cmd` but this one here will have to be a bit more generic.

examples/demo-stf/src/sov-cli/main.rs
```diff
@@ -0,0 +1,200 @@
...
+                SerializedTx::new(&sender_priv_key_path, &module_name, &call_data_path, nonce)
+                    .unwrap_or_else(|e| panic!("Call message serialization error: {}", e));
+
+            let mut bin_path = PathBuf::from(call_data_path);
+            bin_path.set_extension("dat");
+
+            let mut file = File::create(bin_path)
+                .unwrap_or_else(|e| panic!("Unable to crate .dat file: {}", e));
+
+            file.write_all(&vec![serialized.raw.data].try_to_vec().unwrap())
```
bkolad, line 169: We are writing the `borsh` data, should we convert it to hex? Can be done in a separate PR. What do you think @preston-evans98?

module-system/sov-modules-macros/src/dispatch/cmd.rs
```diff
@@ -0,0 +1,102 @@
...
+            let type_path = match &field.ty {
+                Type::Path(type_path) => {
+                    let mut segments = type_path.path.segments.clone();
+                    if let Some(PathSegment { arguments, .. }) = segments.last_mut() {
+                        if let PathArguments::AngleBracketed(_) = arguments {
+                            // Replace with an empty AngleBracketedGenericArguments
+                            *arguments = PathArguments::None;
+                        }
+                    }
+                    Path { segments, ..type_path.path.clone() }
```
bkolad, line 55: This logic can be simplified with
```Rust
let mut segments = path.segments.clone();
let last = segments.last_mut().expect("Impossible happened! A type path has no segments");
last.arguments = PathArguments::None;
segments
```

examples/demo-stf/src/sov-cli/main.rs
dubbelosix: I think we can make output format an option so they can specify bin, hex.. celestia-appd (and some other tools) expect it in hex, so that would definitely be convenient

examples/demo-stf/src/sov-cli/main.rs
```diff
@@ -0,0 +1,223 @@
...
```
preston-evans98, line 192: Separately, it looks like we're wrapping the transaction in an outer `Vec` here, which turns it into a complete `Blob`, not just a transaction. That's ok for now (it's convenient being able to send the output of this tool directly to Celestia), but we should think about how to represent the difference between a single Tx and a Blob in future

module-system/sov-modules-macros/src/lib.rs
```diff
@@ -154,3 +155,20 @@ pub fn expose_rpc(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    // let input = parse_macro_input!(input);
+    let cmd_macro = CmdMacro::new("Cmd");
+
+    handle_macro_error(
+        cmd_macro
+            .derive_cmd(input, context_type)
+            .map(|ok| ok.into()),
+    )
+
+    // handle_macro_error(dispatch::cmd::build_cmd_parser(input, context_type).map(|ok| ok.into()))
```
preston-evans98, line 173: nit: Let's remove this commented code

examples/demo-stf/src/runtime.rs
```diff
@@ -36,6 +38,7 @@ use sov_modules_macros::{DispatchCall, Genesis, MessageCodec};
 /// Similar mechanism works for queries with the difference that queries are submitted by users directly to the rollup node
 /// instead of going through the DA layer.
 
+#[cfg_attr(feature = "native", cmd(DefaultContext))]
```
preston-evans98, line 41: Can we rename the cmd macro to something that makes its functionality more clear?

module-system/module-implementations/examples/sov-election/src/call.rs
```diff
@@ -8,7 +8,12 @@ use sov_modules_api::{CallResponse, Context};
 use sov_state::WorkingSet;
 
 /// Call actions supported byte the module.
-#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq)]
+#[cfg_attr(
+    feature = "native",
+    derive(serde::Serialize),
+    derive(serde::Deserialize)
+)]
```
preston-evans98, line 15: This annotation seems really easy to forget. Let's add a bound on the `CallMessage` associated type of the `Module` trait

examples/demo-stf/src/sov-cli/main.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// Serialize a call to a module.
+    /// This creates a dat file containing the serialized transaction
+    SerializeCall {
+        /// Path to the json file containing the private key of the sender
+        sender_priv_key_path: String,
+        /// Name of the module to generate the call.
+        /// Modules defined in your Runtime are supported.
+        /// (eg: Bank, Accounts)
+        module_name: String,
+        /// Path to the json file containing the parameters for a module call
```
preston-evans98, line 42: How do these json files get generated? Is there any way we can do that automatically as well?

preston-evans98:
This looks pretty nice, but it would be great to have some test coverage. Maybe we can add an integration test which generates transactions using the tool and verifies that they decode properly in the runtime?

examples/demo-stf/src/sov-cli/main.rs
dubbelosix: yeah, we can def separate them with a flag - to prep a celestia ready blob or to just generate a single borsh serialized call - i guess for the borsh serialized call, the user takes on the burden of serializing it into a blob (or we could also additionally have another command that takes a bunch of serialized calls and generates a celestia-ready blob)

examples/demo-stf/src/sov-cli/main.rs
dubbelosix: These json files would be generated manually based on the module the user wants to call. for example:
```
{
    "CreateToken": {
      "salt": 11,
      "token_name": "sov-test-token",
      "initial_balance": 1000,
      "minter_address": "...",
      "authorized_minters": ["..."]
    }
}
```
are you talking about perhaps letting the entire call be clapped so that a user can do something like:
-- module-call create-token 11 sov-test-token --initial-balance=1000 
or something similar?

citizen-stig:
@dubbelosix minor note, could you please use reference to issue with some wording like `Resolves #X` then GitHub automatically link issue and it will be closed when PR is merged.

Also It might be slighly more readable when full description is inside PR description and title is just summary of what it does.

module-system/module-implementations/examples/sov-election/src/call.rs
dubbelosix: just discussed this with @bkolad there are some unintended consequences where the compiler complains and requires even Context/C to have the serde bounds. this is defintiely fixable, but the change seems a bit more involved, so created another issue to track it
[cli_parser](https://github.com/Sovereign-Labs/sovereign-sdk/issues/378)

examples/demo-stf/src/sov-cli/main.rs
dubbelosix: https://github.com/Sovereign-Labs/sovereign-sdk/issues/379
added another issue to track this. could also use some input on the right kinds of options that we want for the cli 
@bkolad

examples/demo-stf/src/sov-cli/main.rs
dubbelosix: one good thing if we decide to go this route is that we wouldn't need json deser to be implemented for CallMessage. @bkolad ? could explore this in a different issue

bkolad: I think it is ok to do it in different PR

dubbelosix:
Resolved #358

bkolad (approved):

Issue #237, opened 2023-05-08
Make the state machine in sov-app-template explicit


preston-evans98:
Currently, the `apply_blob` function has an implicit state machine. At the start of each blob, we make the `batch_workspace` revertable, and then gradually begin applying transactions to it. Since the state machine is implicit, the type checker does not enforce that we leave the batch workspace in a coherent state - so it's easy to forget to commit or revert changes on unhappy code paths. We should make the state machine explicit, so that invalid transitions are rejected at compile time.

PR #384, opened 2023-06-05
Modules: WorkingSet refactoring.

bkolad:
# Description
This PR  splits `WorkingSet` enum into two types `WorkingSet & CommittedWorkingSet`.  This has the following advantages:
1. The `AppTemplate` now stores `CommittedWorkingSet`, ensuring that only the state that has been committed in the `apply_batch` method is stored in the database during the `end_slot` phase

2. The `commit & revert` methods are exclusive to the `WorkingSet` type and return a `CommittedWorkingSet` instance. Additionally, only the `CommittedWorkingSet` type has `to_revertable()` method. This explicit state transition between `WorkingSet` and `CommittedWorkingSet` prevents the accidental reversion of a `CommittedWorkingSet` or the storage of a `WorkingSet` in the database without committing (or reverting) it beforehand.

## Linked Issues
- Related to #237

## Testing
All relevant tests ware updated.
## Docs
Updated docs for `WorkingSet & CommittedWorkingSet`


module-system/sov-state/src/scratchpad.rs
```diff
@@ -40,90 +40,75 @@ impl<S: Storage> Debug for RevertableDelta<S> {
     }
 }
 
-/// A read-write set which can be committed as a unit
-enum ReadWriteSet<S: Storage> {
-    Standard(Delta<S>),
-    Revertable(RevertableDelta<S>),
+/// This structure stores the read-write set, which will be persisted in the database at the
```
citizen-stig, line 43: Could you please extend purpose of these structures, in a way that you did in issue description.
Like when each of those should be used, etc

citizen-stig (approved):

preston-evans98 (approved):
This change looks great overall, but I'm wondering if we can rename some things to clarify roles. (This isn't blocking, since the problems pre-date this PR). 

Currently, the workflow to commit some changes to the DB is `storage.validate_and_commit(working_set.commit().freeze())`, which seems really cumbersome.

Maybe we can rename `CommittedWorkingSet` to `StateCheckpoint` or something. And then `commit()` could get renamed to `checkpoint()`. As for `freeze` - maybe we could rename to `split_for_storage()` or something? 

If anyone has suggestions for renaming any of these items, leave them below! cc @citizen-stig @bkolad @dubbelosix

bkolad:
> This change looks great overall, but I'm wondering if we can rename some things to clarify roles. (This isn't blocking, since the problems pre-date this PR).
> 
> Currently, the workflow to commit some changes to the DB is `storage.validate_and_commit(working_set.commit().freeze())`, which seems really cumbersome.
> 
> Maybe we can rename `CommittedWorkingSet` to `StateCheckpoint` or something. And then `commit()` could get renamed to `checkpoint()`. As for `freeze` - maybe we could rename to `split_for_storage()` or something?
> 
> If anyone has suggestions for renaming any of these items, leave them below! cc @citizen-stig @bkolad @dubbelosix

Added an issue: https://github.com/Sovereign-Labs/sovereign-sdk/issues/388

PR #385, opened 2023-06-05
Remove vendored `temppath`

citizen-stig:
# Description
This PR suggests replacing vendored temppath in favour to [tempfile](https://docs.rs/tempfile/3.5.0/tempfile/index.html) crate.

It also removes `temp` and `temppath` features, as using of these features in does not clearly indicate when temporary directory will be deleted, which can lead either to temp folder pollution or storage corruption.

I understand that this is significant change, but having transparency in the tests has long term benefits.
With new approach, each test controls how long temporary folder lives.

Also less inner dependencies and feature flags to check

## Testing
All tests are adjusted


preston-evans98 (approved):
Nice, hopefully this will bring down compile times for a few packages.

PR #387, opened 2023-06-05
Only run Rust workflow on particular set of files

citizen-stig:
# Description

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,9 +2,17 @@ name: Rust
 
 on:
   push:
-    branches: ["main"]
+    branches: [ "main" ]
+    paths:
+      - '**.rs'
+      - '**.toml'
+      - '**.json'
   pull_request:
-    branches: ["main"]
+    branches: [ "main" ]
+    paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -2,9 +2,17 @@ name: Rust
...
+      - '**.rs'
+      - '**.toml'
+      - '**.json'
   pull_request:
-    branches: ["main"]
+    branches: [ "main" ]
+    paths:
+      - '**.rs'
+      - '**.toml'
+      - '**.json'
```
preston-evans98, line 15: Is there any chance of missing important files down the line with this change? If so, maybe we can exclude paths instead (`*.md`) if Github Actions allows it

citizen-stig: Good point, testing

bkolad (approved):

preston-evans98 (approved):
Looks great to me @citizen-stig! Thanks!

citizen-stig:
It looks like GitHub evaluates changes on PR level, not on commit level.

Issue #388, opened 2023-06-05
Better naming in `sov-state`


bkolad:
See: https://github.com/Sovereign-Labs/sovereign-sdk/pull/384#pullrequestreview-1463055800

```
This change looks great overall, but I'm wondering if we can rename some things to clarify roles. 
(This isn't blocking, since the problems pre-date this PR).
Currently, the workflow to commit some changes to the DB is s
torage.validate_and_commit(working_set.commit().freeze()), which seems really cumbersome.

Maybe we can rename CommittedWorkingSet to StateCheckpoint or something. And then commit() could get renamed to checkpoint(). As for freeze - maybe we could rename to split_for_storage() or something?
```

PR #390, opened 2023-06-06
`AppTemplate::apply_batch()`  refactoring

### LINKED ISSUE #237: Make the state machine in sov-app-template explicit

bkolad:
# Description

1. The  `CommittedWorkingSet` has been changed to `StateCheckpoint`.
2. The `AppTemplate` has been moved to its own file.
3. Some helper methods have been added to `AppTemplate`, and `apply_batch()` has been made more modular.
The helper methods now return the `Result<_, ApplyBatchError>` type, which allows us to use the `?` operator and make the shortcutting logic in `apply_batch()` more readable.

## Linked Issues
- Fixes #237, #388

## Testing
No impact on tests.
## Docs
Relevant docs updated


module-system/sov-modules-stf-template/clippy.toml
```diff
@@ -0,0 +1 @@
+type-complexity-threshold  = 300
```
bkolad, line 1: `clippy` was not happy with this signature (return type):
```

    fn execute_txs(
        &mut self,
        mut batch_workspace: WorkingSet<C::Storage>,
        txs: Vec<TransactionAndRawHash<C>>,
        batch_data_and_hash: &BatchDataAndHash,
    ) -> Result<(WorkingSet<C::Storage>, Vec<TransactionReceipt<TxEffect>>)> {
...
}
```
But I don't thing it is that bad so I increased the `type-complexity-threshold`

module-system/sov-modules-stf-template/src/app_template.rs
```diff
@@ -0,0 +1,281 @@
...
+type Result<T> = std::result::Result<T, ApplyBatchError>;
+
+pub struct AppTemplate<C: Context, RT, Vm> {
+    pub current_storage: C::Storage,
+    pub runtime: RT,
+    pub(crate) checkpoint: Option<StateCheckpoint<C::Storage>>,
+    phantom_vm: PhantomData<Vm>,
+}
+
+pub(crate) enum ApplyBatchError {
```
citizen-stig, line 28: Could you please add public doc comment explaining what kind of hash is in both cases?

module-system/sov-modules-stf-template/src/app_template.rs
```diff
@@ -0,0 +1,281 @@
...
+                    let ctx = C::new(sender_address.clone());
+                    let tx_result = self.runtime.dispatch_call(msg, &mut batch_workspace, &ctx);
+
+                    self.runtime
+                        .post_dispatch_tx_hook(&tx.tx, &mut batch_workspace)
+                        .expect("Impossible happened: error in post_dispatch_tx_hook");
+
+                    let tx_effect = match tx_result {
+                        Ok(_) => TxEffect::Successful,
+                        Err(_e) => {
```
citizen-stig, line 175: Maybe let's log the error?

module-system/sov-modules-stf-template/src/app_template.rs
bkolad: added

module-system/sov-modules-stf-template/src/app_template.rs
bkolad: I put `debug` it is ok for a single tx to fail here.

preston-evans98 (approved):

PR #392, opened 2023-06-06
Clippy fixes after alpha release

citizen-stig:
# Description

In order to unblock developers to use `cargo clippy` current issues needs to be addressed. This PR solves that.
After it is merged, everybody can use `make lint` which will do `cargo check` and `cargo clippy` for all targets and features.

As for now, clippy checks are optional and not part of CI

## Testing
No tests should be affected

## Docs
No changes in documentatio is needed


bkolad (approved):

preston-evans98 (approved):
🚀

PR #394, opened 2023-06-06
Tests for ledger rpc

theochap:
# Description
This PR implements some simple tests to check the behavior of the rpc methods used with ledger db

## Linked Issues
- Related to #342 

## Testing
These tests reproduce the README workflow, ie:
- It creates a ledger db and populates it with transactions
- It launches a rpc server
- It executes queries and checks the content of the answer.


examples/demo-rollup/Cargo.toml
```diff
@@ -10,6 +10,8 @@ publish = false
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+blake3 = "1.3"
+curl = "0.4"
```
preston-evans98, line 14: These should probably be dev-dependencies if we're only using them for tests. That will reduce the binary size and help mitigate the security risks if one of them is buggy.

examples/demo-rollup/rollup_config.toml
```diff
@@ -4,7 +4,7 @@ start_height = 671431
 
 [da]
 # The JWT used to authenticate with the celestia light client. Instructions for generating this token can be found in the README
-celestia_rpc_auth_token = "MY.SECRET.TOKEN"
+celestia_rpc_auth_token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJwdWJsaWMiLCJyZWFkIiwid3JpdGUiLCJhZG1pbiJdfQ._gj_vbeV9QYWhZZpaHHyhbuUwucBiM0ZQK1S6eNpdjI"
```
preston-evans98, line 7:
```suggestion
celestia_rpc_auth_token = "MY.SECRET.TOKEN"
```

examples/demo-rollup/src/main.rs
```diff
@@ -1,5 +1,6 @@
 mod config;
 mod ledger_rpc;
+mod test_rpc;
```
preston-evans98, line 3: Lets feature-gate this so that it only compiles during testing
```suggestion
#[cfg(test)]
mod test_rpc;
```

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "rust-analyzer.linkedProjects": ["./examples/demo-rollup/Cargo.toml"]
+}
```
preston-evans98: In theory, we shouldn't need this config to have rust-analyzer pick up the project. Did you encounter issues without it?

theochap: Deleted the file

examples/demo-rollup/src/test_rpc.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    assert_eq!(easy.response_code().unwrap(), 200);
+    let contents = easy.get_ref();
+    assert_eq!(String::from_utf8_lossy(&contents.0), expected);
+}
+
+#[derive(Serialize, Deserialize, PartialEq, core::fmt::Debug, Clone)]
+struct TestBlockHeader {
+    prev_hash: [u8; 32],
+}
```
preston-evans98, line 41: This looks like it might be useful for other crates as well. Should we move it into the `rollup-interface` crate under the `mocks` feature?

examples/demo-rollup/src/test_rpc.rs
```diff
@@ -0,0 +1,205 @@
...
+
+impl CanonicalHash for TestBlockHeader {
+    type Output = [u8; 32];
+
+    fn hash(&self) -> Self::Output {
+        self.prev_hash
```
preston-evans98, line 47: This implementation seems likely it might cause confusion, because `self.hash()` and `self.prev_hash()` return the same thing - but in a normal blockchain `self.hash()` is computed as `blake3(self.prev_hash() || self.some_other_field) || ... )`. 

I think changing it to `blake3(self.prev_hash)` would alleviate any potential confusion.

examples/demo-rollup/src/test_rpc.rs
theochap: Sure, doing it atm

examples/demo-rollup/src/test_rpc.rs
```diff
@@ -0,0 +1,205 @@
...
+        let expected = r#"{"jsonrpc":"2.0","result":[{"key":[101,118,101,110,116,49,95,107,101,121],"value":[101,118,101,110,116,49,95,118,97,108,117,101]}],"id":1}"#;
+        query_test_helper(data, expected);
+
+        let data = r#"{"jsonrpc":"2.0","method":"ledger_getEvents","params":[2],"id":1}"#.as_bytes();
+        let expected = r#"{"jsonrpc":"2.0","result":[{"key":[101,118,101,110,116,50,95,107,101,121],"value":[101,118,101,110,116,50,95,118,97,108,117,101]}],"id":1}"#;
+        query_test_helper(data, expected);
+
+        let data = r#"{"jsonrpc":"2.0","method":"ledger_getEvents","params":[3],"id":1}"#.as_bytes();
+        let expected = r#"{"jsonrpc":"2.0","result":[null],"id":1}"#;
+        query_test_helper(data, expected);
```
preston-evans98, line 199: nit: It's usually considered good style to have separate `#[test]`s for each logical unit of code - in this case, each RPC method. That way, you get a nice printout showing exactly what's broken when the test fails instead of having to do extra analysis to figure out which endpoint is broken. 

It will cost us a little bit of efficiency, since we'll have to set up and tear down the tokio::runtime with each test, but these tests are still so lightweight that it shouldn't matter.

examples/demo-rollup/Cargo.toml
```diff
@@ -43,3 +43,5 @@ const-rollup-config = { path = "../const-rollup-config" }
 
 [dev-dependencies]
 tempfile = "3.5.0"
+blake3 = "1.3"
```
preston-evans98, line 46: nit: We've been using sha2 for testing in other places. Not a big deal at all, but it would be nice to stay consistent for now.

examples/demo-rollup/src/test_rpc.rs
```diff
@@ -0,0 +1,167 @@
+use std::net::SocketAddr;
+
+use curl::easy::{Easy2, Handler, List, WriteError};
```
bkolad, line 3: Is there a particular reason why you went with `curl`? `reqwest` is much higher level framework and "standard" in rust?

examples/demo-rollup/src/test_rpc.rs
theochap: I divided this test into 4 subtests: one for each rpc function

bkolad (approved):

Issue #316, opened 2023-05-23
Create solution for maintaining static configuration.


citizen-stig:
Previously mentioned in 

#240 

> The DaService trait has a RuntimeConfig type for the dynamic configuration, and the DaSpec has a ChainParams type for the compile-time configuration. Similarly, the StateTransitionRunner has a RuntimeConfig for dynamic configuration and an InitialState type which should be specified at compile time.

citizen-stig:
We use `constants.toml` for this

PR #395, opened 2023-06-07
Adding a local docker setup to run a single node celestia instance

### LINKED ISSUE #383: use a local single node (or containerized) deployment for celestia for demo-rollup in addition to the celestia devnet
currently, blobs are submitted to arabica-6 and and read from the network by demo-rollup. this has a few problems
1. recently arabica-6 is being sunset in favor of arabica-8 (which has some compatibility issues). this can be an on-going issue and cause developer pain
2. syncing the chain upto a certain block takes time for the light client
3. remote devnet sometimes has issues and is in-accessible
4. requires a live connection for developers (not that much of an issue, but developers do prefer having the environment self contained and not require live sync)
5. a somewhat complicated setup process (keys, auth tokens, node versions, downloading go etc)

in order to solve the above, it might be a good idea to also support a containerized / single node deployment of celestia that is local to the developers machine and the demo-rollup is already configured with keys, tokens and other parameters (eg: height) baked into the container. 

since the DA layer interaction is abstracted, we could actually freeze the local development DA layer 0.7.1 (since we can theoretically think of it as a "different" DA layer not unlike avail, eigenDA etc...our abstraction should be able to support this).

Looked at using a newly built docker image for celestia-node 0.7.1 (which jupiter is currently compatible with). we could also support upgrading it (since the end goal is to deploy on celestia), but we would control the upgrade so the dev experience is more seamless - and they don't end up in a situation where the network is upgraded and their devex or demo is impacted due to compatibility issues.

dubbelosix:
create a Makefile in demo-rollup that supports setting up a celestia node locally.
resolves #383 
* user would need to install docker
* make start
* make test-create-token

the goal is to ensure that a developer can quickly have something running end to end with minimal effort and then dig in later into the individual components

### DIFF
--- a/examples/demo-rollup/localsetup.md
+++ b/examples/demo-rollup/localsetup.md
@@ -2,7 +2,7 @@
 
 * Install docker https://www.docker.com
 * switch to the `demo-rollup` directory
-* Start the celestia services locally
+* Start the celestia services locally (the details of what the Makefile does are explained in the next section)
 ```
 make clean
 make start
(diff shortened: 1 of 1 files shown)


examples/demo-rollup/Makefile
```diff
@@ -0,0 +1,100 @@
...
+	done
+
+config: check-container-running 
+ifeq ($(shell uname -s),Darwin)
+	@sed -i '' 's/\(pub const SEQUENCER_DA_ADDRESS: \[u8; 47\] = \*\)b"[^"]*";/\1b"$(get_address)";/' ../const-rollup-config/src/lib.rs
+	@sed -i '' 's/^\(celestia_rpc_auth_token = \)"[^"]*"/\1"$(get_auth)"/' rollup_config.toml
+	@sed -i '' 's#^\(celestia_rpc_address = \)"[^"]*"#\1"http://127.0.0.1:$(RPC_PORT)"#' rollup_config.toml
+	@sed -i '' 's#^\(start_height = \)[0-9]*#\1$(START_HEIGHT)#' rollup_config.toml
+else
+	@sed -i 's/\(pub const SEQUENCER_DA_ADDRESS: \[u8; 47\] = \*\)b"[^"]*";/\1b"$(get_address)";/' ../const-rollup-config/src/lib.rs
```
citizen-stig, line 68: I am fine with modifying this rust file from `make`, because it is another problem:

https://github.com/Sovereign-Labs/sovereign-sdk/issues/316

You can add it as comment there, if somebode else might be confused.

examples/demo-rollup/localsetup.md
```diff
@@ -0,0 +1,47 @@
+## Setting up SDK to run locally
+
+* Install docker https://www.docker.com
+* switch to the `demo-rollup` directory
+* Start the celestia services locally
+```
+make clean
+make start
```
preston-evans98, line 8: @dubbelosix, this looks super easy to use, but hard to understand. Can we add a thorough explanation of what's going on for each of these commands? 

Maybe something like this:

```markdown
## Setting up SDK to run locally

To start, you'll need to configure a Celestia client in your `rollup_config.toml`. The easiest way to get reliable access to a Celestia client is to spin up a local devnet - so we've provided a `makefile` to do that for you. If you want to configure your own Celestia client, simply paste its connection information into `rollup_config.toml`. Otherwise, follow the steps below to get started:


* Install docker https://www.docker.com
* switch to the `demo-rollup` directory
* Start the celestia services locally
...
```

Note - this can wait until after we've updated to `clap` if you want

examples/demo-rollup/localsetup.md
```diff
@@ -0,0 +1,47 @@
...
+2023-06-07T10:03:25.497719Z  INFO jupiter::da_service: Fetching header at height=2...
+2023-06-07T10:03:25.505412Z  INFO sov_demo_rollup: Received 0 blobs
+2023-06-07T10:03:25.505992Z  INFO sov_demo_rollup: Requesting data for height 3 and prev_state_root 0xa96745d3184e54d098982daf44923d84c358800bd22c1864734ccb978027a670
+2023-06-07T10:03:25.506003Z  INFO jupiter::da_service: Fetching header at height=3...
+2023-06-07T10:03:25.511237Z  INFO sov_demo_rollup: Received 0 blobs
+2023-06-07T10:03:25.511815Z  INFO sov_demo_rollup: Requesting data for height 4 and prev_state_root 0xa96745d3184e54d098982daf44923d84c358800bd22c1864734ccb978027a670
+```
+* Run the test transaction command, which creates a token
+```
+make test-create-token
```
preston-evans98, line 36: Same thing here. Eventually (maybe after `clap` support is added) we should try to explain *what* to do, not just how to use the `Makefile` to do it.

bkolad (approved):

examples/demo-stf/src/sov-cli/main.rs
```diff
@@ -225,6 +227,9 @@ pub fn main() {
                 PrivKeyAndAddress::generate_and_save_to_file(priv_key_path.as_ref())
                     .unwrap_or_else(|e| panic!("Create private key error: {}", e));
             }
+            UtilCommands::PrintNamespace => {
+                println!("{}",hex::encode(ROLLUP_NAMESPACE_RAW));
```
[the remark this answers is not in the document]
citizen-stig: Ooh, if it is input for something else, then it is fine.

PR #396, opened 2023-06-07
Submit batch to da layer

citizen-stig:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


examples/demo-rollup/src/main.rs
```diff
@@ -26,12 +26,41 @@ use sov_rollup_interface::stf::StateTransitionFunction;
...
+use risc0_adapter::Risc0MethodId;
 use sov_modules_api::RpcRunner;
+use sov_rollup_interface::services::batch_builder::BatchBuilder;
+use sov_rollup_interface::zk::traits::{Zkvm, ZkvmHost};
+
+/// In order to allow DemoAppRunner to be `Send`, this Zkvm stub is added
+/// It can be used to communicate with a Risc0Host over socket.
+struct RemoteZkHost {
+    _address: SocketAddr,
+}
```
preston-evans98, line 45: I think we can remove this impl entirely if we add a `DemoAppRunner` implementation for the `Standard` config. 

IIRC, we require the NativeAppRunner to have access to a ZkvmHost (meaning that it can verify proofs *and* write data to the guest). But in reality, the only requirement for the block builder is that it can verify proofs. 

To get rid of this ugly remote zk host, I think we just need to rename `NativeAppRunner` to `ProverAppRunner` and add a `StandardAppRunner` which uses just a regular `Risc0Verifier` as its Zkvm. We'll also need to implement a `Risc0Verifier`, but that should be very very simple.

bkolad (approved):

examples/demo-rollup/src/main.rs
citizen-stig: Done!

preston-evans98 (approved):

examples/demo-rollup/src/main.rs
```diff
@@ -113,33 +174,39 @@ async fn main() -> Result<(), anyhow::Error> {
...
         },
-    );
+    ));
+
+    start_batch_producing(demo_runner.clone(), da_service.clone());
+
     // For demonstration,  we also initialize the DaVerifier interface using the DaVerifier interface
     // Running the verifier is only *necessary* during proof generation not normal execution
-    let da_verifier = CelestiaVerifier::new(RollupParams {
+    let da_verifier = Arc::new(CelestiaVerifier::new(RollupParams {
```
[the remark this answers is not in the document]
citizen-stig: It is used, created blob submitted to DA Layer

PR #400, opened 2023-06-08
Allow rpc tests to choose random port

citizen-stig:
# Description

In order to faster find flaky tests `flaky-finder` allows to run tests in parallel. But RPC tests have hard coded port, so they fail and clash.

This PR fixes it.

### DIFF
--- a/examples/demo-rollup/src/test_rpc.rs
+++ b/examples/demo-rollup/src/test_rpc.rs
@@ -1,8 +1,6 @@
 use reqwest::header::CONTENT_TYPE;
 use sov_db::ledger_db::{LedgerDB, SlotCommit};
 use std::net::SocketAddr;
-use std::sync::atomic::{AtomicU16, Ordering};
-use std::sync::Arc;
 
 #[cfg(test)]
 use sov_rollup_interface::mocks::{TestBlock, TestBlockHeader};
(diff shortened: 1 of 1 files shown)


examples/demo-rollup/src/test_rpc.rs
```diff
@@ -98,18 +96,34 @@ fn test_helper(data: String, expected: &str, port: u16) {
 
         let ledger_rpc_module = ledger_rpc::get_ledger_rpc::<i32, i32>(ledger_db.clone());
 
+        let actual_port_placeholder: Arc<AtomicU16> = Arc::new(AtomicU16::new(0));
```
preston-evans98, line 99: (non-blocking, question) With an atomic, just a `&` reference should be safe to share between threads. Does the compiler complain because it's not `'static`?

examples/demo-rollup/src/test_rpc.rs
```diff
@@ -98,18 +96,34 @@ fn test_helper(data: String, expected: &str, port: u16) {
...
             rx_end.await.unwrap();
         });
 
         rx_start.await.unwrap();
 
+        for _ in 0..1000 {
+            if actual_port_placeholder.load(Ordering::Relaxed) != 0 {
+                break;
+            }
+            std::thread::sleep(std::time::Duration::from_millis(10));
```
preston-evans98, line 119: It seems like it would be simpler and faster to use a oneshot channel here. Is there a reason we need to use an atomic with manual synchronization instead?

citizen-stig: I haven't noticed performance improvement, but it is simpler. Fix pushed

Issue #298, opened 2023-05-18
Skip batch/transaction hashing in zk


preston-evans98:
The `Receipt` type requires batch and transaction hashes to be calculated, but those hashes are not consensus-critical. Currently, they're only used for indexing data in the database. So, we don't need to calculate them in the zk context. 

There are a few mechanisms we might use to skip this hashing.
1. Make the `Hash` fields of `BatchReceipt` and `TxReceipt` into `Option<[u8;32]>` and simply leave them as `None` in zk
2. Use conditional compilation to either remove the receipt types from the zk context, or remove the hash field from the receipt in the zk context
3. Add an extra hasher type to context called something like "IndexingHasher", and make it a no-op in the zk context

PR #401, opened 2023-06-08
Partial verification of blob

### LINKED ISSUE #268: Add mechanism for *partial* verification of blobs
## Background
Currently, the Data Availability Layer specs require that the DA must provide the full contents of all "relevant" blobs, along with proofs of completeness and correctness proofs for the data. But, per the spec, the criteria for "relevance" must be stateless - which means that it includes some false positives. We would like to minimize the cost of such false positives. 

As a motivating example, consider a 1 MB blob on Celestia sent by an entity who is *not* a registered sequencer. In the current specification, the Jupiter DaVerifier is required to check the entire blob was provided correctly - which involves opening the Celestia data square at about 2000 locations. That requires roughly 4000 namespaced merkle tree hash calculations or 12,000 sha256::compress invocations. In circuit, that's _very_ expensive. Ideally, we would only prove the inclusion of the first share (which contains the length of the blob) and the sender field. Proving the first share would only require a logarithmic number of merkle hashes - say 20-30 at most - which would translate into about a 100x reduction in proving cost for the invalid blob.

## Suggested Changes
Modify the `extract_relevant_txs_with_proof` in the DaService trait to accept an argument `min_lengths`
```rust
    /// Extract the relevant transactions from a block, along with a proof that the extraction has been done correctly.
    /// For each blob, the implementation *may* truncate the returned blob to the corresponding `min_length`. Any data
    /// which is not returned may also be excluded from the returned proofs. 
    ///
    /// For example, this method might return the senders of all blob transactions in rollup's namespace on Celestia,
    /// together with a range proof against the root of the namespaced-merkle-tree, demonstrating that the entire
    /// rollup namespace has been covered. However, it might *not* open the NMT to the entire contents of some blobs, if the 
    /// min_lengths array specified that such data was not needed. 
    fn extract_relevant_txs_with_proof(
        &self,
        block: Self::FilteredBlock,
        min_lengths: Vec<usize>,
    ) -> (
        Vec<<Self::Spec as DaSpec>::BlobTransaction>,
        <Self::Spec as DaSpec>::InclusionMultiProof,
        <Self::Spec as DaSpec>::CompletenessProof,
    );
```
Modify the `BlobTransaction` trait to allow the omission of some blob data.
```rust
pub enum BlobData<B: Buf> {
  Partial(B),
  Complete(B),
}

pub trait BlobTransactionTrait: Serialize + DeserializeOwned {
    type Data: Buf;
    type Address: AddressTrait;
    /// Returns the address (on the DA layer) of the entity which submitted the blob transaction
    fn sender(&self) -> Self::Address;
    /// The length of the complete blob data. 
    fn full_data_length()
    /// The data from this blob needed by the DaVerifier. This may be a prefix of the full blob.
    fn data(&self) -> BlobData<Self::Data>;
}
```

theochap:
# Description
This PR enables partial verification of blobs. It solves the issue #268 

## Meaningful changes
- The method `hash` has been added to the `BlobTransactionTrait`, it returns a [u8; 32] array, to be compliant with the internal hash representations used in the template.
- The structure `BufWithCounter` has been created, which is a wrapper around `Buf`. This wrapper maintains an internal counter which tracks the number of bytes read from the inner buffer.
- The `get_extraction_proof` function has been added, it produces proofs that the data have been correctly extracted from the block. The `extract_relevant_txs_with_proof` has been kept for legacy reasons but it should simply calls `extract_relevant_txs` followed by `get_extraction_proof`. The `get_extraction_proof` should be called after the rollup has processed the transaction so that we can track how far we've read from the blob (and hence do the partial blob verification optimization).
- This PR simplified the structure of the `apply_blob` function of the module_stf_template. Mainly it avoids reading the whole buffer before checking the headers.
- The Celestia adapters have been modified to take into account the changes.

## Linked Issues
- Fixes #268 

## Tests
The test suite has been ran, no other new tests have been created


adapters/celestia/src/verifier/mod.rs
```diff
@@ -29,12 +33,29 @@ pub const PARITY_SHARES_NAMESPACE: NamespaceId = NamespaceId(hex_literal::hex!("
...
+        }
+
+        // If there is no hash computed yet we have to calculate it using the buf data
+        let mut reader = self.data().inner().reader();
+        let mut batch_data = Vec::new();
+        reader
+            .read_to_end(&mut batch_data)
+            .unwrap_or_else(|e| panic!("Unable to read batch data {}", e));
+
+        sha2::Sha256::digest(&batch_data)
```
bkolad, line 58: Is it ok to hardcode a specific hash function here?

rollup-interface/src/state_machine/da.rs
```diff
@@ -56,14 +54,54 @@ pub trait DaVerifier {
...
+    pub fn new(inner: B) -> Self {
+        BufWithCounter { inner, counter: 0 }
+    }
+
+    // To use with care: if any Buf method is applied to inner directly the counter won't be updated
+    pub fn inner(self) -> B {
+        self.inner
+    }
+
+    pub fn counter(self) -> usize {
```
bkolad, line 74: This doesn't have to consume `self`

rollup-interface/src/state_machine/da.rs
```diff
@@ -56,14 +54,54 @@ pub trait DaVerifier {
...
+    inner: B,
+    counter: usize,
+}
+
+impl<B: Buf> BufWithCounter<B> {
+    pub fn new(inner: B) -> Self {
+        BufWithCounter { inner, counter: 0 }
+    }
+
+    // To use with care: if any Buf method is applied to inner directly the counter won't be updated
```
bkolad, line 69: This method consumes `BufWithCounter ` so there is no more counter after the method is called.

adapters/celestia/src/verifier/mod.rs
```diff
@@ -29,12 +33,29 @@ pub const PARITY_SHARES_NAMESPACE: NamespaceId = NamespaceId(hex_literal::hex!("
...
         self.sender.clone()
     }
 
-    fn data(&self) -> Self::Data {
-        self.blob.clone().into_iter()
+    fn data(&self) -> BufWithCounter<Self::Data> {
+        BufWithCounter::new(self.blob.clone().into_iter())
+    }
+
+    fn hash(&self) -> [u8; 32] {
```
bkolad, line 45: In the current API the `blob` will be cloned in a call to `hash` and `data`. Isn't always to be the same caller for these two methods? If the `data()` and `hash()` are called only once in the same place then we could unify them and save one clone.

adapters/celestia/src/verifier/mod.rs
theochap: Modified with the last PR: for Celestia set the blob hash as the blob commitment

rollup-interface/src/state_machine/da.rs
theochap: Solved in the last PR

adapters/celestia/src/verifier/proofs.rs
```diff
@@ -50,14 +51,18 @@ impl CorrectnessProof {
 
         // Extract (and clone) the position of each transaction
         for tx in relevant_txs.iter() {
-            let commitment = recreate_commitment(block.square_size(), BlobRef::with(&tx.blob.0))
-                .expect("commitment is valid");
+            // We process the transaction only if we read something from it
+            if tx.data().counter() != 0 {
```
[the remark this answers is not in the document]
theochap: Modified the data method to return a reference to the inner data member instead

rollup-interface/src/state_machine/da.rs
theochap: Solved with the last commit

adapters/celestia/src/da_service.rs
```diff
@@ -222,29 +222,49 @@ impl DaService for CelestiaService {
                 .signer
                 .clone();
 
+            let blob: Blob = blob_ref.into();
+
+            // TODO: can't we just use the commitment as a blob hash?
```
preston-evans98, line 227: Can we remove this comment, or is there something else to clean up here?

rollup-interface/src/state_machine/da.rs
```diff
@@ -56,14 +56,68 @@ pub trait DaVerifier {
...
+    }
+}
+
+impl<B: Buf> Buf for BufWithCounter<B> {
+    fn remaining(&self) -> usize {
+        self.inner.remaining()
+    }
+
+    fn chunk(&self) -> &[u8] {
+        self.inner.chunk()
```
preston-evans98, line 96: This function looks like it could cause a soundness problem - using it, an STF could read data from the buffer without advancing the counter. If the prover was malicious, they could tamper with that data and since the counter didn't advance it might not be checked by the DA verifier.

I see a couple of options to fix it:
- add an extra counter to let us track the furthest byte read in addition to the location of the cursor
- Modify the API of the STF to accept `impl Read` instead of `impl Buf` (this will cause a slight efficiency loss)
- (not recommended) Try to guarantee that a `chunk` always corresponds to a share of the underlying DA layer, so that any data returned by `chunk` will automatically get authenticated

preston-evans98 (changes requested):

bkolad (approved):

theochap:
Also, I realized now that merging this PR will very likely solve #298

examples/demo-stf/src/hooks_impl.rs
```diff
@@ -34,7 +36,7 @@ impl<C: Context> ApplyBlobHooks for Runtime<C> {
     fn begin_blob_hook(
         &self,
         sequencer: &[u8],
-        raw_blob: &[u8],
+        raw_blob: &impl Read,
```
preston-evans98, line 39: It seems odd to pass `&impl Read` when (IIRC) the `Read` trait doesn't have any methods which can be called on `&self`. Do we need this argument at all? If so, what do we use it for?

adapters/celestia/src/verifier/proofs.rs
```diff
@@ -45,24 +45,27 @@ impl CompletenessProof {
...
-            let commitment = recreate_commitment(block.square_size(), BlobRef::with(&tx.blob.0))
-                .expect("commitment is valid");
-
-            let (_, position) = block
-                .relevant_pfbs
-                .get(&commitment[..])
-                .expect("commitment must exist in map");
-            needed_tx_shares.push(position.clone());
+        for tx in blobs.iter() {
+            // We process the transaction only if we read something from it
```
preston-evans98, line 56: On closer review there might be one more soundness issue here. In our system, an STF might decide to ignore a transaction based on its sender - so even if the data is unread, we still need to check that the PFB transaction was provided correctly - we just don't need to open the blob commitment

rollup-interface/src/state_machine/da.rs
theochap: Fixed by the last commit, I went with the second approach as the extra counter wasn't feasible (the chunk method doesn't take a mutable version of self)

adapters/celestia/src/verifier/proofs.rs
theochap: After discussion with Preston I have opened the issue #412 to implement the partial blob optimization for Celestia.

examples/demo-stf/src/hooks_impl.rs
theochap: Changed to pass a BufReaderWithCounter which has an accumulator to be able to read again data from the buffer.

examples/demo-rollup/src/txs_rpc.rs
```diff
@@ -137,14 +137,14 @@ mod tests {
 
         fn extract_relevant_txs(
             &self,
-            _block: Self::FilteredBlock,
+            _block: &Self::FilteredBlock,
         ) -> Vec<<Self::Spec as DaSpec>::BlobTransaction> {
             todo!()
         }
 
         fn extract_relevant_txs_with_proof(
```
preston-evans98, line 145: nit: This function isn't really useful anymore. I'd either remove it, or add a default implementation in the trait so that users don't have to do anything to implementing it.

theochap: I added a default implementation

rollup-interface/src/state_machine/da.rs
```diff
@@ -56,14 +58,68 @@ pub trait DaVerifier {
...
+        &self.acc
+    }
+}
+
+impl<B: Buf> Read for BufReaderWithCounter<B> {
+    // Reads the inner buf into the provided buffer, and appends the data read to inner accumulator
+    fn read(&mut self, mut buf: &mut [u8]) -> std::io::Result<usize> {
+        let len_before_reading = self.inner.remaining();
+
+        buf.write_all(&self.inner.copy_to_bytes(min(buf.len(), len_before_reading)))?;
```
preston-evans98, line 95: It looks like we're doing one extra copy here?

rollup-interface/src/state_machine/da.rs
```diff
@@ -56,14 +58,68 @@ pub trait DaVerifier {
     ) -> Result<(), Self::Error>;
 }
 
+// Simple structure that implements the Read trait for a buffer and  counts the number of bytes read from the beginning
+// Useful for the partial blob reading optimization: we know for each blob how many bytes have been read from the beginning
+// Because of soundness issues we cannot implement the Buf trait because the prover could get unproved blob data using the chunk method.
+#[derive(Debug, Clone, Serialize, Deserialize, BorshDeserialize, BorshSerialize, PartialEq)]
+pub struct BufReaderWithCounter<B: Buf> {
+    inner: B,
+    acc: Vec<u8>,
```
preston-evans98, line 67: nit: let's see if we can find a more expressive name for this field.

rollup-interface/src/state_machine/da.rs
```diff
@@ -56,14 +58,68 @@ pub trait DaVerifier {
...
+    counter: usize,
+}
+
+impl<B: Buf> BufReaderWithCounter<B> {
```
preston-evans98, line 71: nit: Let's add doc comments on all of the public methods

rollup-interface/src/state_machine/da.rs
```diff
@@ -56,14 +58,68 @@ pub trait DaVerifier {
...
+impl<B: Buf> Read for BufReaderWithCounter<B> {
+    // Reads the inner buf into the provided buffer, and appends the data read to inner accumulator
+    fn read(&mut self, mut buf: &mut [u8]) -> std::io::Result<usize> {
+        let len_before_reading = self.inner.remaining();
+
+        buf.write_all(&self.inner.copy_to_bytes(min(buf.len(), len_before_reading)))?;
+
+        let num_read = len_before_reading - self.inner.remaining();
+
+        self.acc.append(&mut buf.to_vec());
```
preston-evans98, line 99: We're doing an extra copy here. Let's use `extend_from_slice` so that we don't have to convert the buf to a `vec`

rollup-interface/src/state_machine/da.rs
preston-evans98: Looked at this again after finding the `copy_to_bytes ` docs. I think we want to switch to using `copy_to_slice` instead. 

```rust
let buf_end = min(buf.len(), len_before_reading);
self.inner.copy_to_slice(&mut buf[..buf_end])
```

rollup-interface/src/state_machine/da.rs
```diff
@@ -56,14 +58,68 @@ pub trait DaVerifier {
     ) -> Result<(), Self::Error>;
 }
 
+// Simple structure that implements the Read trait for a buffer and  counts the number of bytes read from the beginning
+// Useful for the partial blob reading optimization: we know for each blob how many bytes have been read from the beginning
+// Because of soundness issues we cannot implement the Buf trait because the prover could get unproved blob data using the chunk method.
+#[derive(Debug, Clone, Serialize, Deserialize, BorshDeserialize, BorshSerialize, PartialEq)]
+pub struct BufReaderWithCounter<B: Buf> {
```
preston-evans98, line 65: nit: Since this type shows up in the public API, let's make sure that it's as clean and well-documented as we can. 

I'd recommend adding a doc comment to each field, and maybe renaming to `CountedBufReader`

preston-evans98 (approved):

rollup-interface/src/state_machine/da.rs
theochap: Great! That's a good way to avoid doing the copies indeed

Issue #359, opened 2023-05-30
Implement basic block builder


preston-evans98:
Once we've added the RPC endpoint for submitting transactions #357 , we should make an easy way to submit them on chain. 

A simple workflow is
- Select all transactions from the node's pending TX pool
- Assemble them into a batch in random order
- In a loop ...
   - Try executing the batch against the latest state. 
   - If any tx fails, remove it from the batch
   - If all txs succeed, break the loop
- If the resulting batch is not empty, post it onto the DA layer with the `send_transaction` interface

This workflow will cause the tx pool to fill up with old pruned transactions. Separately, we need to implement tx pool management (TODO: create issue)

Issue #357, opened 2023-05-29
Implement Send Transaction RPC


preston-evans98:
The Sovereign full node should have an RPC endpoint which accepts transactions on behalf of end-users. This RPC mode should be optional and enabled by a config setting. 

The endpoint should...
- Accept encoded transactions in the format generated by our `bank-cmd` tool (binary?). 
- Run stateless verification on the transaction. If stateless verification fails, return a descriptive error
- (Eventually) Check that the nonce of the transaction has not already been used. if it has, return a descriptive error.
- Add the transaction to the pending pool

citizen-stig:
> Run stateless verification on the transaction. If stateless verification fails, return a descriptive error

Should RPC node has it's own verification or should this verification added to `BatchBuilder?

> (Eventually) Check that the nonce of the transaction has not already been used. if it has, return a descriptive error.

What state should be checked? Should it be nonce from commited state, or nonces in mempool? 

> Add the transaction to the pending pool

Is pending pool same as mempool in `BatchBuilder` or is it separate pool and some another routine moves from one to another?

@preston-evans98 @bkolad @dubbelosix

PR #402, opened 2023-06-09
Txs RPC module

citizen-stig:
# Description
Creates Txs RPS module, which moves creation of the batch to the RPC endpoint.

This is also ground work for #357 

## Linked Issues

- Related to #359  and #357

## Testing

Unit test for RPC module has been added

## Docs

Public docs for TxsRPC handler has been added

### DIFF
--- a/examples/demo-rollup/src/main.rs
+++ b/examples/demo-rollup/src/main.rs
@@ -10,7 +10,6 @@ use anyhow::Context;
 use const_rollup_config::{ROLLUP_NAMESPACE_RAW, SEQUENCER_DA_ADDRESS};
 use demo_stf::app::{DefaultContext, DemoBatchReceipt, DemoTxReceipt};
 use demo_stf::app::{DefaultPrivateKey, NativeAppRunner};
-use demo_stf::batch_builder::FiFoStrictBatchBuilder;
 use demo_stf::genesis_config::create_demo_genesis_config;
 use demo_stf::runner_config::from_toml_path;
 use demo_stf::runtime::GenesisConfig;
@@ -120,12 +119,9 @@ async fn main() -> Result<(), anyhow::Error> {
         .merge(ledger_rpc_module)
         .expect("Failed to merge ledger RPC modules");
 
-    let batch_builder = demo_runner.take_batch_builder()?;
+    let batch_builder = demo_runner.take_batch_builder().unwrap();
 
-    let r = get_txs_rpc::<
-        FiFoStrictBatchBuilder<demo_stf::runtime::Runtime<DefaultContext>, DefaultContext>,
-        CelestiaService,
-    >(batch_builder, da_service.clone());
+    let r = get_txs_rpc(batch_builder, da_service.clone());
 
     methods.merge(r).expect("Failed to merge Txs RPC modules");
 
--- a/examples/demo-rollup/src/main.rs
+++ b/examples/demo-rollup/src/main.rs
@@ -133,9 +133,7 @@ async fn main() -> Result<(), anyhow::Error> {
         start_rpc_server(methods, address).await;
     });
 
-    // start_batch_producing(batch_builder, da_service.clone());
-
-    // For demonstration,  we also initialize the DaVerifier interface using the DaVerifier interface
+    // For demonstration, we also initialize the DaVerifier interface.
     // Running the verifier is only *necessary* during proof generation not normal execution
     let da_verifier = Arc::new(CelestiaVerifier::new(RollupParams {
         namespace: ROLLUP_NAMESPACE,
--- a/examples/demo-rollup/src/main.rs
+++ b/examples/demo-rollup/src/main.rs
@@ -3,12 +3,14 @@ mod ledger_rpc;
 
 #[cfg(test)]
 mod test_rpc;
+mod txs_rpc;
 
 use crate::config::RollupConfig;
 use anyhow::Context;
 use const_rollup_config::{ROLLUP_NAMESPACE_RAW, SEQUENCER_DA_ADDRESS};
 use demo_stf::app::{DefaultContext, DemoBatchReceipt, DemoTxReceipt};
 use demo_stf::app::{DefaultPrivateKey, NativeAppRunner};
+use demo_stf::batch_builder::FiFoStrictBatchBuilder;
 use demo_stf::genesis_config::create_demo_genesis_config;
 use demo_stf::runner_config::from_toml_path;
 use demo_stf::runtime::GenesisConfig;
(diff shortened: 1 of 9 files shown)


examples/demo-rollup/src/main.rs
```diff
@@ -139,23 +119,22 @@ async fn main() -> Result<(), anyhow::Error> {
         ledger_rpc::get_ledger_rpc::<DemoBatchReceipt, DemoTxReceipt>(ledger_db.clone());
     methods
         .merge(ledger_rpc_module)
-        .expect("Failed to merge rpc modules");
+        .expect("Failed to merge ledger RPC modules");
+
+    let batch_builder = demo_runner.take_batch_builder()?;
```
preston-evans98, line 124: Looking at this API, it seems a little odd that the demo-runner is opinionated about only creating a single builder. Is there something that makes it impossible for the demo to create two builders? If not, maybe we should make this API infallible and rename to somethign `new_batch_builder()`

examples/demo-rollup/src/main.rs
```diff
@@ -139,23 +119,22 @@ async fn main() -> Result<(), anyhow::Error> {
...
+
+    let r = get_txs_rpc::<
```
preston-evans98, line 126: I'm surprised we need type annotations here - generally, we want to avoid references to module-system specific structs like `DefaultContext` in the the demo. Do you know why type inference is broken here?

examples/demo-rollup/src/main.rs
```diff
@@ -139,23 +119,22 @@ async fn main() -> Result<(), anyhow::Error> {
...
-        rollup_config.da.clone(),
-        RollupParams {
-            namespace: ROLLUP_NAMESPACE,
-        },
-    ));
-
-    let batch_builder = demo_runner.take_batch_builder()?;
-
-    start_batch_producing(batch_builder, da_service.clone());
+    // start_batch_producing(batch_builder, da_service.clone());
```
preston-evans98, line 137: nit: this looks light it was commented by accident. Should we remove?

examples/demo-rollup/src/txs_rpc.rs
```diff
@@ -0,0 +1,63 @@
...
+/// Single data structure that manages mempool and batch producing.
+pub struct TxsRpcHandler<B: BatchBuilder, T: DaService> {
+    batch_builder: RwLock<B>,
+    da_service: Arc<T>,
+}
+
+impl<B: BatchBuilder + Send + Sync, T: DaService + Send + Sync> TxsRpcHandler<B, T> {
+    pub fn new(batch_builder: B, da_service: Arc<T>) -> Self {
+        Self {
+            batch_builder: RwLock::new(batch_builder),
```
preston-evans98, line 17: Do we need an `RwLock` here? It looks like we only ever call "write" - maybe a `Mutex` is more appropriate?

examples/demo-rollup/src/txs_rpc.rs
```diff
@@ -0,0 +1,63 @@
...
+fn register_txs_rpc_methods<B, D>(
+    rpc: &mut RpcModule<TxsRpcHandler<B, D>>,
+) -> Result<(), jsonrpsee::core::Error>
+where
+    // R: Serialize + DeserializeOwned,
+    // T: Serialize + DeserializeOwned,
+    B: BatchBuilder + Send + Sync + 'static,
+    D: DaService + Send + Sync + 'static,
+{
+    rpc.register_async_method("submit_batch", |_, batch_builder| async move {
```
preston-evans98, line 43: The JsonRPC convention is `{namespace}_{methodName}`. Maybe we want the namespace to be `builder`? So something like `builder_submitBatch`?

examples/demo-rollup/src/txs_rpc.rs
```diff
@@ -0,0 +1,63 @@
...
+pub fn get_txs_rpc<B, D>(batch_builder: B, da_service: Arc<D>) -> RpcModule<TxsRpcHandler<B, D>>
+where
+    // R: Serialize + DeserializeOwned,
+    // T: Serialize + DeserializeOwned,
+    B: BatchBuilder + Send + Sync + 'static,
+    D: DaService + Send + Sync + 'static,
+{
+    let txs_handler = TxsRpcHandler::new(batch_builder, da_service);
+    let mut rpc = RpcModule::new(txs_handler);
+    register_txs_rpc_methods::<B, D>(&mut rpc).expect("Failed to register txs RPC methods");
```
bkolad, line 61: nit: Let's avoid side effects here. `register_txs_rpc_methods` could take `txs_handler ` and  return a new `RpcModule`.  It can even be a method of `TxsRpcHandler` like `self.register_txs_rpc_methods()-> RpcModule`.

examples/demo-rollup/src/main.rs
citizen-stig: I decided to go this way, as it solves problem of internal set of pending transactions, I call it mempool for simplicity.
So it can only takes builder once.

I see several ways how to do it:

1. method signature changes to `fn get_batch_builder(&self) -> Arc<BacthBuilder>`, but since it is shared, and requires mutable references to do any operation, that can be done with `Arc<Mutex<BatchBuilder>>`
2. internal mempool of BatchBuilder is wrapped in `Arc<Mutex<>`
3. We allow several independent BatchBuilder with their own mempools
4. Actual batch builder is started in separate thread, and this method returns a struct that just channels requests to actual mempool.

Each options has its own drawbacks

examples/demo-rollup/src/txs_rpc.rs
citizen-stig: I followed the same logic as in ledger_rpc:

https://github.com/Sovereign-Labs/sovereign-sdk/blob/fef717e281d7466f9cdf7c4187f60e0b7169e86c/examples/demo-rollup/src/ledger_rpc.rs#L21

I mean, for this can, it can be inside single function.

examples/demo-rollup/src/txs_rpc.rs
citizen-stig: I agree, it needs follow that pattern.

Isn't `builder` too generic? What if it is `batchBuilder_submit`? 

I use term `mempool` everywhere, but I am not sure if this is correct in this case, but it can be `mempool_produceBatch` But then whole `BatchBuilder` trait probably needs to be renamed. Which is not a big problem at the moment.

Maybe side question, how do you envision mempool in the full node, would it be some completely separate module?

examples/demo-rollup/src/txs_rpc.rs
bkolad: ok

examples/demo-rollup/src/txs_rpc.rs
citizen-stig: Decided on a call

preston-evans98 (approved):

PR #406, opened 2023-06-13
`expose_rpc` macro works with `Runtime`

bkolad:
# Description
This PR introduces the following changes 
1. `expose_rpc`  works on `Runtime` rather than `StateTransitionRunner`.
2. Adds `rpc_gen` to `Accounts, Sequencer, Bank`

## Linked Issues
- Related to #409 

## Testing
No impact on tests.

## Docs
Relevant docs updated.


preston-evans98 (commented):
LGTM, but would love to see a review from @dubbelosix before we merge since he's most familiar with this part of the codebase.

examples/demo-stf/src/runtime.rs
```diff
@@ -40,17 +51,17 @@ use sov_modules_macros::{DefaultRuntime, DispatchCall, Genesis, MessageCodec};
 /// Similar mechanism works for queries with the difference that queries are submitted by users directly to the rollup node
 /// instead of going through the DA layer.
 
-#[cfg_attr(feature = "native", cli_parser(DefaultContext))]
+#[cfg_attr(
+    feature = "native",
+    cli_parser(DefaultContext),
+    expose_rpc(DefaultContext)
```
preston-evans98, line 57: Is it possible to remove the `DefaultContext` argument now that this annotation lives on the runtime instead of the STF-Runner?

bkolad: I tried doing this but I got a compilation error saying that `Context` does not implement `DeserializeOwned` which is  required by ``jsonrpsee`. I think we should be able to remove the Context but it will require a bit more digging;

citizen-stig (approved):

dubbelosix (approved):
lgtm!

Issue #391, opened 2023-06-06
Based Sequencing Feature


citizen-stig:
# Description

high level feature for tracking based sequencing

https://ethresear.ch/t/based-rollups-superpowers-from-l1-sequencing/15016

PR #410, opened 2023-06-15
Multiple sequencers registration and exit

citizen-stig:
# Description

## Changes

* Sequencer registry maintains list of "allowed sequencers", which is checked in begin blob hook. Potential sequencers registers separately and stakes funds. Sequencer can exit and get its funds back. Those are separated call requests. 
* If sequencer is slashed, it is removed from "allowed sequencers" list and cannot produce more batches or get funds back. But it can register again and stake new amount.
* Sequencer registry has a new enum "Sequencer Outcome", which only contains 2 variants: `Completed` or `Slashed` as only those 2 are relevant to sequencer registry lifecycle. It does not manage rewards.
* (!) End hook for sequencer registry requires sequencer da adress to be passed, so it knows which sequencer to slash. This required changes in `ApplyBatchError` and `SequencerOutcome` enums in `demo-stf`
* (!) Sequencer registry genesis depends on bank to be initialized, so bank should be listed before sequencer registry in the runtime. Also genesis sequencer should have enough funds to stake.
* Minor refactoring in `demo-stf`, duplicated helper function `create_demo_config` is moved into  `genesis_config`, as they it shared between test and cli.

## Fixes

* Removes specific `bech32` type as da address in response in favour of Vec<u8>
* Removes quering balance via sequencer registry

## Potential next features

* Allow exit funds to be sent to another address
* RPC query call to list all sequencers (blocked by traversing statemap)

## Linked Issues
- Related to #391 

## Testing
Unit tests were added or adjusted, both for registration/exit and hooks

## Docs
Sequencer-registry module public docs have peen extended


module-system/module-implementations/sov-sequencer-registry/src/lib.rs
```diff
@@ -26,31 +26,86 @@ pub struct Sequencer<C: sov_modules_api::Context> {
...
+                self.register(da_address, context, working_set)?
+            }
+            call::CallMessage::Exit { da_address } => {
+                self.exit(da_address, context, working_set)?
+            }
+        })
+    }
+}
+
+impl<C: sov_modules_api::Context> SequencerRegistry<C> {
```
bkolad, line 80: Let's move the code to `call.rs`?

bkolad (approved):

module-system/module-implementations/sov-sequencer-registry/src/lib.rs
citizen-stig: Could you please elaborate why? 

Then it makes genesis depending on "call.rs". And `get_coins_to_lock` does not have corresponding call nor query message, it is just helper method.

bkolad: My mistake I meant only `register_sequencer` which is used only in `call.rs`.

citizen-stig: `register_sequencer` is also used in `genesis`. Originally whole logic was just in `call.rs`, but as I adjusted genesis, I've moved it to lib.rs

examples/demo-stf/src/tests/mod.rs
```diff
@@ -60,3 +30,9 @@ pub fn has_tx_events(apply_blob_outcome: &BatchReceipt<SequencerOutcome, TxEffec
 
     events.peekable().peek().is_some()
 }
+
+pub fn new_test_blob(batch: Batch, address: &[u8]) -> TestBlob {
+    let address = Address::try_from(address).unwrap();
+    let data = batch.try_to_vec().unwrap();
+    TestBlob::new(data, address, [0; 32])
+}
```
preston-evans98, line 38: This looks like it's duplicated in sov-cli

preston-evans98 (approved):
Looks good to me overall!

One thing to be aware of is that we may potentially want to have a variable bonding amount in future - where sequencers who lock up more coins can post bigger bundles. But that's not a priority for now, so this seems like a good change.

examples/demo-stf/src/tests/mod.rs
citizen-stig: Moved to https://github.com/Sovereign-Labs/sovereign-sdk/issues/431

PR #415, opened 2023-06-16
ApplyBlobHooks, always call `end_blob_hook`: iteration 2

citizen-stig:
# Description

Now `end_blob_hook` is always called 

* Clearer separation of conserns. Now `apply_batch` method solely controls when and where working set is commited or revereted. Stateless checks are kept in separate stateless methos
* Decoding and dispatching are split into 2 different iteration. That should make it less cost efficient to DoS with bad batches. If last transaction of batch cannot be decoded, but previous txs are expensive to dispatch, for example.
* Transaction pre_dispatch_hook now takes reference instead of consuming transaction. This removes extra cloning.

## Linked Issues
- Related to #391 , because it is required for changes in sequencer registry

## Testing
Existing tests hopefully cover the change.

## Docs
Doc strings were updated

### DIFF
--- a/examples/demo-stf/src/tests/tx_revert_tests.rs
+++ b/examples/demo-stf/src/tests/tx_revert_tests.rs
@@ -3,9 +3,13 @@ use crate::{
     runtime::Runtime,
     tests::{data_generation::simulate_da_with_bad_serialization, has_tx_events},
 };
+use borsh::BorshSerialize;
+use sov_accounts::query::Response;
+use sov_modules_api::transaction::Transaction;
 use sov_modules_api::{
-    default_context::DefaultContext, default_signature::private_key::DefaultPrivateKey,
+    default_context::DefaultContext, default_signature::private_key::DefaultPrivateKey, PublicKey,
 };
+use sov_modules_stf_template::RawTx;
 use sov_modules_stf_template::{Batch, SequencerOutcome, SlashingReason};
 use sov_rollup_interface::{mocks::MockZkvm, stf::StateTransitionFunction};
 use sov_state::{ProverStorage, WorkingSet};
@@ -46,13 +50,14 @@ fn test_tx_revert() {
             None,
         );
 
-        assert!(
-            matches!(apply_blob_outcome.inner, SequencerOutcome::Rewarded(0),),
-            "Unexpected outcome: Batch exeuction should have succeeded"
+        assert_eq!(
+            SequencerOutcome::Rewarded(0),
+            apply_blob_outcome.inner,
+            "Unexpected outcome: Batch execution should have succeeded",
         );
 
         // Some events were observed
-        assert!(has_tx_events(&apply_blob_outcome));
+        assert!(has_tx_events(&apply_blob_outcome), "No events were taken");
 
         StateTransitionFunction::<MockZkvm>::end_slot(&mut demo);
     }
@@ -87,6 +92,96 @@ fn test_tx_revert() {
 }
 
 #[test]
+fn test_nonce_incremented_on_revert() {
+    let tempdir = tempfile::tempdir().unwrap();
+    let path = tempdir.path();
+    let value_setter_admin_private_key = DefaultPrivateKey::generate();
+    let election_admin_private_key = DefaultPrivateKey::generate();
+    let voter = DefaultPrivateKey::generate();
+
+    let config = create_demo_config(
+        SEQUENCER_BALANCE,
+        &value_setter_admin_private_key,
+        &election_admin_private_key,
+    );
+
+    {
+        let mut demo = create_new_demo(path);
+        StateTransitionFunction::<MockZkvm>::init_chain(&mut demo, config);
+        StateTransitionFunction::<MockZkvm>::begin_slot(&mut demo, Default::default());
+
+        let set_candidates_message = Runtime::<DefaultContext>::encode_election_call(
+            sov_election::call::CallMessage::SetCandidates {
+                names: vec!["candidate_1".to_owned(), "candidate_2".to_owned()],
+            },
+        );
+
+        let set_candidates_message = Transaction::<DefaultContext>::new_signed_tx(
+            &election_admin_private_key,
+            set_candidates_message,
+            0,
+        );
+
+        let add_voter_message = Runtime::<DefaultContext>::encode_election_call(
+            sov_election::call::CallMessage::AddVoter(voter.pub_key().to_address()),
+        );
+        let add_voter_message = Transaction::<DefaultContext>::new_signed_tx(
+            &election_admin_private_key,
+            add_voter_message,
+            1,
+        );
+
+        let vote_message = Runtime::<DefaultContext>::encode_election_call(
+            sov_election::call::CallMessage::Vote(100),
+        );
+        let vote_message = Transaction::<DefaultContext>::new_signed_tx(&voter, vote_message, 0);
+
+        let txs = vec![set_candidates_message, add_voter_message, vote_message];
+        let txs = txs
+            .into_iter()
+            .map(|t| RawTx {
+                data: t.try_to_vec().unwrap(),
+            })
+            .collect();
+
+        let apply_blob_outcome = StateTransitionFunction::<MockZkvm>::apply_blob(
+            &mut demo,
+            new_test_blob(Batch { txs }, &DEMO_SEQUENCER_DA_ADDRESS),
+            None,
+        );
+
+        assert_eq!(
+            SequencerOutcome::Rewarded(0),
+            apply_blob_outcome.inner,
+            "Unexpected outcome: Batch execution should have succeeded",
+        );
+        StateTransitionFunction::<MockZkvm>::end_slot(&mut demo);
+    }
+
+    {
+        let runtime = &mut Runtime::<DefaultContext>::default();
+        let storage = ProverStorage::with_path(path).unwrap();
+        let mut working_set = WorkingSet::new(storage);
+
+        // We sent 4 vote messages but one of them is invalid and should be reverted.
+        let resp = runtime.election.number_of_votes(&mut working_set);
+
+        assert_eq!(resp, sov_election::query::GetNbOfVotesResponse::Result(0));
+
+        let nonce = match runtime
+            .accounts
+            .get_account(voter.pub_key(), &mut working_set)
+        {
+            Response::AccountExists { nonce, .. } => nonce,
+            Response::AccountEmpty => 0,
+        };
+
+        // Voter should have its nonce implemented
+        assert_eq!(1, nonce);
+    }
+}
+
+#[test]
 fn test_tx_bad_sig() {
     let tempdir = tempfile::tempdir().unwrap();
     let path = tempdir.path();
@@ -113,8 +208,9 @@ fn test_tx_bad_sig() {
             None,
         );
 
-        assert!(
-            matches!(apply_blob_outcome.inner, SequencerOutcome::Slashed(SlashingReason::StatelessVerificationFailed),),
+        assert_eq!(
+            SequencerOutcome::Slashed(SlashingReason::StatelessVerificationFailed),
+            apply_blob_outcome.inner,
             "Unexpected outcome: Stateless verification should have failed due to invalid signature"
         );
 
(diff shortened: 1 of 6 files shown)


module-system/sov-modules-stf-template/src/lib.rs
```diff
@@ -24,8 +24,11 @@ pub enum TxEffect {
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub enum SequencerOutcome {
+    /// Sequencer receives reward amount in defined token and can withdraw its deposit
     Rewarded(u64),
+    /// Sequencer loses its deposit and receives no reward
     Slashed(SlashingReason),
+    /// TODO: What does it actually mean? No reward, but what is with deposit?
```
citizen-stig, line 31: This needs to be clarified

module-system/sov-modules-api/src/hooks.rs
```diff
@@ -31,14 +31,18 @@ pub trait ApplyBlobHooks {
     type BlobResult;
 
     /// Runs at the beginning of apply_blob, locks the sequencer bond.
+    /// If this hook returns Err, batch is not applied
+    /// TODO: Expectation about what is in working set, if it is "clean"
```
citizen-stig, line 35: This needs to be addresses

module-system/sov-modules-api/src/hooks.rs
```diff
@@ -31,14 +31,18 @@ pub trait ApplyBlobHooks {
...
     fn begin_blob_hook(
         &self,
         sequencer: &[u8],
         raw_blob: &[u8],
         working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
     ) -> anyhow::Result<()>;
 
-    /// Executes at the end of apply_blob and rewards the sequencer. This method is not invoked if the sequencer has been slashed.
+    /// Executes at the end of apply_blob and rewards or slashed the sequencer
+    /// TODO: Expectation about what is in working set, if it is "clean"
```
citizen-stig, line 44: This needs to be addressed

module-system/sov-modules-stf-template/src/app_template.rs
```diff
@@ -163,107 +165,130 @@ where
...
+            let tx_result = self.runtime.dispatch_call(msg, &mut batch_workspace, &ctx);
 
-                    let receipt = TransactionReceipt {
-                        tx_hash: raw_tx_hash,
-                        body_to_save: None,
-                        events: batch_workspace.take_events(),
-                        receipt: tx_effect,
-                    };
+            // TODO: Do we want to call it before revert in case of error?
+            // TODO: Replace expect with proper handling
```
citizen-stig, line 172: Should I address both of these in this PR or offload to separate issue?

module-system/sov-modules-api/src/hooks.rs
preston-evans98: I think the expectation is that a `WorkingSet` is dirty. If it's clean, we should pass a `StateCheckpoint` instead. 

cc @bkolad is that right?

module-system/sov-modules-stf-template/src/app_template.rs
```diff
@@ -59,92 +59,94 @@ where
...
             error!(
-                "Error: The transaction was rejected by the 'enter_apply_blob' hook. Skipping batch without slashing the sequencer: {}",
+                "Error: The batch was rejected by the 'begin_blob_hook' hook. Skipping batch without slashing the sequencer: {}",
                 e
             );
+            // TODO: Should it commit?
             self.checkpoint = Some(batch_workspace.revert());
             return Err(ApplyBatchError::Ignored(batch_data_and_hash.hash));
         }
+        // Point B: sequencer has paid
```
preston-evans98, line 103: nit: It's a little unclear why `Sequencer has paid` is a `point`, but `ApplyBlobHook: begin` isn't. What do the `Point`s correspond to?

module-system/sov-modules-stf-template/src/app_template.rs
```diff
@@ -59,92 +59,94 @@ where
...
-                error!(
-                    "Unable to deserialize batch provided by the sequencer {}",
-                    e
-                );
-                self.checkpoint = Some(batch_workspace.revert());
-                Err(ApplyBatchError::Slashed {
+        let (txs, messages) = match self.pre_process_batch(&batch_data_and_hash) {
+            Ok((txs, messages)) => (txs, messages),
+            Err(slashing_reason) => {
+                // Point B.1: TODO: What is purpose of these revert? The only thing that changed after Point B was take_events()
```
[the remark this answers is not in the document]
preston-evans98: It looks to me like that's just an artifact and should be safe to remove. cc @bkolad

module-system/sov-modules-stf-template/src/lib.rs
preston-evans98: This means that the batch was ignored, so there is no sequencer deposit

module-system/sov-modules-stf-template/src/app_template.rs
```diff
@@ -59,92 +59,94 @@ where
...
         if let Err(e) =
             self.runtime
                 .begin_blob_hook(sequencer, &batch_data_and_hash.data, &mut batch_workspace)
         {
             error!(
-                "Error: The transaction was rejected by the 'enter_apply_blob' hook. Skipping batch without slashing the sequencer: {}",
+                "Error: The batch was rejected by the 'begin_blob_hook' hook. Skipping batch without slashing the sequencer: {}",
                 e
             );
+            // TODO: Should it commit?
```
preston-evans98, line 99: I think reverting is the correct behavior, since the hook failed. 

On that note, though, we should document clearly which hooks (if any) have their state changes committed on error, and which ones (if any) revert on failure.

module-system/sov-modules-stf-template/src/app_template.rs
```diff
@@ -59,92 +59,94 @@ where
...
+        batch: impl Buf,
+    ) -> ApplyBatchResult<BatchReceipt<SequencerOutcome, TxEffect>> {
+        let batch_data_and_hash = BatchDataAndHash::new::<C>(batch);
         debug!(
             "Applying batch from sequencer: 0x{}",
             hex::encode(sequencer)
         );
+
+        // Point A:
+        // Initial batch workspace
```
preston-evans98, line 83: Nit: this comment is a bit confusing. Maybe we could change to `// Initialize batch workspace`?

preston-evans98 (commented):
This looks like an improvement, but the state machine is still so complex that it's hard to be sure we haven't introduced any new bugs and/or changes in behavior. Maybe we need to break the function down into chunks and specify exactly what should happen in each one - something like:

1. Begin blob hook: performs per-blob work like locking the sequencer's balance. If this hook fails, execution is aborted and no state changes are committed. If the hook succeeds.. changes are committed immediately. The end-blob hook is guaranteed to run in that case.
2. Pre-validation: runs stateless checks on the batch. Batch is skipped on error. No additional state changes are made in that case
3. Pre-dispatch tx hook. State changes made by this hook are committed unconditionally. 
4. Tx application. State changes made by the transaction are reverted if the runtime returns an error
5. Post tx hook. State changes made by the post-tx hook are committed unconditionally
6. End blob hook. State changes made here are committed unconditionally.

Once we have an outline agreed on (this one might not be correct, it's just an example!), we should probably implement an explicit state machine with return types that make it clear which state changes are committed and which are not.

module-system/sov-modules-api/src/hooks.rs
bkolad: Where this expectations come from?  The idea is to create `WorkingSet` at the beginning of the `ApplyBatch` and pass it everywhere (why the functions we are passing it to have to expect anything about it?).

module-system/sov-modules-api/src/hooks.rs
bkolad: The same as above.

module-system/sov-modules-stf-template/src/app_template.rs
bkolad: Yes I think we should revert.

module-system/sov-modules-stf-template/src/app_template.rs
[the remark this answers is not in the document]
bkolad: I added it here for future-proofing. We have to revert on slashing and if we ever extend or change the point B (or add some other logic to reverts) we might forget to update the slashing logic. So I added revert even if it is not technically needed now:
 1. to make it obvious that we revert on slashing. 
 2. to ensure reverts still works even if we make changes in code not directly related to line 112.
 
I am fine with removing it if it causes problems.

module-system/sov-modules-api/src/hooks.rs
citizen-stig: > why the functions we are passing it to have to expect anything about it?

Because it receives mutable reference to working set. And in different hooks it gets different state of this working set.

For example, in `post_dispatch_tx_hook` it receives "dirty" working set, and it can be both commited or reverted, depending on transaction status, where in `post_apply_blob_hook` state is clean, which means was either commited or reveted.

Having no expectations about state makes it hard to use the hook. 
For example post dispatch hook for `Accounts` it looks like module assumes that state is clean and won't be reverted, so it increases account nonce on invalid transaction, when in reality it won't happen. I am working on updating tests to consider that case.

> The idea is to create WorkingSet at the beginning of the ApplyBatch and pass it everywhere

This approach is very complicated to implement without bugs, that's why it is good at least to have an assumptions

module-system/sov-modules-stf-template/src/app_template.rs
citizen-stig: Points comment are actually leftovers from drafting and discussion,

citizen-stig:
> Once we have an outline agreed on (this one might not be correct, it's just an example!), we should probably implement an explicit state machine with return types that make it clear which state changes are committed and which are not.

I agree on this approach, even though it increases the scope of original task. 

Do you think it will be easire to first transition to state machine from current main?
Or maybe commit these changes and then push forward for state machine changes.

module-system/sov-modules-api/src/hooks.rs
bkolad: Now, I think we could pass `StateCheckpoint` to the function and convert it to `WorkingSet` inside, so we add type-level information. But I am still not sure why the hooks need to be aware of why the `WorkinSet` was committed or reverted.

citizen-stig: > But I am still not sure why the hooks need to be aware of why the WorkinSet was committed or reverted.

I guess my point could be better formulated. It shouldn't be aware if it was commited or reverted, but it should expected if its changes are going to be commited or reverted. 

In my original comment, "dirty" state means that it will be reverted or commited after hook executed. Which introduces bug in nonce tracking.

examples/demo-stf/src/tests/tx_revert_tests.rs
```diff
@@ -86,6 +91,95 @@ fn test_tx_revert() {
     }
 }
 
+#[test]
+fn test_nonce_incremented_on_revert() {
```
citizen-stig, line 95: @bkolad here's faulty case in current design

module-system/sov-modules-api/src/hooks.rs
preston-evans98: Yep, this makes sense. The more we  can use the type system the better IMO :)

preston-evans98:
> Do you think it will be easire to first transition to state machine from current main? Or maybe commit these changes and then push forward for state machine changes.

Yeah, let's push these changes first and then create a separate task for the refactor

citizen-stig:
Refactor task is created: https://github.com/Sovereign-Labs/sovereign-sdk/issues/421

module-system/sov-modules-stf-template/src/app_template.rs
citizen-stig: Yes, this will be discussed as part of https://github.com/Sovereign-Labs/sovereign-sdk/issues/421

module-system/sov-modules-api/src/hooks.rs
citizen-stig: Yes, offloaded this to https://github.com/Sovereign-Labs/sovereign-sdk/issues/421

module-system/sov-modules-stf-template/src/app_template.rs
citizen-stig: But it means that it reverts any changes from `begin_blob_hook`, which is not what it probably expects.

bkolad (approved):

Issue #377, opened 2023-06-01
Design: Unified Rollup Architecture


preston-evans98:
# Background
In its original design, Sovereign assumed that both the `StateTransitionFunction` and the `DaVerifier` would always be executed in zk. 

## How Proofs Function in a Zk-rollup with Zk proofs of fork choice
 In that design, our logical proof format was pretty straightforward:

```rust
struct ProofOutput { 
  input_state_root: [u8;32],
  da_block_hash: [u8;32]
  output_state_root: [u8;32],
}
```

Every time a block was put onto the DA layer, we would create a SNARK which said "applying the transactions from {da_block_hash} to a rollup with state {input_state_root} yields new state root {output_state_root}". 

By aggregating these proofs with recursion, we could create new proofs with the following structure

```rust 
struct RecursiveProofOutput { 
  initial_state_root: [u8;32], 
  initial_da_block_hash: [u8;32], 
  final_da_block_hash: [u8;32] 
  output_state_root: [u8;32],
}
```
And after enough layers of recursion, we would have a final `RecursiveProofOutput` whose `initial_state_root` was the rollup's genesis state. To sync the chain, all a user would need to do would be:
1. Verify the outer recursive snark
2. Verify that `initial_state_root` was the correct genesis state root
3. Verify that `final_da_block_hash` matched a recent DA layer block

Since this structure was so simple, it made sense to hard-code it in the rollup. 

## Current State
Today, the original assumptions of Sovereign no longer hold. We intend to support optimistic rollups (with zk proofs of *fraud* rather than proofs of validity), and the Avail team has asked for a mode in which fork-choice checks are externalized to the full node rather than running in the Snark. 

Looking ahead, it seems likely that we'll want to support at least four different paradigms: 
- ZK STF with ZK fork choice
- ZK STF with standard fork-choice (meaning that every verifier has to download the full rollup data to check the fork choice. Needed by Avail, likely useful for Ethereum as well)
- Optimistic STF with ZK fork choice (useful for bridging optimistic rollups)
- Optimistic STF with standard fork-choice

## Suggested Action

To accommodate all of these options, we should consider creating a mechanism to express that a SNARK requires additional validity conditions to be checked before causing a state transition. 

[...]

PR #418, opened 2023-06-17
Add support for Validity Conditions

preston-evans98:
# Description
Implement basic Validity Condition traits from #377. Update Jupiter to use ValidityConditions.

## Linked Issues
- Related to #377


rollup-interface/src/state_machine/crypto/simple_hasher.rs
```diff
@@ -0,0 +1,50 @@
+use digest::{generic_array::GenericArray, Digest, OutputSizeUser};
+
+/// A minimal trait representing a hash function. We implement our own
+/// rather than relying on `Digest` for broader compatibility.
+pub trait SimpleHasher: Sized {
```
bkolad, line 5: Cool our own trait! Can you please grep our code and replace any `jmt::SimpleHasher` with this `SimpleHasher `

adapters/celestia/src/verifier/mod.rs
```diff
@@ -102,33 +105,53 @@ pub struct RollupParams {
...
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+/// A validity condition expressing that a chain of DA layer blocks is contiguous and canonical
+pub struct ChainValidityCondition {
+    pub prev_hash: [u8; 32],
+    pub block_hash: [u8; 32],
+}
+
+impl ValidityCondition for ChainValidityCondition {
+    fn combine<H: SimpleHasher>(&self, rhs: Self) -> Result<Self, anyhow::Error> {
```
bkolad, line 116: Nit: Should treat this code more as a library and use `thiserror` and let the caller convert to `anyhow`? The error enums in `thiserror` have better structure for examination by the caller of the api (so it will help with tests).

rollup-interface/src/state_machine/zk/traits.rs
```diff
@@ -31,14 +33,40 @@ pub trait Zkvm {
...
 }
 
-// TODO!
-mod risc0 {
-    #[allow(unused)]
-    struct MethodId([u8; 32]);
+/// This trait expresses that a type can check a validity condition.
+pub trait ValidityConditionChecker<Condition: ValidityCondition> {
+    /// Check a validity condition
+    fn check(&mut self, condition: &Condition) -> Result<(), anyhow::Error>;
```
bkolad, line 67: Maybe `thiserror`?

rollup-interface/src/state_machine/crypto/simple_hasher.rs
preston-evans98: Good thinking. Will do!

preston-evans98: I actually switched to re-exporting `jmt::SimpleHasher` instead of making a custom version. Otherwise, we'll introduce conflicts with the default implementations we want (a `Digest` should be one of our `SimpleHasher`s, but it also has a blanket impl of `jmt::SimpleHasher`).

adapters/celestia/src/verifier/mod.rs
preston-evans98: Switched to an associated `Error` type in the trait and `ThisError` in the implementation.

bkolad (approved):

citizen-stig (approved):

theochap (approved):

Issue #417, opened 2023-06-16
Fix ledger rpc getTransactions bug


theochap:
# Context
The PR #403 designs a series of fuzzing tests for ledger rpc using the protest API. A fuzzing test fails on CI (and not locally) probably due to concurrency issues. 

## How to reproduce the error
In test_rpc.rs in the test `proptest_get_batches`, the following test fails on CI but not locally
```rust
// TODO: Solve this test 
TestExpect{
       data: format!(r#"{{"jsonrpc":"2.0","method":"ledger_getBatches","params":[[{random_batch_num}], "Full"],"id":1}}"#),
       expected: format!(r#"{{"jsonrpc":"2.0","result":[{{"hash":"0x{batch_hash}","tx_range":{{"start":{first_tx_num},"end":{last_tx_num}}},"txs":[{tx_full_data}],"custom_receipt":{batch_receipt}}}],"id":1}}"#)},
```

## Possible causes
It is pretty hard to track down the cause of failure for this test. The CI fails randomly a variable number of test cases, this test also randomly passes.
The failure is due to the fact that the list of transaction hashes expected does not match the list of transaction hashes delivered by rpc. The order of the transactions is modified and sometimes new hashes appear randomly to the transaction list.
It seems that this is probably a concurrency issue in the way rpc handles incoming transactions concurrently. The handlers and the code for the test are all mono threaded. It may also be an error related to CI configuration. This test sends a significant number of requests concurrently to RPC.

PR #403, opened 2023-06-09
Fuzzing tests for ledger rpc

theochap:
# Description
This PR implements fuzzing tests for ledger rpc. The fuzzing api is used to generate random test data for the ledger db and random queries to test the ledger rpc api.

## Linked Issues
- Fixes #397

## Testing
Some tests have been added to the test_rpc.rs folder.

PR #422, opened 2023-06-20
Write slot as atomic batch

preston-evans98:
# Description
This commit updates the ledger DB so that all components of a slot are written as batch. This should both improve efficiency and eliminate some potential concurrency bugs. 

## Linked Issues
- Related to #417 

## Testing
Existing unit tests pass. cc @theochap @citizen-stig - let's see if we can reproduce the issue from #417 after this change or not

### DIFF
--- a/full-node/db/sov-db/src/ledger_db/mod.rs
+++ b/full-node/db/sov-db/src/ledger_db/mod.rs
@@ -9,7 +9,7 @@ use sov_rollup_interface::{
     services::da::SlotData,
     stf::{BatchReceipt, Event},
 };
-use sov_schema_db::{Schema, DB};
+use sov_schema_db::{Schema, SchemaBatch, DB};
 
 use crate::{
     rocks_db_config::gen_rocksdb_options,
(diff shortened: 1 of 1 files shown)


full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -211,6 +218,8 @@ impl LedgerDB {
             // The lock is released here
```
cemozerr, line 218: Can't point this out (because its not in diff) but in line 208, there's a TODO for exposing a batch API, which we seem to implement now. Might want to remove/update that TODO.

cemozerr (approved):
LGTM

full-node/db/sov-db/src/ledger_db/mod.rs
preston-evans98: Good call. Removing!

bkolad (approved):

citizen-stig:
@preston-evans98 Do we have a test in main or this branch which has failing case? Otherwise we are blind.

Maybe this PR should be opened against #403, and merged there first?

PR #423, opened 2023-06-21
add macros to generate clap commands and args for modules in runtime

dubbelosix:
# Description
We currently let the users use sov-cli by passing in a json file that contains a json representation of the CallMessage
```
cargo run --bin sov-cli serialize-call src/sov-cli/test_data/minter_private_key.json Bank src/sov-cli/test_data/create_token.json 1
```
where `src/sov-cli/test_data/create_token.json` is
```
{
    "CreateToken": {
      "salt": 11,
      "token_name": "sov-test-token",
      "initial_balance": 1000,
      "minter_address": "sov15vspj48hpttzyvxu8kzq5klhvaczcpyxn6z6k0hwpwtzs4a6wkvqmlyjd6",
      "authorized_minters": ["sov15vspj48hpttzyvxu8kzq5klhvaczcpyxn6z6k0hwpwtzs4a6wkvqmlyjd6"]
    }
}
```

What this PR does is to enable the CallMessage enum (which contains all the variants for call messages accepted by the module) to built using clap. so the above command becomes
```
cargo run --bin sov-cli module --nonce 1 --private-key src/sov-cli/test_data/minter_private_key.json  bank create-token 11 sov-test-token 1000 sov15vspj48hpttzyvxu8kzq5klhvaczcpyxn6z6k0hwpwtzs4a6wkvqmlyjd6 sov15vspj48hpttzyvxu8kzq5klhvaczcpyxn6z6k0hwpwtzs4a6wkvqmlyjd6
```
and sov-cli would also have documentation based on doc comments in the CallMessage enum for each module
```
% cargo run --features="cli" --bin sov-cli module bank
    Finished dev [unoptimized + debuginfo] target(s) in 0.30s
     Running `/Users/dubbelosix/sovereign/target/debug/sov-cli module bank`
Usage: sov-cli module bank <COMMAND>

Commands:
  create-token  Creates a new token with the specified name and initial balance
  transfer      Transfers a specified amount of tokens to the specified address
  burn          Burns a specified amount of tokens
  mint          Mints a specified amount of tokens
  freeze        Freeze a token so that the supply is frozen
  help          Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help

```

## Linked Issues
- Related to #358 #371  

## Testing
TBD

## Docs
TBD


adapters/celestia/src/celestia.rs
```diff
@@ -335,6 +336,22 @@ impl From<[u8; 32]> for H160 {
     }
 }
 
+impl FromStr for H160 {
+    type Err = anyhow::Error;
```
bkolad, line 340: Let's use `thiserror` here and let the caller convert it ti anyhow if needed.

adapters/celestia/src/verifier/address.rs
```diff
@@ -32,4 +33,15 @@ impl Display for CelestiaAddress {
     }
 }
 
+impl FromStr for CelestiaAddress {
+    type Err = anyhow::Error;
```
bkolad, line 37: Let's use thiserror here and let the caller convert it ti anyhow if needed.

examples/demo-simple-stf/tests/stf_test.rs
```diff
@@ -26,6 +27,13 @@ impl From<[u8; 32]> for DaAddress {
     }
 }
 
+impl FromStr for DaAddress {
+    type Err = anyhow::Error;
```
bkolad, line 31: Let's use thiserror here and let the caller convert it ti anyhow if needed.

examples/demo-simple-stf/tests/stf_test.rs
```diff
@@ -26,6 +27,13 @@ impl From<[u8; 32]> for DaAddress {
...
+    fn from_str(_s: &str) -> Result<Self, Self::Err> {
+        unimplemented!()
```
bkolad, line 33: Is it ok?

examples/demo-stf/src/runtime.rs
```diff
@@ -53,12 +55,13 @@ use sov_value_setter::query::{ValueSetterRpcImpl, ValueSetterRpcServer};
 
 #[cfg_attr(
     feature = "native",
-    cli_parser(DefaultContext),
+    cli_parser(DefaultContext, "sequencer"),
```
bkolad, line 58: I see there is `// #[cli_skip]` bellow. Can we di it this way instead of hardcoding module name here?

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+
+            let raw_contents = hex::encode(serialized.raw.data);
+            println!("{}", raw_contents);
+        }
+    }
+}
+
+#[cfg(not(feature = "native"))]
+pub fn main() {
+    println!("build binary with native");
```
bkolad, line 305: Why do we have `fn main` here and `main.rs` in the same directory?

module-system/module-implementations/examples/sov-election/src/call.rs
```diff
@@ -7,12 +7,16 @@ use anyhow::{anyhow, bail, ensure, Result};
...
+#[cfg(feature = "native")]
+use sov_modules_macros::CustomParser;
+
 /// Call actions supported byte the module.
 #[cfg_attr(
     feature = "native",
     derive(serde::Serialize),
     derive(serde::Deserialize)
 )]
+#[cfg_attr(feature = "native", derive(CustomParser), module_name = "Election")]
```
bkolad, line 19: Would it be possible to add `fn module_name()` to the `ModuleInfo` trait and have the name generated automatically? We could use the same approach for `RPC`

module-system/module-implementations/examples/sov-value-setter/Cargo.toml
```diff
@@ -27,6 +27,7 @@ serde_json = { workspace = true, optional = true }
 thiserror = { workspace = true }
 borsh = { workspace = true, features = ["rc"] }
 jsonrpsee = { workspace = true, features = ["macros", "client-core", "server"], optional = true }
+clap = { workspace = true }
```
bkolad, line 30: We should include clap only when compiled with native feature (for all the modules)

module-system/module-implementations/sov-bank/src/token.rs
```diff
@@ -9,14 +11,46 @@ pub type Amount = u64;
...
+    derive(serde::Deserialize),
+    derive(clap::Parser)
 )]
 #[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq, Clone)]
 pub struct Coins<C: sov_modules_api::Context> {
     pub amount: Amount,
     pub token_address: C::Address,
 }
 
+// pub fn parse_amount(s: &str) -> Result<Amount, anyhow::Error> {
```
bkolad, line 23: Let's remove it

module-system/module-implementations/sov-bank/src/token.rs
```diff
@@ -9,14 +11,46 @@ pub type Amount = u64;
...
+//     match s.parse::<u64>() {
+//         Ok(val) => Ok(val),
+//         Err(e) => Err(anyhow::Error::new(e)),
+//     }
+// }
+
+impl<C: Context> FromStr for Coins<C>
+where
+    C::Address: FromStr<Err = anyhow::Error>,
```
bkolad, line 32: thiserror

module-system/sov-modules-api/src/default_signature.rs
```diff
@@ -181,3 +183,25 @@ fn map_error(e: ed25519_dalek::SignatureError) -> std::io::Error {
 fn map_error(_e: ed25519_dalek::SignatureError) -> std::io::Error {
     std::io::Error::new(std::io::ErrorKind::Other, "Signature error")
 }
+
+impl FromStr for DefaultPublicKey {
+    type Err = anyhow::Error;
```
bkolad, line 188: thiserror

module-system/sov-modules-api/src/lib.rs
```diff
@@ -53,6 +55,16 @@ impl<'a> TryFrom<&'a [u8]> for Address {
     }
 }
 
+impl FromStr for Address {
+    type Err = anyhow::Error;
```
bkolad, line 59: thiserror

module-system/sov-modules-macros/src/cli_parser.rs
```diff
@@ -25,49 +29,111 @@ impl CliParserMacro {
...
+                let mut module_path = type_path.path.clone();
+                if let Some(segment) = module_path.segments.last_mut() {
+                    if let syn::PathArguments::AngleBracketed(angle_bracketed_data) =
+                        &mut segment.arguments
+                    {
+                        let mut new_args = Vec::new();
+                        for gen_arg in &angle_bracketed_data.args {
+                            match gen_arg {
+                                syn::GenericArgument::Type(syn::Type::Path(type_path))
+                                    if type_path.path.is_ident("C") =>
```
bkolad, line 56: We can't hardcode `C` here. Everything should work even when someone writes `<K: Context>` and we need unit tests for this.

bkolad:
Thanks for the work! The module system now makes `clap` firs class citizen which is a bit unfortunate. Would it be possible to convert the module `CallMessages` to some intermediate structures (which we can even serialize to a file, something like abi in ethereum) and generate the clap options from this structures? I looks like we could achieve this by using the clap builder pattern. This way we decouple the dependency + we can use the structures to generate clis in other programming languages or even web clients. 

I understand this might be huge chunk of work so I am fine with merging the PR, but I would like to discuss the intermediate structures approach on the daily.

module-system/module-implementations/examples/sov-election/Cargo.toml
```diff
@@ -24,7 +24,7 @@ thiserror = { workspace = true }
 borsh = { workspace = true, features = ["rc"] }
 hex = { workspace = true }
 jsonrpsee = { workspace = true, features = ["macros", "client-core", "server"], optional = true }
-
+clap = {workspace = true, features = ["derive"]}
```
citizen-stig, line 27: Nit, formatting
```suggestion
clap = { workspace = true, features = ["derive"] }
```

module-system/module-implementations/examples/sov-value-setter/Cargo.toml
citizen-stig, line 30:
```suggestion
clap = { workspace = true, optional = true }
```

module-system/sov-modules-api/src/default_signature.rs
```diff
@@ -181,3 +183,25 @@ fn map_error(e: ed25519_dalek::SignatureError) -> std::io::Error {
...
+    }
+}
+
+impl FromStr for DefaultSignature {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let bytes = hex::decode(s)?;
+        let msg_sig =
+            DalekSignature::from_bytes(&bytes).map_err(|_| anyhow::anyhow!("Invalid signature"))?;
```
citizen-stig, line 204: Having source error to be included at least in form of text can help debugging
```suggestion
            DalekSignature::from_bytes(&bytes).map_err(|e| anyhow::anyhow!("Invalid signature: {:?}", e))?;
```

module-system/sov-modules-api/src/lib.rs
```diff
@@ -141,7 +154,9 @@ pub trait Spec {
         + Serialize
         + for<'a> Deserialize<'a>
         + Send
-        + Sync;
+        + Sync
+        + 'static
```
citizen-stig, line 158: 'static looks like very strict requirement, could you please elaborate, why it is needed?

examples/demo-stf/src/sov-cli/native.rs
preston-evans98, line 305: nit: Let's use the same error message from before in this case

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+    match cli.command {
+        Commands::SerializeCall {
+            sender_priv_key_path,
+            module_name,
+            call_data_path,
+            nonce,
+        } => {
+            let serialized =
+                SerializedTx::new(&sender_priv_key_path, &module_name, &call_data_path, nonce)
+                    .unwrap_or_else(|e| panic!("Call message serialization error: {}", e));
```
preston-evans98, line 224: nit: Since this is a user-facing tool, I don't think we want to panic at all if possible. Why don't we change the signature to `fn main() -> Result<(), anyhow::Error>`

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+
+            let mut bin_path = PathBuf::from(call_data_path);
+            bin_path.set_extension("dat");
+
+            let mut file = File::create(bin_path)
+                .unwrap_or_else(|e| panic!("Unable to crate .dat file: {}", e));
+
+            let raw_contents = hex::encode(serialized.raw.data).as_bytes().to_vec();
+            file.write_all(&raw_contents)
+                .unwrap_or_else(|e| panic!("Unable to save .dat file: {}", e));
```
preston-evans98, line 234: nit: let's avoid panics

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+        }
+        Commands::MakeBlob { path_list } => {
+            let mut data_list: Vec<Vec<u8>> = vec![];
+
+            for path in path_list {
+                let mut f = File::open(path).expect("Unable to open file");
```
preston-evans98, line 240: nit: let's avoid panics

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+                let mut hex_string = String::new();
+                f.read_to_string(&mut hex_string)
+                    .expect("Unable to read the file");
```
preston-evans98, line 243: nit: let's avoid panics

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+
+                // decode the hex string to bytes
+                let bytes = hex::decode(hex_string.trim()).expect("Decoding failed");
```
preston-evans98, line 246: nit: let's avoid panics

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+
+                // push it into data_list
+                data_list.push(bytes);
+            }
+            let blob = data_list.try_to_vec().unwrap();
```
preston-evans98, line 251: nit: I think this panic is ok (IIUC, an error should not actually be possible here), but lets add a comment documenting _why_ a panic is ok: the error is unexpected and unrecoverable, likely indicating a bug

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+                    }),
+                );
+                let token_address =
+                    sov_bank::create_token_address::<C>(&token_name, sender_address.as_ref(), salt);
+                println!("{}", token_address);
+            }
+
+            UtilCommands::ShowPublicKey { private_key_path } => {
+                let sender_priv_key = SerializedTx::deserialize_priv_key(private_key_path)
+                    .expect("Failed to get private key from file");
```
preston-evans98, line 275: nit: let's avoid panics

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+                let sender_address: Address = sender_priv_key.pub_key().to_address();
+                println!("{}", sender_address);
+            }
+
+            UtilCommands::CreatePrivateKey { priv_key_path } => {
+                PrivKeyAndAddress::generate_and_save_to_file(priv_key_path.as_ref())
+                    .unwrap_or_else(|e| panic!("Create private key error: {}", e));
```
preston-evans98, line 282: nit: let's avoid panics

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+            }
+            UtilCommands::PrintNamespace => {
+                println!("{}", hex::encode(ROLLUP_NAMESPACE_RAW));
+            }
+        },
+        Commands::Module(module_args) => {
+            let sender_priv_key_path = &module_args.private_key.unwrap();
+            let nonce = module_args.nonce.unwrap();
```
preston-evans98, line 290: Is it safe to unwrap here?

module-system/module-implementations/examples/sov-election/Cargo.toml
preston-evans98: Can we make this optional and gate it behind the native flag?

module-system/sov-modules-api/src/lib.rs
preston-evans98: @citizen-stig, I think 'static is already enforced in practice by `for<'a> Deserialize<'a>`. 

But if we don't strictly need the extra bound, I'm all for removing it :)

examples/demo-simple-stf/tests/stf_test.rs
dubbelosix: oh yeah, since this is just for a test. and we never call that. I can give a proper implementation as well, but the test case doesn't really use it. I can implement it if we think it's necessary (just need to copy any of the u8;32 addresses that we already have)

examples/demo-stf/src/runtime.rs
dubbelosix: so the thing is - The original method i tried was #[cli_skip] and it worked, but the problem is cargo hack fails when the native feature is not passed. for some reason, restricting cli_skip to also work only for native doesn't seem to work.

#[cfg_attr(feature = "native", cli_skip)]
seems to be ignored and compiler complains that Sequencer doesn;t have the necessary traits implemented.

additionally, i also tried defining cli_skip as a no-op proc macro, but it still doesn't seem to work.

I only did the second attribute to cli_parser after trying for a lot of time to fix the errors for #[cli_skip] with cargo hack. I think its fine if native is supplied, but in other cases, it's causing failures.

examples/demo-stf/src/sov-cli/native.rs
dubbelosix: I did this primarily to ensure sov-cli only runs with the native flag supplied, otherwise it was leading to a lot of conditional annotation on different imports, structs and enums - the code was full #[cfg(feature = "native")] - and since we need native for the entire binary, I just wrapped it in a module and i'm importing that only when native feature is enabled. I tried checking if there are ways in Cargo.toml to restrict that a binary can only be built with specific features but didn't find anything there

module-system/module-implementations/examples/sov-election/src/call.rs
dubbelosix: hmm yeah, that's a good idea. let me look into that. the main reason i passed the module_name is because we didnt have access to it, but we can build that into the trait

module-system/module-implementations/examples/sov-value-setter/Cargo.toml
dubbelosix: yup, this was a mistake. i'll enable it only for native. this is also the reason the prover is failing in CI - there are a few crates i included while testing but didn't restrict them to native

examples/demo-simple-stf/tests/stf_test.rs
bkolad: Someone can start using it in the future, so we should provide the implementation.

examples/demo-stf/src/runtime.rs
bkolad: I see two approaches:
1. Can you check what serde is doing? I think some crates derive serde for they datatypes only when serde feature is enabled (and they can use field attributes)
2. If `CallMessage` is not specified for a module, the type defaults to `pub enum NonInstantiable {}` we could potentially derive the clip machinery for it as well.

We should explore option1 first since it is less invasive on the `sov-modules-api`

module-system/module-implementations/sov-bank/src/token.rs
bkolad: Actually, where is this code used?

module-system/sov-modules-macros/src/cli_parser.rs
```diff
@@ -25,49 +29,111 @@ impl CliParserMacro {
...
+        let mut command_types = vec![];
+        let mut module_commands = vec![];
+        let mut module_args = vec![];
+        let mut match_arms = vec![];
+        let mut command_idents = vec![];
+        let mut parse_match_arms = vec![];
+
+        // Loop over the fields
+        for field in &fields {
+            if skip_fields.contains(&field.ident.to_string()) {
```
bkolad, line 43: This complicated loop, would you mind describing what does it do?

rollup-interface/src/state_machine/mocks.rs
```diff
@@ -98,6 +99,21 @@ pub struct MockAddress {
     addr: [u8; 32],
 }
 
+impl FromStr for MockAddress {
+    type Err = anyhow::Error;
```
bkolad, line 103: thiserror

module-system/sov-modules-macros/src/cli_parser.rs
```diff
@@ -97,6 +181,174 @@ impl CliParserMacro {
...
+
+            impl #impl_generics From<#cli_enum_with_fields_ident #ty_generics> for #enum_name #ty_generics #where_clause {
+                fn from(item: #cli_enum_with_fields_ident #ty_generics) -> Self {
+                    match item {
+                        #(#convert_cases)*
+                    }
+                }
+            }
+
+            impl<C: sov_modules_api::Context> sov_modules_api::AutoClap for #trait_type_ident<C> {
```
bkolad, line 344: Hardcoded "C"

module-system/sov-modules-api/src/lib.rs
dubbelosix: oh i already removed 'static requirement from a bunch of places. Needed it earlier, but we don't anymore. I think i missed this place. removed the bound and it works ok

examples/demo-stf/src/app.rs
```diff
@@ -26,7 +26,7 @@ use sov_state::ProverStorage;
 use sov_state::Storage;
 use sov_state::ZkStorage;
 
-pub struct DemoAppRunner<C: Context, Vm: Zkvm> {
+pub struct DemoAppRunner<C: Context + 'static, Vm: Zkvm> {
```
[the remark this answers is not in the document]
citizen-stig: +1

examples/demo-stf/src/sov-cli/main.rs
```diff
@@ -1,436 +1,12 @@
...
-        println!(
-            "private key written to path: {}",
-            path.into_os_string().into_string().unwrap()
-        );
-        Ok(())
-    }
+#[cfg(feature = "native")]
+mod native;
+#[cfg(feature = "native")]
+#[cfg(feature = "native")]
```
citizen-stig, line 4: I wonder, why is it duplicated on 2 lines?

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+use demo_stf::runtime::cmd_parser;
+
+use sov_modules_api::{
+    default_context::DefaultContext, default_signature::private_key::DefaultPrivateKey,
+    AddressBech32, PublicKey, Spec,
+};
+
+use demo_stf::runtime::{module_parse_helper, ModuleCommands};
+
+type C = DefaultContext;
```
citizen-stig, line 21: Does it mean, that cli is only available for DefaultContext? What if another context is used?

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,487 @@
...
+        /// Modules defined in your Runtime are supported.
+        /// (eg: Bank, Accounts)
+        module_name: String,
+        /// Path to the json file containing the parameters for a module call
+        call_data_path: String,
+        /// Nonce for the transaction
+        nonce: u64,
+    },
+    /// Combine a list of files generated by SerializeCall into a blob for submission to Celestia
+    MakeBlob {
```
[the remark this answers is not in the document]
citizen-stig: I vote for `MakeBatch`

examples/demo-stf/src/sov-cli/native.rs
citizen-stig: +1

examples/demo-stf/src/sov-cli/native.rs
citizen-stig: Yes, it should never panic. It can only panic, if it fails to allocate vec. I think `Result` signature is leaked from borsh serializing into io.

examples/demo-stf/src/sov-cli/main.rs
```diff
@@ -1,436 +1,12 @@
-use anyhow::Context;
```
citizen-stig, line 1: Why sov-cli lives in demo-stf? Can it be its own generic crate?

module-system/module-implementations/examples/sov-election/src/call.rs
preston-evans98: @bkolad If we add a `fn module_name()` to the `ModuleInfo`, we still won't have access to it in this context right? Or am I missing something?

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,546 @@
...
+        }
+        Commands::MakeBatch { path_list } => {
+            let mut hex_encoded_txs = vec![];
+            for path in path_list {
+                let mut file =
+                    File::open(&path).with_context(|| format!("Unable to create file {}", path))?;
+                let mut hex_string = String::new();
+                file.read_to_string(&mut hex_string)
+                    .context("Unable to read the file")?;
+                // push it into data_list
```
bkolad: Looks like this comment is outdated.

examples/demo-stf/src/sov-cli/native.rs
```diff
@@ -0,0 +1,546 @@
...
+                hex_encoded_txs.push(hex_string);
+            }
+
+            let blob = make_hex_blob(hex_encoded_txs.into_iter())?;
+            println!("{}", blob)
```
bkolad, line 301: Let's remove this print or add a message which says what is being printed.

module-system/sov-modules-macros/src/cli_parser.rs
```diff
@@ -14,90 +16,416 @@ impl CliParserMacro {
...
+        // Implement the `CliWalletArg` trait for the original type. This is what allows the original type to be used as a CLI arg.
+        impl #impl_generics sov_modules_api::CliWalletArg for #item_name #ty_generics #where_clause {
+            type CliStringRepr = #clap_type;
+        }
+    };
+    Ok(expanded.into())
+}
+
+/// A struct or enum field with a name and type
+pub struct NamedFieldWithDocs {
```
bkolad: We have a`StructNamedField` type in `common.rs`  which looks similar. Is it possible to extend `StructNamedField` and use it instead of `NamedFieldWithDocs`? Can be done in another PR.

module-system/sov-modules-macros/src/common.rs
```diff
@@ -213,3 +219,148 @@ pub fn get_serialization_attrs(item: &syn::DeriveInput) -> Result<Vec<TokenStrea
...
+/// // We want to extract both the inline bounds, and the bounds from the where clause...
+/// // so that the generics from above definition are equivalent what we would have gotten
+/// // from writing `T: SomeTrait + SomeOtherTrait` inline
+/// let desired_bounds_for_t: syn::TypeParam = syn::parse_quote!(T: SomeTrait + SomeThirdTrait);
+///
+/// // That is exactly what `GenericTypesWithBounds` does
+/// let our_bounds = extract_generic_type_bounds(&test_struct.generics);
+/// assert_eq!(our_bounds.get(T), Some(&desired_bounds_for_t.bounds));
+/// ```
+pub fn extract_generic_type_bounds(
```
bkolad: This method can be crate private?

module-system/sov-modules-macros/src/common.rs
```diff
@@ -213,3 +219,148 @@ pub fn get_serialization_attrs(item: &syn::DeriveInput) -> Result<Vec<TokenStrea
...
+                    }
+                }
+            }
+        }
+    }
+    generics_with_bounds
+}
+
+/// Extract the type ident from a `TypePath`.
+pub fn extract_ident(type_path: &syn::TypePath) -> &Ident {
```
bkolad, line 354: Can we make it crate private?

bkolad (approved):
Thanks a lot, Preston!  I had a couple of old comments about introducing thiserror: https://github.com/Sovereign-Labs/sovereign-sdk/pull/423/files#r1238171744. I think you have mentioned there were some problems with doing this way, right? If this is the case, feel free to close them.

examples/demo-simple-stf/tests/stf_test.rs
preston-evans98: Done!

examples/demo-stf/src/runtime.rs
preston-evans98: This uses an attribute now 👍

module-system/module-implementations/examples/sov-election/src/call.rs
preston-evans98: This issue is also resolved by the new macro

module-system/module-implementations/sov-bank/src/token.rs
preston-evans98: Done

module-system/module-implementations/examples/sov-value-setter/Cargo.toml
preston-evans98: Fixed

module-system/sov-modules-macros/src/cli_parser.rs
preston-evans98: Fixed by the new macro

adapters/celestia/src/celestia.rs
```diff
@@ -335,6 +336,22 @@ impl From<[u8; 32]> for H160 {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // Remove the "0x" prefix, if it exists.
+        let s = s.strip_prefix("0x").unwrap_or(s);
+        let bytes = hex::decode(s)?;
+        if bytes.len() != 20 {
+            return Err(anyhow::anyhow!("Invalid address length"));
```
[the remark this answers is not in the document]
preston-evans98: We changed to a simpler invocation of `hex::decode` and return their error type directly now.

examples/demo-stf/src/sov-cli/native.rs
preston-evans98: The cli is now available for any native Context

examples/demo-stf/src/sov-cli/main.rs
preston-evans98: Yes, we'll move this to a separate crate as part of #552

examples/demo-stf/src/sov-cli/native.rs
preston-evans98: Removed!

module-system/sov-modules-macros/src/cli_parser.rs
preston-evans98: Migrated!

module-system/sov-modules-macros/src/common.rs
preston-evans98: Done

module-system/sov-modules-macros/src/common.rs
preston-evans98: Done!

PR #424, opened 2023-06-21
Implement EvmDb for WorkingSet

bkolad:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/module-system/module-implementations/sov-evm/src/evm/mod.rs
+++ b/module-system/module-implementations/sov-evm/src/evm/mod.rs
@@ -1,4 +1,53 @@
-mod db;
+use sov_state::Prefix;
+
+mod conversions;
+pub(crate) mod db;
+mod db_commit;
+mod db_init;
 mod executor;
 #[cfg(test)]
 mod tests;
+
+pub(crate) type Address = [u8; 20];
+pub(crate) type SovU256 = [u8; 32];
+
+// Stores information about an EVM account
+#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq, Clone, Default)]
+pub(crate) struct AccountInfo {
+    pub(crate) balance: SovU256,
+    pub(crate) code_hash: SovU256,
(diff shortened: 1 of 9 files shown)


module-system/module-implementations/sov-evm/src/evm/mod.rs
```diff
@@ -1,4 +1,54 @@
...
+    // https://github.com/Sovereign-Labs/sovereign-sdk/issues/425
+    pub(crate) code: Vec<u8>,
+    pub(crate) nonce: u64,
+}
+
+/// Stores information about an EVM account and a corresponding account state.
+#[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq, Clone)]
+pub(crate) struct DbAccount {
+    pub(crate) info: AccountInfo,
+    pub(crate) storage: sov_state::StateMap<SovU256, SovU256>,
```
preston-evans98, line 29: (question): It looks like we're saving the `StateMap` into state (which in practice means serializing and storing the prefix for the inner Map). Do we need to do that? Or can we reconstruct the prefix from the Account address whenever we need it?

preston-evans98 (approved):
Looks good overall! Approving to make sure the work stream stays unblocked, but would love to hear back on the comment :)

module-system/module-implementations/sov-evm/src/evm/mod.rs
bkolad: We could do that, I did it this way for 2 reasons:
- tokens in bank already work this way (but we could change that too if we decide to)
- here is how we create the account storage prefix: 
```Rust
let mut prefix = parent_prefix.as_aligned_vec().clone().into_inner();
prefix.extend_from_slice(&address);
```
If we reconstruct the prefix when we need it would incur some additional work on every state access.

preston-evans98: Interesting. I think reconstructing it once (when we load the account) will likely still be a lot cheaper than reading from storage, since storage data gets hashed. But that's an optimization - no need to worry about it for now

PR #430, opened 2023-06-23
Sequencer crate

### LINKED ISSUE #357: Implement Send Transaction RPC

citizen-stig:
# Description

Moves Sequencer to different crate and adds submit transaction RPC method.

So currently Sequencer crate exposes these 2 methods:

1. `sequencer_acceptTx` where input is suppose to be signed and serialized transaction. This transaction is stored in mempool
2. `sequencer_publishBatch` without any input, which builds the batch using batch builder and publishes it on DA layer.

## What is not included in this PR

* Verification of transactions on registration
* Adjusting CLI to work with Sequencer

## Linked Issues
- Fixes #357
- Related to #391

## Testing
Unit tests have been added

## Docs
No new docs have been added


examples/demo-stf/src/batch_builder.rs
```diff
@@ -89,6 +89,7 @@ where
 
             // Execute
             {
+                // TODO: Bug(!), because potential discrepancy with default stf or any other STF
```
preston-evans98, line 92: Let's handle this TODO (either by fixing it or adding an issue) I think this will be easier to resolve after we finish #361, so it's probably fine to just create an issue and link here.

examples/demo-stf/src/batch_builder.rs
```diff
@@ -89,6 +89,7 @@ where
 
             // Execute
             {
+                // TODO: Bug(!), because potential discrepancy. Should be resolved by https://github.com/Sovereign-Labs/sovereign-sdk/pull/361
```
citizen-stig, line 92:
```suggestion
                // TODO: Bug(!), because potential discrepancy. Should be resolved by https://github.com/Sovereign-Labs/sovereign-sdk/issues/434
```

examples/demo-stf/src/batch_builder.rs
citizen-stig: Agree, added link to the issue

rollup-interface/src/node/services/stf_runner.rs
```diff
@@ -20,6 +20,7 @@ use crate::zk::traits::Zkvm;
 ///
 /// and a `impl StateTransitionRunner<ZkConfig, Vm> for MyRunner` which instead uses a state root as its runtime config.
 ///
+/// TODO: Why is it called runner? It only creates. Creator, Factory.
```
citizen-stig, line 23:
```suggestion
/// TODO: Why is it called runner? It only creates. Creator, Factory: https://github.com/Sovereign-Labs/sovereign-sdk/issues/447
```

preston-evans98 (approved):

PR #436, opened 2023-06-28
docs: readme edits to demo-rollup

jcstein:
# Description
- [x] I think that "params: should be the token address created in step 6" should link to step 5, not 6

## Docs


jcstein:
nope, misread it

PR #437, opened 2023-06-28
Fix Jupiter and Sov-CLI bugs

preston-evans98:
# Description
This PR fixes several bugs related to sequencing and transaction inclusion:
1. The genesis block on Celestia does not have a `prev_hash`, which caused a panic when running on local testnets
2. Jupiter had an erroneous `copy_from_slice` which copied into an empty vec with enough capacity but 0 length. This triggered an assertion when processing blobs
3. The makefile command to submit the `create-token` transaction to Celestia submitted an individual transaction instead of a batch, causing the transaction to be ignored
4. The `sov-cli` had a serialization bug, causing all of its batches to be rejected

### DIFF
--- a/adapters/celestia/src/verifier/mod.rs
+++ b/adapters/celestia/src/verifier/mod.rs
@@ -236,8 +236,7 @@ impl da::DaVerifier for CelestiaVerifier {
                 let blob_ref = blob.clone();
 
                 let mut blob_iter = blob_ref.data();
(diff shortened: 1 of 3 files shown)


citizen-stig:
I am nitpicking, but for future it might be a good idea to give branches more descriptive names, like `hotfix/cli_and_jupiter_bugs`

citizen-stig (approved):

PR #439, opened 2023-06-28
Adjust cache for cargo

citizen-stig:
# Short summary

* Add Swatinem/rust-cache@v2 before sccache, so we can combine crates/target cache with recompilation.
* Adjust macro tests to not require native flag
* change `assert!(matches!())` to `assert_eq!()`
* Switch to use latest `sccache` instead of hard coded.

# Description

Currently we use [sccache](https://github.com/mozilla/sccache) for caching compilation. and it helps, especially with rocksdb, as it can cache not only rust compilation artifacts, but also C/C++.

But it looks like cargo download dependencies everytime and scacche does not handle it that well.

So idea is to put sccache on top of standard rust caching mechanism.

| Iteration                              | Check | Test | Coverage | Features | Check Prover |
| -------------------------- | ------ | ----- | --------- | --------- | ------------ |
| 1 (empty cache)                  | 13m   | 23m  | 23m        | 30m         | 45m             |
| 2                                           | 9 m   | 19m  | 20m       | 23m         | 22m             |
| 3 (add cache for `target`) | 12m   | 26m  | 20m       | 25m         | 24m             |
| 4                                           |  2m   |  7m  |   20m       | 3m         | 10m             |
| 5                                           |  1m   | 9m  | 10m       | 4m         | 14m             |
| 6 (adjust macro test)            |  12m   | 15m  | 17m       | 28m         | 16m             |
| 7                                           | 1m   | 9m  | 7m       | 3m         | 10m             |

## Macro tests

Current macro tests use native flag for testing compilation, which brings compilation of rocksdb second time inside macro test. 
I've changed the tests to operate on ZK conteext and ZK storage, as it require less dependencies.

# Future improvements

Things that can be done to make CI even smoother:

1. Set `check` step to be the first and rest depend on it, so it doesn't waste runner's time if it fails
2. Figure out why llvm-cov spends 10 minutes recompiling.
3. tune `SCCACHE_CACHE_SIZE` to get best results
4. Try [cargo-nextest](https://nexte.st/)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,6 +17,7 @@ jobs:
   check:
     runs-on: ubuntu-latest
     name: check
+    timeout-minutes: 60
     env:
       RUSTFLAGS: -D warnings
       SCCACHE_GHA_ENABLED: "true"
@@ -37,6 +38,7 @@ jobs:
   check-demo-prover:
     runs-on: ubuntu-latest
     name: check demo prover
+    timeout-minutes: 90
     env:
       RUSTFLAGS: -D warnings
       SCCACHE_GHA_ENABLED: "true"
@@ -59,6 +61,7 @@ jobs:
   hack:
     runs-on: ubuntu-latest
     name: features
+    timeout-minutes: 60
     env:
       RUSTFLAGS: -D warnings
       SCCACHE_GHA_ENABLED: "true"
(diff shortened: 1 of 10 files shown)


bkolad (approved):

neysofu (approved):
The benchmarks are very impressive and all changes make sense to me. It would be nice to include a timeout (1h?) for our GH jobs, to save compute in case the CI gets stuck (it doesn't look like it happens at all by glancing at our Actions history but it's a good fail-safe to have, especially since we're doing proptests).

citizen-stig:
> The benchmarks are very impressive and all changes make sense to me. It would be nice to include a timeout (1h?) for our GH jobs, to save compute in case the CI gets stuck (it doesn't look like it happens at all by glancing at our Actions history but it's a good fail-safe to have, especially since we're doing proptests).

Agree, done!

PR #442, opened 2023-06-28
add basic demo rollup  + local celestia check to github workflow

dubbelosix:
# Description

we had multiple instances where bugs were introduced that broke the e2e flow for demo-rollup.
to solve this, we want to run the local setup as a github workflow - this would serve as an integration test that would cover the celestia adapter, sov-cli and demo-rollup and would catch unintentional breakage.

what the workflow does
1. launch celestia docker container
2. run the sov demo-rollup which connects to docker celestia
3. submit a create token transaction
4. simple grep check on RPC to verify if the curl output contains "1000" which is the supply of token - could use jq and parse, but considering this won't change, it seems like a simple check

we can add more complex transactions to it - bulk of time is in spinning up the rollup and local celestia, so we can have multiple modules tested - bank, transfers etc


preston-evans98 (approved):
This will be a huge win, thanks @dubbelosix!

bkolad (approved):

.github/workflows/rust.yml
```diff
@@ -114,3 +114,44 @@ jobs:
...
+    runs-on: ubuntu-latest
+    env:
+      SCCACHE_GHA_ENABLED: "true"
+      RUSTC_WRAPPER: "sccache"
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.3
+        with:
+          version: "v0.4.0"
```
citizen-stig, line 127: Nit: in latest main I've removed these 2 lines, so there's always latest sccache

PR #443, opened 2023-06-28
Use big-endian for serializing `u64_wrapper` keys in RocksDB

neysofu:
This PR changes the way we serialize `u64_wrapper` keys to RocksDB from Borsh (little-endian) to big-endian. I wanted to remove the `derive(Borsh::BorshSerialize)` from these types to make misuse harder, but they are needed in other parts of the codebase.  I'm basing my PR off of https://github.com/Sovereign-Labs/sovereign-sdk/pull/403, which is the branch where we discovered https://github.com/Sovereign-Labs/sovereign-sdk/issues/417 in the first place.

## Linked Issues
- Fixes https://github.com/Sovereign-Labs/sovereign-sdk/issues/417.
- Potentially interesting: https://github.com/facebook/rocksdb/issues/3780.

## Testing
I modified `regular_test_helper` in `test_rpc.rs` to generate a second batch with > 256 transactions, which successfully reproduces the fixed bug. I also made two unrelated changes to `test_rpc.rs`:
- Use the `jsonrpc_req!` and `jsonrpc_result!` macros to facilitate JSON string generation inside tests.
- ~~Use `assert_json_diff!` for more user-friendly JSON comparisons.
I would imagine these changes could result in somewhat slower tests. It wasn't noticeable to me during local development but we can reevaluate if it's significant.~~ EDIT: removed after discussion with @preston-evans98 .

## Docs
I've added some doc comments to the added macros.


full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -157,7 +196,7 @@ define_table_with_default_codec!(
     (BatchByHash) DbHash => BatchNumber
 );
 
-define_table_with_default_codec!(
+define_table_with_u64_wrapper_keys!(
     /// The primary source for transaction data
     (TxByNumber) TxNumber => StoredTransaction
 );
```
preston-evans98, line 202: It looks like we're still using the default (little endian) serializer for `EventByNumber` and `SlotNumber`

preston-evans98 (changes requested):

full-node/db/sov-db/src/schema/tables.rs
neysofu: Fixed in https://github.com/Sovereign-Labs/sovereign-sdk/pull/443/commits/ae4c3d894b5f6d0609a26b784b3458fde56f34b7. I also renamed `define_table_with_u64_wrapper_keys` to `define_table_with_seek_key_codec` and switched to big-endian `bincode` to make it more general-purpose, in preparation for using it on other types such as `NodeKey`. In the meantime I'm removing the blanket `SeekKeyEncoder` implementation to avoid misuse as much as possible.

Issue #413, opened 2023-06-16
Implement `evm` module


bkolad:
We will use the following `evm` implementation: https://github.com/bluealloy/revm
List of tasks
- [x] Create a simple `sov-evm` module with a unit test that executes a simple contract in revm using the `CachedDb`
- [x] Implement the `Database` trait (from revm) for `WorkingSet` and update the aforementioned test to work with the `WorkingSet`.
- [x] Transform the second point into a separate module.
- [x] Implement `eth_sendRawTransaction` RPC Method
- [ ] Determine how to handle `block_hash, coinbase, block_number, etc`.
- [ ] Ensure that the RPC functions work with standard wallets.

Related issues:
#425, #425

LukaszRozmej:
Options and issues after doing some research:

# We need to have the functionality of:

* Evm
* Some subset of **eth** RPC module for wallets
* Evm sequencer and TxPool
* Chain of latest blocks and their receipts (for RPC)
* Way to keep Ethereum-like state with some tree-like structure and state root.

Unknown:
* Do we need tracers?
* Do we need archive?

# Way of integrating with rollup

Evm module single transaction could encode:

* Whole Evm block 
   \+ Similar to Ethereum
   \+ Maps to RPC directly
   \- Potentially big data set.

* Single Evm transaction
   \+ simple 1:1 mapping
   \+ fairly small
   \+ EVM works on transaction level
   \- RPC expects blocks, need to create an artificial block-like thing

After executing Evm transaction the rollup would keep:
* Ethereum-like state in terms of Accounts, Storage Slots, contract ByteCode and some kind of state tree
* Blocks (quaryable by number or by hash)
* Receipts (quaryable by number or by block hash)
* Transactions (quaryable by hash)
* AccountChangeSets? (if archive needed)
* StorageSlotsChangeSets? (if archive needed)

# Implementation options

*  Minimal dependency - use [Revm](https://github.com/bluealloy/revm/) and implement rest ourselves.
   * Need to implement:
      * Storage for Blocks, Receipts, Transactions
      * All the needed RPC
      * Sequencer and TxPool
   * Most flexible
      * Can utilize our storage model for state
      * Easily customizable implementation
      * Easy read-only access to the data from other modules (potentially even write access)
* Including [Reth](https://github.com/paradigmxyz/reth) in rollup
   * Only need integrations to implement
[...]

PR #449, opened 2023-06-29
Implement `call` method for the `EVM module`

bkolad:
# Description
- `BlockEnv` & `EvmTransaction` types which are used in the `EVM` module.
-  Conversions between the aforementioned types and relevant `ReVM` types.
-  Implementation of `call` method for the `EVM` module
- `EVM` module tests

## Linked Issues
- Related to #413

## Testing
Added tests for the `EVM` module.

## Docs
The documentation will be provided after the module is fully implemented.

### DIFF
--- a/module-system/module-implementations/sov-evm/src/evm/conversions.rs
+++ b/module-system/module-implementations/sov-evm/src/evm/conversions.rs
@@ -1,8 +1,11 @@
 use bytes::Bytes;
 
+use super::executor::{BlockEnv, EvmTransaction};
 use super::AccountInfo;
-use revm::primitives::{AccountInfo as ReVmAccountInfo, U256};
-use revm::primitives::{Bytecode, B256};
+use revm::primitives::{
+    AccountInfo as ReVmAccountInfo, CreateScheme, TransactTo, TxEnv, B160, U256,
+};
+use revm::primitives::{BlockEnv as ReBlockEnv, Bytecode, B256};
 
 impl From<AccountInfo> for ReVmAccountInfo {
     fn from(info: AccountInfo) -> Self {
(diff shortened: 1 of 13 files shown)


module-system/module-implementations/sov-evm/src/evm/conversions.rs
```diff
@@ -25,3 +30,59 @@ impl From<ReVmAccountInfo> for AccountInfo {
...
+            // TODO: handle difficulty
+            difficulty: U256::ZERO,
+            prevrandao: block_env.prevrandao.map(|r| B256::from_slice(&r)),
+            basefee: U256::from_le_bytes(block_env.basefee),
+            gas_limit: U256::from_le_bytes(block_env.gas_limit),
+        }
+    }
+}
+
+impl From<AccessListItem> for (B160, Vec<U256>) {
```
citizen-stig, line 49: What does this tuple mean? Maybe we should create a regular struct with field names?

module-system/module-implementations/sov-evm/src/evm/test_helpers.rs
```diff
@@ -0,0 +1,39 @@
...
+use revm::primitives::{ExecutionResult, Output, B160};
+use std::path::PathBuf;
+
+pub(crate) fn output(result: ExecutionResult) -> Bytes {
+    match result {
+        ExecutionResult::Success { output, .. } => match output {
+            Output::Call(out) => out,
+            Output::Create(out, _) => out,
+        },
+        _ => panic!("Expected successful ExecutionResult"),
```
citizen-stig, line 13: Is it ok to panic here? Maybe use `TryFrom` ?

citizen-stig: oh, it is foreign type, but maybe we can return result

citizen-stig: It is for tests, nevermind

module-system/module-implementations/sov-evm/src/evm/tests.rs
```diff
@@ -1,48 +1,25 @@
-use crate::{evm::AccountInfo, Evm};
```
citizen-stig, line 1: What is the reason of it to be unit tests and not integration test?

citizen-stig: I mean placing test outside of src and testing crate via public interface

module-system/module-implementations/sov-evm/src/evm/conversions.rs
bkolad: This is why we have `AccessListItem`, but `revm` expects the above tuple (it is part of the interface and we can't change it) this is why we provide this conversions.

module-system/module-implementations/sov-evm/src/evm/test_helpers.rs
bkolad: This is test helper and in testes I prefer to panic early.

module-system/module-implementations/sov-evm/src/evm/tests.rs
bkolad: The core logic of these tests relies on the public `EVM::call` method and could be separated into an integration test. However, to verify the success of the call, we are currently accessing the crate-private  `EVM` state. While we can make the appropriate query method public, it is currently tailored specifically for testing purposes. Instead, I suggest that we postpone changing the test to an integration test until we implement the `RPC` functionality for the `EVM` module. By doing so, we will be able to examine the internal state using the `RPC`, providing a more comprehensive testing approach.

citizen-stig: Ok, as long as we plan to move to integration test eventually, I am good.

citizen-stig (approved):

PR #454, opened 2023-07-04
Move DB-related traits/items from `sov_rollup_interface` into `sov_schema_db`

neysofu:
# Description
We'd like to make the `sov_schema_db` crate more cohesive, and avoid spreading DB-related logic across multiple crates if possible.

## Linked Issues
None.

## Testing
No new features to test. I did move `db_test.rs` and `iterator_test.rs` into their own `tests/` directory, just to separate production logic from tests.

## Docs
We're just moving things around, nothing new to document. I went looking for `.md`'s that would need updating, but I couldn't find any reference to `sov_rollup_interface::db` in the first place. I think we're good.


bkolad (approved):
LGTM

PR #455, opened 2023-07-04
EVM module: `send_transaction` API test

bkolad:
# Description
1. `tx_rlp_encoding_test` shows how to sign and encode raw ethereum transaction.
2. `send_tx_test_to_eth_node` shows how to send transactions to a local ethereum node (`anvil`)
3. Adds `SimpleStorageContract` struct which simplifies call data creation for tests.
4.  `experimental` feature for the `EVM` module.
5. Adds `foundry-rs` github action to CI.

## Linked Issues
- Related to #413

## Testing
Added `tx_rlp_encoding_test & send_tx_test_to_eth_node` tests described above. 

## Docs
No docs changed.


module-system/module-implementations/sov-evm/src/tests/mod.rs
```diff
@@ -0,0 +1,2 @@
+mod call_tests;
```
citizen-stig, line 1: Why won't place tests as integration tests in `sov-evm/tests` ? Then it will be covering public API.

bkolad: We are still accessing private fields, the tests will be moved after we have a convenient rpc methods for accessing the module state

PR #458, opened 2023-07-04
Create benchmark scripts for demo-rollup running in native mode

dubbelosix:
# Description
We want to be able to measure the native performance of the SDK, and to this purpose we're benchmarking demo-rollup running in native mode. the process for the benching is as follows
* Generate 100 blocks containing 1 blob each with 10k transfer transactions (defaults, but can be configured)
* Once the blocks are loaded into memory, we start the timers and measure how long the blocks take to be processed.
* The primary pieces of code we are measure are
  * begin_slot
  * apply_blob
  * end_slot

demo-rollup/benches/README.md contains the detailed documentation

Tools used:
* criterion
* manual timers
* prometheus metrics (metrics only. no http server, no scraping and no visualization tools for now)
* flamegraph
* make

There are 2 scripts for benchmarking - 
* rollup_bench.rs - This uses criterion and pre-populates a "large" amount of data and lets the number of iterations be decided by criterion. we do constrain measurement time to 20 seconds to prevent too many iterations leading in an indexing error
* rollup_coarse_measure.rs
  * This uses coarse grained manual timers and outputs the timings as well the tps
  * Can also output the readings as a histogram for prometheus. the data is currently a json, but can be fed into visualization tools (this will be done in another issue)

Also contains a Makefile for wrapping all the scripts + cargo flamegraph (sample output shown)
Added README.md explaining the bench methodology 

## Linked Issues
- Related to # (356)


examples/demo-rollup/Cargo.toml
```diff
@@ -19,6 +19,8 @@ tracing = { workspace = true }
 hex = { workspace = true }
 bytes = { workspace = true }
 futures = "0.3"
+metrics = "0.17.0"
```
bkolad, line 22: Let's put the dependency in the workspace.

examples/demo-rollup/Cargo.toml
```diff
@@ -39,11 +41,23 @@ sov-modules-api = { path = "../../module-system/sov-modules-api", features = ["n
...
 reqwest = "0.11"
 tendermint = "0.32"
 tempfile = { workspace = true }
 proptest = { workspace = true }
 sov-rollup-interface = { path = "../../rollup-interface", features = ["fuzzing"] }
+prometheus = "0.11.0"
+prettytable-rs = "^0.10"
+glassbench = "0.3"
+criterion = "0.5.1"
+clap = "4.3.10"
```
bkolad, line 55: `clap` is used in a few places, let's put it in the workspace.

examples/demo-rollup/benches/rollup_coarse_measure.rs
```diff
@@ -0,0 +1,222 @@
...
+use prometheus::Histogram;
+use prometheus::HistogramOpts;
+use prometheus::Registry;
+
+#[macro_use]
+extern crate prettytable;
+use prettytable::Table;
+
+fn remove_dir_if_exists<P: AsRef<std::path::Path>>(path: P) -> io::Result<()> {
+    if path.as_ref().exists() {
```
bkolad, line 33: Let's make `test_helper` and put `remove_dir_if_exists` there it is implemented in two places.

examples/demo-rollup/rollup_config.toml
```diff
@@ -1,6 +1,6 @@
 # We define the rollup's genesis to occur at Celestia block number `start_height`. The rollup will ignore
 # any Celestia blocks before this height
-start_height = 671431
+start_height = 1
```
bkolad, line 3: Is this ok for the case when we run with real celestial?

bkolad (approved):

examples/demo-rollup/rollup_config.toml
dubbelosix: Ah, we use a different config file for benchmarks. Changing this part (It was from before when I didn't have a different config file for bench)

examples/demo-rollup/benches/rollup_bench.rs
```diff
@@ -0,0 +1,110 @@
...
+
+mod test_helper;
+
+use test_helper::remove_dir_if_exists;
+
+fn rollup_bench(_bench: &mut Criterion) {
+    let start_height: u64 = 0u64;
+    let mut end_height: u64 = 100u64;
+    if let Ok(val) = env::var("BLOCKS") {
+        end_height = val.parse().unwrap();
```
citizen-stig, line 32: Could we add message that will indicate that `BLOCK`  environment was incorrect number? 

```suggestion
        end_height = val.parse().expect("`BLOCK` environment variable should be positive integer.");
```

examples/demo-rollup/benches/rollup_bench.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    let mut c = Criterion::default()
+        .sample_size(10)
+        .measurement_time(Duration::from_secs(20));
+    let rollup_config_path = "benches/rollup_config.toml".to_string();
+    let rollup_config: RollupConfig = from_toml_path(&rollup_config_path)
+        .context("Failed to read rollup configuration")
+        .unwrap();
+
+    remove_dir_if_exists(&rollup_config.runner.storage.path).unwrap();
```
citizen-stig, line 43: Just a suggestion, that we can use temp folder here and completely ignore this parameter from config.

citizen-stig (approved):

dubbelosix:
> @dubbelosix I am trying to run benchmark on my baremetal machine with 16 cores and it doesn't seem to fit in 20 seconds:
> 
> ```
> $ make criterion
> Running criterion bench with 10000 transactions per block
> Method: Criterion
> Output: Criterion
>     Updating crates.io index
> ....
>     Finished bench [optimized] target(s) in 16m 08s
>      Running benches/rollup_bench.rs (/home/stig/workspace/sovereign-sdk/target/release/deps/rollup_bench-988dda5198a92fac)
>      Running benches/rollup_bench.rs (/home/stig/workspace/sovereign-sdk/target/release/deps/rollup_bench-988dda5198a92fac)
> Benchmarking rollup main loop
> Benchmarking rollup main loop: Warming up for 3.0000 s
> 
> Warning: Unable to complete 10 samples in 20.0s. You may wish to increase target time to 967.9s.
> Benchmarking rollup main loop: Collecting 10 samples in estimated 967.87 s (10 iterations)
> ```
> 
> What can be a problem? `htop` show no CPU activity at all.

so I checked this issue and it seems like you really can't control the exact amount of time criterion runs - it only takes the measurement time as a hint. tried looking to see if there are other frameworks that have more fine grained control, but didn't find too much. that's one of the reasons i also have the other "rollup_coarse_measure.rs"

examples/demo-rollup/src/rng_xfers.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+}
+
+struct MockBatchBuilder {
+    mempool: Vec<Vec<u8>>,
+}
+
+/// It only takes the first byte of the tx, when submits it.
+/// This allows to show effect of batch builder
+impl BatchBuilder for MockBatchBuilder {
```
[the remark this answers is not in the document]
dubbelosix: ah yes, its retained from some copy-pasting of the MockDaService. I removed it

PR #464, opened 2023-07-05
`rustfmt`: more deterministic imports formatting

neysofu:
# Description
I'm adding these options to a new `rustfmt.toml` file:

```
group_imports = "StdExternalCrate"
imports_granularity = "Module"
```

The idea is that we can cut down on the number of future merge conflicts by enforcing a specific imports ordering and style. These `rustfmt` options are unstable, so we'll only enjoy automatic linting of imports while we run Rust nightly, but we can leave the file there indefinitely as a preferred style documentation. Greppability is also improved.

## Linked Issues
None


neysofu:
The very next PRs will inevitably have some conflicts near the top of files if we merge this, no way around that, I'm afraid.

citizen-stig (approved):

PR #479, opened 2023-07-07
fix readme instructions and test data to run correctly when tutorial is followed exactly

dubbelosix:
The current README instructions when followed exactly will lead to a transaction getting reverted. This is because in our README, we have an automated/abstracted transaction that is used for sanity checking the setup and this creates a token with the address configured in `test_data`. Because this is just a sanity check, the subsequent instructions show how a transaction can be serialized manually using sov-cli and then submitted. Unfortunately, we also used `CreateToken` as the example there and this causes a nonce clash as well as an error (even if the nonce is fixed) that the token already exists. To solve this issue, we're modifying the README to use `Transfer` as the call for serializing and submitting manually using `sov-cli` and `make`

Additional fixes: The current test data is not compatible with itself, so modified the addresses to ensure CreateToken, Transfer, Burn are all coherent with each other

## Linked Issues
- Fixes # (476)

### DIFF
--- a/examples/demo-rollup/README.md
+++ b/examples/demo-rollup/README.md
@@ -192,36 +192,32 @@ pub enum CallMessage<C: sov_modules_api::Context> {
     },
 }
 ```
-In the above snippet, we can see that `CallMessage`s in `Bank` support five different types of calls. The `sov-cli` has the ability to parse a JSON file that aligns with any of these calls and subsequently serialize them. The structure of the JSON file, which represents the call, closely mirrors that of the Enum member. Consider the `CreateToken` message as an example:
+In the above snippet, we can see that `CallMessage`s in `Bank` support five different types of calls. The `sov-cli` has the ability to parse a JSON file that aligns with any of these calls and subsequently serialize them. The structure of the JSON file, which represents the call, closely mirrors that of the Enum member. Consider the `Transfer` message as an example:
 ```rust
-    CreateToken {
-        /// Random value use to create a unique token address.
-        salt: u64,
-        /// The name of the new token.
-        token_name: String,
-        /// The initial balance of the new token.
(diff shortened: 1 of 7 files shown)


examples/demo-rollup/README.md
```diff
@@ -192,37 +192,33 @@ pub enum CallMessage<C: sov_modules_api::Context> {
     },
 }
 ```
-In the above snippet, we can see that `CallMessage`s in `Bank` support five different types of calls. The `sov-cli` has the ability to parse a JSON file that aligns with any of these calls and subsequently serialize them. The structure of the JSON file, which represents the call, closely mirrors that of the Enum member. Consider the `CreateToken` message as an example:
+In the above snippet, we can see that `CallMessage`s in `Bank` support five different types of calls. The `sov-cli` has the ability to parse a JSON file that aligns with any of these calls and subsequently serialize them. The structure of the JSON file, which represents the call, closely mirrors that of the Enum member. Consider the `Transfer` message as an example:
```
bkolad, line 195: I just saw that @neysofu also changed this file in this PR: https://github.com/Sovereign-Labs/sovereign-sdk/pull/477

neysofu: Right – I think getting the fix out asap is more important though, @dubbelosix feel free to merge first and I'll fix the conflicts later.

neysofu (changes requested):
The token address must also be replaced in the `curl` example under the "Verify the supply of the new token created" section. As for everything else, LGTM!

dubbelosix:
> The token address must also be replaced in the `curl` example under the "Verify the supply of the new token created" section. As for everything else, LGTM!

good catch. fixed it!

neysofu (approved):

Issue #438, opened 2023-06-28
Cleanup: Add missing docs/tests to existing code


preston-evans98:
For each of the crates listed: 
1. Add all missing docs (use the #[deny(missing_docs)] flag to enforce this)
2. Implement `proptest::Arbitrary` for all relevant types
3. Ensure adequate test coverage
4. Refactor/re-organize if necessary
5. Re-check and update logs (remove extraneous debug info, format nicely, make sure log levels are reasonable)
6. Enable doc tests.
- [ ] adapters
  - [ ] Celestia 
    - [ ] 1 
    - [ ] 2
    - [ ] 3
    - [ ] 4
    -  Skip until rewrite
  - [ ] Risc0 (@citizen-stig)
     - [ ] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
 - [ ] examples
   - [ ] const-rollup-config (@theochap)
     - [x] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
   - [ ] demo-nft-module (@citizen-stig)
     - [ ] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
   - [ ] demo-prover (@dubbelosix)
     - [ ] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
   - [ ] demo-rollup (@theochap)
     - [x] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
   - [ ] demo-stf
     - [ ] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
  - [ ] demo-simple-stf (@bkolad)
     - [ ] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
 - [ ] full-node
   - [x] sov-db 
     - [ ] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
   - [] sov-schema-db
     - [x] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
 - [ ] module-system (prover-incentives + bank @theochap,)
   - [ ] app-template
     - [x] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
   - [ ] sov-accounts
     - [x] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
   - [ ] sov-bank 
     - [x] 1 
     - [ ] 2
     - [ ] 3
     - [ ] 4
   - [ ] prover-incentives
     - [x] 1 
[...]

PR #483, opened 2023-07-10
Add missing docs to rollup-interface. Cleanup module structure.

preston-evans98:
# Description
- Add missing docs to `rollup-interface`
- Cleanup unneeded code instead of documenting it
   - Remove `MerkleHasher` trait
   - Remove `CanonicalHash` trait and instead add a `hash` method to `BlockHeaderTrait`
   - Remove the standalone `traits` module, and move its contents into either `state_machine/mod.rs` or the `state_machine/da.rs` files as appropriate.
   - Remove unnecessary nesting of modules; everywhere that there was a top-level module with no direct contents and exactly one child module, bring replace the parent with the child.

**N.B. This PR makes no functional changes - only minor refactoring and docs improvements**

## Linked Issues
- Related to #438


bkolad (approved):

neysofu (approved):

rollup-interface/src/node/rpc/mod.rs
```diff
@@ -104,110 +129,165 @@ impl Default for QueryMode {
     }
 }
 
+/// The body of a response to a JSON-RPC request for a particular slot.
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub struct SlotResponse<B, Tx> {
+    /// The slot number
     pub number: u64,
+    /// The hex encoded slot hash
```
citizen-stig, line 137: Should we mention in other places that hash suppose to be hex encoded?

rollup-interface/src/node/rpc/mod.rs
```diff
@@ -77,24 +90,36 @@ pub enum EventIdentifier {
...
 }
 
-/// An identifier that specifies a single slot
+/// An identifier that specifies a single slot.
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum SlotIdentifier {
-    Hash(#[serde(with = "rpc_hex")] [u8; 32]), // the hash of a da block
-    Number(u64),                               // the block number of a da block
+    /// The hash of the slot (i.e. the da layer's block hash).
```
citizen-stig, line 103: Should we specify that it is hex encoded as in other places.

rollup-interface/src/node/rpc/mod.rs
preston-evans98: Good idea, I see you flagged one that I missed - I'll do a quick check for others

PR #500, opened 2023-07-17
De-anonymizing BlobTransactionTrait in STF

### LINKED ISSUE #465: `SequencerOutcome` should have generic over address trait for better logging
Currently if sequencer is slashed, sequencer address from enum is printed just as array of bytes. Each DA layer might have it's own human redable address and in order to make it easier to operate SDK, we should print human readable address where it is possible.

citizen-stig:
# Description

Currently STF has anonymous `BlobTransactionTrait`:

```rust

pub trait StateTransitionFunction<Vm: Zkvm> {
  // ...
  
      fn apply_blob(
        &mut self,
        blob: &mut impl BlobTransactionTrait,
        misbehavior_hint: Option<Self::MisbehaviorProof>,
    ) -> BatchReceipt<Self::BatchReceiptContents, Self::TxReceiptContents>;
}
```

 In order to solve #465 , this trait needs to be de-anonymized, so then `Address` associated type can be linked to `BatchReceiptContent` or other enums that need it.

This PR is first step towards this change, by bringing this trait to trait definition of STF

## Linked Issues
- Related to #465 

## Testing
All automated tests have passed

## Docs
No change in the documentation, since change is only cosmetical.

### DIFF
--- a/examples/demo-simple-stf/src/lib.rs
+++ b/examples/demo-simple-stf/src/lib.rs
@@ -15,7 +15,7 @@ pub enum ApplyBlobResult {
     Success,
 }
 
-impl<VM: Zkvm> StateTransitionFunction<VM> for CheckHashPreimageStf {
+impl<Vm: Zkvm, B: BlobTransactionTrait> StateTransitionFunction<Vm, B> for CheckHashPreimageStf {
     // Since our rollup is stateless, we don't need to consider the StateRoot.
     type StateRoot = ();
 
@@ -49,7 +49,7 @@ impl<VM: Zkvm> StateTransitionFunction<VM> for CheckHashPreimageStf {
     // The core logic of our rollup.
     fn apply_blob(
         &mut self,
-        blob: &mut impl BlobTransactionTrait,
+        blob: &mut B,
         _misbehavior_hint: Option<Self::MisbehaviorProof>,
     ) -> BatchReceipt<Self::BatchReceiptContents, Self::TxReceiptContents> {
         let blob_data = blob.data_mut();
--- a/examples/demo-prover/host/src/main.rs
+++ b/examples/demo-prover/host/src/main.rs
@@ -8,6 +8,7 @@ use demo_stf::runner_config::{from_toml_path, Config as RunnerConfig};
 use jupiter::da_service::{CelestiaService, DaServiceConfig};
 use jupiter::types::NamespaceId;
 use jupiter::verifier::RollupParams;
+use jupiter::BlobWithSender;
 use methods::{ROLLUP_ELF, ROLLUP_ID};
 use risc0_adapter::host::Risc0Host;
 use serde::Deserialize;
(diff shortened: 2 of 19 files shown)


bkolad:
I am trying to understand why we need it. The `SequencerOutcome` keeps `sequencer_da_address: Vec<u8>`  so why do we need the link?

examples/demo-simple-stf/src/lib.rs
```diff
@@ -49,7 +49,7 @@ impl<VM: Zkvm> StateTransitionFunction<VM> for CheckHashPreimageStf {
     // The core logic of our rollup.
     fn apply_blob(
         &mut self,
-        blob: &mut impl BlobTransactionTrait,
+        blob: &mut B,
         _misbehavior_hint: Option<Self::MisbehaviorProof>,
     ) -> BatchReceipt<Self::BatchReceiptContents, Self::TxReceiptContents> {
```
bkolad, line 54: I am trying to understand why we need it. The `SequencerOutcome` (Self::BatchReceiptContents) keeps `sequencer_da_address: Vec<u8>`   Are we going to pass it an `Address` at some point?

bkolad: +Have you thought about using `Box<dyn Address> `? This way, we could avoid passing another generic (not saying it is a better option, just something to consider).

citizen-stig: > I am trying to understand why we need it. The SequencerOutcome (Self::BatchReceiptContents) keeps sequencer_da_address: Vec<u8> Are we going to pass it an Address at some point?

Yes, idea is to take address from BlobTransactionTrait and keep type instead of converting to bytes via `as_ref()` 

```
-  let sequencer_da_address = blob.sender().as_ref().to_vec();
+  let sequencer_da_address = blob.sender().clone();
```

> Have you thought about using Box<dyn Address> ? This way, we could avoid passing another generic (not saying it is a better option, just something to consider).

It does not work:

```   
error[E0038]: the trait `AddressTrait` cannot be made into an object
   --> module-system/sov-modules-stf-template/src/app_template.rs:31:35
    |
31  |         sequencer_da_address: Box<dyn AddressTrait<Error = anyhow::Error>>,
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `AddressTrait` cannot be made into an object
    |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
   --> /Users/ngolub/workspace/work/sovereign/sovereign/rollup-interface/src/state_machine/mod.rs:17:5
    |
17  |     PartialEq
    |     ^^^^^^^^^ the trait cannot be made into an object because it uses `Self` as a type parameter
    |
   ::: /Users/ngolub/.rustup/toolchains/nightly-2023-02-25-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/cmp.rs:282:15
    |
282 | pub trait Eq: PartialEq<Self> {
    |               ^^^^^^^^^^^^^^^ the trait cannot be made into an object because it uses `Self` as a type parameter
```    


and this approach follows hiding **already added generic**.

bkolad (approved):

PR #510, opened 2023-07-18
Blob storage module

### LINKED ISSUE #408: Based Sequencing with Soft Confirmations
## Background
The module system currently has work in-progress to enable "based sequencing". That will be great for censorship resistance, but it sacrifices the sequencer's ability to provide near-instant soft confirmations. Since no sequencer has a "lock" on global state, they can't guarantee that no other transaction will come in and mutate the state that they depended on. 

For example, suppose that Alice wants to make a trade on Uniswap swapping one $ETH for 1000 $DAI with a slippage tolerance of 1%. Accordingly, she signs a transaction and sends it to a sequencer, Steve, who gives her a soft-confirmation that the trade will go through. At about the same time, Bob sends a transaction to another sequencer Tom, which swaps 1000 $ETH for $DAI, moving the price by 30%. Now, if Tom's block lands on the DA layer in front of Steve's the soft confirmation for Alice will be voided. 

To provide a competitive UX, we would very much like to fix that flaw. 

## Naive Implementation: Based Sequencing with "Preferred Sequencer"
A naive approach to this problem would be to enshrine some centralized sequencer and guarantee that they will always get first execution rights in each DA layer slot. So, instead of processing blobs in _exactly_ the order they appear on the DA layer, the rollup always processes the blob from the preferred sequencer as if it had appeared at the first index in its slot. 

Unfortunately, this approach is vulnerable to censorship. To be concrete, suppose that at the end of DA slot 100 there's a significant arbitrage between price of $ETH on two Uniswap `v2` vs `v3`. By default, the preferred sequencer will get to extract this arbitrage, since their blob will be processed as if it had appeared at the very beginning of slot 101. But a smart DA layer block proposer will want to extract the arbitrage for themself. So, they'll have an incentive to create their own blob which extracts the arbitrage, and censor the preferred sequencer to prevent their blob from being processed. 

## Getting Smarter: Simulating Multiple Proposers

To eliminate the incentive for censorship, we need to ensure that the DA layer block proposer can't get their transaction processed if they censor the Preferred Sequencer. One possible approach would be to return to centralized sequencing, but that is undesirable because of the risk of censorship. A smarter approach is to require multiple DA layer block proposers to collude in order to censor effectively, while simultaneously encouraging each individual proposer to defect from the cartel. In other words, we want to ensure that non-censorship is the dominant strategy in block proposal creation. 

To do that, we need to alter the rollup's state transition rule. Previously, our proposed rule looked like this:
```rust
let blobs_to_process = vec![];
// first, apply the blob from the preferred sequencer
for blob in blobs {
   if blob.sender() == PREFERRED_SEQUENCER {
     rollup.apply_blob(blob);
  } else { 
     blobs_to_process.push(blob)
  } 
}
//  then, apply all the other blobs
for blob in blobs_to_process {
   rollup.apply_blob(blob);
}
```

We can alter that rule to defer execution of blobs from non-preferred sequencers for an even longer period. Instead of processing blobs from other sequencers within the current slot, we can save them (or, in practice, a commitment to them) in our proof output, and process them as part of the *next slot

```rust
let blobs_to_defer = vec![];
// first, apply the blob from the preferred sequencer
for blob in blobs {
   if blob.sender() == PREFERRED_SEQUENCER {
     rollup.apply_blob(blob);
  } else { 
     blobs_to_process.push(blob)
  } 
}

let deferred_blob_root = make_merkle_tree(blobs_to_defer);
let blobs_to_process = Zkvm::read_from_host();
ensure!(make_merkle_tree(blobs_to_process) == old_deferred_blob_root);

for blob in blobs_to_process {
   rollup.apply_blob(blob);
}
Zkvm::commit(deferred_blob_root);
```

This way, even if the current block proposer censors the preferred sequencer and inserts his own arbitrage transaction onto the DA layer, his transaction won't be processed until the end of the next slot. And since the next DA layer proposer doesn't have anything to gain by censoring the preferred sequencer, we can expect that the sequencer's bundle will land in that slot and extract the arbitrage. 

We can carry this idea even further by deferring blob execution for an even longer time. Instead of processing "based" blobs in the next slot, we can process them 3 or 5 blocks in the future. The more sequencers will have to collude in order to extract their own arbitrage - and the less incentive any of them will have to do so.

citizen-stig:
# Description

This PR introduces a new module "sov-blob-storage" for storing Data Access (DA) layer blobs within the Rollup state. The addition of this module serves as a prerequisite for sequencing with soft confirmations. The default state transition function (STF) will utilize this module to manage deferred blocks.

The module consists of two key methods:

```rust
    pub fn store_blobs<B: BlobTransactionTrait>(
        &self,
        block_number: u64,
        blobs: &[B],
        working_set: &mut WorkingSet<C::Storage>,
    ) -> anyhow::Result<()>
    
    pub fn take_blobs_for_block_number<B: BlobTransactionTrait>(
        &self,
        block_number: u64,
        working_set: &mut WorkingSet<C::Storage>,
    ) -> Vec<B>
```

The State Transition Function (STF) now has the flexibility to decide which blobs to execute and when.

Consider a situation where the STF receives a blob from the preferred sequencer in block 100. It can use the `store_blob` function to hold onto deferred blobs, and later retrieve these blobs at the next block or after several blocks.

## Points to Consider

This PR mainly focuses on the public API for this module. The current implementation is straightforward and does not contain much business logic.

One key consideration is the "defer period", which refers to the number of blocks for which blobs will be deferred. Currently, this value is hard-coded in the STF, but should it be stored in the state instead?

Also, how strict should this module be? Should it limit the number of blobs that can be stored or track the total size of stored blobs? Should it permit the insertion of blobs out of order?

## Linked Issues
- Related to #408

## Testing
Integration tests have been added

## Docs
`#![deny(missing_docs)]`


module-system/module-implementations/sov-blob-storage/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// The address of blob storage module
+    /// Note: this is address is generated by the module framework and the corresponding private key is unknown.
+    #[address]
+    pub(crate) address: C::Address,
+
+    /// Actual storage of blobs
+    /// stored this way (DA block number, blob)
+    /// TODO: StateVec: https://github.com/Sovereign-Labs/sovereign-sdk/issues/33
+    #[state]
+    pub(crate) blobs: StateValue<Vec<(u64, Vec<u8>)>>,
```
bkolad: Should we limit the size of the `Vec` so it is not unbounded?  This could be done just before insertion to the vec. Or is there a guarantee that it will be emptied periodically?

bkolad (approved):

module-system/module-implementations/sov-blob-storage/src/lib.rs
citizen-stig: It should be constant size at worst and approximately empty on average.

Worst case: 

Every L1 block has blob from preferred sequencer. So `x` blobs from other sequencers are deferred by `n` L1 blocks, where `n` is some sane number between 1-5. After `n` L1 blocks, `x` blobs deferred earlier are pulled and executed.
This means that there will `x * n` elements in that vector all the time. 

I would like to defer introducing boundary right now, as whole feature is not yet completed and some things depend on hooks discussion.
When implementation is stabilized, those can be introduced.

preston-evans98: Hmm, this will require the entire Vec to be loaded every time we want to access an element. That's going to be super expensive, because every blob has to be hashed to look it up in the state trie. Can we store a `StateMap<u64, Vec<Vec<u8>>>` instead?

module-system/module-implementations/sov-blob-storage/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        let mut i = 0;
+        while i < blobs.len() {
+            if blobs[i].0 == block_number {
+                let (_, blob) = blobs.remove(i);
+                blobs_for_block.push(blob);
+            } else {
+                i += 1;
+            }
+        }
```
preston-evans98: This also looks like it would be cleaner with a `StateMap`

preston-evans98 (changes requested):

preston-evans98:
It seems like there's some potential for type confusion here - is the stored blob a serialized `BlobTransaction`, or is it just the blob data? Maybe it would be more natural to change from storing a blob as a raw `Vec<u8>` to storing it as a concrete `StoredBlob` type which contains the data (probably as a `Vec<u8>`) and the sender?

module-system/module-implementations/sov-blob-storage/src/lib.rs
citizen-stig: Addressed!

module-system/module-implementations/sov-blob-storage/src/lib.rs
citizen-stig: Addressed!

citizen-stig:
> It seems like there's some potential for type confusion here - is the stored blob a serialized `BlobTransaction`, or is it just the blob data? Maybe it would be more natural to change from storing a blob as a raw `Vec<u8>` to storing it as a concrete `StoredBlob` type which contains the data (probably as a `Vec<u8>`) and the sender?

Yes, I've done via introducing `BlobTransactionTrait` generic, so STF can just pass it without thinking twice.

preston-evans98 (approved):
Looks great, thanks for addressing the comments @citizen-stig!

PR #511, opened 2023-07-18
make `DaService::get_extraction_proof` async

### LINKED ISSUE #507: Make `DaService::get_extraction_proof` an async function
**Is your feature request related to a problem? Please describe.**
"My `get_extraction_proof` is async and requires communication using the client. I solve this by just getting the extraction proof in `get_finalized_at` unconditionally. But I guess that makes some unnecessary work for pure full node mode." This feature request originated in [Discord](https://discord.com/channels/1050059327626555462/1069729783367336010/1130159348631478413). 

**Describe the solution you'd like**
Make `DaService::get_extraction_proof` an async function 

**Additional context**
My (Preston's) response in Discord: "We can definitely make get_extraction_proof async as well. I had initially envisioned it as a mathematical operation - but it makes sense that it would be baked into the remote client."

bkolad:
# Description
1. Makes `DaService::get_extraction_proof` function async
2. Adds `Send and Sync` bounds to `BlobTransactionTrait &  SlotData`( required by point 1). 
3. Introduces mutex for  `CelestiaHeader:: cached_prev_has` so it is `Send & Sync`

## Linked Issues
- Fixes #507

## Testing
All tests are passing.

## Docs
No impact on docs.

### DIFF
--- a/adapters/celestia/src/celestia.rs
+++ b/adapters/celestia/src/celestia.rs
@@ -1,6 +1,6 @@
-use std::cell::RefCell;
 use std::fmt::{Display, Formatter};
 use std::ops::Range;
+use std::sync::{Arc, Mutex};
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use nmt_rs::NamespacedHash;
@@ -223,13 +223,19 @@ pub struct NamespacedSharesResponse {
     pub height: u64,
 }
 
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
 pub struct CelestiaHeader {
     pub dah: DataAvailabilityHeader,
     pub header: CompactHeader,
     #[borsh_skip]
     #[serde(skip)]
-    cached_prev_hash: RefCell<Option<TmHash>>,
+    cached_prev_hash: Arc<Mutex<Option<TmHash>>>,
+}
+
+impl PartialEq for CelestiaHeader {
+    fn eq(&self, other: &Self) -> bool {
+        self.dah == other.dah && self.header == other.header
+    }
 }
 
 impl CelestiaHeader {
@@ -237,7 +243,7 @@ impl CelestiaHeader {
         Self {
             dah,
             header,
-            cached_prev_hash: RefCell::new(None),
+            cached_prev_hash: Arc::new(Mutex::new(None)),
         }
     }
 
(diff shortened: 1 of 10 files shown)


adapters/celestia/src/celestia.rs
```diff
@@ -223,21 +223,27 @@ pub struct NamespacedSharesResponse {
...
     pub header: CompactHeader,
     #[borsh_skip]
     #[serde(skip)]
-    cached_prev_hash: RefCell<Option<TmHash>>,
+    cached_prev_hash: Arc<Mutex<Option<TmHash>>>,
+}
+
+impl PartialEq for CelestiaHeader {
+    fn eq(&self, other: &Self) -> bool {
+        self.dah == other.dah && self.header == other.header
```
bkolad, line 237: Question: do we have to compare `cached_prev_hash`? I assume we don't  (it is derived from header which we already compare in `eq`)so we can avoid taking a lock here.  @preston-evans98

citizen-stig (approved):

neysofu (approved):

adapters/celestia/src/celestia.rs
preston-evans98: Yep, that's true! We can also switch to std::cell:OnceCell now that it's stable to avoid the lock here entirely.

preston-evans98 (approved):

adapters/celestia/src/celestia.rs
bkolad: Moved to `std::sync::OnceLock`

adapters/celestia/src/celestia.rs
```diff
@@ -257,20 +263,18 @@ impl BlockHeader for CelestiaHeader {
...
-        let mut cached_hash = self.cached_prev_hash.borrow_mut();
+
         let hash =
             <tendermint::block::Id as Protobuf<celestia_tm_version::types::BlockId>>::decode(
                 self.header.last_block_id.as_ref(),
             )
             .expect("must not call prev_hash on block with no predecessor")
             .hash;
-        *cached_hash = Some(TmHash(hash));
+        self.cached_prev_hash.set(TmHash(hash));
```
preston-evans98: It looks like `get_or_init()` is made for exactly this. Should we use it instead?

adapters/celestia/src/celestia.rs
```diff
@@ -223,21 +223,27 @@ pub struct NamespacedSharesResponse {
...
     #[borsh_skip]
     #[serde(skip)]
-    cached_prev_hash: RefCell<Option<TmHash>>,
+    cached_prev_hash: OnceLock<TmHash>,
+}
+
+impl PartialEq for CelestiaHeader {
+    fn eq(&self, other: &Self) -> bool {
+        self.dah == other.dah && self.header == other.header
+    }
```
preston-evans98, line 238: Should we compare hashes instead of comparing full headers? I guess the question is - would this ever cause us to compute a hash that would otherwise never be used

adapters/celestia/src/celestia.rs
bkolad: Good catch!

adapters/celestia/src/celestia.rs
bkolad: Discussed offline: `CelestiaHeader::hash()` is not cached so we won't use it.

Issue #481, opened 2023-07-07
rollup_config.toml is stashed by the sov-celestia-local setup and runs a risk of being committed 


dubbelosix:
## Summary

Currently, rollup_config.toml (the default location for demo-rollup config) is modified by the sov-celestia-local setup. Because this is run pretty frequently, it runs a risk of being committed (And as of the time of creating this issue, it has already been committed). This is not necessary and a new rollup config toml file can be generated and used since demo-rollup and demo-prover can both take the config file as an argument. Creating a new file would keep the stash clean and avoid the risk of committing the changes

citizen-stig:
Might be duplicate of #445

PR #513, opened 2023-07-18
EVM: `get_transaction_receipt & eth_call` endpoints implementation. 

bkolad:
# Description
1. Implementation of the `get_transaction_receipt` endpoint. Currently, the receipts are stored in the module state.
2. Implementation of the `eth_call` endpoint.

With these changes, the `send_tx_test_to_eth` test now functions correctly, proving that the EVM module can interact with an external wallet. The current implementation is very basic, and the pull request contains "todo" items that will be addressed in future PRs (together with documentation & tests). 

## Testing
Tests will be added in the following PRs.

## Docs
Docs will be provided in the following PRs.

### DIFF
--- a/examples/demo-rollup/rollup_config.toml
+++ b/examples/demo-rollup/rollup_config.toml
@@ -4,7 +4,7 @@ start_height = 1
 
 [da]
 # The JWT used to authenticate with the celestia light client. Instructions for generating this token can be found in the README
-celestia_rpc_auth_token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJwdWJsaWMiLCJyZWFkIiwid3JpdGUiLCJhZG1pbiJdfQ.jp_hghhRkcur3g0o-QhQnkLt89YPu4oykLtZW52_ZA0"
+celestia_rpc_auth_token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJwdWJsaWMiLCJyZWFkIiwid3JpdGUiLCJhZG1pbiJdfQ.0lh1NrdtePmLPvpf2xR_i5wLWWYyxY_e_jHzHXV7aKE"
 # The address of the *trusted* Celestia light client to interact with
 celestia_rpc_address = "http://127.0.0.1:26658"
 # The largest response the rollup will accept from the Celestia node. Defaults to 100 MB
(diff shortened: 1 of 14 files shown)


examples/demo-rollup/rollup_config.toml
```diff
@@ -4,7 +4,7 @@ start_height = 1
 
 [da]
 # The JWT used to authenticate with the celestia light client. Instructions for generating this token can be found in the README
-celestia_rpc_auth_token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJwdWJsaWMiLCJyZWFkIiwid3JpdGUiLCJhZG1pbiJdfQ.jp_hghhRkcur3g0o-QhQnkLt89YPu4oykLtZW52_ZA0"
+celestia_rpc_auth_token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJwdWJsaWMiLCJyZWFkIiwid3JpdGUiLCJhZG1pbiJdfQ.i6Cx-WziFU3Jcz2INKii2uMCareIZHDDMR0v8zW5zZc"
```
preston-evans98, line 7: I think this should be set to `"MY_SECRET_TOKEN"`, not an actual token, right?

module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -68,15 +63,69 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
         &self,
-        _hash: H256,
-        _working_set: &mut WorkingSet<C::Storage>,
+        hash: H256,
+        working_set: &mut WorkingSet<C::Storage>,
     ) -> Option<TransactionReceipt> {
-        unimplemented!("eth_getTransactionReceipt not implemented")
+        info!("evm module: eth_getTransactionReceipt");
+
+        let receipt = self.receipts.get(&hash.into(), working_set);
```
preston-evans98, line 75: (Non-blocking): Do we want receipts to live in State, or just be (optionally) available via RPC. I'm wondering if we should add another DB interface for modules which stores/retrieves information in native mode and is a no-op in zk mode.

preston-evans98 (approved):

examples/demo-rollup/rollup_config.toml
bkolad: We have an issue for this, so I will let @dubbelosix handle it: https://github.com/Sovereign-Labs/sovereign-sdk/issues/481

module-system/module-implementations/sov-evm/src/query.rs
bkolad: Yes, we will have to think of a better way of doing this, I will work on the design this week and schedule a call to discuss it.

PR #517, opened 2023-07-18
New `ModuleCallJsonSchema` macro to generate JSON Schemas via `schemars`

neysofu:
# Description

This PR defines a new trait `ModuleCallJsonSchema`, which is intended to be used by all modules that wish to expose a JSON Schema for their call messages. This trait can be either implemented manually or via the new `derive` macro. 

As a proof of concept, the trait is derived for `sov_bank::Bank`. If this PR is approved, I'll instrument all other modules with the derive macros and macro attributes necessary for JSON Schema generation.

## How `#[derive(ModuleCallJsonSchema)]` works

The derive for `ModuleCallJsonSchema)` is conceptually very simple, as it offloads the complexity of schema generation to the `schemars` crate. `schemars` itself exposes a derivable trait –`JsonSchema`– which must be derived by `CallMessage` and all its children types. Occasionally, we also need to pollute our types with `schemars`–sepcific macro attributes, specifically to help it generate the right type bounds when dealing with `sov_modules_api::Context`.

The quality of the generated schema(s) seems, at a first glances, very good. `schemars` supports most `serde` attribute macros and has crate features to support well-known crates, so most bases are covered. It's possible we might need to patch the schemas in the future. We'll have to decide how to do that when the need arises. Custom macro attributes or hand-written implementations of `schemars::JsonSchema` could be viable options.

## Linked Issues
I don't think there are any? The team discussed this as part of the July roadmap, though.

## Testing
There's a new "dummy" crate, `sov-module-schemas`, which generates JSON Schemas for all supported modules and stores them in a directory. These are tracked by git so it will be very easy to see if a code change results in a schema change. This way, breakages and changes to call interfaces can be tracked easily.

Also, I refactored the `get_generics_type_param` utility function to generate a better error message when deriving `ModuleInfo`. This is unrelated to the work on JSON Schemas, but it's small and I figured it made sense to include it in the same PR.

## Docs
I added `#![deny(missing_docs)]` to `sov_modules_macros` and I patched some of the doc tests to be runnable instead of `ignore`'d by Cargo.


module-system/sov-modules-macros/src/module_call_json_schema.rs
```diff
@@ -0,0 +1,26 @@
...
+        ident, generics, ..
+    } = input;
+
+    let (impl_generics, type_generics, where_clause) = generics.split_for_impl();
+
+    let tokens = quote::quote! {
+        use ::schemars::JsonSchema;
+
+        impl #impl_generics ::sov_modules_api::ModuleCallJsonSchema for #ident #type_generics #where_clause {
+            fn json_schema() -> String {
```
bkolad: Let's return `::std::string::String` (macro hygiene)

neysofu: Good catch, fixed in https://github.com/Sovereign-Labs/sovereign-sdk/pull/517/commits/04a37bff3a172c5e4c13d07738c532178c3f5c76.

module-system/sov-modules-macros/tests/all_tests.rs
```diff
@@ -3,12 +3,13 @@ fn module_info_tests() {
     let t = trybuild::TestCases::new();
     t.pass("tests/module_info/parse.rs");
     t.pass("tests/module_info/mod_and_state.rs");
+    t.compile_fail("tests/module_info/derive_on_enum_not_supported.rs");
     t.compile_fail("tests/module_info/field_missing_attribute.rs");
+    t.compile_fail("tests/module_info/missing_address.rs");
+    t.compile_fail("tests/module_info/no_generics.rs");
```
bkolad, line 9: Does adding a test for the `ModuleCallJsonSchema` macro make sense?  Maybe we could tests the `Failed to serialize JSON schema; this is a bug in the module` message?

module-system/module-implementations/sov-bank/src/call.rs
```diff
@@ -8,10 +8,15 @@ use crate::{Amount, Bank, Coins, Token};
 #[cfg_attr(
     feature = "native",
     derive(serde::Serialize),
-    derive(serde::Deserialize)
+    derive(serde::Deserialize),
+    derive(schemars::JsonSchema),
+    schemars(bound = "C::Address: ::schemars::JsonSchema", rename = "CallMessage")
```
preston-evans98, line 13: (Question) What name does `schemars` generate without the `rename = CallMessage` annotation?

module-system/sov-modules-api/src/lib.rs
```diff
@@ -235,6 +239,18 @@ pub trait Module {
...
 }
 
+/// A [`Module`] that has a well-defined and known [JSON
+/// Schema](https://json-schema.org/) for its [`Module::CallMessage`].
+///
+/// This trait is intended to support code generation tools, CLIs, and
+/// documentation. You can derive it with `#[derive(ModuleCallJsonSchema)]`, or
+/// implement it manually if your use case demands more control over the JSON
+/// Schema generation.
+pub trait ModuleCallJsonSchema: Module {
```
preston-evans98, line 249: This trait looks good, but I think we also want Json Schemas for RPC query messages. I don't have a strong opinion about whether we should have two separate traits (one for call messages and one for queries), or package them both into one. One argument in favor of two traits is that we may have additional `derives` we want to package together for call messages (`BorshSchema`, `BorshSerialize`, `clap::Parser` etc), while query messages probably only need `serde` and `JsonSchema` to support RPC.

cc @bkolad and @dubbelosix - any thoughts?

module-system/sov-modules-macros/src/common.rs
```diff
@@ -123,9 +123,16 @@ impl<'a> StructDef<'a> {
...
-    let generic_param = match generics.params.first().unwrap() {
+/// Gets the type parameter's identifier from [`syn::Generics`].
+pub(crate) fn get_generics_type_param(
+    generics: &Generics,
+    error_span: Span,
+) -> Result<Ident, syn::Error> {
+    let generic_param = match generics
+        .params
+        .first()
+        .ok_or_else(|| syn::Error::new(error_span, "No generic parameters found"))?
```
preston-evans98, line 134: (Out-of-scope) I know this logic wasn't introduced in this PR, but do we document anywhere that the first generic must always be the `Context` type?

module-system/sov-modules-macros/tests/all_tests.rs
neysofu: I don't that that would be very useful. `to_string_pretty` can only fail if:

> Serialization can fail if T’s implementation of Serialize decides to fail, or if T contains a map with non-string keys.

When deriving `serde::Serialize` like we do that can't happen, and we'd have to specifically manually implement a type that fails to serialize to trigger the panic.

The majority of problems we could see with this macro are when the developer forgets to implement `schemars::JsonSchema`. We could have `compile_fail` tests for that, although we wouldn't really be testing the macro logic.

bkolad (approved):

module-system/sov-modules-api/src/lib.rs
bkolad: I don't have any strong opinion here. I am fine with postponing the decision until we start working on JSON Schemas for  RPC queries and see how it looks.

module-system/module-implementations/sov-bank/src/call.rs
neysofu: Ah, that's a funny story which just got me down a bit of a rabbit hole. When working on the PR, I first tried without `rename`, but I kept getting a compilation error which seemed to indicate that Schemars was requiring `C: JsonSchema` as a type bound. I then noticed that `rename` was used in one of Schemars' examples in combination with a type generic, so I tried that and sure enough it worked. Only now I'm realizing why:

https://github.com/GREsau/schemars/blob/386e3d7f5ac601795fb4e247291bbef31512ded3/schemars_derive/src/lib.rs#L96-L118

Schemars tries to get the schema name for the type parameter, thus assuming that it implements `JsonSchema` even though it may not:

```rust
format!(#schema_name_fmt #(,#type_params::schema_name())*)
```

The resulting schema name would then be something like `CallMessage_for_Context`... it only `Context` implemented `JsonSchema`. So we're effectively using `rename` as a hack to skip the generics' schema name generation logic, which is flawed.

Upstream is not extremely receptive to PRs, and the maintainer seems aware of this issue, so I'm not sure what we can do. We can:
- Keep `rename`.
- Fork.
- Submit a PR and push for it to get merged until we receive a response.

module-system/sov-modules-api/src/lib.rs
bkolad: I don't have any strong opinion here. I am fine with postponing the decision until we start working on JSON Schemas for RPC queries and see how it looks.

module-system/sov-modules-macros/src/common.rs
neysofu: There seems to be some relevant docs here: https://github.com/Sovereign-Labs/sovereign-sdk/blob/04a37bff3a172c5e4c13d07738c532178c3f5c76/module-system/README.md#L12-L36

module-system/module-implementations/sov-bank/src/call.rs
neysofu: One thing is for sure, I'll need to update the docs to show how to implement `ModuleCallJsonSchema` correctly for new modules as it's not covered anywhere. I can do it in the next PR if that's okay, where I'll instrument all the other modules.

preston-evans98: One more potential option: we could wrap `derive(JsonSchema)` in our own macro which (for now), just adds the `JsonSchema` annotation with `rename`. That way, we can also change out our schema generator in future without breaking consumers

module-system/sov-modules-macros/src/common.rs
preston-evans98: I might be missing something, but I don't see anything about the _ordering_ constraint on the generics (i.e. that `C` must come first). Maybe we can add that in here?

module-system/sov-modules-api/src/lib.rs
preston-evans98: That's fine by me 👍

module-system/module-implementations/sov-bank/src/call.rs
neysofu: That's an attractive option, the problem I see with that is it's quite leaky by necessity. Anytime one needs to add a `#[schemars(...)]` attribute on a field or use some of the more advanced features –we didn't need to do it for `sov-bank`, luckily– they will have to go through our proc macro and still inevitably take into consideration Schemars implementation details.

I say we reevaluate after all built-in modules have a schema. If we see there's no need for meddling with Schemars internals, we can then assume most users won't have to care about it and we can wrap it. Otherwise, we evaluate one of the other options. WDYT?

module-system/sov-modules-api/src/lib.rs
neysofu: Marking as resolved.

module-system/sov-modules-macros/src/common.rs
neysofu: I clarified a bit in https://github.com/Sovereign-Labs/sovereign-sdk/pull/517/commits/12035eab403340f32ad61d9b904b3dee611a492f :+1:

module-system/module-implementations/sov-bank/src/call.rs
preston-evans98: That's a good point. Sounds fine to me!

PR #525, opened 2023-07-19
JSON Schemas for all built-in modules

neysofu:
# Description
After the implementation of `ModuleCallJsonSchema` in #517 and the proof-of-concept for `sov-bank`, we're now adding JSON Schemas for all built-in modules.

P.S. the line count is scary, but 95% of it is autogenerated JSON files :stuck_out_tongue:

## Linked Issues
Follow-up to #517.

## Testing
No additional tests.

## Docs
Both example modules have been updated, as well as the module template.


module-system/sov-modules-api/src/lib.rs
```diff
@@ -146,6 +146,7 @@ pub trait Spec {
         + PublicKey
         + Serialize
         + for<'a> Deserialize<'a>
+        + ::schemars::JsonSchema
```
preston-evans98, line 149: Two related questions here :) 
1. Do we need this bound to satisfy the compiler? (Often you can get away without these bounds since you'll always fill in a concrete `Context` implementation before trying to serialize schemas)
2. If not, is there ever a situation where a rollup wouldn't want to support JsonSchema?

bkolad (approved):

bkolad:
Question: Does the `derive(schemars::JsonSchema)` macro generate any public methods/types? If yes then let's generate documentation so a module flagged with `#![deny(missing_docs)]` can compile. 

#edit:
Ok, we are fine it implements `ModuleCallJsonSchema` which already has documentation.

module-system/sov-modules-api/src/lib.rs
neysofu: Good question(s). I'd imagine there are situations where a user doesn't care about generating JSON Schema; maybe when authoring a new module it's annoying to have to deal with this. We not enforce modules to implement `JsonSchema`, but `PublicKey` and `Signature` must implement it. It's not ideal.

We need these bounds to satisfy the compiler because `ModuleCallJson` is implemented for all `C: Context`. I think we could add a macro attribute to specify the type bounds, much like `schemars` already does, and this way we wouldn't have to change any type bounds inside `Spec`. Something of this sort:

```rust
[cfg_attr(
    feature = "native",
    derive(sov_modules_macros::ModuleCallJsonSchema),
    module_call_json_schema(bounds = "C::Signature: schemars::JsonSchema")
)]
#[derive(ModuleInfo, Clone)]
pub struct Election<C: sov_modules_api::Context> {
    #[address]
    pub address: C::Address,
    // ...
}
```

As an alternative, I can try and generate type bounds on `Signature` and `PublicKey` by default and only have the macro attribute to opt-out if the module authors wants to. That seems like the best solution to me.

module-system/module-implementations/sov-accounts/src/call.rs
```diff
@@ -9,7 +9,12 @@ pub const UPDATE_ACCOUNT_MSG: [u8; 32] = [1; 32];
 #[cfg_attr(
     feature = "native",
     derive(serde::Serialize),
-    derive(serde::Deserialize)
+    derive(serde::Deserialize),
+    derive(schemars::JsonSchema),
+    schemars(
+        bound = "C::PublicKey: ::schemars::JsonSchema, C::Signature: ::schemars::JsonSchema",
```
citizen-stig, line 15: What would be fail message, if developer change generic parameter from `C` to `T` for example, but forgets to update attribute here?

citizen-stig (approved):

module-system/module-implementations/sov-accounts/src/call.rs
neysofu:
```
error[E0412]: cannot find type `T` in this scope
  --> module-system/module-implementations/examples/sov-election/src/call.rs:14:22
   |
14 |     schemars(bound = "T: sov_modules_api::Context", rename = "CallMessage")
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a type parameter with a similar name exists: `C`
...
17 | pub enum CallMessage<C: Context> {
   |                      - similarly named type parameter `C` defined here

```

I don't think we are in control of that error message unfortunately, as it happens only after macro expansion.

module-system/sov-modules-api/src/lib.rs
neysofu: Tracked in https://github.com/Sovereign-Labs/sovereign-sdk/issues/539.

PR #538, opened 2023-07-20
Bump timeout for doc job

citizen-stig:
So it won't fail that often


neysofu (commented):
What are the changes to `Cargo.lock` from?

citizen-stig:
They are from risc0 update, I compuled while project.

neysofu (approved):

PR #549, opened 2023-07-24
Sorting modules by dependencies

### LINKED ISSUE #428: `Genesis` cross-module dependencies.
# Description
If my module relies on another module in `genesis` logic, it might fail, as there are no strong guarantee that my depedency will be initialized at the moment of genesis of my module. 
It will work only if I declare modules in correct order in runtime, which implies that I, as developer, know dependencies of each module.

# Solutions

There are no clear the best solution, but only variants:

## Disallow dependencies on other modules

*pros:*

* Simpler logic

*cons:*

* Severely limits what can be done in genesis
* Require more compilcated configuration file
* Harder to enforce.

## Require genesis to be idempotent

This allow module explicitly call genesis for dependencies that really needed for genesis, and if it is already been called, nothing should happen.

*pros:*

* Modules can manage gensis themself

*cons:*

* Require developers to stick to contraint
* Big problem of passing configuration around.

## Keep manual order of modules, but make it more robust

Currently module genesises are called in order they declared in Runtime. 
This might be fine, except that there's possibility if calling module before genesis won't result in explicit error, but rather in implicit one.

For example, bank module and sequencer registry module. If bank module is not initialized, it will error on any operation, as balances are not preset. 
But what if some another module can operate without error before genesis. This means that if developer didn't order modules correctly, genesis of child module will pass, but it's changes might be overwritten by genesis of parent, which will result in broken state, which is hard to debug.

Solution might be similar to "idempotent" approach, where each module prevents calling any of its public methods before some boolean variable in state is initialized.

LukaszRozmej:
# Description
- `ModuleInfo` trait was expanded with the `dependencies` function and returns the addresses of modules it depends on.
- `derive_module_info` macro generates `dependencies` function for each module using `#[module]` attribute.
- `sort_by_dependencies` function allows to sort `ModuleInfo` by their dependencies. It takes modules and additional type tuples and returns sorted vector of this additional type. This done this way to easily map it back to compile time generated constants (see expanded `derive_genesis` macro).
- `derive_genesis` macro changed how `fn genesis` was created. Now it creates a `Vec<ModuleTrait, usize>` that is sorted with `sort_by_dependencies` and then the `usize` are matched to correct module `genesis` call.
- Two errors were introduced: 
 1. Missing module in runtime
 2. Cyclic dependency

# Side effect changes
- Addresses are now hashable. `AddressTrait` and all its implementations was expanded with `Hash` trait. 
- `ModuleInfo` is not subtrait of `Default`, this wasn't required and made the trait not object safe.

# Example of expanded `derive_genesis` macro for demo-app
```rust
    impl<C: Context> sov_modules_api::Genesis for Runtime<C> {
        type Context = C;
        type Config = GenesisConfig<C>;
        fn genesis(
            &self,
            config: &Self::Config,
            working_set: &mut sov_state::WorkingSet<
                <<Self as sov_modules_api::Genesis>::Context as sov_modules_api::Spec>::Storage,
            >,
        ) -> core::result::Result<(), sov_modules_api::Error> {
            let modules: ::std::vec::Vec<
                (
                    &dyn ::sov_modules_api::ModuleInfo<
                        Context = <Self as sov_modules_api::Genesis>::Context,
                    >,
                    usize,
                ),
            > = <[_]>::into_vec(
                #[rustc_box]
                ::alloc::boxed::Box::new([
                    (&self.bank, 0usize),
                    (&self.sequencer_registry, 1usize),
                    (&self.election, 2usize),
                    (&self.value_setter, 3usize),
                    (&self.accounts, 4usize),
                ]),
            );
            let sorted_modules = ::sov_modules_api::sort_values_by_modules_dependencies(
                modules,
            )?;
            for module in sorted_modules {
                match module {
                    0usize => {
                        ::sov_modules_api::Genesis::genesis(
                            &self.bank,
                            &config.bank,
                            working_set,
                        )
                    }
                    1usize => {
                        ::sov_modules_api::Genesis::genesis(
                            &self.sequencer_registry,
                            &config.sequencer_registry,
                            working_set,
                        )
                    }
                    2usize => {
                        ::sov_modules_api::Genesis::genesis(
                            &self.election,
                            &config.election,
                            working_set,
                        )
                    }
                    3usize => {
                        ::sov_modules_api::Genesis::genesis(
                            &self.value_setter,
                            &config.value_setter,
                            working_set,
                        )
                    }
                    4usize => {
                        ::sov_modules_api::Genesis::genesis(
                            &self.accounts,
                            &config.accounts,
                            working_set,
                        )
                    }
                    _ => {
                        Err(
                            ::sov_modules_api::Error::ModuleError(
                                ::anyhow::Error::msg({
                                    let res = ::alloc::fmt::format(
                                        format_args!("Module not found: {0:?}", module),
                                    );
                                    res
                                }),
                            ),
                        )
                    }
                }?
            }
            Ok(())
        }
    }
```

## Linked Issues
- Fixes #428

## Testing
- Unit test were added to cover `derive_module_info` macro and `sort_by_dependencies`, `derive_genesis` macro was already expanded in other tests.

## Docs
- No changes needed.


module-system/sov-modules-api/src/lib.rs
```diff
@@ -265,3 +275,44 @@ pub trait RpcRunner {
...
     fn get_storage(&self) -> <Self::Context as Spec>::Storage;
 }
+
+/// Sorts ModuleInfo objects by their dependencies
+pub fn sort_by_dependencies<C: Context>(
+    modules: Vec<&dyn ModuleInfo<Context = C>>,
+) -> Result<Vec<&dyn ModuleInfo<Context = C>>, anyhow::Error> {
+
+    fn visit_module<'a, C: Context>(
+        module: &'a dyn ModuleInfo<Context = C>,
```
bkolad, line 357: What do you think about making a `ModuleVisitor` struct that holds all the required data and a `visit` method?  This way, we could remove the explicit lifetime `a` as it will be inferred from `&mut self` + it will give a bite more structure to the code (but it is a matter of taste).

module-system/sov-modules-api/src/lib.rs
```diff
@@ -265,3 +275,44 @@ pub trait RpcRunner {
...
+        visited: &mut HashSet<&'a C::Address>,
+        sorted_modules: &mut std::vec::Vec<&'a (dyn ModuleInfo<Context = C> + 'a)>,
+        module_map: &HashMap<&<C as Spec>::Address, &'a (dyn ModuleInfo<Context = C> + 'a)>
+    ) -> Result<(), anyhow::Error> {
```
bkolad, line 359: let's return `syn::Error` and produce a nice compile error if something goes wrong.

module-system/sov-modules-macros/src/module_info.rs
```diff
@@ -249,6 +254,19 @@ fn make_fn_address(
...
 
+fn make_fn_dependencies(modules: Vec<&proc_macro2::Ident>) -> proc_macro2::TokenStream {
+    let address_tokens = modules.iter().map(|ident| {
+        quote::quote! {
+            &self.#ident.address()
+        }
+    });
+
+    quote::quote! {
+        fn dependencies(&self) -> Vec<&<Self::Context as sov_modules_api::Spec>::Address> {
```
bkolad: We need a fully qualified name for `Vec` here (this will prevent name collisions if someone defines `Vec` struct in the file where the macro is used)

module-system/sov-modules-api/src/lib.rs
LukaszRozmej: Impossible, this is called at runtime. Check example expanded genesis macro.

module-system/sov-modules-api/src/lib.rs
LukaszRozmej: Would have to try it, I don't have enough intuition about Rust yet

bkolad (approved):

module-system/sov-modules-api/src/tests.rs
```diff
@@ -44,3 +45,127 @@ fn test_hex_conversion() {
...
+        module_a: module_a_c,
+        module_b: module_b_c,
+    };
+
+    let modules: Vec<(&dyn ModuleInfo<Context = DefaultContext>, &i32)> =
+        vec![(&module_b, &2), (&module_c, &3), (&module_a, &1)];
+
+    let sorted_modules = crate::sort_modules_by_dependencies(&modules).unwrap();
+
+    assert_eq!(sorted_modules, vec![&1, &2, &3]);
```
citizen-stig: Just nitpicking, but for `assert_eq` I usually use (expected, actual) order

citizen-stig (approved):

module-system/sov-modules-api/src/tests.rs
LukaszRozmej: Hm... this is not consistent in the repository. I also find it easier to read in order, we "assert on sorted modules equal something expected"

citizen-stig: Yeah, I don't have strong opinion on that. But we need to unify on this so readability will be the same for everybody.

module-system/sov-modules-api/src/lib.rs
```diff
@@ -290,3 +300,45 @@ pub trait RpcRunner {
...
+) -> Result<Vec<&'a T>, anyhow::Error> {
+    fn visit_module<'a, C: Context, T>(
+        module_tuple: (&'a (dyn ModuleInfo<Context = C> + 'a), &'a T),
+        visited: &mut HashSet<&'a C::Address>,
+        sorted_modules: &mut Vec<&'a T>,
+        module_map: &HashMap<&'a C::Address, (&'a (dyn ModuleInfo<Context = C> + 'a), &'a T)>,
+    ) -> Result<(), anyhow::Error> {
+        let (module, item) = module_tuple;
+        let address = module.address();
+        if visited.insert(address) {
```
preston-evans98: nit: The semantics of `if visited.insert(...)` aren't super obvious at a glance. Can we add a comment explaining? Maybe something like `// If the module hasn't been visited yet, visit it and its dependencies`

module-system/sov-modules-api/src/lib.rs
```diff
@@ -290,3 +300,45 @@ pub trait RpcRunner {
     type Context: Context;
     fn get_storage(&self) -> <Self::Context as Spec>::Storage;
 }
+
+/// Sorts ModuleInfo objects by their dependencies
+pub fn sort_modules_by_dependencies<'a, C: Context, T>(
```
preston-evans98: Question: Why do we use a generic `T` here? Can we pick a suitable integer type (maybe usize) instead, or is there some blocker?

preston-evans98 (commented):
Left a few minor comments, but this looks like it will get the job done. That being said, I'm curious about the approach of assigning each module a number and then `match`ing on the sorted module numbers? I have two minor concerns there. 

First, I think the generated code might be quadratic in the number of modules (I'm not sure if rustc is smart enough to optimize a match on an integer into a jump). That's probably ok, since the number of modules should be small, but it's a little unexpected. 

Second, the `match` just feels a little bit ugly/hacky. Again that's not a blocker - sometimes you need a hack to make things work; but this seems like code smell to me. 

I think we can solve both problems with a small tweak to the API. If we make sorted_modules a `&mut Vec<&'a dyn Genesis>`, then we can just iterate over it directly instead of using the `match`. That also means we'll need to change the signature of the other args to have a `dyn Genesis` in place of `T`. Or, equivalently, we can use a supertrait of `ModuleInfo` and `Genesis` so that we can pass around a single trait object. 

Anyway, I'm curious if there's some drawback to that approach that I'm missing.

module-system/sov-modules-api/src/lib.rs
LukaszRozmej: So I simplified the function to:
```rust
pub fn sort_modules_by_dependencies<C: Context>(
    modules: Vec<&dyn ModuleInfo<Context = C>>,
) -> Result<Vec<&dyn ModuleInfo<Context = C>>, anyhow::Error> 
```
But I also kept the mapping one and it calls the sort_modules_by_dependencies
```rust
pub fn sort_values_by_modules_dependencies<'a, C: Context, TValue>(
    module_value_tuples: Vec<(&dyn ModuleInfo<Context = C>, &'a TValue)>,
) -> Result<Vec<&'a TValue>, anyhow::Error> where TValue: 'a 
```
I want to have the second one because it is easier to call it from the macro generated code rather than generate that code inside the macro.

module-system/sov-modules-api/src/lib.rs
LukaszRozmej: I tried to refactor it to something like:

```rust
struct ModuleVisitor<C: Context> {
    visited: HashSet<<C as Spec>::Address>,
    modules: Vec<Box<dyn ModuleInfo<Context = C>>>,
}

impl<C: Context> ModuleVisitor<C>{
    fn visit_module(
        &mut self, 
        module: &dyn ModuleInfo<Context = C>,
        sorted_modules: &mut std::vec::Vec<&dyn ModuleInfo<Context = C>>,
    ) -> Result<(), anyhow::Error> {
        let address = module.address();
        
        // if the module hasn't been visited yet, visit it and its dependencies
        if self.visited.insert(*address) {
            for dependency_address in module.dependencies() {
                let dependency_module = self.modules.iter().find(|module| module.address() == dependency_address)
                    .ok_or_else(|| anyhow::Error::msg(format!("Module not found: {:?}", dependency_address)))?
                    .as_ref();
                self.visit_module(dependency_module, sorted_modules)?;
            }
    
            sorted_modules.push(module);
        }
    
        Ok(())
    }
}
```
but failed to make it work.

LukaszRozmej: refactor in #570

preston-evans98 (approved):
LGTM! Thanks @LukaszRozmej

Issue #493, opened 2023-07-14
Split `AddressTrait`


citizen-stig:
# Description

Currently address trait looks like this:

```rust
/// A marker trait for addresses.
pub trait AddressTrait:
    PartialEq
    + core::fmt::Debug
    + Clone
    + AsRef<[u8]>
    + for<'a> TryFrom<&'a [u8], Error = anyhow::Error>
    + Eq
    + Serialize
    + DeserializeOwned
    + From<[u8; 32]>
    + Send
    + Sync
    + core::fmt::Display
{
}
```

And it is used for both Data Availability Layer addresses and Rollup Addresses.

There are 2 issues:

* Not all DA addresses might fit into 32 bytes
* `+ for<'a> TryFrom<&'a [u8], Error = anyhow::Error>` is misused as `FromStr`.
* Nit: `PartialEq` can be removed as there's `Eq` already

So I suggest following traits structure:

```rust
pub trait Address:
    Clone
    + Eq
    + Send
    + Sync
    + core::fmt::Debug
    + core::fmt::Display
    + std::str::FromStr
    + for<'a> TryFrom<&'a [u8], Error = thiserror::Error>
    + AsRef<[u8]>
    + Serialize
    + DeserializeOwned
{
}

pub trait RollupAddress: Address + From<[u8; 32]> {}
```

**Notes**:

This change adds `FromStr` bound and I strongly believe this is essential, because it will simplify receiving addresses via RPC, since we won't need to worry about bytes encoding.

## Things to consider

* Should we bound "Copy" as we assume that addresses are bytes under the hood essentially.
* Shoulw we replce `anyhow::Error` with `thiserror::Error`?

PR #551, opened 2023-07-24
Refactor `CelestiaAddress`

### LINKED ISSUE #469: Refactor `CelestiaAddress` to fix `From<[u8; 32]>`
# Problem

Current implmentation `CelestiaAddress`:

```rust
pub struct CelestiaAddress(pub Vec<u8>);
```

We store raw ASCII string of celestia address "as is", with HRP and without any decoding:

```rust
pub const SEQUENCER_DA_ADDRESS: [u8; 47] = *b"celestia1w7wcupk5gswj25c0khnkey5fwmlndx6t5aarmk";
```

Length of this string is 47 bytes, and even without HRP("celesita") it is 39 bytes. 

Which means that `impl From<[u8; 32]> for CelestiaAddress {` required by `AddressTrait` cannot be done with current representation.

# Suggested solution 1

Change internal represenation of celestia address to store only decoded `Vec<u5>` as `Vec<u8>`. Here's example how it might look like:

TBD: 

```rust
const HRP: &str = "celestia";
#[derive(Debug, PartialEq, Clone, Eq, Serialize, Deserialize, BorshDeserialize, BorshSerialize)]
struct CelestiaAddressNew([u8; 32]);

impl AsRef<[u8]> for CelestiaAddressNew {
    fn as_ref(&self) -> &[u8] {
        self.0.as_ref()
    }
}

impl<'a> TryFrom<&'a [u8]> for CelestiaAddressNew {
    type Error = anyhow::Error;

    fn try_from(value: &'a [u8]) -> Result<Self, Self::Error> {
        match value.len() {
            // Different cases of ASCII
            39 | 47 => {
                todo!()
                // let raw = if value.len() == 47 {
                //    value[HRP.len()..]
                // } else {
                //     value
                // };
                // let mut result: [u8; 32] = [0; 32];
                // for (i, &item) in value.iter().enumerate() {
                //     let v = bech32::u5::try_from_u8(item)?;
                //     result[i] = v.to_u8();
                // }
                // Ok(Self(result))
            }
            // Raw data
            32 => {
                for &item in value {
                    bech32::u5::try_from_u8(item)?;
                }
                // Should not panic, because len is always 32
                let value: [u8; 32] = value.try_into().unwrap();
                Ok(Self(value))
            }
            l => anyhow::bail!("Unexpected length of slice: {}, expect 32, 47 or 39", l),
        }
    }
}

impl Display for CelestiaAddressNew {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        // Should not panic, as we sanitize input before
        let data: Vec<bech32::u5> = self
            .0
            .iter()
            .map(|&i| bech32::u5::try_from_u8(i).unwrap())
            .collect();
        bech32::encode_to_fmt(f, HRP, &data, Variant::Bech32)
            .map_err(|e| std::fmt::Error::custom(e.to_string()))?
    }
}

impl From<[u8; 32]> for CelestiaAddressNew {
    fn from(value: [u8; 32]) -> Self {
        // TODO: Add checks and panic?
        Self(value)
    }
}

impl FromStr for CelestiaAddressNew {
    type Err = anyhow::Error;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let (hrp, raw_address_u5, variant) = bech32::decode(s)?;
        // TODO: Check hrp against hrp
        // TODO: Probably store variant in struct
        let mut value: [u8; 32] = [0; 32];
        println!("FromStr LEN: {}", raw_address_u5.len());
        for (i, raw) in raw_address_u5.iter().enumerate() {
            value[i] = raw.to_u8();
        }
        Ok(Self(value))
    }
}

impl AddressTrait for CelestiaAddressNew {}

#[test]
fn play_compact_address() {
    let raw_address_str = "celestia1w7wcupk5gswj25c0khnkey5fwmlndx6t5aarmk";
    let (hrp, raw_address_u5, variant) = bech32::decode(raw_address_str).unwrap();
    // 32
    println!("L1: {} {:?}", raw_address_u5.len(), raw_address_u5);
    let raw_address_u8 = Vec::<u8>::from_base32(&raw_address_u5).unwrap();
    // 20
    println!("L2: {} {:?}", raw_address_u8.len(), raw_address_u8);
    // 32
    let my_u8: Vec<u8> = raw_address_u5.iter().map(|&u| u.to_u8()).collect();
    println!("L3: {} {:?}", my_u8.len(), my_u8);
    let raw_u8: [u8; 32] = my_u8.try_into().unwrap();

    let address1 = CelestiaAddressNew::from_str(raw_address_str).unwrap();
    let raw_bytes = raw_address_str.as_bytes().to_vec();
    // let address2 = CelestiaAddressNew::try_from(&raw_bytes[..]).unwrap();
    // let address3 = CelestiaAddressNew::try_from(&raw_bytes[HRP.len()..]).unwrap();
    let address4 = CelestiaAddressNew::from(raw_u8);

    // assert_eq!(address1, address2);
    // assert_eq!(address2, address3);
    assert_eq!(address1, address4);

    let output_1 = format!("{}", address1);
    // let output_2 = format!("{}", address2);
    // let output_3 = format!("{}", address3);
    let output_4 = format!("{}", address4);

    assert_eq!(raw_address_str, output_1);
    // assert_eq!(raw_address_str, output_2);
    // assert_eq!(raw_address_str, output_3);
    assert_eq!(raw_address_str, output_4);
}
```

# Suggested solution 2

Change `AddressTrait` to replace ` + for<'a> TryFrom<&'a [u8], Error = anyhow::Error>` with `FromStr`

Current usage of this trait essentially using it from string. This way we empathis that AddressTrait can be only 32bytes in "raw" format, or it should be possible to convert it from string.

Another improvement would be to replace `From<[u8; 32]>` to `TryFrom<[u8; 32]>` so incorrect conversion would not cause panic.

Second suggestion does not contradict first and they both can be implemented

citizen-stig:
# Description

This PR modifies how Celestia addresses are stored, now only containing the data payload of the address.

We've introduced the use of `unsafe` in this PR, with clear documentation explaining its necessity and usage. If this raises concerns, safe version, but with slight performance overhead can be used instead.

Here are the key changes:

* `TryFrom<&[u8]>` now supports either ASCII strings encoded as bytes or the output from `as_ref()`. The only prerequisite is that the byte slice must contain valid `u5` elements. Additionally, if an address is decoded from a `bech32` string, serialized via `as_ref()`, and then decoded from bytes again, it will be decoded as `bech32`. This means the string representation won't remain the same.
* `FromStr` now invokes `bech32::decode` and stores the result.
* `From<[u8;32]>` will trigger a panic if any of the input elements exceed 32 (u5), as this is the internal representation of bech32.
* `Display` implementation avoids allocation and uses writer directly

## Linked Issues
- Fixes #469

## Testing
Proptests have been added

## Docs
Documentation is updated

### DIFF
--- a/adapters/celestia/src/verifier/address.rs
+++ b/adapters/celestia/src/verifier/address.rs
@@ -1,19 +1,24 @@
 use std::fmt::{Display, Formatter};
 use std::str::FromStr;
 
+use bech32::WriteBase32;
 use borsh::{BorshDeserialize, BorshSerialize};
 use serde::{Deserialize, Serialize};
 use sov_rollup_interface::AddressTrait;
 use thiserror::Error;
 
+/// Human Readable Part: "celestia" for Celestia network
 const HRP: &str = "celestia";
+/// Bech32 variant is used for Celestia and CosmosSDK
+const VARIANT: bech32::Variant = bech32::Variant::Bech32;
 
+/// Representation of the address in the Celestia network
+/// https://github.com/celestiaorg/celestia-specs/blob/e59efd63a2165866584833e91e1cb8a6ed8c8203/src/specs/data_structures.md#address
+/// Spec says: "Addresses have a length of 32 bytes.", but in reality it is 32 `u5` elements, which can be compressed as 20 bytes.
 #[derive(
     Debug, PartialEq, Clone, Eq, Serialize, Deserialize, BorshDeserialize, BorshSerialize, Hash,
 )]
-// Raw ASCII bytes, including HRP
-// TODO: https://github.com/Sovereign-Labs/sovereign-sdk/issues/469
-pub struct CelestiaAddress(Vec<u8>);
+pub struct CelestiaAddress([u8; 32]);
 
 impl AsRef<[u8]> for CelestiaAddress {
     fn as_ref(&self) -> &[u8] {
@@ -21,36 +26,65 @@ impl AsRef<[u8]> for CelestiaAddress {
     }
 }
 
+/// Decodes slice of bytes into CelestiaAddress
+/// Treats it as string if it starts with HRP and the rest is valid ASCII
+/// Otherwise just checks if it contains valid `u5` elements and has the correct length.
 impl<'a> TryFrom<&'a [u8]> for CelestiaAddress {
     type Error = anyhow::Error;
 
     fn try_from(value: &'a [u8]) -> Result<Self, Self::Error> {
-        Ok(Self(value.to_vec()))
+        if value.starts_with(HRP.as_bytes()) && value.is_ascii() {
+            // safety, because we checked that it is ASCII
+            let s = unsafe { std::str::from_utf8_unchecked(value) };
+            return CelestiaAddress::from_str(s).map_err(|e| anyhow::anyhow!("{}", e));
+        }
+        if value.len() != 32 {
+            anyhow::bail!("An address must be 32 u5 long");
+        }
+        let mut raw_address = [0u8; 32];
+        for (idx, &item) in value.iter().enumerate() {
+            bech32::u5::try_from_u8(item)
+                .map_err(|e| anyhow::anyhow!("Element at {} is not u5: {}", idx, e))?;
+            raw_address[idx] = item;
+        }
+        Ok(Self(raw_address))
     }
 }
 
+/// Panics if any element is not in range 0..32 (u5)
+/// TODO: Will be removed after https://github.com/Sovereign-Labs/sovereign-sdk/issues/493
 impl From<[u8; 32]> for CelestiaAddress {
     fn from(value: [u8; 32]) -> Self {
-        // TODO: This is completely broken with current implementation.
-        // https://github.com/Sovereign-Labs/sovereign-sdk/issues/469
-        Self(value.to_vec())
+        for item in value {
+            bech32::u5::try_from_u8(item).unwrap();
+        }
+        Self(value)
     }
 }
 
 impl Display for CelestiaAddress {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        let ascii_string = String::from_utf8_lossy(&self.0);
-        write!(f, "{}", ascii_string)
+        let mut w = bech32::Bech32Writer::new(HRP, VARIANT, f)?;
+        for elem in self.0.iter() {
+            // It is ok to unwrap, because we always sanitize data
+            w.write_u5(bech32::u5::try_from_u8(*elem).unwrap())?;
+        }
+        w.finalize()
     }
 }
 
-#[derive(Clone, Debug, Error)]
+#[derive(Clone, Debug, Error, PartialEq)]
 /// An error which occurs while decoding a `CelestialAddress` from a string.
 pub enum CelestiaAddressFromStrError {
-    /// The address has an invalid human readable prefix. Valid addresses must start with the prefix 'celestia'.
-    #[error("The address has an invalid human readable prefix. Valid addresses must start with the prefix 'celestia', but this one began with {0}")]
+    /// The address has an invalid human-readable prefix.
+    /// Valid addresses must start with the prefix 'celestia'.
+    #[error("The address has an invalid human-readable prefix. Valid addresses must start with the prefix 'celestia', but this one began with {0}")]
     InvalidHumanReadablePrefix(String),
-    /// The address could note be decoded as valid bech32
+    /// The address has an invalid human-readable prefix.
+    /// Valid addresses must start with the prefix 'celestia'.
+    #[error("The address has an invalid bech32 variant. Valid addresses must be encoded in Bech32, but this is encoded in Bech32m")]
+    InvalidVariant,
+    /// The address could not be decoded as valid bech32
     #[error("The address could not be decoded as valid bech32: {0}")]
     InvalidBech32(#[from] bech32::Error),
 }
@@ -59,12 +93,24 @@ impl FromStr for CelestiaAddress {
     type Err = CelestiaAddressFromStrError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        // This could be the way to save memory:
-        let (hrp, _raw_address_u5, _variant) = bech32::decode(s)?;
+        let (hrp, raw_address_u5, variant) = bech32::decode(s)?;
         if hrp != HRP {
             return Err(CelestiaAddressFromStrError::InvalidHumanReadablePrefix(hrp));
         }
-        let value = s.as_bytes().to_vec();
+        if variant != VARIANT {
+            return Err(CelestiaAddressFromStrError::InvalidVariant);
+        }
+        if raw_address_u5.len() != 32 {
+            return Err(CelestiaAddressFromStrError::InvalidBech32(
+                bech32::Error::InvalidLength,
+            ));
+        }
+
+        let mut value: [u8; 32] = [0; 32];
+
+        for (idx, &item) in raw_address_u5.iter().enumerate() {
+            value[idx] = item.to_u8();
+        }
         Ok(Self(value))
     }
 }
@@ -73,11 +119,16 @@ impl AddressTrait for CelestiaAddress {}
 
 #[cfg(test)]
 mod tests {
+    use std::hint::black_box;
+
+    use bech32::ToBase32;
+    use proptest::prelude::*;
+
     use super::*;
 
     #[test]
     fn test_address_display_from_string() {
-        let raw_address_str = "celestia1w7wcupk5gswj25c0khnkey5fwmlndx6t5aarmk";
+        let raw_address_str = "celestia1hvp2nfz3r6nqt8mlrzqf9ctwle942tkr0wql75";
         let address = CelestiaAddress::from_str(raw_address_str).unwrap();
         let output = format!("{}", address);
         assert_eq!(raw_address_str, output);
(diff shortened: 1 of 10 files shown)


adapters/celestia/src/verifier/address.rs
```diff
@@ -1,64 +1,116 @@
...
+/// Representation of the address in the Celestia network
 #[derive(Debug, PartialEq, Clone, Eq, Serialize, Deserialize, BorshDeserialize, BorshSerialize)]
-// Raw ASCII bytes, including HRP
-// TODO: https://github.com/Sovereign-Labs/sovereign-sdk/issues/469
-pub struct CelestiaAddress(Vec<u8>);
+pub struct CelestiaAddress {
+    // Storing vector of u5 each as u8
+    data: Vec<u8>,
+    // indication if it is Bech32m or not
+    m: bool,
```
preston-evans98: Does Celestia dynamically select between the standard and `m` variants of bech32? I thought it always used the standard variant

adapters/celestia/src/verifier/address.rs
```diff
@@ -1,64 +1,116 @@
...
 const HRP: &str = "celestia";
 
+/// Representation of the address in the Celestia network
 #[derive(Debug, PartialEq, Clone, Eq, Serialize, Deserialize, BorshDeserialize, BorshSerialize)]
-// Raw ASCII bytes, including HRP
-// TODO: https://github.com/Sovereign-Labs/sovereign-sdk/issues/469
-pub struct CelestiaAddress(Vec<u8>);
+pub struct CelestiaAddress {
+    // Storing vector of u5 each as u8
+    data: Vec<u8>,
```
preston-evans98: Celestia addresses are [u8;32] bech32 encoded. Can we use the `[u8;32]` as our internal representation the same way we do in our `Bech32Address` type?

adapters/celestia/src/verifier/address.rs
```diff
@@ -1,64 +1,116 @@
...
+        if value.len() >= HRP.len() && &value[..HRP.len()] == HRP.as_bytes() && value.is_ascii() {
+            // safety, because we checked that it is ASCII
+            let s = unsafe { std::str::from_utf8_unchecked(value) };
+            return CelestiaAddress::from_str(s);
+        }
+        let mut data = Vec::with_capacity(value.len());
+        for (idx, &item) in value.iter().enumerate() {
+            bech32::u5::try_from_u8(item)
+                .map_err(|e| anyhow::anyhow!("Element at {} is not u5: {}", idx, e))?;
+            data.push(item);
```
preston-evans98: Going along with my previous comment, can we use a [u8;32] internal representation and just use the `bech32::decode` function here?

adapters/celestia/src/verifier/address.rs
```diff
@@ -1,64 +1,116 @@
...
+        }
+        Ok(Self { data, m: true })
     }
 }
 
+/// Panics if any element is not in range 0..32 (u5)
+/// TODO: Will be removed after https://github.com/Sovereign-Labs/sovereign-sdk/issues/493
 impl From<[u8; 32]> for CelestiaAddress {
```
preston-evans98, line 57: If we do change the internal representation, this will also need to be updated

adapters/celestia/src/verifier/address.rs
bkolad: Is it possible to make it two variant enum instead of having a flag?

bkolad (approved):

adapters/celestia/src/verifier/address.rs
citizen-stig: That's good point. I think I overgeneralized implementation. I will double check Celestia codebase to see which variant they use and how many bytes this it.

citizen-stig: Addressed this

adapters/celestia/src/verifier/address.rs
citizen-stig: `bech32::decode` has following syntax:

```rust
pub fn decode(s: &str) -> Result<(String, Vec<u5>, Variant), Error>
```

Which decodes string. 
we store array of 32 u5s internally, so we should make sure, that each element actually valid u5.

adapters/celestia/src/verifier/address.rs
citizen-stig: @preston-evans98 , you are correct, currently celestia and cosmos sdk use `bech32`. When(if) they upgrade, we will do as well. But I believe that this upgrade might be impossible, as string representation of address will change.

adapters/celestia/src/verifier/address.rs
citizen-stig: I believe that according to https://github.com/Sovereign-Labs/sovereign-sdk/issues/493 it is better to be removed completely.

adapters/celestia/src/verifier/address.rs
```diff
@@ -1,56 +1,90 @@
...
 
+/// Decodes slice of bytes into CelestiaAddress
+/// Treats it as string if it starts with HRP and the rest is valid ASCII
+/// Otherwise just checks if it contains valid `u5` elements and has the correct length.
 impl<'a> TryFrom<&'a [u8]> for CelestiaAddress {
     type Error = anyhow::Error;
 
     fn try_from(value: &'a [u8]) -> Result<Self, Self::Error> {
-        Ok(Self(value.to_vec()))
+        if value.len() >= HRP.len() && &value[..HRP.len()] == HRP.as_bytes() && value.is_ascii() {
```
preston-evans98: nit: Can we use `.starts_with(HRP)` instead of `value.len() >= HRP.len() && &value[..HRP.len()] == HRP.as_bytes()`?

adapters/celestia/src/verifier/address.rs
```diff
@@ -1,56 +1,90 @@
...
+        }
+        if value.len() != 32 {
+            anyhow::bail!("An address must be 32 u5 long");
+        }
+        let mut raw_address = [0u8; 32];
+        for (idx, &item) in value.iter().enumerate() {
+            bech32::u5::try_from_u8(item)
+                .map_err(|e| anyhow::anyhow!("Element at {} is not u5: {}", idx, e))?;
+            raw_address[idx] = item;
+        }
```
preston-evans98, line 50: I think this decoding is incorrect. An address is just a sha256 hash of the pubkey, so its bytes are not secretly `u5`s. If we get passed a slice of 32 bytes, we shouldn't do any decoding - just a copy from slice.

adapters/celestia/src/verifier/address.rs
```diff
@@ -1,56 +1,90 @@
...
+/// TODO: Will be removed after https://github.com/Sovereign-Labs/sovereign-sdk/issues/493
 impl From<[u8; 32]> for CelestiaAddress {
     fn from(value: [u8; 32]) -> Self {
-        // TODO: This is completely broken with current implementation.
-        // https://github.com/Sovereign-Labs/sovereign-sdk/issues/469
-        Self(value.to_vec())
+        for item in value {
+            bech32::u5::try_from_u8(item).unwrap();
+        }
+        Self(value)
```
preston-evans98, line 62: I think this decoding is also incorrect

adapters/celestia/src/verifier/address.rs
```diff
@@ -1,56 +1,90 @@
...
 
 impl Display for CelestiaAddress {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        let ascii_string = String::from_utf8_lossy(&self.0);
-        write!(f, "{}", ascii_string)
+        let mut w = bech32::Bech32Writer::new(HRP, VARIANT, f)?;
+        for elem in self.0.iter() {
+            // It is ok to unwrap, because we always sanitize data
+            w.write_u5(bech32::u5::try_from_u8(*elem).unwrap())?;
+        }
```
preston-evans98, line 72: I think this encoding is also incorrect. We'll need to call `to_u5` on the inner `[u8;32]`

adapters/celestia/src/verifier/address.rs
```diff
@@ -59,12 +93,24 @@ impl FromStr for CelestiaAddress {
...
         }
-        let value = s.as_bytes().to_vec();
+        if variant != VARIANT {
+            return Err(CelestiaAddressFromStrError::InvalidVariant);
+        }
+        if raw_address_u5.len() != 32 {
+            return Err(CelestiaAddressFromStrError::InvalidBech32(
+                bech32::Error::InvalidLength,
+            ));
+        }
```
preston-evans98, line 108: I think this check is incorrect. The final (`u8`) length should be 32, not the intermediate length in (`u5`)

adapters/celestia/src/verifier/address.rs
```diff
@@ -91,4 +142,90 @@ mod tests {
         let output = format!("{}", address);
         assert_eq!(raw_address_str, output);
     }
+
```
preston-evans98, line 146: Let's add a test with a real celestia address and make sure it encodes/decodes correctly. I stole this one from the block explorer: `celestia1hvp2nfz3r6nqt8mlrzqf9ctwle942tkr0wql75`

preston-evans98 (changes requested):

adapters/celestia/src/verifier/address.rs
citizen-stig: Good point!

adapters/celestia/src/verifier/address.rs
citizen-stig: Discussed over zoom

adapters/celestia/src/verifier/address.rs
citizen-stig: Discussed over zoom

adapters/celestia/src/verifier/address.rs
citizen-stig: Discussed over zoom

adapters/celestia/src/verifier/address.rs
citizen-stig: Discussed over zoom

adapters/celestia/src/verifier/address.rs
citizen-stig: Discussed over zoom

citizen-stig:
So summary after call. Current implementation is correct.

The remaining question is how do we want to have "internal" representation.
Should it be u8 that stores u5 or just u5.

preston-evans98 (approved):
Let's open an issue to update the bech32 crate so that we can switch to `[u5;32]` for the internal representation. No need to block this for now though!

PR #553, opened 2023-07-24
Remove hack for prev_block_hash in Celestia genesis

### LINKED ISSUE #441: Remove hack for prev_block_hash in Celestia genesis
Since we added the `prev_block_hash` method to DA layer headers, we need to add special casing for Celestia (and other Tendermint chains) which don't provide a prev_block_hash on for their genesis. Currently, we simply return `[255;32]` if no prev_block_hash is found. We _should_ check if the block is genesis and only return the marker value in that case - other blocks should cause a panic if they're missing the hash.

preston-evans98:
# Description
Tendermint's `Hash` type is an enum with two variants: `sha256` or `None`, where `None` is only used by the Genesis block. This representation violates the Sovereign SDK's requirement that all blocks be uniquely identifiable by a [u8;32]. To fix it, we just need to hardcode a special "placeholder" hash for the genesis block. 

We had previously done this inside our `TmHash` type, but that type was missing block height information - so it could potentially be fooled into returning the placeholder for blocks other than the first. This PR fixes that issue.

## Linked Issues
- Fixes #441

## Testing
I tested this using a local Celestia devnet to ensure that the hack worked if and only if the height was `1`.

### DIFF
--- a/adapters/celestia/src/celestia.rs
+++ b/adapters/celestia/src/celestia.rs
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 use sov_rollup_interface::da::{BlockHeaderTrait as BlockHeader, CountedBufReader};
 use sov_rollup_interface::AddressTrait;
 pub use tendermint::block::Header as TendermintHeader;
+use tendermint::block::Height;
 use tendermint::crypto::default::Sha256;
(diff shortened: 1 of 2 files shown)


bkolad (approved):

citizen-stig (approved):

neysofu:
I rebased to –hopefully– fix CI.

PR #558, opened 2023-07-25
[TEST]: Verify license checking works

neysofu:
I'm adding a dependency to [`nettle`](https://gitlab.com/sequoia-pgp/nettle-rs), which is GPL-versioned, just to see if `dependency-review` will trigger an error. Also testing with https://github.com/rnd-ash/ecu_diagnostics. **This PR will not be merged.**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -468,6 +468,26 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
+version = "0.63.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "36d860121800b2a9a94f9b5604b332d5cffb234ce17609ea479d723dbc9d3885"
+dependencies = [
+ "bitflags 1.3.2",
+ "cexpr",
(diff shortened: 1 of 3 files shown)


neysofu:
Conclusion: https://github.com/actions/dependency-review-action seems to be very poor at detecting dependencies' licenses and doesn't raise any issues when adding a GPL-licensed crate to the workspace. We should evaluate switching to [`cargo-deny`](https://github.com/EmbarkStudios/cargo-deny), which is much more stringent.

cc @bkolad @citizen-stig

citizen-stig:
@neysofu `cargo-udeps` looks like a good choice, I had pretty good experience with it.

neysofu:
`cargo-udeps` is great – it can't do license auditing though, or can it?

citizen-stig:
Sorry, I meant `cargo-deny` =)

neysofu:
Gotcha! Agreed, `cargo-deny` is best for this. Let's switch.

PR #560, opened 2023-07-25
Change STF to work on block level instead of batch level

### LINKED ISSUE #408: Based Sequencing with Soft Confirmations

citizen-stig:
# Summary of Proposed Changes

We propose a significant refactor to improve the logic and efficiency of how blobs are processed in each block, specifically in the context of sequencing with soft confirmation. These changes impact several key components and introduce new types. This write-up will detail the adjustments and their motivation, aiming for clarity and comprehension.

## Changes to `StateTransitionFunction`

I've condensed the three existing methods - `begin_slot`, `apply_blob`, and `end_slot` - into a single method called `apply_slot`. This change hands over the block handling logic to individual StateTransitionFunction (STF) implementations.

New design for `StateTransitionFunction` is as follows:

```rust
pub trait StateTransitionFunction<Vm: Zkvm, B: BlobReaderTrait> {
    type StateRoot;
    type TxReceiptContents: Serialize + DeserializeOwned + Clone;
    type BatchReceiptContents: Serialize + DeserializeOwned + Clone;
    type Witness: Default + Serialize + 'static;
    
    fn apply_slot<'a, I>(
        &mut self,
        witness: Self::Witness,
        blobs: I,
    ) -> SlotResult<
        Self::StateRoot,
        Self::BatchReceiptContents,
        Self::TxReceiptContents,
        Self::Witness,
    >
    where
        I: IntoIterator<Item = &'a mut B>;
}
```

###  Removing `MisbehaviorHint`

As it is currently unused and makes changes of interface more convoluted, we've decided to remove it for now,  and introduce it as part of the `Witness` type

### Introduction of `SlotResult`

`SlotResult` is a new container type, designed to streamline the output from the previous methods `apply_blob` and `end_slot`.

Previously, these methods looked like this:

```rust

    fn apply_blob(
        &mut self,
        blob: &mut B,
        misbehavior_hint: Option<Self::MisbehaviorProof>,
    ) -> BatchReceipt<Self::BatchReceiptContents, Self::TxReceiptContents>;

    fn end_slot(&mut self) -> (Self::StateRoot, Self::Witness);
```

Now, I introduce `SlotResult`:

```rust
///  - S - generic for state root
///  - B - generic for batch receipt contents
///  - T - generic for transaction receipt contents
///  - W - generic for witness
pub struct SlotResult<S, B, T, W> {
    pub state_root: S,
    pub batch_receipts: Vec<BatchReceipt<B, T>>,
    pub witness: W,
}
```

I decided to introduce `SlotResult` instead of using a type alias for a three-element tuple to increase readability.

### `BlobTransactionTrait` renamed to `BlobReaderTrait`

I've renamed `BlobTransactionTrait` to `BlobReaderTrait` to better reflect its actual role. This trait doesn't carry any transaction-specific aspects and might be used for Attestations or Proofs in the future.

### 'static bound for some traits

Traits like `BlobReaderTrait`, `AddressTrait` and `STF::Witness` associated type  now carry a `'static` bound. This constraint implies that only owned types can implement these traits, which I consider a reasonable requirement, though we're open to feedback on this matter.

### Other Noteworthy Changes

* Benchmarks now solely measure apply_slot. While this reduces granularity, it better represents our new design choice.
* Tests that previously checked whether a result is saved if end_slot isn't called are now ignored. I plan to transition these to a new testing strategy, where execution fails mid-way through applying blobs or even mid-transaction.

## Linked Issues
- Related to #408 

## Testing
Automated tests have been updated

## Docs
Doc strings have been updated

### DIFF
--- a/examples/demo-prover/host/src/main.rs
+++ b/examples/demo-prover/host/src/main.rs
@@ -73,9 +73,10 @@ async fn main() -> Result<(), anyhow::Error> {
         demo.init_chain(genesis_config);
     }
 
-    demo.begin_slot(Default::default());
-    let (prev_state_root, _) = demo.end_slot();
-    let mut prev_state_root = prev_state_root.0;
+    let mut prev_state_root = {
+        let res = demo.apply_slot(Default::default(), []);
+        res.state_root.0
+    };
 
     for height in rollup_config.start_height.. {
         let mut host = Risc0Host::new(ROLLUP_ELF);
@@ -88,41 +89,31 @@ async fn main() -> Result<(), anyhow::Error> {
         let filtered_block = da_service.get_finalized_at(height).await?;
         let header_hash = hex::encode(filtered_block.header.header.hash());
         host.write_to_guest(&filtered_block.header);
-        let (blob_txs, inclusion_proof, completeness_proof) = da_service
+        let (mut blobs, inclusion_proof, completeness_proof) = da_service
             .extract_relevant_txs_with_proof(&filtered_block)
             .await;
 
+        info!(
+            "Extracted {} relevant blobs at height {} header 0x{}",
+            blobs.len(),
+            height,
+            header_hash,
+        );
+
         host.write_to_guest(&inclusion_proof);
         host.write_to_guest(&completeness_proof);
+        host.write_to_guest(&blobs);
 
-        demo.begin_slot(Default::default());
-        if blob_txs.is_empty() {
-            info!(
-                "Block at height {} with header 0x{} has no batches, skip proving",
-                height, header_hash
-            );
-            continue;
-        }
-        info!("Block has {} batches", blob_txs.len());
-        for mut blob in blob_txs.clone() {
-            let receipt = demo.apply_blob(&mut blob, None);
-            info!(
-                "batch with hash=0x{} has been applied",
-                hex::encode(receipt.batch_hash)
-            );
-        }
-        // Write txs only after they been read, so verification can be done properly
-        host.write_to_guest(&blob_txs);
+        let result = demo.apply_slot(Default::default(), &mut blobs);
 
-        let (next_state_root, witness) = demo.end_slot();
-        host.write_to_guest(&witness);
+        host.write_to_guest(&result.witness);
 
         info!("Starting proving...");
         let receipt = host.run().expect("Prover should run successfully");
         info!("Start verifying..");
         receipt.verify(ROLLUP_ID).expect("Receipt should be valid");
 
-        prev_state_root = next_state_root.0;
+        prev_state_root = result.state_root.0;
         info!("Completed proving and verifying block {height}");
     }
 
--- a/examples/demo-prover/methods/guest/src/bin/rollup.rs
+++ b/examples/demo-prover/methods/guest/src/bin/rollup.rs
@@ -36,13 +36,13 @@ pub fn main() {
     env::write(&"Prev root hash read\n");
     // Step 1: read tx list
     let header: CelestiaHeader = guest.read_from_host();
-    env::write(&"header read\n");
+    env::write(&"header has been read\n");
     let inclusion_proof: <CelestiaSpec as DaSpec>::InclusionMultiProof = guest.read_from_host();
-    env::write(&"inclusion proof read\n");
+    env::write(&"inclusion proof has been read\n");
     let completeness_proof: <CelestiaSpec as DaSpec>::CompletenessProof = guest.read_from_host();
-    env::write(&"completeness proof read\n");
+    env::write(&"completeness proof has been read\n");
     let mut blobs: Vec<BlobWithSender> = guest.read_from_host();
-    env::write(&"txs read\n");
+    env::write(&"blobs have been read\n");
 
     // Step 2: Apply blobs
     let mut demo_runner = <ZkAppRunner<Risc0Guest, BlobWithSender> as StateTransitionRunner<
@@ -53,21 +53,18 @@ pub fn main() {
     let demo = demo_runner.inner_mut();
 
     let witness: ArrayWitness = guest.read_from_host();
-    env::write(&"Witness read\n");
+    env::write(&"Witness have been read\n");
 
-    demo.begin_slot(witness);
-    env::write(&"Slot has begun\n");
-    for blob in &mut blobs {
-        demo.apply_blob(blob, None);
-        env::write(&"Blob applied\n");
-    }
-    let (state_root, _) = demo.end_slot();
-    env::write(&"Slot has ended\n");
+    env::write(&"Applying slot...\n");
+    let result = demo.apply_slot(witness, &mut blobs);
+
+    env::write(&"Slot has been applied\n");
 
     // Step 3: Verify tx list
     let verifier = CelestiaVerifier::new(jupiter::verifier::RollupParams {
         namespace: ROLLUP_NAMESPACE,
     });
+
     let validity_condition = verifier
         .verify_relevant_tx_list::<NoOpHasher>(&header, &blobs, inclusion_proof, completeness_proof)
         .expect("Transaction list must be correct");
@@ -75,7 +72,7 @@ pub fn main() {
 
     let output = StateTransition {
         initial_state_root: prev_state_root_hash,
-        final_state_root: state_root.0,
+        final_state_root: result.state_root.0,
         validity_condition,
     };
     env::commit(&output);
--- a/module-system/sov-modules-stf-template/src/lib.rs
+++ b/module-system/sov-modules-stf-template/src/lib.rs
@@ -8,10 +8,11 @@ pub use app_template::AppTemplate;
 pub use batch::Batch;
 use sov_modules_api::hooks::{ApplyBlobHooks, TxHooks};
 use sov_modules_api::{Context, DispatchCall, Genesis, Spec};
-use sov_rollup_interface::da::BlobTransactionTrait;
-use sov_rollup_interface::stf::{BatchReceipt, StateTransitionFunction};
+use sov_rollup_interface::da::BlobReaderTrait;
+use sov_rollup_interface::stf::{SlotResult, StateTransitionFunction};
 use sov_rollup_interface::zk::Zkvm;
 use sov_state::{StateCheckpoint, Storage};
+use tracing::info;
 pub use tx_verifier::RawTx;
 
 /// The receipts of all the transactions in a batch.
@@ -54,7 +55,25 @@ pub enum SlashingReason {
     InvalidTransactionEncoding,
 }
 
-impl<C: Context, RT, Vm: Zkvm, B: BlobTransactionTrait> StateTransitionFunction<Vm, B>
+impl<C: Context, RT, Vm: Zkvm, B: BlobReaderTrait> AppTemplate<C, RT, Vm, B> {
+    fn begin_slot(&mut self, witness: <<C as Spec>::Storage as Storage>::Witness) {
+        self.checkpoint = Some(StateCheckpoint::with_witness(
+            self.current_storage.clone(),
+            witness,
+        ));
+    }
+
+    fn end_slot(&mut self) -> (jmt::RootHash, <<C as Spec>::Storage as Storage>::Witness) {
+        let (cache_log, witness) = self.checkpoint.take().unwrap().freeze();
+        let root_hash = self
+            .current_storage
+            .validate_and_commit(cache_log, &witness)
+            .expect("jellyfish merkle tree update must succeed");
+        (jmt::RootHash(root_hash), witness)
+    }
+}
+
+impl<C: Context, RT, Vm: Zkvm, B: BlobReaderTrait> StateTransitionFunction<Vm, B>
     for AppTemplate<C, RT, Vm, B>
 where
     RT: DispatchCall<Context = C>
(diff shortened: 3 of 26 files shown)


examples/demo-prover/host/src/main.rs
```diff
@@ -88,41 +90,36 @@ async fn main() -> Result<(), anyhow::Error> {
...
-        receipt.verify(ROLLUP_ID).expect("Receipt should be valid");
+        host.write_to_guest(&blobs);
+        host.write_to_guest(&result.witness);
 
-        prev_state_root = next_state_root.0;
+        info!("Starting proving...");
+        let session = host
+            .run_without_proving()
+            .expect("Prover should run successfully");
+        info!("Verification skipped");
```
citizen-stig: Temporarely for faster debugging

examples/demo-prover/host/src/main.rs
```diff
@@ -73,12 +73,14 @@ async fn main() -> Result<(), anyhow::Error> {
...
-    let (prev_state_root, _) = demo.end_slot();
-    let mut prev_state_root = prev_state_root.0;
+    let mut prev_state_root = {
+        let res = demo.apply_block(0, Default::default(), []);
+        res.state_root.0
+    };
 
     for height in rollup_config.start_height.. {
         let mut host = Risc0Host::new(ROLLUP_ELF);
+        host.write_to_guest(height);
```
preston-evans98: This seems risky. Let's compute it and store it in state instead. (I think Theo already has code to do this)

examples/demo-prover/methods/guest/src/bin/rollup.rs
```diff
@@ -32,17 +32,21 @@ pub fn main() {
...
     let guest = Risc0Guest;
 
+    let block_height: u64 = guest.read_from_host();
+    env::write(&"Block height read\n");
     let prev_state_root_hash: [u8; 32] = guest.read_from_host();
     env::write(&"Prev root hash read\n");
     // Step 1: read tx list
     let header: CelestiaHeader = guest.read_from_host();
-    env::write(&"header read\n");
+    env::write(&"header has ben read\n");
```
preston-evans98: nit: "has been"

module-system/sov-state/src/witness.rs
```diff
@@ -7,7 +7,7 @@ use serde::{Deserialize, Serialize};
 
 // TODO: Refactor witness trait so it only require Serialize / Deserialize
 //   https://github.com/Sovereign-Labs/sovereign-sdk/issues/263
-pub trait Witness: Default + Serialize {
+pub trait Witness: Default + Serialize + 'static {
```
preston-evans98: Question: Where did the new `'static` requirement come from?

rollup-interface/src/state_machine/stf.rs
```diff
@@ -91,39 +106,41 @@ pub trait StateTransitionFunction<Vm: Zkvm, B: BlobTransactionTrait> {
...
-        blob: &mut B,
-        misbehavior_hint: Option<Self::MisbehaviorProof>,
-    ) -> BatchReceipt<Self::BatchReceiptContents, Self::TxReceiptContents>;
-
-    /// Called once at the *end* of each DA layer block (i.e. after all rollup blobs have been processed)
-    /// Commits state changes to the database
     ///
-    fn end_slot(&mut self) -> (Self::StateRoot, Self::Witness);
+    /// Commits state changes to the database
+    fn apply_block<'a, I>(
```
preston-evans98: This represents a change in terminology. Before we called a da layer block a slot. Let's either change this to slot, or make sure we've updated all of our docs to use the new term.

rollup-interface/src/state_machine/stf.rs
```diff
@@ -91,39 +106,41 @@ pub trait StateTransitionFunction<Vm: Zkvm, B: BlobTransactionTrait> {
...
+    /// The concrete blob type is defined by the DA layer implementation,
+    /// which is why we use a generic here instead of an associated type.
+    ///
     /// Misbehavior hint allows prover optimizations - the sequencer can be slashed
     /// for including a transaction which fails stateless checks (i.e. has an invalid signature) -
     /// and in that case we ignore his entire batch.
     /// This method lets you give a hint to the prover telling
     /// it where that invalid signature is, so that it can skip signature checks on other transactions.
-    /// (If the misbehavior hint is wrong, then the host is malicious so we can
+    /// (If the misbehavior hint is wrong, then the host is malicious, so we can
```
preston-evans98: nit: outdated doc

rollup-interface/src/state_machine/stf.rs
```diff
@@ -91,39 +106,41 @@ pub trait StateTransitionFunction<Vm: Zkvm, B: BlobTransactionTrait> {
...
-    /// Apply a blob/batch of transactions to the rollup, slashing the sequencer who proposed the blob on failure.
-    /// The concrete blob type is defined by the DA layer implementation, which is why we use a generic here instead
-    /// of an associated type.
+    ///
+    /// Applies batches of transactions to the rollup,
+    /// slashing the sequencer who proposed the blob on failure.
+    /// The concrete blob type is defined by the DA layer implementation,
+    /// which is why we use a generic here instead of an associated type.
+    ///
     /// Misbehavior hint allows prover optimizations - the sequencer can be slashed
```
preston-evans98: nit: outdated doc

rollup-interface/src/state_machine/stf.rs
```diff
@@ -91,39 +106,41 @@ pub trait StateTransitionFunction<Vm: Zkvm, B: BlobTransactionTrait> {
...
-    ) -> BatchReceipt<Self::BatchReceiptContents, Self::TxReceiptContents>;
-
-    /// Called once at the *end* of each DA layer block (i.e. after all rollup blobs have been processed)
-    /// Commits state changes to the database
     ///
-    fn end_slot(&mut self) -> (Self::StateRoot, Self::Witness);
+    /// Commits state changes to the database
+    fn apply_block<'a, I>(
+        &mut self,
+        height: u64,
```
preston-evans98: Let's remove this argument, since if the prover lies we don't have a way to detect it. We can still get access to the height inside the STF using the chain state module

preston-evans98 (commented):
Looks pretty good overall. Just two issues  - first the height isn't authenticated, so the prover could like about it and cause a fork. And second the docs/specs still need to be updated with the new STF API.

module-system/sov-modules-stf-template/src/lib.rs
```diff
@@ -87,30 +104,53 @@ where
...
+        Self::Witness,
+    >
+    where
+        I: IntoIterator<Item = &'a mut B>,
+    {
+        self.begin_slot(witness);
+
+        let mut batch_receipts = vec![];
+        for (blob_idx, blob) in blobs.into_iter().enumerate() {
+            let batch_receipt = match self.apply_blob(blob) {
```
bkolad: You can use `?` instead of pattern matching.

bkolad (approved):
LGTM once @preston-evans98 comments are addressed.

module-system/sov-state/src/witness.rs
citizen-stig: Resolved!

module-system/sov-modules-stf-template/src/lib.rs
citizen-stig: That's a good idea.

It didn't work originally, because of :

```
124 |               let batch_receipt = self.apply_blob(blob)?;
    |                                                        ^ cannot use the `?` operator in a method that returns `BlockResult<RootHash, SequencerOutcome, TxEffect, <<C as Spec>::Storage as Storage>::Witness>`
    |
    = help: the trait `FromResidual<Result<Infallible, ApplyBatchError>>` is not implemented for `BlockResult<RootHash, SequencerOutcome, TxEffect, <<C as Spec>::Storage as Storage>::Witness>`
    
```

But I can implement needed trait.

rollup-interface/src/state_machine/stf.rs
citizen-stig: Yes, and I think it is not blocking this PR, it will only be needed for soft confirmations PR.

examples/demo-prover/host/src/main.rs
citizen-stig: Yes, I've removed it from this PR, because it is not necessary now.

It will be added when blobs deferring will start to be used

module-system/sov-modules-stf-template/src/lib.rs
citizen-stig: That's actually interesting observation.

If single blob in slot has failed to be applied, should we abort whole slot? If so, then `?` can be used.

But if slot should be continued, then just simple `.into()` is fine.

citizen-stig: Actually in current logic, we keep going. @preston-evans98 please confirm if this is desired behaviour.

Also, I've get rid of `match` for something more compact.

rollup-interface/src/state_machine/mod.rs
```diff
@@ -27,5 +27,6 @@ pub trait AddressTrait:
     + Sync
     + core::fmt::Display
     + std::hash::Hash
+    + 'static
```
preston-evans98, line 30: Question: do we still need this `'static`?

preston-evans98 (commented):
Looks great!

preston-evans98 (approved):

rollup-interface/src/state_machine/mod.rs
citizen-stig: Yes, because BlobReaderTrait require it:

```rust
pub trait BlobReaderTrait: Serialize + DeserializeOwned + Send + Sync + 'static
```

and we need it there, as mentioned in PR description

module-system/sov-modules-stf-template/src/lib.rs
preston-evans98: Yes, we want to keep going. Otherwise, it would be pretty easy to DOS the rollup by posting invalid blobs in every slot

PR #569, opened 2023-07-25
Preston/sorted dependencies

preston-evans98:
# Description
This PR shows how we can simplify the interface of the sorting function. It's still a WIP - needs a few comments to be cleaned up, and a test for the `panic` case with a cyclic dependency.

### DIFF
--- a/module-system/sov-modules-api/src/lib.rs
+++ b/module-system/sov-modules-api/src/lib.rs
@@ -29,7 +29,7 @@ pub mod macros {
 }
 
 use core::fmt::{self, Debug, Display};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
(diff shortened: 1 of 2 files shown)


preston-evans98:
This PR makes the choice to sort in-place which definitely isn't necessary, and it goes a little overboard with the comments. I'd clean it up, but I ran out of time :)

Anyway, feel free to accept and modify it or just close it out - it's just meant to show how we can clean up the interface.

LukaszRozmej:
I think my solution of breaking down the function is actually simpler. Maybe you could update your PR based on those changes?
I see you are keen on sorting in-place which is fine, but I don't see it as crucial as the number of elements will be small by design.

LukaszRozmej:
refactor to struct suggested by @bkolad #570

PR #570, opened 2023-07-26
sorted dependencies struct refactor

LukaszRozmej:


module-system/sov-modules-api/src/lib.rs
```diff
@@ -301,46 +301,66 @@ pub trait RpcRunner {
...
+/// Sorts ModuleInfo objects by their dependencies
+pub fn sort_modules_by_dependencies<C: Context>(
+    modules: Vec<&dyn ModuleInfo<Context = C>>,
+) -> Result<Vec<&dyn ModuleInfo<Context = C>>, anyhow::Error> {
+    let mut module_visitor = ModuleVisitor::<C>::new();
+    module_visitor.visit_modules(modules)?;
+    Ok(module_visitor.sorted_modules)
 }
 
 /// Accepts Vec<> of tuples (&ModuleInfo, &TValue), and returns Vec<&TValue> sorted by mapped module dependencies
```
preston-evans98, line 380: It looks to me like we don't rely on the fact that `TValue` is passed by reference. Am I missing something? If not, we should use this signature: 
```rust
pub fn sort_values_by_modules_dependencies<C: Context, TValue>(
    module_value_tuples: Vec<(&dyn ModuleInfo<Context = C>, TValue)>,
) -> Result<Vec<TValue>, anyhow::Error>
```
That signature allows you to pass types by reference, but lets you pass by value as well if you want and removes  the lifetime constraint.

Also, I know I pushed for the rename from `T` which became `TValue` - but with the new signature I don't think we need it anymore. Now it's very clear that T is not related to the module in any way except for being part of the tuple - so `T` is idiomatic.

preston-evans98 (approved):
Looks very good - just a small nit on the new function signature

preston-evans98 (approved):

PR #571, opened 2023-07-26
EVM: Remove hardcoded private key form `Ethereum` RPC

### LINKED ISSUE #514: EVM: Pass `sov` private key as a dpendency in `eth_sendRawTransaction` endpoint
## Description
The EVM module is a "standard" SOV module. As a result, call messages sent to the DA layer adhere to a standard format that allows them to be processed by the rollup. However, we also want to provide users with the same API as a standard Ethereum node. For instance, the `eth_sendRawTransaction` endpoint should be able to receive messages from typical Ethereum wallets.

To achieve compatibility with the `Sov-rollup` system, the exposed `eth_sendRawTransaction` endpoint accepts a standard Ethereum transaction. Internally, it wraps this transaction within a rollup transaction, which is then submitted to the DA layer. The rollup transaction must be signed (the Ethereum wallet already signs the Ethereum transaction). Consequently, the `eth_sendRawTransaction` endpoint implementation stores a private key and uses it to sign the message.

First, we need to remove the hardcoded key from the `eth_sendRawTransaction`  endpoint and pass it as a dependency. Once this is accomplished, we can reevaluate the design and determine a better way to integrate the Ethereum message format with the rollup message format.

## Acceptance Criteria
The private key for signing sov-transactions is passed as a parameter in the `eth_sendRawTransaction` endpoint logic.

bkolad:
# Description
1. Removes a hardcoded private key from the `Ethereum` rpc (now the key is passes as parameter)
2. Reorganizes `demo-rollup/main.rs`: the RPC registration logic was moved to separate methods. 

## Linked Issues
- Fixes #514

## Testing
No tests were added.

## Docs
No changes in the docs.


examples/demo-rollup/src/main.rs
```diff
@@ -12,6 +12,8 @@ use demo_stf::runner_config::from_toml_path;
 use demo_stf::runtime::{get_rpc_methods, GenesisConfig};
 use jsonrpsee::core::server::rpc_module::Methods;
 use jupiter::da_service::CelestiaService;
+#[cfg(feature = "experimental")]
+use jupiter::da_service::DaServiceConfig;
```
citizen-stig, line 16: Why this config is experimental?

citizen-stig (approved):

examples/demo-rollup/src/main.rs
bkolad: It is not, I just use the type in experimental code and need to guard it with the flag so we don't get an "unused import" warning.

Issue #188, opened 2023-04-21
Tracking Issue: Improve CI


preston-evans98:
## Comprehensiveness
- [x] Add coverage (see https://lib.rs/crates/cargo-tarpaulin)
- [x] Consider adding license check (cc @citizen-stig) https://github.com/Sovereign-Labs/sovereign-sdk/pull/559
- [x] Consider adding clippy
- [x] Check docstrings
- [x] Introduce job dependencies. Run `check` first, and other jobs run only if it succeeds. So if PR is completely broken, not a lot of resources will be wasted. 

## Performance
CI runtime has gone from 30 minutes to 10 minutes. (Thanks @citizen-stig for switching to `sccache`in #197). We would like to reduce this time even further.

## Some Ideas (TODO: expand, evaluate tradeoffs)
- Separate sov-state into two crates: one which provides the API and one which provides the implementation. This would let us avoid depending on RocksDB everywhere. 
- Move the sov-modules-macros tests into their own crate instead of relying on `trybuild` so that they can build in parallel with other items instead of sequentially.

Thanks @bkolad for the suggestion

preston-evans98:
@citizen-stig Suggests that `sccache` might alleviate the compilation bottleneck.

citizen-stig:
Checked running time of the tests and we are fine, except for 2 tests, which are big offenders:

<img width="478" alt="Screenshot 2023-04-24 at 20 29 54" src="https://user-images.githubusercontent.com/3115300/234084729-c80f2d4e-3d7f-4b25-8da0-6a5c079ad1c9.png">

But still they should not lead to 20-30 minutes time of the CI

citizen-stig:
Some tips recommended by @bkolad :

https://www.youtube.com/watch?v=xUH-4y92jPg

citizen-stig:
Also there were some comments in #439 in description

citizen-stig:
@neysofu you might be interested in this one.

neysofu:
I would propose we close this –it's outdated at this point– and create single issues for more specific improvements we want to make.

preston-evans98:
Agreed. Closing.

PR #572, opened 2023-07-26
CI cache improvements

neysofu:
## Description
- Automatically removes caches associated with a branch, once its PR gets merged. This helps GHA's cache eviction algorithm to perform better.
- Use [`mold`](https://github.com/rui314/mold) to reduce compilation times.
- Use [`cargo-nextest`](https://nexte.st/) to run tests faster.
- **Completely remove all usage of zsccache`.** After some analysis of our CI logs, it looks like `sccache` does not play well with GHA for projects as chunky as ours. It sends way too many cache restore requests, which result in 429s and slowdowns.
- As an alternative to `sccache`, I'm doubling down on our usage of `Swatinem/rust-cache`. It now also caches `examples/demo-prover` compilation artifacts, and caches are shared between jobs when possible. Some more tweaks may be needed.

## Linked Issues
Related to #188.

## Testing
CI passes.


bkolad (approved):

.github/workflows/rust.yml
```diff
@@ -1,5 +1,40 @@
...
+#   a cache key.
+# - When caching Rust compilation artifacts, keep in mind that different `cargo`
+#   commands will use different profiles
+#   (https://doc.rust-lang.org/cargo/reference/profiles.html). Learn what you
+#   can reuse between one job and another and don't assume two commands will
+#   just share caches without conflicts.
+# - Be extremely aware of cache thrashing a.k.a. churning. GitHub Actions' cache
+#   allows for 10GiB of data which is easily exceeded if not careful.
+#   Sometimes it's better not to cache than cache excessively.
+#   We might even want to disable caching writes for non-default branches
```
citizen-stig: I think this is a good idea!!!! Cache write only on nightly/stable.

maybe let's do it right away?

.github/workflows/rust.yml
```diff
@@ -1,5 +1,40 @@
...
+# - Learn cache invalidation rules of `Swatinem/rust-cache` before making
+#   changes, e.g. what happens when `rustc --version` changes or `Cargo.lock`
+#   changes (or is missing).
+# - The jobs dependency tree is the way it is to accomodate for sharing caches,
+#   not necessarily because it makes logical sense to run one job after the
+#   other. This is due to the fact that we can't share caches between jobs that
+#   run in parallel.
+# - `sccache` is a good alternative to `Swatinem/rust-cache`, but it behaves
+#   poorly with GHA and often incurs into cache requests rate limits. We should
+#   probably explore `sccache` with a different backend.
```
citizen-stig, line 32: Can you please create an issue for this? Maybe next CI person can look into it.

.github/workflows/rust.yml
```diff
@@ -21,128 +57,88 @@ concurrency:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
     name: check
+    runs-on:
+      group: 8-cores_32GB_Ubuntu Group
     timeout-minutes: 60
-    env:
-      RUSTFLAGS: -D warnings
```
citizen-stig, line 28: Why we've removed `RUSTFLAGS: -D warnings` ? 

It will allow warnings to sneak in

.github/workflows/rust.yml
```diff
@@ -21,128 +57,88 @@ concurrency:
...
+    # `cargo-hack` uses the same profile as `cargo check` and doesn't require
+    # building dependencies, only chceking them, so we can share caches
+    # effectively.
+    needs: check
     runs-on:
       group: 8-cores_32GB_Ubuntu Group
-    name: features
     timeout-minutes: 60
-    env:
-      RUSTFLAGS: -D warnings
```
citizen-stig, line 73: Same for warnings, it will allow to have "unused imports" when feature gated incorrectly

.github/workflows/rust.yml
neysofu: I've enabled it globally instead, under `env:`

citizen-stig: Oh, I see! 

I think that's good. Maybe a little strict, since bad formatting will fail tests, even if they are correct, giving a less information. But on the other hand having it global prevents forgetting it if new job added.

citizen-stig (approved):

.github/workflows/rust.yml
neysofu: Tracked in https://github.com/Sovereign-Labs/sovereign-sdk/issues/584.

.github/workflows/rust.yml
neysofu: Oh, you're right, I didn't even realize. I think it doesn't matter much, but we can remove it from tests if the next CI person has a preference!

.github/workflows/rust.yml
neysofu: Sure, let's try! https://github.com/Sovereign-Labs/sovereign-sdk/pull/572/commits/b604d67de446f5a856d1a6e9993e91106394b4a0.

.github/workflows/rust.yml
```diff
@@ -1,5 +1,40 @@
...
+#   a cache key.
+# - When caching Rust compilation artifacts, keep in mind that different `cargo`
+#   commands will use different profiles
+#   (https://doc.rust-lang.org/cargo/reference/profiles.html). Learn what you
+#   can reuse between one job and another and don't assume two commands will
+#   just share caches without conflicts.
+# - Be extremely aware of cache thrashing a.k.a. churning. GitHub Actions' cache
+#   allows for 10GiB of data which is easily exceeded if not careful.
+#   Sometimes it's better not to cache than cache excessively.
+#   Disabling cache writes for non-default branches altogether if cache churning
```
citizen-stig, line 21: I think it is a very good idea!

What do you think about doing it right away?

PR #582, opened 2023-07-31
Change query methods return type from `SomeType` to `RpcResult<SomeType>`

bkolad:
# Description
1. Query methods now return `RpcResult<SomeType>` instead of just `SomeType`. This change allows queries to return `jsonrpsee_types::error::Error` if something goes wrong (before the `RpcResult` wrapping was done on a macro level so the querirs couldn't throw an error).
2. The `wrap_in_jsonprsee_result`  method  which wrapped `RpcResult` inside a macro, has been removed.

## Testing
Tests have been updated accordingly.

## Docs
Relevant documentation has been updated.


dubbelosix (approved):
lgtm!

citizen-stig (approved):

PR #592, opened 2023-07-31
Add missing docs for bank, prover-incentives, demo-rollup and const-rollup-config

theochap:
# Description
This PR adds missing docs for bank, prover-incentives, demo-rollup and const-rollup-config.

## Linked Issues
- Related to  #438 

## Testing
No testing needed as it is only documentation


module-system/module-implementations/sov-bank/src/lib.rs
```diff
@@ -5,24 +5,35 @@ mod query;
 mod token;
```
bkolad, line 7: Let's add the following lints:
```
#![deny(missing_docs)]
```
So it will fail compilation if a public doc is missing and add README as crate level doc.

module-system/module-implementations/sov-prover-incentives/src/lib.rs
```diff
@@ -8,7 +8,9 @@ mod tests;
 mod query;
```
bkolad, line 10: Let's add the following lints:
```
#![deny(missing_docs)]
```
So it will fail compilation if a public doc is missing and add README as crate level doc.

bkolad (approved):

PR #593, opened 2023-07-31
Refactoring buf reader

### LINKED ISSUE #462: Refactor `CountedBufReader::acc()` for better usability
Currently callers of `CountedBufReader::acc()` method have assumptions about state of the reader.
For example `DaVerifier::verify_relevant_tx_list` assumes that accumulator has been consumed, when it is not always the case.

So I suggest following changes:

1. rename `acc()` to `accumulator()`
2. this method return enum with explicitly stating 

```rust

enum AccumulatorStatus<'a> {
    Completed(&'a [u8]),
    InProgress(&'a [u8]),
}

impl<B: Buf> CountedBufReader<B> {
    pub fn accumulator(&self) -> AccumulatorStatus {
        match self.inner.remaining() {
            0 => AccumulatorStatus::Completed(&self.reading_acc),
            _ => AccumulatorStatus::InProgress(&self.reading_acc),
        }
    }
}
```

This way callers would be explicitly decide what they are going to do with accumulated data and if it matches their expectation.

For example this how verification part could be changed after that:

```rust

let mut blob_iter = blob_ref.data();
let mut blob_data = vec![0; blob_iter.remaining()];
blob_iter.copy_to_slice(blob_data.as_mut_slice());
let tx_data = tx.data().acc();

let tx_data = match tx.data().accumulator() {
    AccumulatorStatus::Completed(d) => d,
    AccumulatorStatus::InProgress(_) => {
        panic!("Blob wasn't read completely")
    }
}

assert_eq!(blob_data, *tx_data);
```

Or it can consume buffer and try again

theochap:
# Description
This refactors the accumulator from the `CountedBufReader` to indicate a reading status. The spec given by the #462 has been applied. The field `read_acc` has been renamed `accumulator` and an `AccumulatorStatus` enum has been added. 
In the Celestia adapter, if a given blob has not been read entirely, a new error `ValidationError::IncompleteData` is returned.

## Linked Issues
- Fixes #462 

## Testing
The existing test suite has been run successfully

## Docs
The public methods and structures have been documented


rollup-interface/src/state_machine/da.rs
```diff
@@ -68,6 +68,27 @@ pub trait DaVerifier {
...
+    /// The underlying buffer has been completely read and [`Vec<u8>`] contains the result
+    Completed(Vec<u8>),
+    /// The underlying buffer still contains elements to be read. [`Vec<u8>`] contains the
+    /// accumulated elements.
+    InProgress(Vec<u8>),
+}
+
+impl AccumulatorStatus {
+    /// Returns the accumulator vector contained in the wrapper
+    pub fn inner_vec(&mut self) -> &mut Vec<u8> {
```
citizen-stig: I believe this function can return slice, so then we don't need a mutable self, and in the code it is used in `from_slice` anyway.

citizen-stig: Also, what is use case for this method?

rollup-interface/src/state_machine/da.rs
```diff
@@ -68,6 +68,27 @@ pub trait DaVerifier {
     ) -> Result<Self::ValidityCondition, Self::Error>;
 }
 
+/// [`AccumulatorStatus`] is a wrapper around an accumulator vector that specifies
+/// whether a [`CountedBufReader`] has finished reading the underlying buffer.
+#[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub enum AccumulatorStatus {
```
citizen-stig: I would like to suggest not owned wrapper, but referenced:

```rust
pub enum AccumulatorStatus<'a> {
    Completed(&'a[u8]),
    InProgress(&'a[u8])
}
```

and remove `pub inner_vec`, so callers have to deal with matching themself.

Because API with `inner_vec` can be misused easily, when dev can always call it like that:

```
let i_need_data_now = counted_buf_reader.accumulator().inner_vec()
```

And then it just does not care if it was read or not.

citizen-stig: Oh, now I see that `inner_vec` is used by CountedBufReader itself.

I think that `CountedBufReader` can maintain internal vec as it was doing before, just forbid direct access to it, only via `AccumulatorStatus` (which can be just renamed to`Accumulator`)

citizen-stig (approved):

PR #596, opened 2023-08-01
Based Sequencer with Soft Confirmations: deferring blob from execution

### LINKED ISSUE #654: Based Sequencer with Soft Confirmations: handle preferred sequencer exit
Rollup with based sequencing can function in 2 modes:

1. If rollup has preferred sequencer, blobs execution from all other sequencers are always deferred to next slot.
2. If there's no preferred sequencer, blobs are executed immediately. 

The problem is when preferred sequencer exits. In the next slot after sequencer exits, there potentially can be executed 2x number of blobs: all deferred and all current blobs. This can make this slot fail, as it won't fit into gas.

Potential solutions

1. in case of exit, execute only half of current number of blobs, and defer the rest, until there's only few
2. If rollup had preferred sequencer at genesis, always defer blobs
3. Do not allow preferred sequencer to exit (almost the same as the previous)

citizen-stig:
# Description

## Introducing `BlobSelector` rollup capability

So I went ahead with rollup capability approach and added `BlobSelector` trait.

It takes iterator of mutable references to `BlobReaderTrait` and returns some of them back (or all) plus some previously stored blobs.

In order to fully use this trait, `chain_state` module is needed, so default stf can pull correct slot numbers

```rust
pub trait BlobSelector {
    type Context: Context;

    /// It takes 2 arguments:
    /// 1. `current_blobs` - blobs that were received from the network for the current slot.
    /// 2. `working_set` - the working to access storage.
    /// It returns vector containing mix of borrowed and owned blobs.
    fn get_blobs_for_this_slot<'a, I, B>(
        &self,
        current_blobs: I,
        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
    ) -> Vec<BlobRefOrOwned<'a, B>>
    where
        B: BlobReaderTrait,
        I: IntoIterator<Item = &'a mut B>;
}


pub enum BlobRefOrOwned<'a, B: BlobReaderTrait> {
    /// Mutable reference
    Ref(&'a mut B),
    /// Owned blob
    Owned(B),
}
```

## `BlobStorage` module depends on `SequencerRegistry`

`BlobStorage` queries if there's preferred sequencer.

## `BlobStorage` module depends on `ChainState` module

`BlobStorage` queries current slot number.

* https://github.com/Sovereign-Labs/sovereign-sdk/pull/598

## `BlobStorage` has immutable constant on how many slots blobs are deferred

Pros:
 * No hard coded value
 * Extensible, it can be decided separately how this value is going to change
 
Cons:
 * Extra state access on every blob.

Deferring blobs - page 1 copy

## Linked Issues
- Related to #408

## Testing

## Docs


module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::BlobStorage;
+
+impl<C: Context> BlobSelector for BlobStorage<C> {
+    type Context = C;
+
+    fn get_blobs_for_this_slot<'a, I, B>(
+        &self,
+        current_slot: u64,
+        current_blobs: I,
+        pull_from_slot: u64,
```
preston-evans98: Question: I had assumed that the blob_selector would be responsible for this internally - it seems like whatever is invoking it won't have enough context to make very smart decisions here. How did you envision this being used?

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -0,0 +1,57 @@
...
+        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
+    ) -> (Vec<&'a mut B>, Vec<B>)
```
preston-evans98: It seems a little odd that we return two separate `Vec`s here; feels like the trait matches very specifically with our initial use case - which makes me wonder if it should be a trait at all. As it is, when we're not using the deferred blob feature we'll always return `(current_blobs, vec[]!)` - which seems a little odd.

Maybe we can make a more general interface like:

```
fn get_blobs_for_this_slot<'a, I, B>(
        &self,
        current_slot: u64,
        current_blobs: I,
        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
    ) -> Vec<B>
```
    
  I see there's a slight mismatch between the borrowed and owned `Blob` types, but we can hide that behind an enum or a trait.

module-system/sov-modules-api/src/capabilities.rs
```diff
@@ -0,0 +1,35 @@
+#![deny(missing_docs)]
+
+//! Rollup capabilities rust module defines rollup specific capabilities,
+//! which can be implemented by different module
```
preston-evans98: nit: this comment is a little hard to parse. How about

```
//! The rollup capabilities module defines "capabilities" that rollups must
//! provide if they wish to use the standard app template. If you don't 
//! want to provide these capabilites, you can bypass the Sovereign module-system completely and write a 
//! state transition function from scratch. See here for docs (TODO: insert link)
```

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
citizen-stig: That's good point.

I think this is related to the question about dependencies. Now I realize, that it is better to encapsulated logic in blob storage.

This means following:

1. `BlobStorage` depends on `SequencerRegistry`, as in this PR. For getting preferred sequencer
2. `BlobStorage` depends on `ChainState` module, so it can pull slot number.
3. `BlobStorage` has "number of blocks to defer" as state variable, which is defined during genesis. How it is changed can be defined later.
4. `BlobStorage` implements capability `BlobSelector`, which only accepts `current_slots` and `working_set`. So completely different logic can be done using different blob selector.

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
citizen-stig: > I see there's a slight mismatch between the borrowed and owned Blob types, but we can hide that behind an enum or a trait.

This was motivation to have a tuple. If we return `Vec<B>` means that we clone borrowed slots, which is big memory operation and I believe it might impact ZK performance. 

What is `enum` or `trait` for return type do have in mind for return type?

citizen-stig: I decided that `enum` is a way to go, quite straighforward, also it can implement public method to return mutable reference, so

module-system/sov-modules-api/src/capabilities.rs
citizen-stig: Yes, it needs some love.

I decided to always add `#![deny(missing_docs)]` to any new rust module I write, so debt is not accumulated.

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -0,0 +1,71 @@
...
+
+        if preferred_sequencer.is_none() {
+            return Ok(current_blobs
+                .into_iter()
+                .map(Into::into)
+                .chain(past_deferred.into_iter().map(Into::into))
+                .collect());
+        }
+
+        let preferred_sequencer = preferred_sequencer.unwrap();
```
bkolad: this and line 28 can be replaced with `if let  preferred_sequencer = Some(..) else {..}`

module-system/sov-modules-api/src/capabilities.rs
```diff
@@ -0,0 +1,62 @@
...
+//! and write a state transition function from scratch.
+//! [See here for docs](https://github.com/Sovereign-Labs/sovereign-sdk/blob/nightly/examples/demo-stf/README.md)
+
+use sov_rollup_interface::da::BlobReaderTrait;
+use sov_state::WorkingSet;
+
+use crate::{Context, Spec};
+
+/// Container type for mixing borrowed and owned blobs.
+pub enum BlobRefOrOwned<'a, B: BlobReaderTrait> {
```
bkolad, line 16: Is it possible to use https://doc.rust-lang.org/std/borrow/enum.Cow.html instead?

module-system/sov-modules-api/src/capabilities.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// It takes 2 arguments:
+    /// 1. `current_blobs` - blobs that were received from the network for the current slot.
+    /// 2. `working_set` - the working to access storage.
+    /// It returns a vector containing a mix of borrowed and owned blobs.
+    fn get_blobs_for_this_slot<'a, I, B>(
+        &self,
+        current_blobs: I,
+        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
+    ) -> anyhow::Result<Vec<BlobRefOrOwned<'a, B>>>
```
bkolad, line 58: Let's use `thiserror`?

module-system/module-implementations/sov-blob-storage/src/lib.rs
```diff
@@ -2,12 +2,21 @@
```
bkolad: Maybe not for this PR but this module is missing `README`

bkolad (approved):

module-system/sov-modules-api/src/capabilities.rs
citizen-stig: Unfortunately using `Cow` wouldn't work, because we need to access `&'a mut` which this enum implements, but Cow doesn't. Also `Cow` has unnecessary for us trait bound for `Clone`

module-system/sov-modules-api/src/capabilities.rs
citizen-stig: As discussed over call.

This is similar to hooks, in terms that modules implement this, and we treat module as "apps", means that they work with `anyhow`

module-system/module-implementations/sov-blob-storage/src/lib.rs
citizen-stig: Done!

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
citizen-stig: Done

module-system/module-implementations/sov-sequencer-registry/src/call.rs
```diff
@@ -33,7 +33,7 @@ impl<C: sov_modules_api::Context> SequencerRegistry<C> {
         Ok(CallResponse::default())
     }
 
-    pub(crate) fn exit(
+    pub fn exit(
```
citizen-stig: FIX

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -0,0 +1,88 @@
...
+            .unwrap_or_default();
+        let preferred_sequencer = self.get_preferred_sequencer(working_set);
+
+        let preferred_sequencer = if let Some(sequencer) = preferred_sequencer {
+            sequencer
+        } else {
+            return Ok(past_deferred
+                .into_iter()
+                .map(Into::into)
+                .chain(current_blobs.into_iter().map(Into::into))
```
preston-evans98, line 37: This line means that we'll process all the blobs from both the previous block and the current block at once if the sequencer exits. Are we sure that's safe?

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -0,0 +1,88 @@
...
+                priority_blobs.push(blob);
+            } else {
+                to_defer.push(blob);
+            }
+        }
+
+        // TODO: chain state module: https://github.com/Sovereign-Labs/sovereign-sdk/pull/598/
+        self.slot_number.set(&(current_slot + 1), working_set);
+
+        if !to_defer.is_empty() {
```
preston-evans98, line 55: nit: I think this if is unnecessary since we use iterator combinators everywhere

module-system/module-implementations/integration-tests/src/chain_state/helpers.rs
```diff
@@ -73,6 +74,22 @@ impl<C: Context, Cond: ValidityCondition> SlotHooks<Cond> for TestRuntime<C, Con
...
+    fn get_blobs_for_this_slot<'a, I, B>(
+        &self,
+        current_blobs: I,
+        _working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
+    ) -> anyhow::Result<Vec<BlobRefOrOwned<'a, B>>>
+    where
+        B: BlobReaderTrait,
+        I: IntoIterator<Item = &'a mut B>,
+    {
+        Ok(current_blobs.into_iter().map(Into::into).collect())
```
preston-evans98, line 89: I think this blob selector should still filter out blobs from unregistered senders, right?

citizen-stig: It can, but then it will be double work for all honest sequencers.

Sequencer registry has apply_blob hook implemented, where check and all appropriate logic is implemented.

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
citizen-stig: That's a good point.

So no blob from unregistered sequencer is going to be actually executed, because `apply_blob_hook` from sequencer registry will figure them out.

But overall gas usage might be 2x normal block size. And that might be a problem. 

But what would be solution?

The code can chain only half of the current blobs and save the rest for the next block, and repeat it until there's only 4 blobs saved (for instance).

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
citizen-stig: but we allocate inside.

It is single reference, but still an allocation. Also if we have any other logic related to saving, it is inside this branch.

Previous messages made me think about potential attach vector.

For simplicity, imagine there are preferred sequencer, but he is inactive at the moment. So all blobs just moved to next blob.

Malicious sequencer can submit 3 blobs: A, B and C in single slot number 100. 

The sequencer is registered at the beginning of the slot 100. But blob A contains exit transaction, which makes blobs B and C to be sent from unregistered sequencer.

ApplySlotHooks only run once per slot and all blobs are checked against the state at the beginning of the slot. Slo ApplySlotHooks will save all 3 blobs: A, B and C to be executed in slot 101, but blobs B and C will be discarded by `ApplyBlobHooks`. Considering that storing blobs is expensive operation, this might be costly.

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
citizen-stig: As discussed on the call, let's move this issue out of the scope of this PR.

Tracking issue: https://github.com/Sovereign-Labs/sovereign-sdk/issues/654

module-system/module-implementations/integration-tests/src/chain_state/helpers.rs
preston-evans98: Hmm, that's a bit confusing. Can we move that logic into the blob selector?

citizen-stig: `BlobSelector` is run once per slot. 
`ApplyBlobHooks` are run before every blob. 

It will be quite big of change to merge these 2 functionalities.

preston-evans98: Synced with @citizen-stig - we're going to merge this as-is, and then move all of the sequencer billing logic into BlobSelector in a follow-up.

preston-evans98 (approved):

PR #608, opened 2023-08-01
Remove `jmt::SimpleHasher` from rollup interface; Add PrivateKey trait

### LINKED ISSUE #225: Remove `jmt` dependency from SDK package
Currently, the Sovereign SDK re-exports the `jmt` package and exposes one trait which depends on it (`TreeWitnessReader`). Since the SDK is supposed to be agnostic about what state commitment is used by implementers, we should remove that dependency. We should be able to move all of the relevant functionality into the `sov-state` package.

### LINKED ISSUE #552: Tracking Issue: `sov-cli` improvements
## Background
`sov-cli` grew organically from a simple transaction generator for the bank module into a complex command line tool. Unfortunately, this organic growth has made it more and more unwieldy over time. At Modular summit, one of the biggest pieces of feedback we received was that the cli was unintuitive and hard to use. This is completely fair. Now that #423, #525, and #446  have landed, we have the scaffolding to re-architect the CLI into a standalone crate and provide a much better UX. 

## Tasks
- [x] Move sov-cli into its own crate
- [ ] Refactor `sov-cli` interface to make it easier to use
- [x] Add persistance (i.e. for private keys) to `sov-cli`
- [x] Add RPC capabilities to `sov-cli`

preston-evans98:
# Description
Remove `jmt::SimpleHasher` from the rollup interface and replace it with `Digest` from the `digest` crate.  (See #225)

Add PrivateKey trait and implement it for `DefaultPrivateKey`. In the process, change our `sign/verify` APIs to work on messages rather than message hashes. This reduces the risk of accidental misuse, since not all hashes can be combined with all signature schemes.

## Linked Issues
- Fixes #225 
- Related to #552 

## Testing
Tests have been adjusted. I added one test to ensure that a "roundtrip" sign and verify cycle worked as expected using the new traits.

## Docs
Docs have been updated to reflect the API changes.

### DIFF
--- a/module-system/sov-modules-api/src/lib.rs
+++ b/module-system/sov-modules-api/src/lib.rs
@@ -10,6 +10,7 @@ pub mod hooks;
 mod prefix;
 mod response;
 mod serde_address;
+pub mod test_utils;
 #[cfg(test)]
 mod tests;
 pub mod transaction;
@@ -17,6 +18,8 @@ pub mod transaction;
 #[cfg(feature = "macros")]
 extern crate sov_modules_macros;
 
+use digest::typenum::U32;
+use digest::Digest;
 #[cfg(feature = "macros")]
 pub use sov_modules_macros::{
     DispatchCall, Genesis, MessageCodec, ModuleCallJsonSchema, ModuleInfo,
@@ -40,8 +43,7 @@ pub use error::Error;
 pub use prefix::Prefix;
 pub use response::CallResponse;
 use serde::{Deserialize, Serialize};
-pub use sov_rollup_interface::crypto::SimpleHasher as Hasher;
-pub use sov_rollup_interface::AddressTrait;
+pub use sov_rollup_interface::{digest, AddressTrait};
 use sov_state::{Storage, Witness, WorkingSet};
 use thiserror::Error;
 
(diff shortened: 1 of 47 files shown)


module-system/sov-modules-api/src/lib.rs
```diff
@@ -138,6 +136,19 @@ pub trait PublicKey {
     fn to_address<A: AddressTrait>(&self) -> A;
 }
 
+/// PublicKey used in the Module System.
```
bkolad: should be *PrivateKey*

module-system/sov-modules-api/src/lib.rs
```diff
@@ -183,6 +194,14 @@ pub trait Spec {
         + Sync
         + FromStr<Err = anyhow::Error>;
 
+    /// The public key used for digital signatures
+    #[cfg(feature = "native")]
+    type PrivateKey: Debug
+        + Send
+        + Sync
+        + for<'a> TryFrom<&'a [u8], Error = anyhow::Error>
```
bkolad, line 204: Let's use `thiserror` instead of `anyhow`?

bkolad (approved):

citizen-stig (approved):

module-system/sov-modules-api/src/lib.rs
preston-evans98: I think this is the same issue we ran into before - if we want to use `thiserror`, we have to add an extra associated type to the `Spec` trait for the Error

module-system/sov-modules-api/src/lib.rs
preston-evans98: Good catch!

PR #614, opened 2023-08-03
StateTransitionRunner refactoring. 

bkolad:
# Description 
The goal of this PR is to restructure the logic for running the rollup, enabling it to function with various combinations of `Runtime/Data-Availability/State-Transition-Function`. This modification has several benefits:

1. It makes the logic independent of a particular rollup (currently it is hardcoded in `demo-rollup`).
2. Simplifies testing (point 1).
3. Facilitates the implementation of in-memory DA.

After this PR, the responsibilities of the relevant components are as follows:

- `sov-stf-runner`: Abstracts the logic for executing the STF.
- `demo-stf`: Contains code specific to a particular `Runtime`, including `hooks`, `configs`, and a concrete `STFR`.
- `demo-rollup`: Uses both `sov-stf-runner` and `demo-stf`, acting as the glue code to initiate the rollup process.

### The changes introduced in this PR are as follows:

1. `StateTransitionRunner(STFR)` is transformed into a concrete type and moved from the `rollup-interface` to a separate crate `full-node/sov-stf-runner`. By making it a concrete type, we were able to eliminate `DemoAppRunner, NativeAppRunner, and DemoApp`, all of which essentially represent the same type beneath the surface.  To simplify matters, we removed `DaVerifier` from the `STFR` since it was initially included solely for demonstration purposes. However, it can easily be reintroduced if necessary. The  logic for running the rollup lives in the `run` method.

2. We eliminated the `RpcRunner` trait. As there were no methods or types reliant on this trait, so we didn't really make use of it.

3. All the `ledger_XXX ` rpc methods ware moved to `sov-stf-runner` because they are common to any kind of rollup.
 
### Future work:
1. Implement `MockDA` and test `sov-stf-runner`.
2. The `demo-prover/host` also has blob processing logic which is similar to the one in `STFR`, but it simultaneously writes to the host.  If we hide the `host` behind a trait and make it a noop for the `node` we will be able to extend `STFR` and use it both in the `prover` & `node`
3. The code in `sov-stf-runner` was mostly copied from `demo-stf` & `demo-rollup` and it needs to be cleanup further. For example we have many different `Configs` with the same name. The code is mostly internal to the crate and don't leak outside.

## Linked Issues
- Related to # (issue) 

## Testing
Ci passes.

## Docs
All related docs to `StateTransitionRunner` updated.


full-node/sov-stf-runner/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+    DA: DaService,
+    Vm: Zkvm,
+    ST: StateTransitionFunction<Vm, <<DA as DaService>::Spec as DaSpec>::BlobTransaction>,
+{
+    start_height: u64,
+    da_service: DA,
+    app: ST,
+    ledger_db: LedgerDB,
+    state_root: StateRoot<ST, Vm, DA>,
+    socket_address: SocketAddr,
```
citizen-stig: I assume it is `listen_address`

citizen-stig:
@bkolad, does design in this PR accommodates running a prover? Or is it **only** for a full node?

bkolad:
> 2\. The `demo-prover/host` also has blob processing logic which is similar to the one in `STFR`, but it simultaneously writes to the host.  If we hide the `host` behind a trait and make it a noop for the `node` we will be able to extend `STFR` and use it both in the `prover` & `node`

Future work:
"The demo-prover/host also has blob processing logic which is similar to the one in STFR, but it simultaneously writes to the host. If we hide the host behind a trait and make it a noop for the node we will be able to extend STFR and use it both in the prover & node"

full-node/sov-stf-runner/src/lib.rs
bkolad: Done

examples/demo-rollup/rollup_config.toml
```diff
@@ -4,7 +4,7 @@ start_height = 1
 
 [da]
 # The JWT used to authenticate with the celestia light client. Instructions for generating this token can be found in the README
-celestia_rpc_auth_token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJwdWJsaWMiLCJyZWFkIiwid3JpdGUiLCJhZG1pbiJdfQ.YO1sa3cHeTsZx-NR06sIXSRNrLAJvr4oKuvFq3MvaJo"
```
preston-evans98, line 7: This looks like it got added by mistake in some previous PR. Can we rever to the placeholder?

module-system/sov-modules-stf-template/src/lib.rs
```diff
@@ -15,6 +15,15 @@ use sov_state::{StateCheckpoint, Storage};
 use tracing::info;
 pub use tx_verifier::RawTx;
 
+///TODO
```
preston-evans98: What's the TODO here?

preston-evans98 (approved):
Looks amazing overall. Just two small things!

examples/demo-rollup/rollup_config.toml
bkolad: Fixed

module-system/sov-modules-stf-template/src/lib.rs
bkolad: Missing comment, just fixed.

full-node/sov-stf-runner/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+    Vm: Zkvm,
+    ST: StateTransitionFunction<Vm, <<DA as DaService>::Spec as DaSpec>::BlobTransaction>,
+{
+    /// Creates a new `StateTransitionRunner` runner.
+    pub fn new(
+        rollup_config: RollupConfig,
+        da_service: DA,
+        ledger_db: LedgerDB,
+        mut app: ST,
+        is_storage_empty: bool,
```
citizen-stig: Just a suggeston, should we name this as `should_init_chain`? 

Because what if we have different condition for initialization independent from storage.

full-node/sov-stf-runner/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+                .unwrap();
+
+            info!("Starting RPC server at {} ", server.local_addr().unwrap());
+            let _server_handle = server.start(methods).unwrap();
+            futures::future::pending::<()>().await;
+        });
+    }
+
+    /// Runs the rollup.
+    pub async fn run(&mut self) -> Result<(), anyhow::Error> {
```
citizen-stig, line 113: It is out of scope of this PR, but how stf runner shuts down? 

I guess it can do graceful shutdown after processing slot.

bkolad: Now it just runs, let's implement shutdown in another PR.

module-system/sov-modules-stf-template/src/app_template.rs
```diff
@@ -64,10 +63,7 @@ impl From<ApplyBatchError> for BatchReceipt<SequencerOutcome, TxEffect> {
 
 impl<C: Context, RT, Vm, B: BlobReaderTrait> AppTemplate<C, RT, Vm, B>
 where
-    RT: DispatchCall<Context = C>
-        + Genesis<Context = C>
-        + TxHooks<Context = C>
-        + ApplyBlobHooks<Context = C, BlobResult = SequencerOutcome>,
```
citizen-stig, line 70: Nice

citizen-stig (approved):

full-node/sov-stf-runner/src/lib.rs
bkolad: done

PR #631, opened 2023-08-09
Add Avail Adapter for Sovereign SDK Integration and Examples

vibhurajeev:
# Description

## Overview
This pull request introduces the Avail Adapter to the Sovereign SDK. Additionally, examples demonstrating how to utilize this new adapter have been included.

## Changes Made
- Added the Avail Adapter to the Sovereign SDK.
- Included examples showcasing how to utilize the Avail Adapter.

## Documentation
The Avail Adapter and its integration into the Sovereign SDK are documented in the following locations:
- Updated the SDK documentation to include information about the new Avail Adapter and its usage.
- The examples repository contains detailed documentation on how to use the Avail Adapter for various use cases.

## Additional Changes
A minor enhancement was made to the sov-stf-runner:
- Separated da-config and rollup-config for improved clarity and maintainability.


adapters/avail/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "presence"
```
aterentic-ethernal, line 2: Do we want to rename it?

examples/demo-rollup-avail/src/ledger_rpc.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// Extract the args from an RPC query, being liberal in what is accepted.
+    /// To query for a list of items, users can either pass a list of ids, or tuple containing a list of ids and a query mode
+    pub fn extract_query_args<I: DeserializeOwned>(
+        params: jsonrpsee::types::Params,
+    ) -> Result<QueryArgs<I>, jsonrpsee::core::Error> {
+        if let Ok(args) = params.parse() {
+            return Ok(args);
+        }
+        let ids: Vec<I> = params.parse()?;
+        Ok(QueryArgs(ids, Default::default()))
```
aterentic-ethernal, line 78: It seems that this never extracts ids, since line 78 is never reached, is that an issue or just obsolete code?

examples/demo-rollup-avail/src/ledger_rpc.rs
```diff
@@ -0,0 +1,80 @@
...
+///    Example Query: `curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"ledger_getBatches","params":[[1, 2], "Standard"],"id":1}' http://127.0.0.1:12345`
+/// - ledger_getTransactions
+///    Example Query: `curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"ledger_getBatches","params":[[1, 2], "Full"],"id":1}' http://127.0.0.1:12345`
+/// - ledger_getEvents
+///    Example Query: `curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"ledger_getBatches","params":[1, 2],"id":1}' http://127.0.0.1:12345`
+fn register_ledger_rpc_methods<B: Serialize + DeserializeOwned, T: Serialize + DeserializeOwned>(
+    rpc: &mut RpcModule<LedgerDB>,
+) -> Result<(), jsonrpsee::core::Error> {
+    rpc.register_method("ledger_getHead", move |_, db| {
+        db.get_head::<B, T>().map_err(|e| e.into())
```
aterentic-ethernal, line 26: We can replace this and following `map_err` with `?` since `?` is doing implicite `into`

adapters/avail/src/verifier.rs
```diff
@@ -0,0 +1,62 @@
...
+        _inclusion_proof: <Self::Spec as DaSpec>::InclusionMultiProof,
+        _completeness_proof: <Self::Spec as DaSpec>::CompletenessProof,
+    ) -> Result<Self::ValidityCondition, Self::Error> {
+        let validity_condition = ChainValidityCondition {
+            prev_hash: *block_header.prev_hash().inner(),
+            block_hash: *block_header.hash().inner(),
+        };
+
+        Ok(validity_condition)
+    }
```
preston-evans98: @vibhurajeev, this implementation breaks soundness.

The only way for a rollup to work without a smart contract is to allow the end-user (light client) to verify that the rollup's fork choice rule was followed. If the verifier can't check this, then a malicious prover can create a *valid* proof which extends a non-canonical chain, ultimately causing honest light clients to accept the wrong rollup state. 

In this implementation, the SNARK doesn't commit to the transactions that it's processed. It commits to the block header, but since this function doesn't tie the actual bytes of data that the rollup processes back to the block header, that doesn't constrain the prover. Given a block header, the prover can pretend that the block contained whatever transactions he wants and the SNARK won't stop him. 

As we discussed offline, it's ok to push the work of verifying the transaction list outside of the SNARK - that's exactly what `ValidityConditions` were added for. But, the SNARK still has to commit to the actual set of blobs that it received from the prover, not just the alleged block header.

I think the correct way to implement this is roughly...

1. Add an additional field to the `ChainValidityCondition` for `tx_root`, and put the hash of the `_txs: &[<Self::Spec as DaSpec>::BlobTransaction]` argument in that field
2. Add a `check`er for the validity condition which downloads the data from a light client and enforces that it matches those nodes. 
3. Change the `combine` rule for validity conditions to ensure that when you check the validity condition for block N, you're also forced ot recursively check back to Genesis. (This is necessary since just because the prover gave the correct tx list in block N doesn't mean he gave the correct one in block N-1).

preston-evans98 (changes requested):

preston-evans98:
@vibhurajeev, finally had a chance to start working on this! I've PR'd the first chunk (the adapter itself as #905). It needed a few modifications to pass CI, so I've opened a PR against your fork to show what the changes are. No need to merge that PR - just give an approval on #905 if it looks good to you.

cemozerr:
Given that #905 is merged, can we close this PR? @preston-evans98

PR #648, opened 2023-08-11
Allow custom (de)serialization logic for `#[state]`

### LINKED ISSUE #550: Allow storage of `#[state]` values with alternate encodings
## Background
The `jmt` crate implements a byte-oriented Key/Value store. The Sovereign SDK transforms that byte store into a typed key/value store using the `StateValue` and `StateMap` abstractions.  When a `#[state]` value is set, the underlying implementation uses the `StorageKey::new()` and `StorageValue::new()` constructors to serialize the key and value with `borsh` before inserting them into the current `WorkingSet`. 

While this system works very well for keys and values which implement `borsh`, it has a few limitations:
1. Items which don't implement the borsh serialize/deserialize traits can't be stored, so many foreign types have to be wrapped in a newtype
2. It adds an additional layer of serialization when storing pre-encoded types
3. It prevents us from supporting IBC, since IBC requires that the chain's authenticated state tree must store the raw protobuf encodings of certain messages.

## Desired Outcome
To support IBC, we need to be able to take the Rust struct representing an IBC message and store its protobuf encoding into the state tree without introducing any additional serialization. There are at least two designs which will support this pattern. 

### Option 1: Add a (phantom) serializer to the StateValue
In this option, we change the API of StateValue from `StateValue<T>` to `StateValue<T, C>`. A rough sketch of the APIs is as follows: 

```rust 
pub struct StateValue<V, C> { 
    _phantom: PhantomData<(V, C)>,
    prefix: Prefix,
}

impl<T, C: SomeCodecTrait> StateValue<T, C> { 
  fn set(&self, value: &T, working_set: WorkingSet<...>) { 
    let value = StorageValue::new::<C>::(value);
    working_set.set(self.prefix(), value)
  } 
} 

impl StorageValue { 
  fn new<C: SomeCodecTrait>(value: &C) -> Self { 
     let encoded_value = C::encode(value);
        Self {
            value: Arc::new(encoded_value),
        }
  }
} 

```

Note that this description is slightly inaccurate, since it pretends that the `WorkingSet` accepts a `StorageValue` as an argument. In reality, the `WorkingSet` accepts the raw type and calls the `StorageValue` constructor internally - so we would need to add generics to the `WorkingSet::set` function as well.

### Option 2: Add a new `RawStateValue` type; add a  `from_bytes` implementation to `StorageKey` and `StorageValue`.
```
pub struct RawStateValue { 
    prefix: Prefix,
}

impl<T, C: SomeCodecTrait> RawStateValue<T, C> { 
  fn set(&self, value: Vec<u8> working_set: &mut WorkingSet<...>) { 
    let value = StorageValue::from_bytes(value);
    working_set.set(self.prefix(), value)
  } 
} 

impl StorageValue { 
  fn from_bytes(value: Vec<u8>) -> Self { 
        Self {
            value: Arc::new(value),
        }
  }
} 

```

neysofu:
# Description
This PR adds support for custom encoding & decoding logic within `#[state]` items other than Borsh.

We're achieving that by adding a default type parameter ("codec") to `StateValue` and `StateMap`, as well as an additional generic to some `WorkingSet` and `StorageKey` methods. The default type parameter is `sov_state::codec::BorshCodec`, which is a wrapper around `borsh` (duh) and makes it so previously written modules don't require any changes.

Three traits are involved with the codec mechanisms introduced here:
- `StateKeyCodec`;
- `StateValueCodec`;
- `StateCodec`, which is a marker trait for types that implement both of the above.

These traits are defined in `sov_state::codec`, where two pre-built implementors are also found:
- `BorshCodec`, the default codec;
- `PairOfCodecs`, which is a utility codec used internally by `StateValue` and could also be of use to module authors.

# How to use this feature

To switch the codec used by a `#[state]` item, the module authors needs to make two changes:
- change the type parameter of their `State[Map|Value]` from the default `BorshCodec` into another type of their choosing;
- add a new nested argument to `#[state]`, like so:

  ```rust
  #[derive(ModuleInfo)]
  struct Module {
      // ...
      #[state(codec_builder = "my_module::CustomCodec::build")]
      state: StateValue<u32, CustomCodec>
  }
  ```

  this last step is optional; only if the builder function differs from `Default::default`.

This feature required some changes to the `derive(ModuleInfo)` macro, which has been extensively rewritten to improve code clarity (hopefully!). Some of the tests were modified to accommodate for slight changes in macro behavior, but they are largely the same. 

## Linked Issues
Closes https://github.com/Sovereign-Labs/sovereign-sdk/issues/550.

## Testing
- New `trybuild` tests.
- New integration test `custom_codec_must_be_used.rs`.

## Docs
All newly introduced traits and codec implementors have been documented. I also added docstrings to some other methods.


module-system/sov-modules-macros/tests/custom_codec_must_be_used.rs
```diff
@@ -0,0 +1,46 @@
...
+#[derive(ModuleInfo)]
+struct TestModule<C>
+where
+    C: Context,
+{
+    #[address]
+    pub address: C::Address,
+
+    #[state(codec_builder = "Default::default")]
+    pub state_value: StateValue<u32, CustomCodec>,
```
bkolad, line 16: Are we getting a user-friendly error if `CustomCodec` doesn't implement `StateValueCodec`? Let's add a test for it.

bkolad (approved):
Thx, looks great. I will use this functionality for the `RLP` encoding in the EVM module.

module-system/sov-modules-macros/tests/custom_codec_must_be_used.rs
neysofu: Good thinking, I'll do that

neysofu: Unfortunately we're not in control of type bounds -related error messages within the proc macro. I have modified the type bounds to imporve the error message, though: https://github.com/Sovereign-Labs/sovereign-sdk/pull/648/commits/11f13dc0a8bcd8b08a791689f39d447ea57e534f

Issue #419, opened 2023-06-19
Design: Optimistic Light Clients


preston-evans98:
# Design: Optimistic Sovereign Rollups

## Background

Currently, Sovereign only supports full nodes. Although we have support for zk-proof generation, our node implementation always re-executes all transactions starting from rollup genesis. As we move toward production implementations, we'll need support for faster syncing modes.

### Efficiency

- Sync *time* should *not* depend on cumulative gas used on the rollup.
    - In other words, the sync should be constant-time *per block*
- Sync *bandwidth* requirements should be less than syncing a full node. In other words, nodes should *not* download every transaction during sync (unless the underlying DA layer requires it).

### Correctness

The following assumptions may be required to ensure that a node ends up at the correct state:

- The node has access to a correct view of the DA layer's consensus (i.e. it's header chain)
- (ORUs only) All invalid state transition claims are challenged within some period (say, 24 hrs)
- (Non-succinct DA layers only) The node has (untrusted) access to the DA chain's history

The following additional assumptions *may* be introduced if necessary to speed up sync in the "happy case":

- Sovereign Labs (or another centralized entity) can operate *untrusted* but relatively available (95%+ percent uptime) centralized infrastructure to aid light clients
- Most rollup nodes are connected to at least one honest peer via the p2p network.

The following assumptions are insecure and may not be relied on:

- All nodes are connected to an honest peer via the *rollup* p2p network.
- Rollup nodes know which of their peers are honest/malicious
- Sovereign Labs (or any other centralized entity) is able to run *trusted* infrastructure and/or provide honest peer connections

### Basic Design: Attestations and Challenges

Sovereign Optimistic Rollups should separate their data into two namespaces, a "syncing" namespace and a "transaction (tx)" namespace.  ***********************************Rollups must process all of the data from both namespaces.*********************************** As in the existing Sovereign design, blobs can be posted to the tx namespace in any order - so the latest rollup state can only be computed after a DA layer block is accepted. 

In addition to "sequencers", we introduce a new role "attesters". The job of an attester is to inform light clients that a state transition has occurred by posting a special `Attestation` blob to the "syncing" namespace. Like sequencers, attesters must register by staking tokens.

```rust
/// An attestation that a particular DA layer block transitioned the rollup state to some value
struct Attestation<StateProof> {
  initial_state_root: [u8;32],
  /// The hash of the block in which the transition occurred
  da_block_hash: [u8;32]
  /// The alleged post-state root
[...]

PR #650, opened 2023-08-11
Optimistic

theochap:
# Description

This PR introduces the attester incentives module, which is the main component of the optimistic rollup that handles the relations with attesters and challengers - as well as the attestation processing workflow in itself.

## Linked Issues

Related to https://github.com/Sovereign-Labs/sovereign-sdk/issues/419

## Testing

Completely Unit tested. An end to end integration test will be added with the next PR that plugs this module to the app template interface.

## Docs

The macro `#[deny(missing_docs)] have been added to the module.


module-system/module-implementations/sov-attester-incentives/src/helpers.rs
```diff
@@ -0,0 +1,183 @@
...
+pub const DEFAULT_ROLLUP_FINALITY: u64 = 3;
+pub const INIT_HEIGHT: u64 = 0;
+
+/// Consumes and commit the existing working set on the underlying storage
+/// `storage` must be the underlying storage defined on the working set for this method to work.
+pub(crate) fn commit_get_new_working_set(
+    storage: &ProverStorage<DefaultStorageSpec>,
+    working_set: WorkingSet<<C as Spec>::Storage>,
+) -> WorkingSet<<C as Spec>::Storage> {
+    let (reads_writes, witness) = working_set.checkpoint().freeze();
```
bkolad: Why do we need to commit inside a module? `AppTemplate` should take care of it. Now I think we should not expose the `checkpoint` method in `working_set`.

module-system/module-implementations/sov-attester-incentives/src/lib.rs
```diff
@@ -0,0 +1,219 @@
+pub mod call;
```
bkolad, line 5: #![deny(missing_docs)]

module-system/module-implementations/sov-attester-incentives/src/helpers.rs
bkolad: Ahh if this is for tests then't let's make `/test` folder and move it there.

module-system/module-implementations/sov-attester-incentives/src/lib.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/helpers.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,730 @@
...
+        &self,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> C::Address {
+        self.reward_token_supply_address
+            .get(working_set)
+            .expect("The reward token supply address should be set at genesis")
+    }
+
+    /// A helper function that simply slashes an attester and returns a reward value
+    fn slash_user_helper(
```
bkolad: let's remove `_helper`?

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,730 @@
...
+    pub(crate) fn unbond_challenger(
+        &self,
+        context: &C,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<sov_modules_api::CallResponse> {
+        // Get the user's old balance.
+        if let Some(old_balance) = self.bonded_challengers.get(context.sender(), working_set) {
+            // Transfer the bond amount from the sender to the module's address.
+            // On failure, no state is changed
+            self.reward_sender(old_balance, context, working_set)?;
```
bkolad: We follow the convention that `context` is always the first argument to a method (after `self`)

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,730 @@
...
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<sov_modules_api::CallResponse> {
+        // We have to ensure that the attester is unbonding, and that the unbonding transaction
+        // occurred at least `finality_period` blocks ago to let the attester unbond
+        if let Ok(begin_unbond_height) = self
+            .unbonding_attesters
+            .get_or_err(context.sender(), working_set)
+        {
+            // These two constants should always be set beforehand, hence we can panic if they're not set
+            let curr_height = self.light_client_finalized_height.get(working_set).unwrap();
```
bkolad: let's panic with `expect("...")`

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,730 @@
...
+        Ok(CallResponse::default())
+    }
+
+    /// The bonding proof is now a proof that an attester was bonded during the last `finality_period` range.
+    /// The proof must refer to a valid state of the rollup. The initial root hash must represent a state between
+    /// the bonding proof one and the current state.
+    fn check_bonding_proof(
+        &self,
+        attestation: &Attestation<StorageProof<<C::Storage as Storage>::Proof>>,
+        context: &C,
```
bkolad, line 390: let's put `context` just after self.

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,738 @@
...
+        if public_outputs.slot_hash != transition.da_block_hash() {
+            return Err(SlashingReason::TransitionInvalid);
+        }
+
+        // TODO: Should we compare the validity conditions of the public outputs with the ones of the recorded transition?
+        if condition_checker
+            .check(&public_outputs.validity_condition)
+            .is_err()
+        {
+            return Err(SlashingReason::TransitionInvalid);
```
bkolad, line 674: you can use `map_err(..)?` here

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,738 @@
...
+        reward
+    }
+
+    fn slash_burn_reward(
+        &self,
+        user: &C::Address,
+        role: Role,
+        reason: SlashingReason,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<sov_modules_api::CallResponse, AttesterIncentiveErrors> {
```
bkolad: This function always returns `Err` but the type is `Result`. This forces the caller to unnecessarily call `unwrap_err` sometimes (which he knows won't panic but is not obvious by reading the code). We should do the opposite return `AttesterIncentiveErrors` and wrap it in Err when necessary.

module-system/module-implementations/sov-attester-incentives/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    phantom_data: PhantomData<Cond>,
+}
+
+/// A wrapper around a code commitment which implements borsh serialization
+#[derive(Clone, Debug)]
+pub struct StoredCodeCommitment<Vm: Zkvm> {
+    commitment: Vm::CodeCommitment,
+}
+
+impl<Vm: Zkvm> BorshSerialize for StoredCodeCommitment<Vm> {
```
bkolad: let's move the serialisation logic to a separate file.

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,738 @@
...
+
+        // This proof checks that the attester was bonded at the given transition num
+        let (attester_key, bond_opt) = working_set
+            .backing()
+            .open_proof(bonding_root, attestation.proof_of_bond.proof.clone())
+            .map_err(|_err| AttesterIncentiveErrors::InvalidBondingProof)?;
+
+        // We have to check that the storage key is the same as the sender's
+        if attester_key
+            != StorageKey::new(
```
bkolad: The `StorageKey` is a low level concept and we should avoid using it  in the module code. Would it work if instead of `open_proof` we have `verify_proof` which does this check and the one below (for bond)? This way `StorageKey` and deserialisation logic won't leak to the module code.

bkolad (approved):
LGTM but it would be good if @preston-evans98 review it as well.

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/lib.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This module implements the logic for processing proof transactions. Such
+logic is necessary if you want to reward provers or do anything else that's "aware" of proof
+generation inside you state transition function.
+
+Currently, this module allows provers to register and de-register, and allows the on-chain validation
+of proofs from registered provers. If proof validation fails, the offending prover is slashed.
+
+This module does _not_ reward provers - incentives for provers will depend on gas metering, which has
+yet to be implemented.
```
preston-evans98: This doesn't look quite right :)

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+use sov_rollup_interface::zk::{
+    StateTransition, ValidityCondition, ValidityConditionChecker, Zkvm,
+};
+use sov_state::storage::StorageProof;
+use sov_state::{Storage, WorkingSet};
+use thiserror::Error;
+
+use crate::AttesterIncentives;
+
+/// This enumeration represents the available call messages for interacting with the `ExampleModule` module.
```
preston-evans98: nit: This comment doesn't look quite right

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+    /// No invalid transition to challenge.
+    NoInvalidTransition,
+
+    #[error("User is not trying to unbond at the time of the transaction")]
+    /// User is not trying to unbond at the time of the transaction
+    AttesterIsNotUnbonding,
+
+    #[error("First phase unbonding has not been finalized")]
+    /// The attester is trying to finish the two phase unbonding too soon
+    UnbondingNotFinalized,
```
preston-evans98: I don't know if this should be a slashable offense. Is there a reason we should slash here?

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+
+    #[error("Invalid bonding proof")]
+    /// The bonding proof was invalid
+    InvalidBondingProof,
+
+    #[error("The sender key doesn't match the attester key provided in the proof")]
+    /// The sender key doesn't match the attester key provided in the proof
+    InvalidSender,
+
+    #[error("The bond is not a 64 bits number")]
```
preston-evans98: nit: `bits` -> `bit`

module-system/module-implementations/sov-attester-incentives/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub reward_token_supply_address: C::Address,
+    /// The minimum bond for an attester.
+    pub minimum_attester_bond: u64,
+    /// The minimum bond for a challenger.
+    pub minimum_challenger_bond: u64,
+    /// A code commitment to be used for verifying proofs
+    pub commitment_to_allowed_challenge_method: Vm::CodeCommitment,
+    /// A list of initial provers and their bonded amount.
+    pub initial_attesters: Vec<(C::Address, u64)>,
+    /// The finality period of the rollup (constant)
```
preston-evans98: Can we add a note on the unit? Is it DA layer slots?

module-system/module-implementations/sov-attester-incentives/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub rollup_finality_period: u64,
+    /// The current maximum attested height
+    pub maximum_attested_height: u64,
```
preston-evans98: Is this ever not zero?

module-system/module-implementations/sov-attester-incentives/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub validity_condition_checker: Checker,
+    /// Phantom data that contains the validity condition
+    phantom_data: PhantomData<Cond>,
+}
+
+/// The information about an attester's unbonding
+#[derive(BorshDeserialize, BorshSerialize, Clone, Debug)]
+pub struct UnbondingInfo {
+    /// The height at which an attester is allowed to withdraw their tokens
+    pub unbonding_initiated_height: u64,
```
preston-evans98: The comment and the variable name don't match here. Is this the height they started unbonding or the height they finish unbonding?

module-system/module-implementations/sov-attester-incentives/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    #[state]
+    pub reward_token_supply_address: sov_state::StateValue<C::Address>,
+
+    /// The code commitment to be used for verifying proofs
+    #[state]
+    pub commitment_to_allowed_challenge_method: sov_state::StateValue<StoredCodeCommitment<Vm>>,
+
+    /// Constant validity condition checker for the module.
+    #[state]
+    pub validity_cond_checker: sov_state::StateValue<Checker>,
```
preston-evans98, line 107: It doesn't look like we ever use this checker. It's initialized, but not read AFAICT. What am I missing?

module-system/module-implementations/sov-attester-incentives/src/query.rs
```diff
@@ -0,0 +1,69 @@
...
+
+// TODO: implement rpc_gen macro
+impl<C, Vm, Cond, Checker> AttesterIncentives<C, Vm, Cond, Checker>
+where
+    C: sov_modules_api::Context,
+    Vm: Zkvm,
+    Cond: ValidityCondition,
+    Checker: ValidityConditionChecker<Cond> + BorshDeserialize + BorshSerialize,
+{
+    /// Queries the state of the module.
```
preston-evans98, line 24: We should probably add a TODO for making the "unbonding_amount" queryable as well.

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+    /// The bond is not a 64 bits number
```
preston-evans98: nit: `bits` -> `bit`

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+    /// Try to unbond the requested amount of coins with context.sender() as the beneficiary.
+    pub(crate) fn unbond_challenger(
+        &self,
+        context: &C,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<sov_modules_api::CallResponse> {
+        // Get the user's old balance.
+        if let Some(old_balance) = self.bonded_challengers.get(context.sender(), working_set) {
+            // Transfer the bond amount from the sender to the module's address.
+            // On failure, no state is changed
```
preston-evans98, line 293: I think this comment is backwards.  This transfers from the module to the sender, right?

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+            .get(context.sender(), working_set)
+            .is_some()
+        {
+            // Then add the bonded attester to the unbonding set, with the current finalized height
+            let finalized_height = self
+                .light_client_finalized_height
+                .get(working_set)
+                .expect("Must be set at genesis");
+            self.unbonding_attesters
+                .set(context.sender(), &finalized_height, working_set);
```
preston-evans98: ## Important
Don't we need to remove the attester from the bonded attesters set? Otherwise, they'll still be able to create valid-looking storage proofs that they're bonded during the unbonding period, which I think defeats the purpose of the 2-phase protocol

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+    pub(crate) fn end_unbond_attester(
+        &self,
+        context: &C,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<sov_modules_api::CallResponse, AttesterIncentiveErrors> {
+        // We first have to ensure that the attester is bonded
+        if self
+            .bonded_attesters
+            .get(context.sender(), working_set)
+            .is_some()
```
preston-evans98: If we do remove the attester at the beginning of unbonding, we'll need to change this logic too

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+                    .get(working_set)
+                    .expect("Should be defined at genesis");
+
+                if begin_unbond_height + finality_period > curr_height {
+                    return Err(self.slash_burn_reward(
+                        context.sender(),
+                        Role::Attester,
+                        SlashingReason::UnbondingNotFinalized,
+                        working_set,
+                    ));
```
preston-evans98: I don't know that we need to slash here. Can we just fail the transaction?

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+                }
+
+                // Get the user's old balance.
+                if let Some(old_balance) = self.bonded_attesters.get(context.sender(), working_set)
+                {
+                    // Transfer the bond amount from the sender to the module's address.
+                    // On failure, no state is changed
```
preston-evans98: I think this comment is backwards

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+                        &format!("amount_withdrawn: {old_balance:?}"),
+                    );
+                }
+            } else {
+                return Err(self.slash_burn_reward(
+                    context.sender(),
+                    Role::Attester,
+                    SlashingReason::AttesterIsNotUnbonding,
+                    working_set,
+                ));
```
preston-evans98: I also don't think we need to slash here. We can just revert

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,763 @@
...
+        let bonding_root = {
+            // If we cannot get the transition before the current one, it means that we are trying
+            // to get the genesis state root
+            if let Some(transition) = self
+                .chain_state
+                .historical_transitions
+                .get(&(attestation.proof_of_bond.transition_num - 1), working_set)
+            {
+                transition.post_state_root()
+            } else {
```
preston-evans98: I  think there's an edge case here where the attester gives an invalid `transition_num` that's in the future, which will cause us to use the genesis root instead of failing

module-system/module-implementations/sov-attester-incentives/src/lib.rs
preston-evans98: Ah, looks like the reference just didn't show up in search. Nevermind!

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Fixed with the last commit

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Fixed with the last commit

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+                .get(working_set)
+                .expect("Should be defined at genesis");
+
+            if unbonding_info.unbonding_initiated_height + finality_period > curr_height {
+                return Err(self.slash_burn_reward(
+                    context.sender(),
+                    Role::Attester,
+                    SlashingReason::UnbondingNotFinalized,
+                    working_set,
+                ));
```
preston-evans98: I don't think we need to slash here - we can just revert

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+                let amount = unbonding_info.amount;
+                &format!("amount_withdrawn: {:?}", amount)
+            });
+        } else {
+            return Err(self.slash_burn_reward(
+                context.sender(),
+                Role::Attester,
+                SlashingReason::AttesterIsNotUnbonding,
+                working_set,
+            ));
```
preston-evans98: Same here - I don't think that this needs to be slashed

rollup-interface/src/state_machine/optimistic.rs
```diff
@@ -0,0 +1,46 @@
...
+use crate::zk::StateTransition;
+
+/// A proof that the attester was bonded at the transition num `transition_num`.
+/// For rollups using the `jmt`, this will be a `jmt::SparseMerkleProof`
+#[derive(
+    Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize, Serialize, Deserialize, Default,
+)]
+pub struct ProofOfBond<StateProof: BorshSerialize> {
+    /// The transition number for which the proof of bond applies
+    pub transition_num: u64,
```
preston-evans98: Can't we recover the transition number from the `da_block_hash`?  Having the attester post it feels like an easy way to confuse trusted for untrusted data. If not, maybe we should rename it to `claimed_transition_num` so that we remember that it might be falsified.

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+                bonding_root,
+                attestation.proof_of_bond.proof.clone(),
+                context.sender(),
+                &self.bonded_attesters,
+            )
+            .map_err(|_err| AttesterIncentiveErrors::InvalidBondingProof)?;
+
+        let bond = bond_opt.unwrap_or_default();
+        let bond: u64 = BorshDeserialize::deserialize(&mut bond.value())
+            .map_err(|_err| AttesterIncentiveErrors::InvalidBondFormat)?;
```
preston-evans98: nit: This will cause a slightly confusing error message if `bond_opt` is `None`. The returned error will be `InvalidBondFormat` when it should be `UserNotBonded`. We can fix it by using roughly `bond_opt.map(BorshDeserialize).ok_or(Err(AttesterIncentiveErrors::UserNotBonded))??;`

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+                ));
+            }
+            Ok(CallResponse::default())
+        } else {
+            Err(self.slash_burn_reward(
+                attester,
+                Role::Attester,
+                SlashingReason::TransitionNotFound,
+                working_set,
+            ))
```
preston-evans98, line 470: ## Important
This looks like we're slashing the attester if they try to attest to a transition that's already been attested - am I reading that right? That doesn't seem like the expected behavior; I think those attestations should just be ignored, right?

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+
+        // Update the max_attested_height in case the blocks have already been finalized
+        let new_height_to_attest = max(last_attested_height, current_finalized_height) + 1;
+
+        // Minimum height at which the proof of bond can be valid
+        let min_height = if new_height_to_attest > finality {
+            new_height_to_attest - finality
+        } else {
+            0
+        };
```
preston-evans98: nit: We can just use `new_height_to_attest.saturating_sub(finality)`

module-system/module-implementations/sov-attester-incentives/src/call.rs
preston-evans98: Nevermind - it looks like this is handling the case where the attestation is false because the transition hasn't happened yet. So this seems fine! Let's add a comment to the else explaining what case it handles

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+        if let Some(curr_tx) = self
+            .chain_state
+            .historical_transitions
+            .get(&max_attested_height, working_set)
+        {
+            // We first need to compare the initial block hash to the previous post state root
+            if !curr_tx.compare_hashes(&attestation.da_block_hash, &attestation.post_state_root) {
+                // It is the right attestation, we have to compare the initial block hash to the
+                // previous post state root
+                // Slash the attester
```
preston-evans98: nit: Let's try to clarify these comments. How about:

```
// Check if the attestation has the same da_block_hash and post_state_root as the actual transition that we found in state. If not, slash the attester. 
// If so, the attestation is valid, so return Ok

File: module-system/module-implementations/sov-attester-incentives/src/call.rs [hunk 35]
```diff
@@ -0,0 +1,764 @@
...
+            .light_client_finalized_height
+            .get(working_set)
+            .expect("The light client finalized height should be set at genesis");
+        let finality = self
+            .rollup_finality_period
+            .get(working_set)
+            .expect("The rollup finality period should be set at genesis");
+
+        // Update the max_attested_height in case the blocks have already been finalized
+        let new_height_to_attest = max(last_attested_height, current_finalized_height) + 1;
```
preston-evans98: Can the `last_attested_height` ever be less than the `current_finalized_height`? If so, that seems like a cause for concern - light clients will have accepted a state transition which was never attested to!

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+        if self
+            .bonded_attesters
+            .get(context.sender(), working_set)
+            .is_none()
+        {
+            return Err(AttesterIncentiveErrors::UserNotBonded);
+        }
+
+        // We first need to check the bonding proof (because we can't slash the attester if he's not bonded)
+        self.check_bonding_proof(context, &attestation, working_set)?;
```
preston-evans98, line 556: At first glance, this check looks like it has already been handled by ` if self.bonded_attesters.get(context.sender(), working_set).is_none() { return Err(_)}`. On a closer look, I think there is an edge that we need to handle where the attester is bonded, but he provides an invalid proof of bond - is that what this check is intended to do? If so, we can just change the comment to say "If the bonding proof in the attestation is invalid, light clients will ignore the attestation. In that case, we should too. "

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+        {
+            if transition.post_state_root() != attestation.initial_state_root {
+                // The initial root hashes don't match, just slash the attester
+                return Err(self.slash_burn_reward(
+                    attester,
+                    Role::Attester,
+                    SlashingReason::InvalidInitialHash,
+                    working_set,
+                ));
+            }
```
preston-evans98: Since the param is `max_attested_height`, I think we end up slashing any attesters who attest to old transitions. But it's possible for multiple attestations to the same transition to land on-chain at the same time - so whoever gets ordered second by the DA layer will get slashed. 

I think we need to change this to work on the attestation's claimed `transition_num`, not on the `max_attested_height`. That way, we'll slash attesters who attest to an invalid root regardless of whether they're the first attestation for that transition, but we won't penalize people just for being the second to attest to a transition.

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+        max_attested_height: u64,
+        attester: &C::Address,
+        attestation: &Attestation<StorageProof<<C::Storage as Storage>::Proof>>,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<sov_modules_api::CallResponse, AttesterIncentiveErrors> {
+        // Normal state
+        if let Some(transition) = self
+            .chain_state
+            .historical_transitions
+            .get(&(max_attested_height - 1), working_set)
```
preston-evans98: Can this underflow? It looks like it to me - but let's use `saturating_sub` here in any case so that it's easier to tell at a glance.

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+                // The initial root hashes don't match, just slash the attester
+                return Err(self.slash_burn_reward(
+                    attester,
+                    Role::Attester,
+                    SlashingReason::InvalidInitialHash,
+                    working_set,
+                ));
+            }
+        } else {
+            // Genesis state
```
preston-evans98, line 497: Are we sure that this must be the genesis state? It seems safer to explicitly check that the transition number is 0. That way, we don't introduce a hidden invariant that the argument to this function must be less than the current block height. 

Alternatively, we can use a newtype to make the invariant explicit

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+
+        if public_outputs.initial_state_root != initial_hash {
+            return Err(SlashingReason::InvalidInitialHash);
+        }
+
+        if public_outputs.slot_hash != transition.da_block_hash() {
+            return Err(SlashingReason::TransitionInvalid);
+        }
+
+        // TODO: Should we compare the validity conditions of the public outputs with the ones of the recorded transition?
```
preston-evans98, line 677: Yes, we should!

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+            .set(&(new_height_to_attest), working_set);
+
+        // Reward the sender
+        self.reward_sender(
+            context,
+            self.minimum_attester_bond
+                .get(working_set)
+                .expect("Should be defined at genesis"),
+            working_set,
+        )?;
```
preston-evans98: I think we need to add a conditional to differentiate between the first attestation for a new transition and redundant attestations. 

If the sender is bonded and the attestation is in the right window (from `last_finalized_height` to `current_height`), we should *always* check the attestation and slash the attester if it lies about the transition. 
- When I say "Check the attestation", I mean checking that the da_block_hash and pre_state_root, and post_state_root are correct *for `proof.claimed_transition_num`*.

If the attestation is redundant, we should just stop after that check (we don't slash, but we also don't reward). 
Otherwise, the attestation is the first one for the given transition, so the sender should be rewarded. 

Does that sound right?

preston-evans98 (changes requested):
Overall, great work on this PR. This is a huge and complex feature and the code quality looks high. 

At a high level, there are basically just two things we need to change:
- Attesters shouldn't get slashed for redundant attestations
- The zkVM can't check validity conditions, so we'll need to think about how we handle those. 

For the short term, I think we can just leave all of the `validity_condition` stuff as a TODO - I'll pick that up after you leave. Let's go ahead and fix the slashing for redundant attestations if we can, and then merge this and make an issue for dealing with validity conditions. 


Update: I noticed one more thing to change. I think we want to make the attester bonding work on DA layer addresses rather than L1 addresses. That will let us avoid a bunch of signature checks that we otherwise would have to make. Let's also open an issue for that rather than trying to change it in this PR.

module-system/module-implementations/sov-attester-incentives/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// A list of initial provers and their bonded amount.
+    pub initial_attesters: Vec<(C::Address, u64)>,
+    /// The finality period of the rollup (constant)
+    pub rollup_finality_period: u64,
+    /// The current maximum attested height
+    pub maximum_attested_height: u64,
+    /// The light client finalized height
+    pub light_client_finalized_height: u64,
+    /// The validity condition checker used to check validity conditions
+    pub validity_condition_checker: Checker,
```
preston-evans98, line 56: In general, you can't run the validity_condition_checker inside of the zkvm. These checks usually rely on either a trusted light client of the DA layer, or a ton of very SNARK unfriendly operations - so we'll need to think about how to handle them a little differently. 

I think the way to do it will be to add the validity_conditions as an argument to `apply_slot` so that they can get saved as part of the historical_transitions in the chain state module. That way, we can just do an equality check to ensure that the validity conditions from a challenge match the expected ones - and the actual checking of the conditions will get pushed to the node implementations. But that's way too much to try to do in this PR, so let's just leave this for now and open an issue.

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -0,0 +1,764 @@
...
+        if bond < minimum_bond {
+            return Err(AttesterIncentiveErrors::UserNotBonded);
+        }
+
+        Ok(())
+    }
+
+    fn check_transition(
+        &self,
+        max_attested_height: u64,
```
preston-evans98: This `height` variable is used in a bunch of these checks, and it looks to me like it's hiding an invariant. Let's think about using a newtype to ensure safety here.

rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -43,10 +43,40 @@ pub trait Zkvm {
...
+        Ok(BorshDeserialize::deserialize_reader(&mut output)?)
+    }
+}
+
+/// A wrapper around a code commitment which implements borsh serialization
+#[derive(Clone, Debug)]
+pub struct StoredCodeCommitment<Vm: Zkvm> {
+    /// The inner field of the wrapper that contains the code commitment.
+    pub commitment: Vm::CodeCommitment,
+}
```
preston-evans98, line 64: Out-of-scope: we can get rid of this wrapper now that the JMT supports arbitrary serialization

module-system/module-implementations/sov-attester-incentives/src/lib.rs
theochap: Closing the conversation

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Yes but in that case, we fail the invariant test that comes after

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: So I have adopted the following rule to slash attesters/challengers: we first show that they are bonded. If that's not the case, then we can just revert and return an error. If that's the case, we slash the faulty users. I am going to remove this offense from the slashable set.

module-system/module-implementations/sov-attester-incentives/src/lib.rs
theochap: This should be almost always not zero as it tracks the maximum height that is currently attested.

module-system/module-implementations/sov-attester-incentives/src/query.rs
theochap: See: Make the unbonding amount queryable in the attester_incentives module #675

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Yes that's a typo

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Fixed

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Fixed

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Fixed

rollup-interface/src/state_machine/optimistic.rs
theochap: We need to have the transition number to recover the root hash from the chain state. We would be able to store the block hash and recover it from the chain state if needed, but I felt that the transition number is more straightforward. 
This is basically a hint that is used to verify the root hash claimed by the proof, we can definitely rename it `claimed_transition_num`

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: That was meant to deal with the case where the state gets updated without any attestation being posted during the finalizing period. But I think now I remember you saying that we have to revert in that case.

Is that the case? Then I will update the check to make sure that we only take the current_finalized_height

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Yes that's exactly what this is meant to do, I updated the code after your review but did not change the comment!

rollup-interface/src/state_machine/zk/mod.rs
theochap: Let's add a small issue for that

theochap: Remove StoredCodeCommitment wrapper #676

theochap:
> Overall, great work on this PR. This is a huge and complex feature and the code quality looks high.
> 
> At a high level, there are basically just two things we need to change:
> 
> * Attesters shouldn't get slashed for redundant attestations
> * The zkVM can't check validity conditions, so we'll need to think about how we handle those.
> 
> For the short term, I think we can just leave all of the `validity_condition` stuff as a TODO - I'll pick that up after you leave. Let's go ahead and fix the slashing for redundant attestations if we can, and then merge this and make an issue for dealing with validity conditions.
> 
> Update: I noticed one more thing to change. I think we want to make the attester bonding work on DA layer addresses rather than L1 addresses. That will let us avoid a bunch of signature checks that we otherwise would have to make. Let's also open an issue for that rather than trying to change it in this PR.

@preston-evans98 since the `validity_condition` stuff is rather big and you'll pick that up later, I suggest that you open the issue so that you can put the information you need in it

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Sounds right to me, and that's a small fix!

module-system/module-implementations/sov-attester-incentives/src/lib.rs
theochap: Makes a lot of sense. See my comment below for the issue

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Corrected with the last commit

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Added a newtype

module-system/module-implementations/sov-attester-incentives/src/lib.rs
preston-evans98: Sure, but this is the config, not the state value. It should never be greater than zero *at genesis*, right?

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -329,62 +345,46 @@ impl<
...
+                return Err(AttesterIncentiveErrors::UnbondingNotFinalized);
             }
+
+            // Get the user's old balance.
+            // Transfer the bond amount from the sender to the module's address.
+            // On failure, no state is changed
+            self.reward_sender(context, unbonding_info.amount, working_set)?;
+
+            // Update our internal tracking of the total bonded amount for the sender.
+            self.bonded_attesters.remove(context.sender(), working_set);
```
preston-evans98, line 370: I think this remove is redundant. We should have already removed the sender when they initiated the unbonding process

module-system/module-implementations/sov-attester-incentives/src/call.rs
```diff
@@ -550,11 +558,14 @@ impl<
             .expect("The rollup finality period should be set at genesis");
 
         // Update the max_attested_height in case the blocks have already been finalized
-        let new_height_to_attest = max(last_attested_height, current_finalized_height) + 1;
+        let new_height_to_attest =
+            TransitionHeight(max(last_attested_height, current_finalized_height).inner() + 1);
 
         // Minimum height at which the proof of bond can be valid
         let min_height = if new_height_to_attest > finality {
```
preston-evans98: nit: We don't need this `if`. Saturating sub covers both cases here

module-system/module-implementations/sov-chain-state/src/lib.rs
```diff
@@ -89,6 +94,17 @@ impl<Cond> TransitionInProgress<Cond> {
     }
 }
 
+/// The information about a transition height
+#[derive(BorshDeserialize, BorshSerialize, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TransitionHeight(pub u64);
+
+impl TransitionHeight {
+    /// Returns the inner content of [`TransitionHeight`]
```
preston-evans98: We're not actually getting any extra invariant safety from `TransitionHeight`. I think we need to change the definition to `pub struct TransitionHeight(u64);` and change the construction logic to make it impossible to create a `TransitionHeight` bigger than the `current_height` stored in the chain state module. 

I think the easiest way to do this is to have a fallible constructor which takes the `max_height` as an argument and fails if the requested height is too large. `pub fn new(requested_height: u64, max_height: u64) -> Result<Self>;`

preston-evans98 (approved):
Approving with a few nits. So that you won't be blocked on re-review.

module-system/module-implementations/sov-attester-incentives/src/call.rs
preston-evans98: In that case, I think we need to panic! The rollup got into a bad state and light clients won't be able to recover on their own

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Done

module-system/module-implementations/sov-attester-incentives/src/lib.rs
theochap: So this depends on the init height that is set in the chain set module. Since this is a parameter in the chain state, this should be also a parameter in this one to make it consistent

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Ok, so I will add a panic and remove the max

module-system/module-implementations/sov-attester-incentives/src/call.rs
theochap: Added a `genesis_height` to the chain state module so that we can check the transition against that height.

theochap: That prevents us from using an extra new type to make the invariant explicit.

module-system/module-implementations/sov-chain-state/src/lib.rs
theochap: We managed to get around that extra new type by adding an explicit check of the invariant. See above

PR #651, opened 2023-08-12
feat: add fuzzing

vlopes11:
# Description
This commit introduces a generative model for the accounts module
implementation that will generate valid messages to update the account
address.

It also fixes a bug where a prefix with an unaligned vector could be
created by declaring an API module, combining its prefixes, and
generating a vec would result in potentially unaligned vectors.


vlopes11:
Blocked by #666

adapters/celestia/src/shares.rs
```diff
@@ -279,6 +279,11 @@ impl std::error::Error for ShareParsingError {}
 impl NamespaceGroup {
     pub fn from_b64(b64: &str) -> Result<Self, ShareParsingError> {
         let mut decoded = Vec::with_capacity((b64.len() + 3) / 4 * 3);
+        if b64.is_empty() {
+            error!("Empty input");
+            return Err(ShareParsingError::ErrWrongLength);
+        }
+
```
neysofu, line 285: I'd put this before `let mut decoded`, it seems more natural to guard against the empty input before anything deserialization-related. No fix needed though unless you choose to, feel free not to address this and close the thread.

fuzz/Cargo.lock
neysofu, line 1: This file is in `.gitignore`, so I'm assuming it was committed by accident :smile:

module-system/sov-state/Cargo.toml
neysofu, line 1: It seems the `arbitrary` feature is missing here in `sov-state`.

module-system/sov-modules-macros/tests/genesis.rs
neysofu, line 1: Empty file, it should be removed.

README.md
```diff
@@ -106,6 +106,10 @@ to have adapters for almost all Data Availability Layers and LLVM-compatible pro
 maintain adapters for [`Risc0`](https://www.risczero.com) (a ZKVM) and [`Celestia`](https://www.celestia.org) a (DA layer).
 The Avail project also maintains an adapter for their DA layer, which can be found [here](https://github.com/availproject/avail-sovereign-da-adapter).
 
+## Testing
+
+An implementation of LLVM's libFUZZER is available under [fuzz/README.md](./fuzz/README.md).
```
neysofu, line 111: I feel like this would be better in a `CONTRIBUTING.md` or something of that sort once we'll have it, but we don't so it's fine in the main tutorial. However, the section should be more comprehensive and describe the rest of our testing infrastructure to give a complete overview rather than just a one-liner about fuzzing. So I'd either expand this or remove it, WDYT?

module-system/module-implementations/sov-accounts/src/lib.rs
```diff
@@ -75,3 +84,62 @@ impl<C: sov_modules_api::Context> sov_modules_api::Module for Accounts<C> {
...
+            accounts: sov_state::StateMap::new(prefix),
+        };
+
+        accounts
+            .genesis(&config, working_set)
+            .map_err(|_| arbitrary::Error::IncorrectFormat)?;
+
+        Ok(accounts)
+    }
+}
```
neysofu: `Arbitrary` for modules is super cool to see, it would simplify certain testing patterns –not just fuzzing– quite a lot.  🎉

We have to be especially careful around prefix conflicts for `State[Map|Value|Vec]`. In this case, the two maps `public_keys` and `accounts` use the same prefix which can cause key conflicts. I'd propose we always use `Default::default` when initializing modules inside `arbitrary_workset`, and only then we call `genesis` with arbitrary configuration.

module-system/sov-state/src/map.rs
```diff
@@ -115,3 +115,32 @@ where
         working_set.delete_value(self.prefix(), &self.codec, key);
     }
 }
+
+#[cfg(feature = "arbitrary")]
+impl<'a, K, V> StateMap<K, V>
+where
+    K: arbitrary::Arbitrary<'a> + borsh::BorshSerialize + borsh::BorshDeserialize,
+    V: arbitrary::Arbitrary<'a> + borsh::BorshSerialize + borsh::BorshDeserialize,
+{
```
neysofu: You'll have to update this to use `Hash + Eq` for keys when merging `nightly`.

```suggestion
impl<'a, K, V> StateMap<K, V>
where
    K: arbitrary::Arbitrary<'a> + Hash + Eq,
    V: arbitrary::Arbitrary<'a>,
    BorshCodec: StateValueCodec<V>
```

(Having this generic over all codecs and not just `BorshCodec` is likely not worth it and would just complicate things)

I haven't tested this change locally so I'm not sure these type bounds are 100% correct, but should be close.

module-system/module-implementations/sov-accounts/src/call.rs
```diff
@@ -72,3 +72,23 @@ impl<C: sov_modules_api::Context> Accounts<C> {
...
+        use sov_modules_api::PrivateKey;
+
+        let secret = C::PrivateKey::arbitrary(u)?;
+        let public = secret.pub_key();
+
+        let payload_len = u.arbitrary_len::<u8>()?;
+        let payload = u.bytes(payload_len)?;
+        let signature = secret.sign(payload);
+
+        Ok(Self::UpdatePublicKey(public, signature))
```
neysofu, line 92: How about generating some invalid signatures as well? We can choose based on an arbitrary `bool`.

neysofu (changes requested):
Great work, some of the fuzzing targets are very well done. We'll have to do more work to take advantage of this fuzzing infrastructure but this is an awesome start, thank you for picking it up.

I left some comments – mostly suggestions or small fixes. The only significant issue I've found is with `Accounts:: arbitrary_workset` which has conflicting prefixes for different state items.

adapters/celestia/src/shares.rs
vlopes11: Makes sense, done!

fuzz/Cargo.lock
vlopes11: It was actually intentional. The goal is to always run the fuzzing with the latest available libraries. Do you think its an ok idea? The reasoning is to always leave the fuzzing one step ahead of our actual release, but might be worthy to keep the versions in sync

module-system/sov-state/Cargo.toml
vlopes11: Since `arbitrary` is declared as `optional`, a feature is automatically created; otherwise, this dependency would fail:

https://github.com/Sovereign-Labs/sovereign-sdk/blob/337f40e4dc36d2867243df8cd3b007bd081ecad8/module-system/module-implementations/sov-accounts/Cargo.toml#L35

Do you think its best to be done explicitly? It would be smth like

```toml
[features]
arbitrary = ["dep:arbitrary"]
```

module-system/sov-modules-macros/tests/genesis.rs
vlopes11: Removed!

README.md
vlopes11: I tried to create a step-by-step description under `fuzz/README.md` to not pollute the main readme too much. Do you think it's ok to create this split?

Regarding the `CONTRIBUTING`; agreed, as this is more for developers of the library than users of the library (who will likely never fuzz)

fuzz/Cargo.lock
neysofu: Yeah I think it's fine to keep `fuzz/Cargo.lock`, I see no problem with it. But if so, then we should remove the entry from `.gitignore`

module-system/sov-state/Cargo.toml
neysofu: Oh wow, I had no idea about implicit features through optional dependencies. I thought one had to always specify those under `[features]`. We're good then!

module-system/module-implementations/sov-accounts/src/lib.rs
vlopes11: That makes sense! I did like that due to lack of knowledge of the code base; I was initially under the misconception that we need to share these prefixes across different maps, but now I know that they *must* be different.

module-system/module-implementations/sov-accounts/src/call.rs
vlopes11: I think we should create a different structure to wrap randomly correct signatures. Otherwise, we can't consume the generated structs from here with confidence, and that would not allow us to expect certain behavior from arbitrary signatures. Wdyt?

module-system/sov-state/src/map.rs
vlopes11: Indeed the change was required with a slight difference from the suggested snippet. Updated!

fuzz/Cargo.lock
vlopes11: I meant the opposite: if we don't gitignore it, it will stall to outdated versions as users would have to explicitly enter the directory and run a `cargo update`.

module-system/module-implementations/sov-accounts/src/call.rs
neysofu: `CallMessage` has no invariant that the contents of the `struct` are correct, as it represents untrusted user inputs which is part of what we want to test against. I don't think we are supposed to consume the generated struct with confidence because validating against user input is part of the logic to test.

That said, we can have two different functions here, one with always valid inputs and the other which sometimes will generate invalid ones.

README.md
neysofu: Yes, the split is very much okay. Nvm, I'll resolve this :)

module-system/module-implementations/sov-accounts/src/call.rs
neysofu: Tracked here to unblock this PR: https://github.com/Sovereign-Labs/sovereign-sdk/issues/747 🎉

neysofu (approved):

PR #667, opened 2023-08-17
Bump dalek-ed25519 to 2.0.0

citizen-stig:
# Description

Upgrades dalek-ed25519 to 2.0.0, which has more suitable feature set for our use cases.

It also includes some changes in sov-cli and removes temporary used `HexKey` struct from `demo-rollup`

## Testing
Tests have been added or updated

## Docs
Documentation have been updated


module-system/sov-modules-api/Cargo.toml
```diff
@@ -28,20 +28,20 @@ hex = { workspace = true, optional = true }
 clap = { workspace = true, optional = true }
 schemars = { workspace = true, optional = true, features = [] }
 
-# TODO: try update 2.x
-ed25519-dalek = { version = "1.0.1", default-features = false, features = ["u64_backend"] }
-rand = { version = "0.7", optional = true }
+ed25519-dalek = { version = "2.0.0", default-features = false }
```
neysofu, line 33: I noticed the `fast` feature while glancing at the 2.0 manifest. Could it be of interest to us?

https://github.com/dalek-cryptography/ed25519-dalek/blob/02001d8c3422fb0314b541fdb09d04760f7ab4ba/Cargo.toml#L65

module-system/sov-modules-api/src/default_signature.rs
```diff
@@ -2,70 +2,88 @@
...
+        KeyPairError(#[from] ed25519_dalek::SignatureError),
+        #[error("Invalid private key length: {actual}, expected {expected_1} or {expected_2}")]
+        InvalidPrivateKeyLength {
+            expected_1: usize,
+            expected_2: usize,
+            actual: usize,
+        },
+        // This error is thrown only if private methods are misused.
+        #[error("Potential bug: failed to convert from slice")]
+        SliceError(#[from] std::array::TryFromSliceError),
```
neysofu: I think these should be a panic via `expect` instead of returning an error, what do you think? If this ever happens, it means a serious invariant was broken in our code and we must work on a patch.

neysofu (approved):
LGTM! I left two minor comments.

module-system/sov-modules-api/Cargo.toml
citizen-stig: Interesting! @dubbelosix have you tried this one? I am not sure if it is going to be faster in ZK mode, but maybe in native?

CC @preston-evans98

module-system/sov-modules-api/src/default_signature.rs
citizen-stig: Agree!

neysofu: Nice, I guess we can also remove the invariant altogether :100:

module-system/sov-modules-api/Cargo.toml
neysofu: I'll turn this into an issue, so we can unblock this PR.

module-system/sov-modules-api/src/default_signature.rs
citizen-stig: Done!

module-system/sov-modules-api/src/default_signature.rs
```diff
@@ -3,70 +3,87 @@ use std::hash::Hash;
...
-        #[error("PrivateKey deserialization error")]
-        PrivateKeyError(#[from] SignatureError),
+        #[error("KeyPairError deserialization error")]
+        KeyPairError(#[from] ed25519_dalek::SignatureError),
+        #[error("Invalid private key length: {actual}, expected {expected_1} or {expected_2}")]
+        InvalidPrivateKeyLength {
+            expected_1: usize,
+            expected_2: usize,
+            actual: usize,
+        },
```
neysofu, line 32: I think this is not needed anymore, by the way

examples/demo-rollup/src/lib.rs
```diff
@@ -43,21 +36,19 @@ pub struct HexKey {
...
-    let sequencer_private_key = DefaultPrivateKey::from_hex(&hex_key.hex_priv_key).unwrap();
-    assert_eq!(
-        sequencer_private_key.default_address().to_string(),
-        hex_key.address,
-        "Inconsistent key data",
-    );
+    let key_and_address: PrivateKeyAndAddress<DefaultContext> = hex_key
+        .try_into()
+        .expect("Failed to parse sequencer private key and address");
+    assert!(key_and_address.is_matching(), "Inconsistent key data");
```
preston-evans98: Why do we deserialize an address in the genesis config if we aren't willing to use it to override the "default" one that would be generated from the private key? Why not just use a serialized private key with no address?

module-system/sov-cli/src/wallet_state.rs
```diff
@@ -52,6 +53,72 @@ impl<Tx: Serialize + DeserializeOwned, Ctx: sov_modules_api::Context> WalletStat
...
+/// where the private key is represented as a hex string
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct HexPrivateAndAddress {
+    /// Private key is hex encoded bytes, without leading 0x
+    pub hex_priv_key: String,
+    /// Address is in canonical string format
+    pub address: String,
+}
+
+impl<Ctx: sov_modules_api::Context> TryFrom<HexPrivateAndAddress> for PrivateKeyAndAddress<Ctx> {
```
preston-evans98, line 103: Should we add a conversion going the other way (`From<PrivateKeyAndAddress<Ctx>> for HexPrivateAndAddress`)?

citizen-stig: Yes, that would be super simple. I didn't add it, because there was no use case. But those types are interchangable.

examples/demo-rollup/src/lib.rs
citizen-stig: We can do that, but if there's incorrect address in the "key" file. It can bring a confusion, if somebody wants to use those files.

I've been in this situation. I wanted to play with sending transaction one to another, but in one file address was not updated, so it took me a while to identify that problem is there. 

This extra check makes sure, that our key files are always correct.

module-system/sov-cli/src/wallet_state.rs
citizen-stig: So this actually wouldn't work without updating `PrivateKey` trait. Currently it does not have `ToBytes` or `AsRef<[u8]>` or something that will allow to get bytes and encode them.

module-system/sov-cli/src/wallet_state.rs
```diff
@@ -52,6 +53,72 @@ impl<Tx: Serialize + DeserializeOwned, Ctx: sov_modules_api::Context> WalletStat
...
+pub struct PrivateKeyAndAddress<Ctx: sov_modules_api::Context> {
+    /// Private key of the address
+    pub private_key: Ctx::PrivateKey,
+    /// Address associated from the private key
+    pub address: Ctx::Address,
+}
+
+impl<Ctx: sov_modules_api::Context> PartialEq for PrivateKeyAndAddress<Ctx> {
+    fn eq(&self, other: &Self) -> bool {
+        self.private_key.pub_key() == other.private_key.pub_key() && self.address == other.address
```
vlopes11: This is not very obvious. Usually, private keys can have different representation, especially if with projective points, and secret keys have canonical representation. Here we go the other way around: we generate a pub key and compare equality.

Is the reason for that a limination on `dalek-ed25519`? If so, can we add a comment here explaining that? Or other comment explaining the reasoning for this

module-system/sov-cli/src/wallet_state.rs
```diff
@@ -52,6 +53,72 @@ impl<Tx: Serialize + DeserializeOwned, Ctx: sov_modules_api::Context> WalletStat
...
+        let address = private_key.to_address::<Ctx::Address>();
+        Self {
+            private_key,
+            address,
+        }
+    }
+}
+
+/// A simplified struct representing private key and associated address
+/// where the private key is represented as a hex string
```
vlopes11: nit: both are represented as string

module-system/sov-modules-api/src/default_signature.rs
```diff
@@ -3,70 +3,87 @@ use std::hash::Hash;
...
         }
-    }
 
-    impl serde::Serialize for DefaultPrivateKey {
-        fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-            serializer.serialize_bytes(&self.key_pair.to_bytes())
+        fn try_from_private_key(value: &[u8]) -> Self {
+            let value: [u8; SECRET_KEY_LENGTH] = value
+                .try_into()
+                .expect("incorrect usage of `try_from_private_key`, check input length");
```
vlopes11, line 56: I'm not sure its safe to panic here as user input might lead to inconsistent states. If its safe, can we add a comment on why is that?

vlopes11 (commented):
Looks good overall! A couple of small questions & nits

module-system/sov-modules-api/src/default_signature.rs
citizen-stig: Idea behind this panic is that it is private method only used by `try_from`, which currently checks length of user input. So if this panics, means that there's a bug in our code. 

But comment explaining this is definitely must have

module-system/sov-cli/src/wallet_state.rs
citizen-stig: I am not sure, that we ever had comparing private keys. To be honest, I've added this trait for testing purposes, so maybe it will be safer just to remove it and do explicit comparison in test!

module-system/sov-cli/src/wallet_state.rs
citizen-stig: Plan is to remove HexPrivateKey: https://github.com/Sovereign-Labs/sovereign-sdk/issues/766

module-system/sov-cli/src/wallet_state.rs
citizen-stig: Removed this implementation

PR #673, opened 2023-08-17
Optimistic module integration

theochap:
# Description
This PR aims to add an end-to-end integration workflow for the optimistic module. The main changes are concentrated in `SlotData` and the `ChainState` module. The slot data is now equipped with a namespace method that returns the slot's namespace which is used to update the chain-state module and skip the transactions that did not come from the default namespace. The `Namespace` is also integrated in the context as an associated type. The rest of the workflow is left unchanged.

## Testing
An end-to-end integration test is going to be added.


cemozerr:
Can we close this PR now @preston-evans98?

PR #680, opened 2023-08-18
Rollup strcture in demo-rollup

bkolad:
# Description
This PR introduces the `Rollup` structure in `demo-rollup` containing dependencies needed to run the rollup. 
The idea is that the creation of the `Rollup` is tailored to the DA layer and external configuration, while the execution part is independent on the DA.

## Linked Issues
- Related to ##585

## Testing
Tested with local CI.


examples/demo-rollup/src/rollup.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    Ok(Rollup {
+        app,
+        da_service,
+        ledger_db,
+        runner_config: rollup_config.runner,
+    })
+}
+
+impl<Vm: Zkvm, DA: DaService<Error = anyhow::Error>> Rollup<Vm, DA> {
```
neysofu: How about we put the `Clone` bound here instead of the trait? `DA: DaService<Error = anyhow::Error> + Clone`

examples/demo-rollup/src/rollup.rs
```diff
@@ -0,0 +1,94 @@
...
+    let rollup_config: RollupConfig<celestia::DaServiceConfig> =
+        from_toml_path(rollup_config_path).context("Failed to read rollup configuration")?;
+
+    // Initializing logging
+    let subscriber = tracing_subscriber::fmt()
+        .with_max_level(Level::INFO)
+        .finish();
+    tracing::subscriber::set_global_default(subscriber)
+        .map_err(|_err| eprintln!("Unable to set global default subscriber"))
+        .expect("Cannot fail to set subscriber");
```
neysofu: This should probably be inside a `init_logging` function or similar, and called directly from `main` since it modifies global logging settings.

neysofu (approved):
Really nice refactor! Just two suggestions below.

examples/demo-rollup/src/rollup.rs
bkolad: done

examples/demo-rollup/src/rollup.rs
bkolad: done

Issue #33, opened 2023-02-07
sov-modules: Add `StateVec` type to storage


bkolad:
Currently, the module system supports maps only. We should add `StateVec` type that stores a `Vec`. We might need a separate commitment scheme, but there may be a way to avoid it.

PR #685, opened 2023-08-21
Implement State Vectors

eyusufatik:
# Description
This PR adds Vector functionality to `state-vec` module.

`StateVec` is implemented as follows:
Uses key `0` to store current length of the vector on the underlying storage.
Uses `index + 1` to store values on the underlying storage.
Random get and sets are supported.
Since underlying storage is a map based structure, arbitrary removes/inserts are not supported as shifting the vector would be too costly.

`StateVec` can do 5 operations:
- Get
- Set
- Push
- Pop
- Set all

## Linked Issues
- Related to [`sov-state/README.md`](https://github.com/Sovereign-Labs/sovereign-sdk/blob/fd2ff3949ec5d18f65b719b01a07b54e3e706518/module-system/sov-state/README.md?plain=1#L12)

## Testing
Added unit tests to `module-system/sov-state/src/state_tests.rs`
Implemented example under `module-system/module-implementations/examples/sov-vec-setter`. Tests will be added.

## Docs
Documentation will be added in `module-system/sov-state/README.md`


neysofu (approved):
This is greatly appreciated, thank you! It's been on our TODO list for a while: https://github.com/Sovereign-Labs/sovereign-sdk/issues/33.

There's a few changes I'd make, but we can address those in a separate PR: https://github.com/Sovereign-Labs/sovereign-sdk/pull/688. In particular, I'd like us to manage internal state of the `StateVec` re-using `StateValue` and `StateMap` to simplify this logic (and avoid index arithmetic).

Let's also wait for @bkolad 's feedback –and anybody else who wants to take a look– before merging.

neysofu:
The `set_all` behavior of not removing elements beyond the new `len()` is dubious IMO. To be fair, one would never realize if they stay within the API surface of `StateVec`. I think I'm okay with it, but interested in others' feedback.

module-system/sov-state/src/vec.rs
```diff
@@ -0,0 +1,226 @@
...
+                working_set.remove_value(self.prefix(), &self.internal_codec(), &IndexKey(len));
+            self.set_len(len - 1, working_set);
+            elem
+        } else {
+            None
+        }
+    }
+
+    /// Sets all values in the [`StateVec`].
+    /// If the length of the provided values is less than the length of the [`StateVec`], the remaining values stay in storage but are inaccessible.
```
preston-evans98: > /// If the length of the provided values is less than the length of the [`StateVec`], the remaining values stay in storage but are inaccessible.

This is a bit of an odd feature - in general, you really don't want the state trie getting filled up with inaccessible values. What's the motivating use case for this behavior?

preston-evans98:
Agree with @neysofu regarding the behavior of `set_all` - we're happy to consider implementing that API if there's a good reason for it, but we also want to be very careful not to introduce footguns 😅. 

Other than that, this change (+ #688) looks great to me. Thanks for the PR @eyusufatik!

module-system/sov-state/src/vec.rs
neysofu: Doesn't performance benefit from this pattern? This way you're not paying the price of removing all that data.

neysofu:
Hey @eyusufatik, are you okay with us merging the PR or do you plan on adding anything else? If not, I'll go ahead and merge, and #688 after that.

module-system/sov-state/src/vec.rs
preston-evans98: Performance does benefit for the person clearing the map, but at the expense of everyone else's performance forever. Those stale values still have to be stored by all full nodes, and they fill up the JMT making merkle paths longer 😬

eyusufatik:
> Hey @eyusufatik, are you okay with us merging the PR or do you plan on adding anything else? If not, I'll go ahead and merge, and #688 after that.

yes, we can merge this and the follow up.

I'll create an issue for `set_all` storage vs performance trade off discussion.

neysofu:
@eyusufatik No need, I just pushed two commits to:
1. Add tests to both `StateVec` and `sov-vec-setter`, and
2. Change the behavior or `set_all` to discard all old elements.

Feel free to take a look and let me know what you think!

eyusufatik:
@neysofu didn't realize you pushed changes before I sent the comment.

everything looks fine however inside `clear` you delete values from 0 to len - 1 (inclusive)

assume the vector holds 5, 7, 9

then on the underlying storage we have 
index:     0 1 2 3
values:   3 5 7 9

for 0..len will delete indices 0 1 and 2

so it should be 0..len+1

neysofu:
Nice catch! Fixed it, and added a test condition for that. Thank you. I  think we're good to merge once CI passes :smile:

module-system/sov-state/src/vec.rs
neysofu: Fixed, we now clean up after `set_all`.

PR #686, opened 2023-08-21
DA mock implementation

bkolad:
# Description
1. `MockDaService` implementation based on tokio channels. The current implementation ignores height in `get_finalized_at` and blobs are always processed in the order they are sent. This can be changed in the future.
2. Removes `new` function from ` DaService`. During the the DA creation we know the exact type of the `DaService` so putting `new` in the `DaService` puts unnecessary constraints on the implementers.  
3. Moves EVM integration tests for `sov-evm` to `demo-rollup` and runs the test with the `MockDaService`

## Testing
CI passes.

## Docs
No doc changes.


cemozerr:
Awesome. Does this PR remove the need to run the Celestia docker for the demo-rollup?

preston-evans98 (approved):
Looks like a great change. I hadn't thought about removing `new` from `DaService`, but it makes sense in retrospect! Nice!

PR #687, opened 2023-08-21
fix: typos

omahs:
fix: typos

### DIFF
--- a/examples/demo-rollup/README.md
+++ b/examples/demo-rollup/README.md
@@ -166,7 +166,7 @@ use sov_bank::Amount;
 pub enum CallMessage<C: sov_modules_api::Context> {
     /// Creates a new token with the specified name and initial balance.
     CreateToken {
-        /// Random value use to create a unique token address.
+        /// Random value used to create a unique token address.
         salt: u64,
         /// The name of the new token.
         token_name: String,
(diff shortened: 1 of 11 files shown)


examples/demo-rollup/README.md
```diff
@@ -208,7 +208,7 @@ pub enum CallMessage<C: sov_modules_api::Context> {
 }
 ```
 
-In the above snippet, we can see that `CallMessage` in `Bank` support five different types of calls. The `sov-cli` has the ability to parse a JSON file that aligns with any of these calls and subsequently serialize them. The structure of the JSON file, which represents the call, closely mirrors that of the Enum member. Consider the `Transfer` message as an example:
+In the above snippet, we can see that `CallMessage` in `Bank` supports five different types of calls. The `sov-cli` has the ability to parse a JSON file that aligns with any of these calls and subsequently serialize them. The structure of the JSON file, which represents the call, closely mirrors that of the Enum member. Consider the `Transfer` message as an example:
```
neysofu, line 211: The original was correct in this case, let's revert this.

examples/demo-rollup/benches/README.md
```diff
@@ -71,4 +71,4 @@ The Makefile is located in the demo-rollup/benches folder and supports the follo
 
 The Makefile supports setting number of blocks and transactions per block using BLOCKS and TXNS_PER_BLOCK env vars. Defaults are 100 blocks and 10,000 transactions per block when using the Makefile
 
-![Flamgraph](flamegraph_sample.svg)
\ No newline at end of file
+![Flamgraph](flamegraph_sample.svg)
```
neysofu: While we're at it, let's fix this one as well:

```suggestion
![Flamegraph](flamegraph_sample.svg)
```

neysofu (changes requested):
Thanks a bunch for all these fixes! Greatly appreciated.

I provided ~two suggestions~ one suggestion below :smile:

examples/demo-rollup/README.md
neysofu: Nevermind, I got confused while reviewing :P

neysofu (approved):

bkolad (approved):

PR #688, opened 2023-08-21
`StateVec` PR followup

### LINKED ISSUE #33: sov-modules: Add `StateVec` type to storage

neysofu:
# Description
This is a followup PR to https://github.com/Sovereign-Labs/sovereign-sdk/pull/685, making some small changes:

- Instead of storing both the vector's length and elements inside a single key space, this PR splits all state management of `StateVec` into a `StateValue` (for the length) and a `StateMap` (for vector elements). Performance impact should be negligible and we can always optimize things later. The resulting code is simpler and less bug-prone.
- Codec-related logic is simpler as we rely on `StateValue`/`StateMap` APIs.
- Iterator support. It's not yet a `DoubleEndedIterator`, but it can be supported if necessary.

## Linked Issues
Closes https://github.com/Sovereign-Labs/sovereign-sdk/issues/33.

cc @eyusufatik.

### DIFF
--- a/module-system/sov-state/src/vec.rs
+++ b/module-system/sov-state/src/vec.rs
@@ -1,16 +1,22 @@
+use std::iter::FusedIterator;
 use std::marker::PhantomData;
 
-use borsh::{BorshDeserialize, BorshSerialize};
 use thiserror::Error;
 
-use crate::codec::{BorshCodec, StateKeyCodec, StateValueCodec};
-use crate::{Prefix, Storage, WorkingSet};
+use crate::codec::{BorshCodec, PairOfCodecs, StateValueCodec};
+use crate::{Prefix, StateMap, StateValue, Storage, WorkingSet};
 
-#[derive(Debug, PartialEq, Eq, Clone, BorshDeserialize, BorshSerialize)]
-pub struct StateVec<V, C = BorshCodec> {
+type InternalCodec<C> = PairOfCodecs<BorshCodec, C>;
+
+#[derive(Debug, Clone)]
+pub struct StateVec<V, VC = BorshCodec>
+where
+    VC: StateValueCodec<V>,
+{
     _phantom: PhantomData<V>,
-    codec: C,
     prefix: Prefix,
+    len_value: StateValue<usize>,
+    elems: StateMap<usize, V, InternalCodec<VC>>,
 }
 
 /// Error type for `StateVec` get method.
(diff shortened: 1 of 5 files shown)


bkolad (approved):

module-system/sov-state/src/vec.rs
```diff
@@ -0,0 +1,218 @@
...
+
+#[derive(Debug, Clone)]
+pub struct StateVec<V, C = BorshCodec>
+where
+    C: StateValueCodec<V>,
+{
+    _phantom: PhantomData<V>,
+    prefix: Prefix,
+    len_value: StateValue<usize>,
+    elems: StateMap<usize, V, InternalCodec<C>>,
```
preston-evans98: (non-blocking) It seems possible that some use cases will also want control over the key codec. Will it be trivial to switch from a `StateValueCodec` to a `StateCodec` if that turns out to be the case?

module-system/sov-state/src/vec.rs
```diff
@@ -0,0 +1,218 @@
...
+        Self::with_codec(prefix, BorshCodec)
+    }
+}
+
+impl<V, C> StateVec<V, C>
+where
+    C: StateValueCodec<V>,
+{
+    /// Creates a new [`StateVec`] with the given prefix and codec.
+    pub fn with_codec(prefix: Prefix, codec: C) -> Self {
```
preston-evans98: It's a little surprising to me that we pass a codec *instance* here. Is there a reason for that?

preston-evans98 (approved):
Left a few minor comments, but looks good overall. Thanks for the quick followup @neysofu!

module-system/sov-state/src/vec.rs
neysofu: It would be trivial, yes. We'd swap the generic, change some type bounds and `InternalCodec`.

module-system/sov-state/src/vec.rs
neysofu: I'm not sure what the alternative would be 🤔  You mean a codec instance as opposed to only relying on the codec type `C` and storing `PhantomData<C>` internally? Codecs require `self` to encode and decode values, as we assume they may depend on some internal state or settings e.g. endianess, compression. It just so happens that most don't.

PR #689, opened 2023-08-21
Encoding-preserving `StateMap` borrowee keys

neysofu:
# Description

This PR updates #644 with encoding-preserving borrows (which fixes the issues noted in the PR discussion) and fixes all merge conflicts with `nightly` (especially #550), so all merge conflicts between this branch and `vlopes11/feature/state-map-borrow` should get fixed automatically by choosing this branch's files.


module-system/sov-state/src/map.rs
```diff
@@ -1,17 +1,26 @@
...
+/// # Type parameters
+/// [`StateMap`] is generic over:
+/// - a key type (`K`);
+/// - a value type (`V`);
+/// - a [`StateCodec`] (`C`).
 #[derive(borsh::BorshDeserialize, borsh::BorshSerialize, Debug, PartialEq, Clone)]
-pub struct StateMap<K, V> {
+pub struct StateMap<K, V, C = BorshCodec>
+where
+    C: StateCodec<K, V>,
```
vlopes11, line 20: Small, optional nit: we use `C` elsewhere, usually for `Context`. Wdyt of using a different letter for `StateCodec` so we preserve a quickly understandable distinction? Could be `K`, `D`, or any other

module-system/sov-state/src/codec.rs
```diff
@@ -0,0 +1,200 @@
+//! Serialization and deserialization -related logic.
+
+use std::borrow::Borrow;
+use std::rc::Rc;
+use std::sync::Arc;
+
+pub trait StateKeyEncode<K: ?Sized> {
```
vlopes11, line 7: Question: should we use a `type` trait attribute instead? If no, what would be the reasoning to go attribute generic?

module-system/sov-state/src/codec.rs
```diff
@@ -154,3 +185,11 @@ where
         self.value_codec.encode_value(value)
     }
 }
+
+impl<K, Q, KC, VC> StateKeyEncodePreservingBorrow<K, Q> for PairOfCodecs<KC, VC>
+where
+    Q: ?Sized,
+    K: Borrow<Q>,
+    KC: StateKeyEncodePreservingBorrow<K, Q>,
```
bkolad, line 198: @neysofu can `StateKeyEncodePreservingBorrow` and `Borrow` be merged into single trait?

module-system/sov-state/src/codec.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// [`Result`] with either the deserialized key or an error.
+    fn try_decode_key(&self, bytes: &[u8]) -> Result<K, Self::KeyError>;
+
+    /// Deserializes a key from a bytes slice.
+    ///
+    /// # Panics
+    /// Panics if the call to [`StateKeyCodec::try_decode_key`] fails. Use
+    /// [`StateKeyCodec::try_decode_key`] if you need to gracefully handle
+    /// errors.
+    fn decode_key(&self, bytes: &[u8]) -> K {
```
vlopes11, line 31: I'm a bit unsure why would we `unwrap` on arbitrary input as this can likely panic. In case the user is sure the input is valid, he can just unwrap the `try_decode_key`

module-system/sov-state/src/codec.rs
```diff
@@ -0,0 +1,200 @@
...
+                    "Failed to decode key 0x{}, error: {:?}",
+                    hex::encode(bytes),
+                    err
+                )
+            })
+            .unwrap()
+    }
+}
+
+pub trait StateKeyEncodePreservingBorrow<Borrower, Borrowed>
```
vlopes11, line 44: Optinal nit: wdyt of a shorter name? Like `StateKeyEncode` as the `PreservingBorrow` is explicit under the `where` clause

module-system/sov-state/src/codec.rs
```diff
@@ -0,0 +1,200 @@
...
+where
+    Borrowed: ?Sized,
+    Borrower: Borrow<Borrowed>,
+    Self: StateKeyEncode<Borrower> + StateKeyEncode<Borrowed>,
```
vlopes11, line 48: Question: wouldn't the user be able to perform a different encoding strategy depending on the borrowed type? Meaning: I think we can still `Borrow<[u8]>` from both `Vec<u8>` and `[u8; 32]`, but still have different serialization strategy.

To prevent that, maybe the `Borrowed` and `Borrower` would have to be the same and `StateKeyEncode` would have to be for one type only.

module-system/sov-state/src/codec.rs
```diff
@@ -0,0 +1,200 @@
...
+/// A [`StateCodec`] that uses [`borsh`] for all keys and values.
+#[derive(Debug, Default, PartialEq, Eq, Clone, borsh::BorshDeserialize, borsh::BorshSerialize)]
+pub struct BorshCodec;
+
+impl<K> StateKeyEncode<K> for BorshCodec
+where
+    K: ?Sized + borsh::BorshSerialize,
+{
+    fn encode_key(&self, key: &K) -> Vec<u8> {
+        key.try_to_vec().expect("Failed to serialize key")
```
vlopes11, line 105: Should we open an issue to make a trait for infallible borsh serialization in the future? While I agree serialization in general should be infallible (except for constrained environments where memory is limited), an implementer of borsh might fail his serialization and we would panic here

vlopes11 (changes requested):
Looks good overall! Before we merge, let's double check a couple of items

module-system/sov-state/src/codec.rs
neysofu: Are you referring to associated types? If so, that wouldn't work as we need `StateKeyEncode` to be generic over many `K`: a single key codec can serialize several key types. This is precisely what allows for borrowed keys (multiple equivalent key types interchangeable with each other). It *might* if we make extensive modifications to the other traits... although I'm not seeing a huge benefit here.

module-system/sov-state/src/codec.rs
neysofu: Agreed – this is the same thing, but with a more helpful message than a simple `.unwrap()`.

module-system/sov-state/src/codec.rs
neysofu: Interesting – if there is a way to make this even more type safe, I'm all for it. This is the best I could manage and yes, it's up to the user to make sure that `StateKeyEncodePreservingBorrow` is only ever implemented for the right types. What pattern did you have in mind specifically here?

module-system/sov-state/src/codec.rs
neysofu: I don't think we can, unfortunately. I remember trying. But there is a valid alternative: drop altogether all references to `Borrow` and decouple `StateKeyEncodePreservingBorrow` from the borrowing concept entirely. We'd end up with something akin to `StateKeyEquivalence`. Less type bounds as well (`K: Borrow<Q>` would get dropped), but the API is not as close to `HashMap`. @vlopes11 WDYT? I think I'd prefer that.

module-system/sov-state/src/map.rs
neysofu: Agreed. I'll make an issue for this, as changing it right now would cause many conflicts!

module-system/sov-state/src/codec.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+}
+
+impl<K> StateKeyEncodePreservingBorrow<K, K> for BorshCodec where K: borsh::BorshSerialize {}
+impl<K> StateKeyEncodePreservingBorrow<Arc<K>, K> for BorshCodec where K: borsh::BorshSerialize {}
+impl<K> StateKeyEncodePreservingBorrow<Box<K>, K> for BorshCodec where K: borsh::BorshSerialize {}
+impl<K> StateKeyEncodePreservingBorrow<Rc<K>, K> for BorshCodec where K: borsh::BorshSerialize {}
+// Note: we're specifically *not* implementing it for `<[u8; _], [u8]>` because
+// arrays and slices have different representations in Borsh.
+impl<K> StateKeyEncodePreservingBorrow<Vec<K>, [K]> for BorshCodec where K: borsh::BorshSerialize {}
```
bkolad, line 141: Can this panic after upgrade to new borsh: https://github.com/near/borsh-rs/blob/master/borsh/src/ser/mod.rs#L275 ?

module-system/sov-state/src/codec.rs
vlopes11: But I think the main problem is for an incautious user reading the docs that will display only the function signature - this would lead him to believe the function is infallible, if he doesn't read the implementation as well.

I don't have a strong case here, but I'd suggest to make the signature reflect the implementation in entirety.

neysofu:
Superseded by https://github.com/Sovereign-Labs/sovereign-sdk/pull/718, closing.

PR #708, opened 2023-08-21
Refactor `sov-cli` into a library

preston-evans98:
# Description
This PR makes refactors `sov-cli` into its a library crate, and updates the binary from demo-stf to use the library. In the process, it simplifies the API exposed by the CLI. 

Improvements include:
- Store private keys between invocations, so most calls don't need to pass a private key
- Auto-fill nonces during batch submission, so users never have to think about nonces
- Allow batch submission in one step, instead of first serializing and then submitting the batch
- Allow querying of account balances and nonces

In addition, this PR significantly refactors the CLI generation macro. Now, the CLI is split into two pieces - an intermediate representation (which is turned into `CallMessage`s), and a `FrontEnd` (think `clap`). By separating these APIs, we're able to defer the decision of whether or not to use `clap` from the `CliWallet` derive macro to a higher level of the stack. In other words, the CliWallet macro now only has to generate the IR, and then the ultimate consumer simply has to figure out how to create the IR from the command line. As a bonus, this change means that `cli_skip` should now work as expected even if the struct has complex generics. 

## Linked Issues
- Related to #552 

## Testing
Most workflows have been tested. Additional tests will be added in the coming days.

## Docs
All relevant README's have been updated.


examples/demo-stf/src/runtime.rs
```diff
@@ -1,11 +1,11 @@
 #[cfg(feature = "native")]
-use sov_accounts::{AccountsRpcImpl, AccountsRpcServer};
+use sov_accounts::query::{AccountsRpcImpl, AccountsRpcServer};
```
bkolad, line 2: Why do we reexport `query` from the modules? Making it private hides the internal representation of the module.

module-system/sov-modules-macros/src/cli_parser.rs
```diff
@@ -1,7 +1,7 @@
 use quote::{format_ident, quote};
-use syn::{Data, DataEnum, DeriveInput, Fields, Ident, Type};
+use syn::{Data, DataEnum, DeriveInput, Fields, Ident};
 
-use crate::common::{extract_ident, StructFieldExtractor};
+use crate::common::StructFieldExtractor;
 
 pub(crate) struct CliParserMacro {
```
bkolad, line 6: Do we have a unit test for this macro?

module-system/sov-cli/src/workflows/transactions.rs
```diff
@@ -1,107 +1,129 @@
...
-pub enum ImportTransaction {
+pub enum ImportTransaction<Json: Subcommand, File: Subcommand> {
     /// Import a transaction from a JSON file at the provided path
-    #[command(arg_required_else_help(true))]
-    FromFile {
-        /// The expected format of the file contents is: {"module_name": {"call_name": {"field_name": "field_value"}}}
-        path: PathBuf,
-    },
+    #[clap(subcommand)]
+    FromFile(Json),
```
bkolad, line 81: Is it possible to use `Box< Subcommand>` for  `FromFile(Json)`, and for ` FromString(File)`? This way, we could eliminate generics from `ImportTransaction`. Currently, we will need to keep adding the generics as we add new subcommands.

module-system/sov-modules-api/src/cli.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl TryFrom<FileNameArg> for JsonStringArg {
+    type Error = std::io::Error;
+    fn try_from(arg: FileNameArg) -> Result<Self, Self::Error> {
+        let json = std::fs::read_to_string(arg.path)?;
+        Ok(JsonStringArg { json })
+    }
+}
+
+pub trait CliFrontEnd<RT>
```
bkolad, line 27: The `RT` is not used anywhere. Do we need it?

module-system/sov-cli/src/workflows/transactions.rs
```diff
@@ -1,107 +1,129 @@
...
+    where
+        Json: CliFrontEnd<RT>,
+        File: CliFrontEnd<RT>,
+        Json: TryInto<RT::CliStringRepr<U>, Error = E1>,
+        File: TryInto<RT::CliStringRepr<U>, Error = E2>,
+        RT::CliStringRepr<U>: TryInto<RT::Decodable, Error = E3>,
+        RT::Decodable: Serialize,
+        E1: Into<anyhow::Error> + Send + Sync,
+        E2: Into<anyhow::Error> + Send + Sync,
+        E3: Into<anyhow::Error> + Send + Sync,
```
bkolad, line 110: Would it be ok to hardcode errors to anyhow in the upstream code? This way, we would reduce the number of generics needed by `run.`

bkolad (approved):
LGTM, if possible I would prefer to keep `mod query` private in modules.

module-system/sov-cli/src/workflows/transactions.rs
preston-evans98: Great question. I think there should be a way to get rid of these, but it was taking a long time to satisfy the type checker so I wanted to merge this PR to unblock Nikolai first. I'll keep following up on this as I have time.

module-system/sov-modules-api/src/cli.rs
preston-evans98: I don't think we need it *yet*, but we'll have to have it for the version that removes the extra generics. Otherwise, we'll start running into issues with specialization

module-system/sov-cli/src/workflows/transactions.rs
preston-evans98: I think there should also be a way to get rid of these error types without hardcoding when we switch the generics. If not, I'll try to hardcode at least some of these to reduce the number of generics here.

examples/demo-stf/src/runtime.rs
preston-evans98: Synced with @bkolad  offline. Summarizing here: 

The motivation for this change is that we were forgetting to export a bunch of the generated types that users need when they want to actually query the module.  (i.e. _RpcClient, AccountsResponse). Since some of those types are macro generated, I thought it would be cleaner to export the query module instead of expecting module developers export those types manually. 

After talking it through, we think it will be even cleaner to  have the rpc macro generate a new mod  named something like generated_public_types and add a pub use query::generated_public_types::*  instead.

preston-evans98: Opened #713

module-system/sov-modules-macros/src/cli_parser.rs
preston-evans98: Updated the old test to work with the new structure!

examples/demo-rollup/README.md
```diff
@@ -229,88 +225,76 @@ Here's an example of a JSON representing the above call:
...
-$ cat ./examples/test-data/requests/tx_blob
-01000000b60000005ef848746e8d2b9c27ee46210e185dc9f3b690d5cef42a13fb9c336bd40c798210bf7af613997f7af57c9681a242f5fe4121a1539ba4f5f32f14c49f978b990a7b758bf2e7670fafaf6bf0015ce0ff5aa802306fc7e3f45762853ffc37180fe64a0000000001fea6ac5b8751120fb62fff67b54d2eac66aef307c7dde1d394dea1e09e43dd44c800000000000000135d23aee8cb15c890831ff36db170157acaac31df9bba6cd40e7329e608eabd0000000000000000
-```
+This output indicates that the wallet has saved the transaction details for later signing.
 
-#### 4. Submit the Transaction
+#### 3. Submit the Transaction(s)
 
-You now have a blob with one serialized transaction in `./examples/test-data/requests/tx_blob`. Switch back to the `examples/demo-rollup` directory and use the Makefile to submit it:
+You now have a batch with a single transaction in your wallet. If you want to submit any more transactions as part of this
```
cemozerr, line 285: This PR is great. 

Just a small nit, not sure, but wondering if  the usage of word "wallet" in this tutorial might be confusing to folks. 

Because previously we refer to the same tool as `sov-cli` everywhere. Thoughts @preston-evans98?

PR #711, opened 2023-08-22
Modify risc0 tracer util to print function stack to enable better debugging

dubbelosix:
## Enhancements to Tracer Output and Functionality

With this Pull Request, we introduce several key modifications to the Tracer:

* **Function-specific Tracing**: Users can now optionally pass a function name, enabling them to print the entire stack specific to that function.
* **Cycle Counting**: We've incorporated a function to retrieve actual cycle counts as opposed to mere instruction counts. It's noteworthy that while the majority of instructions run at 1 cycle, there are specific ones that necessitate 2 cycles.
* **Makefile Integration**: Introduced a Makefile process that allows for the generation of the trace, the creation of rollup.elf, and subsequently runs the tracer. This is needed because minor changes to the code can cause significant changes to the addresses in the symbol table and completely disrupt the construction of the stack and profiling the cycle counts when the tracer is run. The Makefile ensures that running `make` would generate the new ELF, a new trace file (with the program counter and the instruction name) and then run the tracer util

This PR aims to augment both the granularity and accuracy of the Tracer, offering users a broader spectrum of insights into their execution flows.


examples/demo-rollup/rollup_config.toml
```diff
@@ -1,6 +1,6 @@
 [da]
 # The JWT used to authenticate with the celestia light client. Instructions for generating this token can be found in the README
-celestia_rpc_auth_token = "MY.SECRET.TOKEN"
+celestia_rpc_auth_token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJwdWJsaWMiLCJyZWFkIiwid3JpdGUiLCJhZG1pbiJdfQ.MhcEa3-fyoZvrf5bJ-sqUnrJi1cHhfKBq0W4lGT-oso"
```
neysofu: This was probably committed by accident.

utils/zk-cycle-utils/tracer/src/main.rs
```diff
@@ -38,24 +38,67 @@ struct Args {
...
+
+        // Don't see this in the risc0 code base, but MUL, MULH, MULSU, and MULU all take 1 cycle,
+        // so going with that for MULHU as well.
+        "MULHU" => Ok(1),
+
+        "XORI" | "ORI" | "ANDI" | "SRLI" | "SRAI" | "DIV" | "DIVU" | "REM" | "REMU" => Ok(2),
+
+        _ => Err("Decode error"),
+    }
+}
```
neysofu: Can we add a comment here with a reference to the docs/upstream code where this is explained? If there is any

neysofu (approved):
LGTM! I think a comment is needed in the instruction count function and `rollup-config.toml` got updated by mistake, that's all

examples/demo-rollup/rollup_config.toml
dubbelosix: ah will remove it

utils/zk-cycle-utils/tracer/src/main.rs
dubbelosix: will do!

neysofu: Nice, thank you!

PR #716, opened 2023-08-22
Add create-token integration test

preston-evans98:
# Description
Add integration test using the vanilla module system with MockDA. Refactor EVM integration test to use random RPC port, which allows unlimited concurrency.


examples/demo-rollup/tests/module_system_integration_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+use demo_stf::app::DefaultPrivateKey;
+use demo_stf::runtime::RuntimeCall;
+use jsonrpsee::core::client::ClientT;
+use jsonrpsee::http_client::{HttpClient, HttpClientBuilder};
+use sov_modules_api::default_context::DefaultContext;
+use sov_modules_api::transaction::Transaction;
+use sov_modules_api::{PrivateKey, Spec};
+use test_helpers::start_rollup;
+
+struct TestClient {
```
bkolad: This is a generic code that other rollups could reuse. Let's rename it to `SimpleRollupClient` or something similar and move it to `full-node` or even `rollup-interface`?

examples/demo-rollup/tests/module_system_integration_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+        minter_address: address,
+        authorized_minters: vec![],
+    });
+    let tx = Transaction::<DefaultContext>::new_signed_tx(&key, msg.try_to_vec().unwrap(), 0);
+
+    let client = TestClient::new(&format!("http://localhost:{}", rpc_address.port()));
+
+    client.send_transaction(tx).await
+}
+
```
bkolad: I propose to restructure a bit the file structure:
```
/tests
    -/evm
        - evm_integration_test.rs
        - test_helpers.rs
    -/bank
        - tests.rs
```
This way, it will be clear how to add tests for the rest of the modules in `demo-rollup`.

examples/demo-rollup/tests/test_helpers.rs
```diff
@@ -31,6 +39,49 @@ fn make_contract_from_abi(path: PathBuf) -> BaseContract {
...
+        runner_config: rollup_config.runner,
+        genesis_config,
+    }
+}
+
+pub async fn start_rollup(rpc_reporting_channel: oneshot::Sender<SocketAddr>) {
+    let mut mock_path = PathBuf::from("tests");
+    mock_path.push("test_data");
+    mock_path.push("tmp");
+    mock_path.push("mocks");
```
bkolad: I think the test fails here https://github.com/Sovereign-Labs/sovereign-sdk/actions/runs/5944938901/job/16123134157?pr=716#step:7:132 because we share the same db path. Let's add another string argument, `test_name`, and push it to the `mock_path`. This way tests will have a unique db.

examples/demo-rollup/tests/module_system_integration_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+    let (port_tx, port_rx) = tokio::sync::oneshot::channel();
+
+    let rollup_task = tokio::spawn(async {
+        start_rollup(port_tx).await;
+    });
+
+    // Wait for rollup task to start:
+    let port = port_rx.await.unwrap();
+
+    send_test_create_token_tx(port).await?;
```
bkolad: Let's use the `bank::supply_of` query to verify the token's creation. We can determine the token's address using the `bank::get_token_address` function. The challenge here is that the test should wait until the token is actually created. As a temporary solution, we can introduce a `sleep(1000)` delay and open a separate issue to enhance the `MockDaService` functionality, allowing it to signal tests when a blob has been processed.

bkolad (approved):
Thanks a lot!

examples/demo-rollup/tests/module_system_integration_tests.rs
preston-evans98: I moved it to the `sov-sequencer` crate since...
1. `rollup-interface` doesn't currently take a dependency on jsonrpsee
2. It's implementation is coupled with the sequencer rpc implementation

examples/demo-rollup/tests/test_helpers.rs
preston-evans98: I switched to using `tempfile` instead, both to avoid needing to pick an argument value and for consistency with other tests.

examples/demo-rollup/tests/module_system_integration_tests.rs
preston-evans98: Unfortunately, `cargo` won't recursively look for tests in sub-directories, so all of our test files have to live at the top level.

bkolad: What about this?
```
/tests
    -/evm
        - mod.rs
        - test_helpers.rs
    -/bank
        - mod.rs
 all_tests.rs
```
and in `all_tests.rs`
```
mod evm
mod bank
```
But if it gets too complex then let's stick to what we have now.

examples/demo-rollup/tests/module_system_integration_tests.rs
preston-evans98: I went a little overboard and added a whole websocket subscription rpc to the rollup so that clients can be notified when blocks are processed. Now we wait to be notified that a slot has been run by the rollup before querying!

examples/demo-rollup/tests/bank_integration_tests.rs
```diff
@@ -0,0 +1,81 @@
...
+        token_name: TOKEN_NAME.to_string(),
+        initial_balance: 1000,
+        minter_address: user_address,
+        authorized_minters: vec![],
+    });
+    let tx = Transaction::<DefaultContext>::new_signed_tx(&key, msg.try_to_vec().unwrap(), 0);
+
+    let port = rpc_address.port();
+    let http_client = SimpleClient::new(&format!("http://localhost:{}", port));
+    let ws_client = WsClientBuilder::default()
```
bkolad: This is great! What about moving it inside  `SimpleClient` and adding a relevant method there?

examples/demo-rollup/tests/module_system_integration_tests.rs
bkolad: Great!

examples/demo-rollup/tests/module_system_integration_tests.rs
preston-evans98: Going to merge as-is since ci, but happy to make the small change in a fast-follow

examples/demo-rollup/tests/bank_integration_tests.rs
preston-evans98: Will do!

examples/demo-rollup/tests/module_system_integration_tests.rs
preston-evans98: Looks like we need one more commit anyway, so will do!

PR #715, opened 2023-08-22
[TEST] Run prover in ci

preston-evans98:
# Description
Run the demo-prover in CI

## Linked Issues
- Fixes # (issue, if applicable)
- Related to # (issue) 

## Testing
Describe how these changes were tested. If you've added new features, have you added unit tests?

## Docs
Describe where this code is documented. If it changes a documented interface, have the docs been updated?

PR #717, opened 2023-08-22
Use JMT update proofs

### LINKED ISSUE #246: Update jmt to remove honest treereader assumption.
Currently, the JMT supports trustless proof verification but it does *not* support trustless updates. Specifically, there's currently no way to enforce that every sibling hash in an updated merkle tree is the same as the sibling hash in the old merkle tree *in the absence of a trusted treereader`. To fix this, we need to prove the correctness of every sibling hash used to compute the new merkle root against the old merkle root.

preston-evans98:
# Description
Update zk_storage to use jmt::update proofs. This fixes the soundness bug in our proofs and may improve performance.

## Linked Issues
- Fixes #246 

## Testing
This PR should be exercised by the demo-prover tests. We should wait to merge it until #715 is complete. We should also add integration tests exercising the whole "prover" workflow outside of Risc0.


preston-evans98:
Implemented by #908

PR #718, opened 2023-08-23
Hash-encode keys in `StateMap` instead of serializing them with a codec

neysofu:
# Description
This PR supersedes #689 and #644. This is the third iteration of the design:
1. Allow `Borrow` targets as keys as long as they serialize with Borsh. This, as noted, can cause different key types to serialize differently. #644 
2. Allow all `StateKeyEncodePreservingBorrow` target types as keys. This was a step in the right direction and maintained the invariants that we're interested in, but it's too easily prone to misuse by SDK users which may inadvertently produce subtle bugs if they're not too careful. #689
3. (This PR) Remove the `StateKeyCodec`, `StateCodec`, and `StateKeyEncodePreservingBorrow` traits, and use `Hash` to serialize all keys. `Borrow` targets are always supposed to `Hash` the same way, and this invariant is relied upon by `HashMap` of all types so it tends to be respected by all 3rd-party crates. It seems safer than `StateKeyEncodePreservingBorrow`.

I also made minor changes to `StorageKey` and `CacheKey` to allow use as reference, just like #644.

## Linked Issues
Related to #689 and #644.

## Testing
I'm relying on the current existing tests as no new interfaces were added, only modified.

## Docs
I modified the relevant docs for traits and `StateMap`.


module-system/sov-modules-api/src/default_signature.rs
```diff
@@ -123,6 +124,12 @@ pub struct DefaultPublicKey {
     pub(crate) pub_key: DalekPublicKey,
 }
 
+impl Hash for DefaultPublicKey {
```
vlopes11, line 127: To double check: why do we opt to implement hash manually with the serialized bytes instead of relying on the derive macro that will implement it outside the scope of serialization?

For further context: I believe the main advantage of bringing `Hasher` to the table was to separate the two domains of serialization & in-memory hash table. We are having problems because we intersect the two, when we have no formal guarantee that a "keyable" type will have uniform serialization implementation whereas they might have the same memory representation.

It might be that we have strict requirements in our ZK tree that will require these items to use the storage implementation. Is that the reason? If so, I think its best to add an explicit comment describing that so future developers will immediately understand the decision

vlopes11 (commented):
The approach looks good! I have one question on why we still use serialization, even after the introduction of the `Hasher` of stdlib

module-system/sov-modules-api/src/default_signature.rs
neysofu: We can't derive `Hash` for this type because `DalekPublicKey` doesn't implement `Hash`, so we have to write our own. In this case, it just so happens that the type implements `Hash` re-using some of its custom serialization logic under the hood. I don't see an issue with this, but maybe I'm just not quite getting what you're saying.

> we have no formal guarantee that a "keyable" type will have uniform serialization implementation whereas they might have the same memory representation.

I believe we do, don't we? `Eq` and `Hash` guarantee consistency with each other. In this case, `DalekPublicKey` implements `Eq` as well as `AsRef<[u8]>` and we can assume that `pubkey1 == pubkey2` iff `pubkey1.as_bytes() == pubkey2.as_bytes()`.

vlopes11: Sounds good! From what I understand the [CompressedEdwardsY](https://docs.rs/curve25519-dalek/4.0.0/curve25519_dalek/edwards/struct.CompressedEdwardsY.html) is affine, so the bytes should be deterministic!

LGTM

vlopes11 (approved):
Very good, will help a lot with the ergonomics

PR #719, opened 2023-08-23
EVM remove receipt type

bkolad:
# Description
1. Adds `BcsCodec` in `sov-state`. `bincode & postcard` have problems with serde `[serde(flatten)]` (https://github.com/bincode-org/bincode/issues/245) used by `etheres::Receipt` struct.
```
panicked at 'called `Result::unwrap()` on an `Err` value: SequenceMustHaveLength'
```
2. Replaces custom `Receipt` from the `EVM` module with `etheres::Receipt`. 

## Testing
Ci passes.

## Docs
No doc changes.


neysofu (commented):
Just one comment about `codec` file organization :100:

module-system/sov-state/src/codec.rs
neysofu, line 1: Three files for barely 200 LoCs seems like spreading it too thin. Do you think just one file is cluttered? as an alternative, maybe `codec/` is better and we won't pollute the root source directory as much.

bkolad: I like that different codecs are in different files, the separation is very clear. I added `codec/`

neysofu: 👍

neysofu (approved):

PR #720, opened 2023-08-23
Split AddressTrait

### LINKED ISSUE #493: Split `AddressTrait`

citizen-stig:
# Description

1. Splits `AddressTrait` into 2 traits:

 * `BasicAddress` - general address trait, that we require for address to be nicely used. Has no requirement on underlying structure
 
 * `RollupAddress` - trait that is build on top of `BasicAddress`, but adds requirements it to be convertable from 32 bytes array without error.

2. Adds `FromStr` bound to `BasicAddress`

All addresses needs to be displayed anyway, so having a way to initiate it from string will simplify configs, and call messages

3. Remove `DaAddress` and use `MockAddress` from `sov-rollup-inteface` instead

## Linked Issues
- Fixes #493

## Testing

Existing test are passing

## Docs

Updated

### DIFF
--- a/adapters/celestia/src/celestia.rs
+++ b/adapters/celestia/src/celestia.rs
@@ -1,6 +1,4 @@
-use std::fmt::{Display, Formatter};
 use std::ops::Range;
-use std::str::FromStr;
 use std::sync::{Arc, Mutex};
 
 use base64::engine::general_purpose::STANDARD as B64_ENGINE;
@@ -12,7 +10,6 @@ use prost::Message;
 use serde::{Deserialize, Serialize};
 use sov_rollup_interface::da::{BlockHeaderTrait as BlockHeader, CountedBufReader};
 use sov_rollup_interface::services::da::SlotData;
-use sov_rollup_interface::BasicAddress;
 pub use tendermint::block::Header as TendermintHeader;
 use tendermint::block::Height;
 use tendermint::crypto::default::Sha256;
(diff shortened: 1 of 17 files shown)


module-system/sov-modules-api/src/lib.rs
```diff
@@ -53,7 +53,9 @@ pub use error::Error;
 pub use prefix::Prefix;
 pub use response::CallResponse;
 use serde::{Deserialize, Serialize};
-pub use sov_rollup_interface::{digest, AddressTrait};
+use sov_rollup_interface::BasicAddress;
+// TODO: Check this.
+pub use sov_rollup_interface::{digest, RollupAddress};
```
citizen-stig: Why do we need to re-export?

adapters/celestia/src/celestia.rs
```diff
@@ -370,14 +370,6 @@ impl<'a> TryFrom<&'a [u8]> for H160 {
     }
 }
 
-impl From<[u8; 32]> for H160 {
```
preston-evans98, line 373: I think `H160` might be is never used as an address (now that we have the correct Celestia address type). Can we remove it entirely?

preston-evans98 (approved):

adapters/celestia/src/celestia.rs
citizen-stig: Sure! Done!

PR #733, opened 2023-08-25
Fix `expose_rpc` macro with associated types

citizen-stig:
# Description

`expose_rpc` macro didn't handle generics with associated types. This PR fixes this, which should allow to have runtime with DaSpec generic parameter

## Testing
Tests have been added

## Docs
Documentation have been updated


module-system/sov-modules-macros/src/rpc/expose_rpc.rs
```diff
@@ -33,6 +33,7 @@ impl ExposeRpcMacro {
             })
             .ok_or(syn::Error::new_spanned(
                 &generics,
+                // CliWallet?
```
citizen-stig: @preston-evans98 is it copy paste error or do I miss something?

bkolad (approved):

PR #739, opened 2023-08-28
CLEANUP: remove sov-election module since we don't need it and modify tests using it to use sov-bank

dubbelosix:
### Summary

This PR removes the legacy `sov-election` module in favor of the more full-featured `sov-bank` module. The goal is to maintain a clean and streamlined codebase.

### Changes

- **Remove `sov-election` Module**: To avoid clutter, the older `sov-election` module has been removed.
  
- **Convert Existing Tests**: Tests that were relevant to `sov-election`, such as those for transaction revert, have been adapted to work with `sov-bank`.

- **Update Data Generators**: The `election` data generator has been removed, and the `bank` data generator has been extended to include additional test scenarios.

### Test Scenarios moved to `sov-bank`

1. **Transaction with Invalid Serialization**: Tests to ensure that transactions with invalid serialization are handled correctly.
2. **Transaction with Invalid Signature**: Tests to verify that transactions with an invalid signature are rejected.
3. **Transaction revert (Fails Business Logic)**: Tests to confirm that transactions which do not meet business logic are reverted. Specifically, in the context of the Bank module, scenario where an attempt is made to transfer out more tokens than have been minted.
4. **Reverted transaction increments nonce**: Tests to confirm that even failed and reverted transactions will increment the nonce.


examples/demo-stf/src/tests/da_simulation.rs
```diff
@@ -14,43 +13,44 @@ use crate::runtime::Runtime;
...
 }
-
-pub fn simulate_da_with_bad_serialization(election_admin: DefaultPrivateKey) -> Vec<RawTx> {
-    let election = BadSerializationElectionCallMessages::new(election_admin);
-    election.create_raw_txs::<Runtime<C>>()
+//
+// // TODO: Remove once we fix test with bad nonce
+// //   https://github.com/Sovereign-Labs/sovereign-sdk/issues/235
+// #[allow(unused)]
+// pub fn simulate_da_with_bad_nonce(election_admin: DefaultPrivateKey) -> Vec<RawTx> {
```
citizen-stig: Is there anything right now stops us from testing this?

citizen-stig (approved):

examples/demo-stf/src/tests/da_simulation.rs
dubbelosix: ah good catch, i missed that one. the function isn't used, but right now I think we can add a test for that since we have a bad nonce error message and txn failure!

Issue #737, opened 2023-08-28
EVM: introduce tx-block that is compatible with the rollup batching meechanism.


bkolad:
## Description
Currently, the EVM module processes transactions sequentially, without the notion of blocks. To enable the `getBlockByNumber` endpoint, we have to group  EVM transactions into blocks in a manner that aligns with the rollup batching mechanism.

Issue #615, opened 2023-08-03
Allow Modules to Store and Serve Data outside of the JMT


preston-evans98:
## Background
Currently, modules have access to `#[state]` variables. Under the hood, these variables are stored in the authenticated data store backed by the JMT. 

Recently, the EVM and Optimistic rollup workstreams have both encountered the need to serve data via RPC which does not need to be authenticated. (For the EVM, this is receipt and log data; for ORUs, this is storage proofs). While we always *can* store extra data in state, it's significantly preferable to avoid the overhead in these cases. Instead, we should add a new field type to modules which is write-only within the state machine, but can be read in RPC calls.

## Desired Outcome
Modules should be able to efficiently store and retrieve arbitrary auxiliary data using a non-provable data store. The store should be accessible for reading in RPC handlers but *must not* be readable during execution.

## Potential Designs
Off the top of my head, I can see at least three possible designs: 

### Option 1: Add a new API for registering module-defined DBs
In this design, we create a new trait which allows modules to bring their own (non-provable) database and have it opened/initialized during node startup. The motivating example for this design is to allow the EVM module to use Reth's existing LMDB/MDBX database rather than attempting to port their storage model to RocksDB. 

A very rough sketch of the design is like this:

```rust
// sov-modules-api/lib.rs
pub trait UnprovableStore: Clone { 
  type Config: DeserializeOwned;
  fn open(Config) -> Self;
  fn put<S: Schema>(&self, S::Key, S::Value) -> Result<()>;
  #[cfg(feature = "native")]
  fn get<S: Schema>(&self, S::Key) -> Result<Option<S::Value>>;
} 

#[derive(Default)]
pub struct DbRegistry { 
  dbs: HashMap<Address, Box<dyn UnprovableStore>>
}

//my-modules/lib.rs
/// A new initailization function to create the module store. This is invoked
/// during node startup
fn init(config: Config) -> Option<UnprovableStore>;

//runtime.rs
/// A new (macro-generated?) function to initalize all module dbs.
fn init(config: Config) ->  {
  let db_registry = Default::default();
  if let Some(db) = MyModule::init(config.my_module) { 
    registry.add(MyModules::new().address(), db)
  } 
}

//scratchpad.rs
pub struct WorkingSet<S> { 
  // ...
  module_dbs: DbRegistry,
} 

impl<S> WorkingSet<S> { 
  pub fn get_unprovable(...);
  pub fn put_unprovable(...);
} 

```

[...]

PR #742, opened 2023-08-28
Update `end_slot_hook`

bkolad:
# Description
This PR adds `state_root` to the `end_slot_hook` (which the EVM module requires). We do it as a separate PR because it influences the #615 discussion. 

## Linked Issues
- Related to #737 

## Testing
Ci passes.

## Docs
No update.

### DIFF
--- a/examples/demo-stf/src/runtime.rs
+++ b/examples/demo-stf/src/runtime.rs
@@ -119,8 +119,10 @@ impl<C: Context, Cond: ValidityCondition> SlotHooks<Cond> for Runtime<C> {
 
     fn end_slot_hook(
         &self,
-        _working_set: &mut sov_state::WorkingSet<<Self::Context as sov_modules_api::Spec>::Storage>,
+        working_set: &mut sov_state::WorkingSet<<Self::Context as sov_modules_api::Spec>::Storage>,
+        root_hash: [u8; 32],
     ) {
+        // self.evm.end_slot_hook(working_set, root_hash);
     }
 }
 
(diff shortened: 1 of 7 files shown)


module-system/sov-modules-stf-template/src/lib.rs
```diff
@@ -99,8 +99,7 @@ where
             .expect("jellyfish merkle tree update must succeed");
 
         let mut working_set = WorkingSet::new(self.current_storage.clone());
-
-        self.runtime.end_slot_hook(&mut working_set);
+        self.runtime.end_slot_hook(root_hash, &mut working_set);
```
bkolad, line 102: @neysofu we will need to update the `non-JMT` data in the `end_slot_hook` but with the latest state root hash. This raises the question of when we save the `non-JMT` data.

neysofu: Yup, as agreed during our call. I think we have a plan to tackle this.

examples/demo-stf/src/runtime.rs
```diff
@@ -119,8 +119,13 @@ impl<C: Context, Cond: ValidityCondition> SlotHooks<Cond> for Runtime<C> {
 
     fn end_slot_hook(
         &self,
-        _working_set: &mut sov_state::WorkingSet<<Self::Context as sov_modules_api::Spec>::Storage>,
+        #[allow(unused_variables)] root_hash: [u8; 32],
```
neysofu, line 122: I didn't know you could add argument-specific attributes. Nice!

neysofu (approved):
LGTM

PR #748, opened 2023-08-28
Migrate to buildjet

preston-evans98:
# Description
Switch to CI to buildjet, which is cheaper, faster, and allows larger caches

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -62,15 +62,14 @@ concurrency:
 jobs:
   check:
     name: check
-    runs-on:
-      group: 4-cores_16GB_Ubuntu Group
+    runs-on: buildjet-4vcpu-ubuntu-2204
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@v3
       - uses: rui314/setup-mold@v1
       - name: Install Rust
         run: rustup show
-      - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -62,15 +62,14 @@ concurrency:
...
-      group: 4-cores_16GB_Ubuntu Group
+    runs-on: buildjet-4vcpu-ubuntu-2204
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@v3
       - uses: rui314/setup-mold@v1
       - name: Install Rust
         run: rustup show
-      - uses: Swatinem/rust-cache@v2
+      - uses: astriaorg/buildjet-rust-cache@v2.5.1
```
neysofu: I wonder if we can continue using Swatinem's action by using the new `buildjet` backend – I'm not sure it's equivalent, but it looks like something was upstreamed (https://github.com/Swatinem/rust-cache/pull/154) so it's worth a try. Like this:

```yml
- uses: Swatinem/rust-cache@v2
  with:
    # ...
    cache-provider: "buildjet"
```

neysofu:
That was fast! Thank you for looking into it, let's hope we can see some improvement.

.github/workflows/rust.yml
preston-evans98: Nice! I didn't see that his had been upstreamed already, I'll try switching back

neysofu (approved):

PR #752, opened 2023-08-29
Split `Storage::validate_and_commit` into two separate methods `Storage::calculate_root_hash...` and `Storage::commit`

bkolad:
# Description
In the `EVM` module, the end hook will need to pass the `state_root_hash` to the `EVM` block-building logic. The existing `Storage::validate_and_commit` method performs both the computation of the new state and the database commit as a single step. This PR divides this process into two methods. This change enables the `end_slot_hook` to access the state root hash after the computation step.

The `end_slot_hook` will be further revised after #615 is implemented.

## Linked Issues
- Related to # (issue) 

## Testing
Relevant tests updated.

## Docs
Relevant docs updated.


module-system/sov-modules-stf-template/src/lib.rs
```diff
@@ -93,14 +92,15 @@ where
...
+        let (root_hash, authenticated_node_batch) = self
             .current_storage
-            .validate_and_commit(cache_log, &witness)
+            .calculate_state_root_and_node_batch(cache_log, &witness)
             .expect("jellyfish merkle tree update must succeed");
 
         let mut working_set = WorkingSet::new(self.current_storage.clone());
         self.runtime.end_slot_hook(root_hash, &mut working_set);
 
+        self.current_storage.commit(&authenticated_node_batch);
```
bkolad, line 103: @neysofu we will need to commit the non-jmt data here as well.

LukaszRozmej (approved):

module-system/sov-state/src/prover_storage.rs
```diff
@@ -79,11 +79,11 @@ impl<S: MerkleProofSpec> Storage for ProverStorage<S> {
             .map(|root| root.0)
     }
 
-    fn validate_and_commit(
+    fn calculate_state_root_and_node_batch(
```
preston-evans98: nit: This is name is a tiny bit verbose. Maybe something like "compute_state_update" would be a little easier to read?

bkolad: fixed

module-system/sov-state/src/prover_storage.rs
```diff
@@ -79,11 +79,11 @@ impl<S: MerkleProofSpec> Storage for ProverStorage<S> {
             .map(|root| root.0)
     }
 
-    fn validate_and_commit(
+    fn compute_state_update(
         &self,
         state_accesses: OrderedReadsAndWrites,
         witness: &Self::Witness,
-    ) -> Result<[u8; 32], anyhow::Error> {
+    ) -> Result<([u8; 32], NodeBatch), anyhow::Error> {
```
preston-evans98: Not necessarily a blocker, but it's unfortunate that we leak the `NodeBatch` type into the interface now. Up until now, the `Storage` trait was decoupled from the jmt crate (IIRC), but NodeBatch is a jmt type.

This isn't a huge deal, because we our *database* is jmt specific anyway - so we would always have had to fix that before adding another `Storage` implementation - but this is moving us a little further from the idea end state.

bkolad: Very good point. Added `StateUpdate` associated type to the `Storage`.

PR #755, opened 2023-08-29
Implement `StateTransitionVerifier`. Refactor generics

preston-evans98:
# Description
This PR creates the `StateTransitionVerifier`, a struct exactly analogous the the `StateTransitionRunner` but intended for use in the zkvm. In order to create `StateTransitionVerifier` and allow it to be used in integration tests, a fairly significant refactor of the SDK's generics was required. The changes are as follows:
- Replace the `Cond: ValidityCondition` generic with a reference to `DA::ValidityCondition` wherever it appeared
- Replace the `B: BlobTransactionTrait`  generic with a reference to `DA::BlobTransaction` wherever it appeared
- Add a `Verifier` associated type to the `DaService` trait
- Refactor the `Zkvm` trait into a `ProofSystem` with associated `host`, `guest` and `verifier` trait.
```
trait ProofSystem { 
  type Guest: ZkvmGuest;
  #[cfg(feature = "native")]
  type Host: ZkvmHost;
}

trait ZkvmHost: ZkVerifier { .. }
trait ZkvmGuest: ZkVerifier { .. }
trait ZkVerifier { .. }
```


examples/demo-rollup/src/rollup.rs
```diff
@@ -22,10 +26,21 @@ use crate::register_rpc::register_ethereum;
...
+    Vm,
+>;
+
 /// Dependencies needed to run the rollup.
-pub struct Rollup<Vm: Zkvm, DA: DaService + Clone> {
+pub struct Rollup<Vm: ProofSystem, DA: DaService + Clone> {
     /// Implementation of the STF.
-    pub app: App<Vm, DA::Spec>,
+    pub app: App<Vm::Host, DA::Spec>,
     /// Data availability service.
```
bkolad, line 43: Just for uniformity, should we put `Option<(<Vm as ProofSystem>::Host, AppVerifier<DA, Vm>)>` as a field in Rollup? This way, `App` & `AppVerifier` are goe together.

full-node/sov-stf-runner/src/lib.rs
```diff
@@ -147,13 +187,30 @@ where
...
+                &validity_condition,
+                &mut blobs,
+            );
             for receipt in slot_result.batch_receipts {
                 data_to_commit.add_batch(receipt);
             }
-            let next_state_root = slot_result.state_root;
+            let next_state_root: ST::StateRoot = slot_result.state_root;
+            if let Some((prover_instance, guest)) = self.prover.as_mut() {
+                let (inclusion_proof, completeness_proof) = self
```
bkolad, line 211: The `run`, `demo-prover/host/main`& `demo-prover/guest/rollup` overlaps maybe we could share some code. 

*Between `run` & `demo-prover/host/main`*
There is:
1. For loop over heights.
2. `da_service.extract_relevant_txs`
3. `app.apply_slot`
4. `get_extraction_proof(&filtered_block, &blobs)` but here this is optional.
etc..

So, the "shape" of the code  is very similar.
It looks like the only differences are:
1.  The line 200 here is optional + in 206 we do `prover_instance.write_to_guest(&next_state_root)` while in host we do `host.write_to_guest(prev_state_root);` is that ok? 
2. We run the prover on `host/main` but this logic could be injected via a trait and set to no-op in the case where we don't prove. 

*Between `run` & `demo-prover/guest/rollup`*
It looks like with some adaptation we could use `guest.run_block()`  in `demo-prover/guest/rollup`?
`StateTransitionData` looks like  the code here [link](https://github.com/Sovereign-Labs/sovereign-sdk/pull/755/files/1394c18f89e1d8de680d5e0dd1a1c35524b47508#diff-1efa560a095d4f82e5954829120c2b595962a27c77e4aa0aa85480360c22c537L41)

bkolad: Another way would be to extract the logic from the `for loop` and reuse it here, and in  `demo-prover/host/main` + the `demo-prover` would add the proving step at the end.

examples/demo-rollup/src/rollup.rs
```diff
@@ -67,16 +82,24 @@ pub async fn new_rollup_with_celestia_da(
...
     /// Runs the rollup. Reports rpc port to the caller using the provided channel.
     pub async fn run_and_report_rpc_port(
+        self,
+        channel: Option<oneshot::Sender<SocketAddr>>,
+    ) -> Result<(), anyhow::Error> {
+        self.run_with_prover_opt(channel, None).await
+    }
+
+    /// Runs the rollup. Reports rpc port to the caller using the provided channel.
+    pub async fn run_with_prover_opt(
```
bkolad: It looks like this method is never called with `prover: Some(_)` is it going to be used later?

preston-evans98: Yep, this is a placeholder for now just to keep the diff manageable/ The follow up will add integ tests using this method

full-node/sov-stf-runner/src/lib.rs
preston-evans98: Yep, that's the goal! I left that change for a follow up to keep the diff smaller, but the idea is to basically remove all of the code from demo-prover/host and demo-prover/guest and replace it with `StateTransitionRunner.run()` and `StateTransitionVerifier.run_block()`

preston-evans98: Good catch on `next_state_root`, by the way! That's a bug. Fixing now!

examples/demo-prover/methods/guest/src/bin/rollup.rs
```diff
@@ -50,30 +47,30 @@ pub fn main() {
...
     env::write(&"blobs have been read\n");
 
-    // Step 2: Apply blobs
+    // Step 2: Verify tx list
+    let verifier = CelestiaVerifier::new(celestia::verifier::RollupParams {
+        namespace: ROLLUP_NAMESPACE,
+    });
+
+    let validity_condition = verifier
+        .verify_relevant_tx_list(&header, &blobs, inclusion_proof, completeness_proof)
```
citizen-stig, line 56: Does this reordering works? Have we tested it with actual prover?

`verify_relevant_tx_list` will consume blob reader, so then there will be "no transactions" in apply_slot.

rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -89,7 +101,15 @@ pub trait ZkvmGuest: Zkvm {
...
 /// This trait is implemented on the struct/enum which expresses the validity condition
 pub trait ValidityCondition:
-    Serialize + DeserializeOwned + BorshDeserialize + BorshSerialize + Debug + Clone + Copy + PartialEq
+    Serialize
+    + DeserializeOwned
+    + BorshDeserialize
+    + BorshSerialize
+    + Debug
+    + Clone
+    + Copy
```
citizen-stig, line 115: As a lot of work already done, could we make them can you please remove `Serialize + DeserializeOwned`, because it will make harder to integrate chain state module ?

preston-evans98: This does need to implement `Serialize + DeserializeOwned` though - otherwise we can't pass it from host to guest. I think maybe the issue was when we tried to use `Cond` directly instead of as an associated type? So maybe this will still work?

examples/demo-prover/methods/guest/src/bin/rollup.rs
preston-evans98: Hmm, that seems like a serious issue. If so, the type system is missing a very important invariant here - testing now

rollup-interface/src/state_machine/zk/mod.rs
citizen-stig: Yeah, let's test it when I start integrating!

examples/demo-prover/methods/guest/src/bin/rollup.rs
preston-evans98: Yep this is broken. Seems like a potentially major issue - deep diving now

examples/demo-simple-stf/src/lib.rs
```diff
@@ -26,7 +33,7 @@ pub enum ApplySlotResult {
     Success,
 }
 
-impl<Vm: Zkvm, Cond: ValidityCondition, B: BlobReaderTrait> StateTransitionFunction<Vm, B>
+impl<Vm: ZkVerifier, Cond: ValidityCondition, Da: DaSpec> StateTransitionFunction<Vm, Da>
```
citizen-stig, line 36: DaSpec has associated ValidityCondition, so it is possible to decrease used generics:

```suggestion
impl<Vm: ZkVerifier, Da: DaSpec> StateTransitionFunction<Vm, Da>
```

examples/demo-simple-stf/src/lib.rs
```diff
@@ -26,7 +33,7 @@ pub enum ApplySlotResult {
...
     for CheckHashPreimageStf<Cond>
```
citizen-stig, line 37:
```suggestion
    for CheckHashPreimageStf<DA::ValidityCondition>
```

citizen-stig: This would also ensure that used validity condition are the same type as in DaSpec

examples/demo-simple-stf/src/lib.rs
```diff
@@ -49,13 +56,14 @@ impl<Vm: Zkvm, Cond: ValidityCondition, B: BlobReaderTrait> StateTransitionFunct
...
-        // Do nothing
+        // Do nothing and return an empty state root
     }
 
-    fn apply_slot<'a, I, Data>(
+    fn apply_slot<'a, I>(
         &mut self,
         _witness: Self::Witness,
-        _slot_data: &Data,
+        _slot_header: &Da::BlockHeader,
```
citizen-stig, line 65: Nit: I think we should be consistent in using `Da` vs `DA` capitalizing. I believe it is better to stick to single one.

citizen-stig: And Now I look at it and naming convention, we should probably convert all `DA` to `Da`

examples/demo-stf/src/runtime.rs
```diff
@@ -124,27 +125,28 @@ impl<C: Context, Cond: ValidityCondition> SlotHooks<Cond> for Runtime<C> {
...
 where
     C: Context,
-    Cond: ValidityCondition,
-    B: BlobReaderTrait,
+    Da: DaSpec,
 {
 }
 
-impl<C: Context> BlobSelector for Runtime<C> {
+impl<C: Context, Da: DaSpec> BlobSelector<Da> for Runtime<C> {
```
citizen-stig, line 135: Ouch, I only now noticed, that I introduced similar, but different change: https://github.com/Sovereign-Labs/sovereign-sdk/pull/759 . It looks like having `Da` as generic is a little better, since anything else from this trait can be also accessed

fuzz/artifacts/accounts_call/crash-da39a3ee5e6b4b0d3255bfef95601890afd80709
citizen-stig, line 1: is this file needed?

preston-evans98:
Merged in pieces:  #780, #838, and #842

PR #764, opened 2023-08-31
Fix `expose_rpc` macro to handle all cases of generics from Runtime

### LINKED ISSUE #757: `expose_rpc` generic does not work with associated types 2 level deep.
If there's test struct like this with more then 1 level of associated types, expose_rpc macro do not generate correct `RpcStorage`:

```rust
struct TestStruct<T: FirstTrait, U: SecondTrait, V>
                where
                V: ThirdTrait,
                <U as SecondTrait>::Message: Message,
                <<U as SecondTrait>::Message as Message>::Caller: std::fmt::Display,
                T::Error: std::fmt::Debug,
            {
                field_1: Foo #path_arguments_1 ,
                field_2: Bar #path_arguments_2 ,
                field_3: Bar #path_arguments_3 ,
            }

```

There's ignored test case in common.rs: `test_generics_for_field_associated_types_nested_as`

citizen-stig:
# Description

Previously, generated `RpcStorage` was generic over only `C: sov_modules_api::Context` and then it was matching generics for each module RpcImpl and RpcServer, like that:

```rust
struct Runtime<C: Context, Da: DaSpec> {
   module_1: Module1<C, Da::Address>,
}


// Simplyfing here, basically module's RPC implementations match module generics, not rutime
trait Module1RpcImpl<C: Context, A: Address> {}

struct RpcStorage<C: Context> {
    storage: C::Storage
}

impl<C: Context, A: Address>  Module1RpcImpl<C, A> for RpcStorage<C> {
}
```

Problem with this approach, that is very difficult to match generics from runtime to impls for storage, especially with associated types of several nested levels.

Fortunately, there's solution, now RpcStorage is generic of over same things as Rutime:

```rust
struct Runtime<C: Context, Da: DaSpec> {
   module_1: Module1<C, Da::Address>,
}

struct RpcStorage<C: Context, Da: DaSpec>  {
    storage: C::Storage,
    phantom: PhantomData<Runtime<C, Da>
}
```

This approach just allows copy type parameters for each module.

The only disadvantage, is that all top level Runtime generics needs to be bound by `'static` which raises from jsonrpsee and tokio.

## Linked Issues
- Fixes #757

## Testing
More tests have been added

## Docs
Documentation has been updated


Cargo.toml
```diff
@@ -9,6 +9,7 @@ members = [
     "examples/demo-simple-stf",
     "examples/demo-rollup",
     "examples/demo-nft-module",
+    "examples/expose_rpc_debug",
```
citizen-stig: This is leftover

module-system/sov-modules-macros/src/common.rs
```diff
@@ -286,461 +280,3 @@ pub(crate) fn get_serialization_attrs(
...
-        //     expected_3.generics.to_token_stream(),
-        //     expected_3
-        //         .generics
-        //         .where_clause
-        //         .map_or("".to_string(), |w| w.to_token_stream().to_string())
-        // );
-
-        assert_eq!(expected_3.generics, actual_generics_3);
-    }
-}
```
citizen-stig, line 746: Add unit tests to existing functions

module-system/sov-modules-macros/src/rpc/expose_rpc.rs
```diff
@@ -84,23 +100,26 @@ impl ExposeRpcMacro {
 
             // We don't need to render where, because all bounds are inside `field_generics`
             let rpc_trait_impl = quote! {
-                impl #field_generics #rpc_trait_ident #matched_field_path_args for RpcStorage<#context_type> {
+                // impl #field_generics #rpc_trait_ident #matched_field_path_args for RpcStorage<#context_type> {
```
citizen-stig: remove comment

preston-evans98 (approved):

PR #773, opened 2023-09-01
Non-JMT "accessory" state

### LINKED ISSUE #615: Allow Modules to Store and Serve Data outside of the JMT

neysofu:
# Description
With this PR we're adding support for "accessory" state, i.e. key-value pairs only accessible from the native execution context which don't contribute to the state root hash. These are intended for simplifying serving some data over JSON-RPC and other tooling.

## Linked Issues
- Fixes https://github.com/Sovereign-Labs/sovereign-sdk/issues/615


module-system/module-implementations/examples/sov-accessory-state/src/lib.rs
```diff
@@ -0,0 +1,76 @@
+#[cfg(feature = "native")]
```
bkolad, line 4: let's add 
```
#![deny(missing_docs)]
```

neysofu: On it

module-system/module-implementations/examples/sov-accessory-state/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl<C: Context> AccessorySetter<C> {
+    /// Sets a value in the JMT state.
+    fn set_value(&self, s: String, working_set: &mut WorkingSet<C::Storage>) {
+        self.latest_value.set(&s, working_set);
+    }
+
+    /// Sets a value in the non-JMT state.
+    fn set_value_accessory(&self, s: String, working_set: &mut WorkingSet<C::Storage>) {
+        let prefix = Prefix::new(self.address.as_ref().to_vec());
```
bkolad: It would be grat to hide all this `prefix` handling from a user. Have you considered adding another field type in the module? For example we have `StateValue<String>` and we could add `NotAuthenticatedValue<String>` or something like that. Then would generate `prefix` the same way as for the `StateXX` fields. 
@preston-evans98 any thoughts about the API?

neysofu: Yes, adding non-JMT versions of `StateValue`, `StateMap` (and possibly `StateVec`) would be simple. I'm considering it, but I wanted to get some feedback first. The main drawback would be code duplication. We'd effectively almost copy-paste much of that code, which I'd like to avoid. The main alternative I see is to add a flag to the state containers we have already, and depending on the value of the flag we set/get/delete from JMT data or from accessory data. WDYT?

bkolad: Dumping some ideas:
Another approach would be to add 2 type level flags:
`struct Authenticated`
`struct NotAuthenticated`
Then `StateXX` takes it as a generic argument with default value `Authenticated` but we can opt out with 
```
value = StateValue<u32, SomeCodec, NotAuthenticated>
```
Maybe it is too complex tough.

neysofu: I'm unsure what I prefer myself. Both approaches have pros and cons.

module-system/sov-state/src/prover_storage.rs
```diff
@@ -16,27 +17,28 @@ use crate::{MerkleProofSpec, Storage};
 
 pub struct ProverStorage<S: MerkleProofSpec> {
     db: StateDB,
+    native_db: NativeDB,
```
bkolad, line 20: Now we have two `DBs`. Is it possible to guarantee atomic update between the two? If not then let's use the same DB with different column families and `atomic flush` from rocks db?

module-system/module-implementations/examples/sov-accessory-state/src/lib.rs
neysofu: Fixed by https://github.com/Sovereign-Labs/sovereign-sdk/pull/773/commits/419b2c1bd1f1c9ea48f0e82984784ab60648b40e.

module-system/module-implementations/examples/sov-accessory-state/src/lib.rs
neysofu: I implemented some `AccessoryState*` state containers, I'll resolve this conversation and we can address any comments about those separately :+1:

module-system/sov-state/src/prover_storage.rs
neysofu: Tracked in https://github.com/Sovereign-Labs/sovereign-sdk/issues/791!

module-system/sov-modules-stf-template/src/lib.rs
```diff
@@ -137,7 +138,8 @@ where
             .compute_state_update(log, &witness)
             .expect("Storage update must succeed");
 
-        self.current_storage.commit(&node_batch);
+        self.current_storage
+            .commit(&node_batch, &Default::default());
```
bkolad: Here (and in line 104), we also need to retrieve `AccessoryUpdate` from the working set and pass it to the commit, right?

module-system/module-implementations/examples/sov-accessory-state/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    fn call(
+        &self,
+        msg: Self::CallMessage,
+        _context: &Self::Context,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<sov_modules_api::CallResponse, Error> {
+        match msg {
+            CallMessage::SetValueAccessory(new_value) => {
+                self.accessory_value.set(&new_value, working_set);
```
bkolad: What about passing `working_set.accessorry_state_update` instead of the whole `working_set`?  This way it is clear that we won't modify the state.

neysofu: Good point, applied a fix in https://github.com/Sovereign-Labs/sovereign-sdk/pull/773/commits/f755953b921cb21d50c9712e5f108cb7cd39d4d7.

neysofu: (I meant to reply to the other comment.)

This is *also* a good point, though :smile:

neysofu: Addressed in https://github.com/Sovereign-Labs/sovereign-sdk/pull/773/commits/51bd05b53d4b48025febdee58a1084ee1d884d49.

bkolad (approved):

PR #789, opened 2023-09-04
Add utils for rollup with mock da

### LINKED ISSUE #784: Allow running `demo-rollup` with `MockDa`
We should provide a configuration to run `demo-rollup` with `MockDa`. The integration tests are already using `MockDa` so most of the code can be adapted from there, the main task will be to update steps `README`.

preston-evans98:
# Description
Make it easy to run demo-rollup with MockDa

## Linked Issues
- Fixes #784


citizen-stig (approved):

examples/demo-rollup/src/rollup.rs
```diff
@@ -82,6 +86,43 @@ pub async fn new_rollup_with_celestia_da(
     })
 }
 
+/// Creates MockDa based rollup.
+pub fn new_rollup_with_mock_da(
```
bkolad, line 90: Should we add a runtime flag to `demo-rollup/main` so we can choose between celesta and mock da?
If yes then:
1. Which one should be the default?
2. We will need to update the readme accordingly.

bkolad (approved):
LGTM

examples/demo-rollup/src/rollup.rs
preston-evans98: Hmm, yeah. In general, I think we want using different da layers to be a compile time change - but for the demo maybe it's nice to be able enable mock da at runtime. Maybe we should save this for a follow up?

preston-evans98: Chatted with @bkolad offline - we'll save this for a follow-up

PR #790, opened 2023-09-04
Remove unused dependencies 2023/09 edition

citizen-stig:
# Description

cargo udeps + cargo machete.

Only removed those which I felt comfortable removing. If something breaks, this is good opportunity to add tests

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -57,7 +57,7 @@ env:
 # Source: <https://stackoverflow.com/a/72408109/5148606>
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  # Except in `nightly` and `stable` branches! Any canceled job will cause the
+  # Except in `nightly` and `stable` branches! Any cancelled job will cause the
   # CI run to fail, and we want to keep a clean history for major branches.
   cancel-in-progress: ${{ (github.ref != 'refs/heads/nightly') && (github.ref != 'refs/heads/stable') }}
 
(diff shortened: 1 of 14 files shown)


neysofu (approved):

Makefile
```diff
@@ -21,16 +21,18 @@ install-dev-tools:  ## Installs all necessary cargo helpers
...
 	cargo install cargo-nextest --locked
+	cargo install cargo-risczero
+	cargo risczero install
 
 lint:  ## cargo check and clippy. Skip clippy on guest code since it's not supported by risc0
 	## fmt first, because it's the cheapest
-	cargo fmt --all --check
-	cargo check --all-targets --all-features
+	cargo +nightly fmt --all --check
+	cargo +nightly check --all-targets --all-features
```
neysofu: We only want `+nightly` for `cargo fmt`, all the other commands need to use stable.

neysofu (changes requested):
(Just to make sure the PR doesn't get auto-merged with the `+nightly` usage)

Makefile
citizen-stig: Done!

.github/workflows/rust.yml
```diff
@@ -57,7 +57,7 @@ env:
 # Source: <https://stackoverflow.com/a/72408109/5148606>
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  # Except in `nightly` and `stable` branches! Any cancelled job will cause the
+  # Except in `nightly` and `stable` branches! Any canceled job will cause the
```
citizen-stig:
```suggestion
  # Except in `nightly` and `stable` branches! Any cancelled job will cause the
```

neysofu (approved):

Issue #786, opened 2023-09-04
`sovereign.toml` for SDK config


vlopes11:
**Is your feature request related to a problem? Please describe.**
We need a way to parametrize the module compilation at compile time so the macros can be generated accordingly.

One of the problems this will solve is the gas config. We may set gas configuration on a `sovereign.toml` file, read it as literal string when generating the macros, and extract the relevant content.

**Describe the solution you'd like**
At compile time, read the same directory of the source file being worked by the macro, and search for a `sovereign.toml` file. It should have scoped sections so it can be used for different purposes. Example:

```toml
[gas]
base_cost = 15
variable_cost = 25

[foo]
bar = baz
```

The implementation should recurse to the parent dir and search for `sovereign.toml`, repeating the process until the root dir. If no `sovereign.toml` is found, then panic the compilation.

On the example above, let's take as example the `Account` `ModuleInfo` compilation. Let's assume the source path is located at `/home/vlopes11/sovereign-sdk/module-system/module-implementations/sov-accounts/src/lib.rs`, then, it would, in order, search for a toml file in the following paths:

```
/home/vlopes11/sovereign-sdk/module-system/module-implementations/sov-accounts/src/sovereign.toml
/home/vlopes11/sovereign-sdk/module-system/module-implementations/sov-accounts/sovereign.toml
/home/vlopes11/sovereign-sdk/module-system/module-implementations/sovereign.toml
/home/vlopes11/sovereign-sdk/module-system/sovereign.toml
/home/vlopes11/sovereign-sdk/sovereign.toml
/home/vlopes11/sovereign.toml
/home/sovereign.toml
/sovereign.toml
```

And finally fail, if none found.

There is, however, a problem. Currently, it is a nightly only feature to get the source path from a macro.

https://github.com/dtolnay/proc-macro2/blob/45ef770a5189ceca3adae9d4d6117a782da30a2a/src/lib.rs#L452-L459

This `procmacro2_semver_exempt` flag exists to shield backwards compatibility with older toolchains. It is not a feature, intentionally, and the user will have to set the `RUSTFLAGS` manually if he wants such functionality to be available. For more information, check the [Unstable features](https://docs.rs/proc-macro2/1.0.66/proc_macro2/index.html#unstable-features) documentation from `proc-macro2`.

With that said, it would be very inconvenient to require all downstream users to add the flags to their compilation as such flags will not replicate downstream and must be explicitly declared every time.

```sh
[...]

PR #792, opened 2023-09-04
feat: add sovereign.toml manifest file

vlopes11:
# Description

This commit introduces a manifest file that can be parsed by derive macros and have its contents read at compile-time.

It will read recursively the directory of the call site of the macro until it either finds a `sovereign.toml` file, or the path is depleted.

The implementation for the source file location of proc macros depends on `procmacro2_semver_exempt`, and it will fallback to the manifest path of the `sov-modules-macros` library. The fallback may contain multiple edge cases as cargo will cache the downloaded crate into a `$HOME/.cargo` dir, making a recursion tree impossible to reach the workspace under normal circumstances.

The aforementioned problem will be handled on the issue https://github.com/Sovereign-Labs/sovereign-sdk/issues/786

## Linked Issues
- Related to #786 

## Testing
A unit test will read a dummy `sovereign.toml` file and parse it.

## Docs
N/A


module-system/sov-modules-macros/sovereign.toml
```diff
@@ -0,0 +1,4 @@
+# this file exists only for test purposes
```
bkolad, line 1: If this is only for tests le's move it to /tests?

bkolad (approved):

module-system/sov-modules-macros/sovereign.toml
vlopes11: It has to be on the path from the macros src to the root dir (i.e. somewhere in `sovereign-sdk/module-system/sov-modules-macros/src/`)

PR #795, opened 2023-09-04
feature: gas meter

vlopes11:
Blocked by: #792

### DIFF
--- a/module-system/module-implementations/examples/sov-value-setter/src/lib.rs
+++ b/module-system/module-implementations/examples/sov-value-setter/src/lib.rs
@@ -15,10 +15,7 @@ pub use query::*;
 use sov_modules_api::{Error, ModuleInfo, WorkingSet};
 
 /// Initial configuration for sov-value-setter module.
-#[cfg_attr(
-    feature = "native",
-    derive(serde::Serialize, serde::Deserialize, Debug, PartialEq)
-)]
+#[derive(serde::Serialize, serde::Deserialize, Debug, PartialEq)]
 pub struct ValueSetterConfig<C: sov_modules_api::Context> {
     /// Admin of the module.
     pub admin: C::Address,
--- a/module-system/sov-modules-api/src/default_context.rs
+++ b/module-system/sov-modules-api/src/default_context.rs
@@ -31,7 +31,7 @@ impl Spec for DefaultContext {
 
 #[cfg(feature = "native")]
 impl Context for DefaultContext {
-    type GasUnit = TupleGasUnit<1>;
+    type GasUnit = TupleGasUnit<2>;
 
     fn sender(&self) -> &Self::Address {
         &self.sender
--- a/module-system/sov-modules-api/src/default_context.rs
+++ b/module-system/sov-modules-api/src/default_context.rs
@@ -9,7 +9,7 @@ use sov_state::{ArrayWitness, DefaultStorageSpec, ZkStorage};
 #[cfg(feature = "native")]
 use crate::default_signature::private_key::DefaultPrivateKey;
 use crate::default_signature::{DefaultPublicKey, DefaultSignature};
-use crate::{Address, Context, PublicKey, Spec};
+use crate::{Address, Context, PublicKey, Spec, TupleGasUnit};
 
 #[cfg(feature = "native")]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
@@ -31,6 +31,8 @@ impl Spec for DefaultContext {
 
 #[cfg(feature = "native")]
 impl Context for DefaultContext {
+    type GasUnit = TupleGasUnit<1>;
+
     fn sender(&self) -> &Self::Address {
         &self.sender
     }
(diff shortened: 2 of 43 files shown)


examples/demo-nft-module/tests/nft_test.rs
```diff
@@ -70,7 +76,11 @@ fn transfer() {
         owners: vec![(0, admin), (1, owner1), (2, owner2)],
     };
     let tmpdir = tempfile::tempdir().unwrap();
-    let mut working_set = WorkingSet::new(ProverStorage::with_path(tmpdir.path()).unwrap());
+    let mut working_set = WorkingSet::new(
+        todo!(),
```
bkolad: Let's put `gas & gas_price` in a single struct so the `WorkingSet::new` takes less arguments.

module-system/module-implementations/examples/sov-value-setter/src/lib.rs
```diff
@@ -30,6 +30,10 @@ pub struct ValueSetter<C: sov_modules_api::Context> {
     #[address]
     pub address: C::Address,
 
+    /// The gas configuration for the module
+    #[gas]
+    pub gas: DefaultGasConfig,
```
bkolad: is `[gas]` mandatory? If a module does not do any gas calculation can we skip it?

module-system/module-implementations/examples/sov-election/src/lib.rs
```diff
@@ -55,11 +59,16 @@ impl<C: sov_modules_api::Context> sov_modules_api::Module for Election<C> {
...
+
+    fn genesis<G>(
         &self,
         config: &Self::Config,
-        working_set: &mut WorkingSet<C::Storage>,
-    ) -> Result<(), Error> {
+        working_set: &mut WorkingSet<C::Storage, G>,
+    ) -> Result<(), Error>
+    where
+        G: GasUnit,
```
bkolad: We have to find a more ergonomic way to pass `GasUnit` around. What about putting it in `C: Context` this way `WorkingSet` won't need an explicit `G`

vlopes11: Very good suggestion; onto that

vlopes11:
Blocked by #938

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use core::fmt;
+
+/// A gas unit that provides scalar conversion from complex, multi-dimensional types.
+pub trait GasUnit: fmt::Debug + Clone {
+    /// A zeroed instance of the unit.
+    const ZEROED: Self;
+
+    /// Creates a unit from a multi-dimensional unit with arbitrary dimension.
+    fn from_arbitrary_dimensions(dimensions: &[u64]) -> Self;
```
vlopes11, line 11: This function will allow us to convert arbitrary arrays from the JSON file into the concrete, chosen type.

It is designed to have const-like performance, as in the link below
https://rust.godbolt.org/z/rPhaxnPEY

module-system/sov-modules-api/src/lib.rs
```diff
@@ -305,6 +310,17 @@ pub trait Module {
     ) -> Result<CallResponse, Error> {
         unreachable!()
     }
+
+    /// Attempts to charge the provided amount of gas from the working set.
+    ///
+    /// The scalar gas value will be computed from the price defined on the working set.
+    fn charge_gas(
+        &self,
```
vlopes11, line 324: We take `self` to allow the user to implement his own gas config, if he wants to

module-system/sov-modules-macros/src/manifest.rs
```diff
@@ -188,42 +189,29 @@ impl Manifest {
...
-        })?;
-
-        let def = quote::quote! {
-            #[allow(missing_docs)]
-            #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-            pub struct #ty {
-                #(#fields,)*
-            }
-        };
+        // remove generics, if any
```
vlopes11, line 224: Rust will not allow us to declare the type if it is defined via generics (i.e. reduce `Bank<C>` to `Bank`)

preston-evans98 (approved):

module-system/sov-modules-api/src/lib.rs
bkolad: Why do wee need this function in the `Context`? The caller can directly call `working_set.charge_gas(gas)`.

bkolad (approved):

module-system/sov-modules-api/src/default_context.rs
```diff
@@ -31,6 +31,8 @@ impl Spec for DefaultContext {
 
 #[cfg(feature = "native")]
 impl Context for DefaultContext {
+    type GasUnit = TupleGasUnit<1>;
```
bkolad: It should be `TupleGasUnit<2>` the same as for the `ZkContext` it is because we need to charge for both executions.

vlopes11: Makes sense, fixed!

PR #798, opened 2023-09-05
EVM: migrate to reth types

bkolad:
# Description
1. Removes `ethers` types from `query.rs` in the  `EVM` module
2. Implements `JsonCodec`

## Linked Issues
- Related to #413

## Testing
CI Passes.

## Docs
No doc changes.


module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -88,14 +88,43 @@ mod experimental {
...
+        // TODO JsonCodec: This is a workaround for https://github.com/bincode-org/bincode/issues/245 which affects all
+        // binary serialization formats.
+        // 1. Implement custom types for Block, Transaction etc.. with borsh derived.
+        // 2. Remove JsonCodec.
         #[state]
-        pub(crate) receipts: sov_state::StateMap<
-            ethereum_types::H256,
-            TransactionReceipt,
-            sov_state::codec::BcsCodec,
+        pub(crate) current_block: sov_state::StateValue<reth_rpc_types::Block, JsonCodec>,
```
bkolad, line 98: I think we should keep using `block_env` instead of `current_block` and just create a block in the `end_slot_hook`

LukaszRozmej: if we add pending_transactions, then just Header would suffice. Or if we want to accumulate transactions in it than full block (or our own type containing header and transactions)

module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -29,52 +28,54 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+            transaction_index: Some(reth_primitives::U256::from(0)),
             // TODO https://github.com/Sovereign-Labs/sovereign-sdk/issues/504
             block_hash: Default::default(),
             // TODO https://github.com/Sovereign-Labs/sovereign-sdk/issues/504
-            block_number: Some(0u64.into()),
-            from: evm_tx_recovered.signer().into(),
-            to: evm_tx_recovered.to().map(|to| to.into()),
-            // TODO https://github.com/Sovereign-Labs/sovereign-sdk/issues/504
-            cumulative_gas_used: Default::default(),
+            block_number: Some(reth_primitives::U256::from(0)),
```
LukaszRozmej, line 58: I know it is placeholder but there is a `reth_primitives::U256::ZERO`

LukaszRozmej (approved):
Approved overall to move forward.

Open questions, do we need those types at all:
a) EvmTransactionSignedEcRecovered
b) BlockEnv - IMO non-sealead header from reth is good enough?

module-system/module-implementations/sov-evm/src/call.rs
bkolad: Discussed offline we will replace it with a real value

PR #799, opened 2023-09-05
Expose `ModuleInfo::prefix(&self)`

preston-evans98:
# Description
IBC needs to be able to identify the prefix where a module's data lives in the state trie. This PR exposes the necessary machinery.

### DIFF
--- a/module-system/sov-modules-api/src/lib.rs
+++ b/module-system/sov-modules-api/src/lib.rs
@@ -348,7 +348,7 @@ pub trait ModuleInfo {
     /// Returns address of the module.
     fn address(&self) -> &<Self::Context as Spec>::Address;
 
-    /// Returns address of the module.
+    /// Returns the prefix of the module.
     fn prefix(&self) -> Prefix;
 
     /// Returns addresses of all the other modules this module is dependent on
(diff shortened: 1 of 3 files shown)


module-system/sov-modules-api/src/lib.rs
```diff
@@ -348,6 +348,9 @@ pub trait ModuleInfo {
     /// Returns address of the module.
     fn address(&self) -> &<Self::Context as Spec>::Address;
 
+    /// Returns address of the module.
```
bkolad: Should be "Returns prefix"

bkolad: Do we know why they need it?

bkolad:
Do we know why IBC needs it?

bkolad (approved):

module-system/sov-modules-api/src/lib.rs
preston-evans98: Fixed!

preston-evans98:
> Do we know why IBC needs it?

Yeah, good question. The way IBC works is that messages from chain A to chain B get stored in a pre-defined region of the state trie of chain A. To receive messages, chain B checks a merkle proof that the message is at the expected path in A's state trie. But, IBC is meant to be pretty flexible - so chains can store the messages in whatever prefix of the state trie they want, and there's a "handshake" when you open a new connection where you tell the other chain your prefix.

PR #804, opened 2023-09-06
Bug fix: StateMap::Keys are not consistent across platforms

preston-evans98:
# Description
This PR removes our StateMap key encoding based on std::Hash, which was not consistent across platforms. This caused the demo-prover to fail, since the generated witness used different storage keys than the guest execution. To fix this issue without sacrificing flexibility. PR introduces the `EncodeLike<Ref, Target>` trait which marks that Ref can be encoded like Target by the implementing codec.

This PR also removes the SingletonKey type, which required special handling in codecs. Instead, of using this placeholder, this PR implements auxiliary methods on the working set for dealing with singletons.

### Update: September 07, 2023
After discussion with @neysofu, we decided that it was desirable to allow implementers to use different codecs for keys and values.

The new design is as follows:

```rust
pub trait StateCodec {
    type KeyCodec;
    type ValueCodec;

    fn key_codec(&self) -> &Self::KeyCodec;
    fn value_codec(&self) -> &Self::ValueCodec;
}

pub trait StateKeyCodec<K> {
    fn encode_key(&self, key: &K) -> Vec<u8>;
}

pub trait StateValueCodec<V> {
    type Error: std::fmt::Debug;
    fn encode_value(&self, value: &V) -> Vec<u8>;
    fn try_decode_value(&self, bytes: &[u8]) -> Result<V, Self::Error>;
}

pub trait EncodeKeyLike<Ref: ?Sized, Target> {
    fn encode_key_like(&self, borrowed: &Ref) -> Vec<u8>;
}
```

All three of our state storage types (`StateMap`, `StateKey`, and `StateVec`) are defined with an associated `StateCodec`, which delegates to its own `KeyCodec` and `ValueCodec` as appropriate. This allows us to avoid introducing an extra generic in the `State*` types, while allowing separate key and value codecs. 

This PR also introduces the `SplitCodec<KC, VC>` type which allows anyone to combine a `KeyCodec` and `ValueCodec` into a `StateCodec`. The current trait design allows this `SplitCodec` to directly inherit the functionality of its associated key and value codecs rather. Without the `StateCodec` trait, this would be not be feasible. Instead, `SplitCodec` would have to try to blanket impl the `KeyCodec` and `ValueCodec` traits, which is not always possible without specialization.

### Future Work
This design leaves room for an additional `EncodeValueLike` trait to be implemented in the future if desired. Such a trait would could allow the storage of borrowed types, potentially removing some unnecessary `clone`s.

## Linked Issues
- Related to # (issue) 

## Testing
I ran the demo-rollup before on the same input before and after introducing this fix: 

### Before
``` bash
2023-09-06T22:34:33.069691Z  INFO sov_demo_prover_host: Completed proving and verifying block 2
2023-09-06T22:34:33.069709Z  INFO sov_demo_prover_host: Requesting data for height 3 and prev_state_root 0xa5ee72743963a012e754215fbd135ccd4d1a75915fb7044d605df6fa305c4926
2023-09-06T22:34:33.087281Z  INFO sov_demo_prover_host: Extracted 1 relevant blobs at height 3 header 0x1e90259c5a48c6dd20f1dd190bc21381e0ba8e9a26f833ec5d2845a79843f059
2023-09-06T22:34:33.087338Z  INFO sov_modules_stf_template: Selected 1 blob(s) for execution in current slot
2023-09-06T22:34:33.087433Z  INFO sov_modules_stf_template: blob #0 from sequencer celestia1w7wcupk5gswj25c0khnkey5fwmlndx6t5aarmk with blob_hash 0x24939693604755bf82bcdcda7682be007e639899b82341a98a61ab3077468f6e has been applied with #1 transactions, sequencer outcome Rewarded(0)
2023-09-06T22:34:33.087438Z  INFO sov_modules_stf_template: tx #0 hash: 0x1d0d746984613c0874fd2515d5de82e717c3705284f00ea8fecda756feadabed result Successful

Start guest
Prev root hash read
header has been read
inclusion proof has been read
!completeness proof has been read
blobs have been read
Witness have been read
Applying slot...
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `252`: Batch deserialization failed and some data was not provided. The prover might be malicious', ./src/app_template.rs:270:17
thread 'main' panicked at 'Prover should run successfully: illegal instruction at 0x839e4: 11000000000000000001000001110011 (1110011, 001, 00000, 1100000)', host/src/main.rs:142:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### After
```bash
2023-09-06T22:40:18.793261Z  INFO sov_demo_prover_host: Completed proving and verifying block 2
2023-09-06T22:40:18.793279Z  INFO sov_demo_prover_host: Requesting data for height 3 and prev_state_root 0x15f8d68a89fd36c8dd22058a1a0becf12f27791c74a59a6fdbb55bcc533837f3
2023-09-06T22:40:18.813870Z  INFO sov_demo_prover_host: Extracted 1 relevant blobs at height 3 header 0x1e90259c5a48c6dd20f1dd190bc21381e0ba8e9a26f833ec5d2845a79843f059
2023-09-06T22:40:18.813901Z  INFO sov_modules_stf_template: Selected 1 blob(s) for execution in current slot
2023-09-06T22:40:18.813989Z  INFO sov_modules_stf_template: blob #0 from sequencer celestia1w7wcupk5gswj25c0khnkey5fwmlndx6t5aarmk with blob_hash 0x24939693604755bf82bcdcda7682be007e639899b82341a98a61ab3077468f6e has been applied with #1 transactions, sequencer outcome Rewarded(0)
2023-09-06T22:40:18.813994Z  INFO sov_modules_stf_template: tx #0 hash: 0x1d0d746984613c0874fd2515d5de82e717c3705284f00ea8fecda756feadabed result Successful

Start guest
Prev root hash read
header has been read
inclusion proof has been read
!completeness proof has been read
blobs have been read
Relevant txs verified
Witness have been read
Applying slot...
Slot has been applied
new state root committed
```
## Docs


module-system/sov-state/src/codec/mod.rs
```diff
@@ -42,3 +43,38 @@ pub trait StateValueCodec<V> {
...
+        self.encode_value(borrowed)
+    }
+}
+
+// In borsh, a slice is encoded the same way as a vector
+impl<T> EncodeLike<[T], Vec<T>> for BorshCodec
+where
+    T: BorshSerialize,
+{
+    fn encode_like(&self, borrowed: &[T]) -> Vec<u8> {
```
vlopes11: We probably shouldn't make bigger changes now - especially under the bugfix context - but an alternative that might require less cloning is if we define a serializing interface, such as:

```rust
fn encode_like<S, B>(&self, serializer: &mut S, value: B)
where
  S: Serializer,
  B: AsRef<[T]>
```

This way we avoid forcing owned vecs and reduce allocation

module-system/sov-state/src/codec/mod.rs
```diff
@@ -42,3 +43,38 @@ pub trait StateValueCodec<V> {
...
+        borrowed.try_to_vec().unwrap()
```
vlopes11, line 96: I'm a bit uncomfortable with implicit unwrap. I understand that is a restriction from borsh (i.e. their serialization is fallible, when it probably shouldn't be as there is no I/O - except when we run out of memory and we can't alloc, but in that case it is not recoverable error so we should panic anyway), but then a type implementation that for some reason could be fallible would cause runtime panic.

Should we consider, in the future, wrapping the borsh traits? Something like

```rust
pub trait Serializable: BorshSerializable {
  fn to_vec(&self) -> Vec<u8> {
    borrowed.try_to_vec().unwrap()
  }
}
```

This way, we put everything into the same place so we make it easier to maintain in the future as if we ever have to change that, we just change the trait.

vlopes11 (approved):
Looks good! I left two comments that are more potential future improvements than something to be done now

bkolad (approved):

module-system/sov-state/src/codec/mod.rs
neysofu: (in response to @vlopes11)

The `Serializer`/`Writer` pattern is interesting, it might be a good fit here and reduce cloning and `Vec<u8>` allocations. I think we should keep it in mind for future performance optimizations.

As for `AsRef<[T]>` (why `[T]` and not `T`? That's not clear to me) we have to be really careful not to run into [these](https://github.com/Sovereign-Labs/sovereign-sdk/pull/644#discussion_r1294482223) problems, which apply to `AsRef` just as much as `Borrow`: https://doc.rust-lang.org/src/core/array/mod.rs.html#159

module-system/sov-state/src/codec/mod.rs
```diff
@@ -42,3 +43,38 @@ pub trait StateValueCodec<V> {
...
+where
+    C: StateValueCodec<T>,
+{
+    fn encode_like(&self, borrowed: &T) -> Vec<u8> {
+        self.encode_value(borrowed)
+    }
+}
+
+// In borsh, a slice is encoded the same way as a vector
+impl<T> EncodeLike<[T], Vec<T>> for BorshCodec
```
neysofu: I'm okay with this, but let's document here that vectors and slices don't have *exactly* the same serialization logic starting from Borsh 1.0. Borsh 1.0 disallows zero-sized element types for all collections including `Vec`, but slices are exempt.

https://github.com/near/borsh-rs/blob/09c9295000ced1933e067d094a3cd8fe5507044d/borsh/src/ser/mod.rs#L283

Originally noticed by @bkolad here: https://github.com/Sovereign-Labs/sovereign-sdk/pull/689#discussion_r1301253205

I think we can consider it almost a pathological edge case and still implement the trait.

module-system/sov-state/src/codec/mod.rs
```diff
@@ -42,3 +43,38 @@ pub trait StateValueCodec<V> {
             .unwrap()
     }
 }
+
+pub trait EncodeLike<Ref: ?Sized, Target> {
+    fn encode_like(&self, borrowed: &Ref) -> Vec<u8>;
```
neysofu: I'm wondering if `EncodeLike` should be a marker trait instead. If both `Ref` and `Target` serialize the same way, why is a third method `encode_like` needed? Making this is a marker trait would then require `EncodeLike<..., ...> + StateValueCodec<...>` (or whatever codec trait we must use) in all places where we'll need to actually encode types. So it's more verbose.

This is non-blocking, feel free not to address this suggestion if you prefer the current design.

module-system/sov-state/src/containers/accessory_map.rs
```diff
@@ -57,8 +56,8 @@ where
     /// map’s key type.
     pub fn set<Q, S: Storage>(&self, key: &Q, value: &V, working_set: &mut AccessoryWorkingSet<S>)
     where
-        K: Borrow<Q>,
-        Q: Hash + Eq + ?Sized,
+        VC: EncodeLike<Q, K>,
```
neysofu: The type `VC` is a `StateValueCodec`, but here (and in other places) we're using it for keys instead. I see two problems with this:

1. `StateValueCodec` is only intended to (de)serialize values, and not keys, so it's a misnomer at this point.
2. Besides the misnomer nit, using the same (de)serialization methods for keys and values doesn't allow codec authors to use different encodings for keys and values, which is an important feature we want to preserve.

I see two main solutions:

- Two different traits, `StateKeyCodec` and `StateValueCodec` (similar to what we had before moving to `Hash` for keys).
- A single codec trait `StateCodec` with different methods for keys and values.

neysofu (changes requested):
Looks good, but it also removes support for having different codecs for keys and values by using `StateValueCodec` everywhere, which can't differentiate between keys and values.

I left a few other minor comments, otherwise LGTM!

module-system/sov-state/src/codec/mod.rs
preston-evans98: Good question!

The motivation for this is that a type might have multiple encodings. For example `[u8;N]` should not ordinarily be encoded with a `length`, but it *can* be encoded that way. In this design, a StateCodec<T> specifies the canonical encoding for a type, and `EncodeLike<T, U>` specifies that T can be encoded using a (potentially different) format which is guaranteed to match the canonical format of `U`.

module-system/sov-state/src/containers/accessory_map.rs
preston-evans98: Hmm, interesting. Do we have a motivating use case for separating the encodings of keys and values?

module-system/sov-state/src/codec/mod.rs
neysofu: Ah, makes sense! Big brain stuff.

module-system/sov-state/src/containers/accessory_map.rs
neysofu: Just saw this comment, sorry. Responded on Slack.

preston-evans98: Added!

module-system/sov-state/src/codec/mod.rs
preston-evans98: Done!

module-system/sov-state/src/codec/mod.rs
preston-evans98: This is a good idea, but let's leave it for a follow-up since it's not on the critical path.

module-system/sov-state/src/codec/mod.rs
preston-evans98: I think borsh fixed this in the 1.0 release. We'll take care of these when we migrate.

neysofu (approved):

PR #817, opened 2023-09-07
Release `v0.2.0-alpha`

neysofu:
## Description
This PR makes the following changes:
- Updates a few outdated README tutorials which contained the wrong commands to run.
- Revamps the release process docs to make them more comprehensive and more detailed. Feedback is greatly appreciated here, I'm sure I've missed something.
- Fixes all Cargo manifests of crates that couldn't be published due to issues in their manifests.
- Updates `packages_to_publish.txt`.
- Finally, bumps the version number of all crates to 0.2.0 in preparation for the `v0.2.0-alpha` release.

## Testing
I've tested the release steps locally as much as I could. There's a good chance we'll still find problems along the way, as this is our first release in quite some time.

There is room for improvement in CI to make sure that crates are ready to be published at all times. Commit https://github.com/Sovereign-Labs/sovereign-sdk/commit/b71151461db1ff3af3605a4ae74f166c8287bf4b includes a fix for Cargo manifests that lacked version numbers for some dependencies, which caused errors when I run `cargo publish --dry-run -p ...`. Short of running `cargo publish --dry-run` on CI, I'm not sure there's a way for us to catch those beforehand. I'll try and think of something.

## Docs
I've updated `Releases.md`, `packages_to_publish.txt`, and a few outdated READMEs.

PR #818, opened 2023-09-07
Weekly CI pre-release check

neysofu:
# Description
`cargo publish --dry-run` doesn't benefit from caching and tends to be quite fragile, so it's a poor candidate for CI. There's one specific check that we'd like to run periodically, though: whether a crate can't be released because we haven't properly specified its dependencies' versions.


neysofu:
This check currently fails on `nightly`, fix is in #817.

.github/workflows/cleanup_caches.yml
neysofu, line 1: We don't create new caches for feature branches anymore, only on `nightly`. This workflow is not useful anymore.

.github/workflows/check_missing_dependency_versions.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: buildjet-4vcpu-ubuntu-2204
+    steps:
+      - uses: actions/checkout@v3
+      - run: |
+          while read -r line; do
+            # Skip empty lines and comments.
+            [[ -z "$line" || "$line" =~ ^# ]] && continue
+
+            echo "Checking crate $line..."
+            output=$(cargo publish --allow-dirty --quiet --dry-run -p "$line" 2>&1)
```
vlopes11: It's unfortunate we don't have the `--message-format=JSON` available for `cargo publish`, so it seems the only option is indeed to check stdout.

.github/workflows/check_missing_dependency_versions.yml
```diff
@@ -0,0 +1,37 @@
...
+            #
+            # `cargo publish` may fail for many other reasons, but it'd be too
+            # hard to reason about them all. This is the only one we can expect
+            # to catch reliably.
+            echo "$output" | grep -q "all dependencies must have a version specified when publishing."
+            if [ $? -eq 0 ]; then
+                echo "Error: Found problematic output for crate $line."
+                exit 1
+            fi
+          done < packages_to_publish.txt
```
vlopes11: One alternative is to fetch the packages that are not publishable automatically instead of relying on the file. That way, we also force all the packages that are not expected to be published to have a `publish = false`. The following command would do the job

```sh
grep -riLE 'publish\s*=\s*false' **/Cargo.toml | xargs sed -n 's/name\s*=\s*"\(.*\)".*/\1/p'
```

vlopes11 (approved):
Looks good! Two optional discussions

.github/workflows/check_missing_dependency_versions.yml
neysofu: Agreed! It would be nice to have it.

.github/workflows/check_missing_dependency_versions.yml
neysofu: I love the idea. What do you think about turning this into a dedicated job, and checking that all and only the crates with `publish = true` are present in the `packages_to_publish.txt` file, so we know it's up to date for sure?

neysofu: One-liner to get all publishable crates: `cargo metadata --format-version 1 --no-deps | jq '.packages[] | select(.publish == null) | .name'`. It's similar to what you shared, but it uses `cargo metadata` output and `jq` so it also supports workspace inheritance logic.

I'll try putting together a GH Actions job!

neysofu:
Blocked by #817.

.github/workflows/check_missing_dependency_versions.yml
neysofu: Addressed with the new pre-release check scripts!

PR #824, opened 2023-09-08
Prepend `sov-` prefix to all releasable crates

### LINKED ISSUE #819: Rename crates to avoid name conflicts when publishing to https://crates.io
## Description

Most of our crate names are prefixes by `sov-*` which makes name conflicts quite unlikely, but there's a few outliers and we should rename them to maintain consistency across the SDK:

- `celestia` -> `sov-celestia-adapter`
- `risc0-adapter` -> `sov-risc0-adapter`
- `zk-cycle-macros/utils` -> `sov-zk-cycle-macros/utils`

Context: https://github.com/Sovereign-Labs/nmt-rs/issues/17

neysofu:
Fixes #819.


bkolad (approved):
LGTM, question do we need to change any docs in readmes?

neysofu:
@bkolad I modified the docs wherever I could find references to the old names – I wouldn't be surprised if I missed some though, I'll check again before merging.

PR #829, opened 2023-09-08
Fix benchmark genesis address to match block production address

dubbelosix:
### Summary
This PR addresses an issue with our benchmarking script that used a sequencer address without first registering it with the rollup. This caused all blocks to have their transactions skipped, leading to misleadingly high throughput measurements.

### Changes
1. **Register correct address**: Now the RNG block generator used in benchmarking produces blocks that match the sequencer address registered during genesis. This ensures the benchmark properly processes the blocks.
2. **Benchmark Output**: A new row has been added to the benchmark output that shows the number of successfully processed transactions. This serves as an additional layer of verification for the user

### Root Cause
- The benchmark script and the main code were using different types (and addresses) for sequencer addresses in genesis.
- Type checking was not in place, allowing the discrepancy to go unnoticed.
- Benchmark script was using CelestiaAddress in the genesis and rng_xfers.rs was using a default MockAddress [99u8;32] in the headers for the blocks it produced

### Test Plan
1. Run the benchmark script to ensure that it no longer rejects all blocks.
2. Verify that the throughput measurement is now accurate/reasonable.
3. Confirm that the new row in the benchmark output shows the correct number of successfully processed transactions.

### DIFF
--- a/examples/demo-rollup/benches/rollup_coarse_measure.rs
+++ b/examples/demo-rollup/benches/rollup_coarse_measure.rs
@@ -117,7 +117,7 @@ async fn main() -> Result<(), anyhow::Error> {
     // data generation
     let mut blobs = vec![];
     let mut blocks = vec![];
-    for height in start_height..end_height + 1 {
+    for height in start_height..=end_height {
         let num_bytes = height.to_le_bytes();
         let mut barray = [0u8; 32];
         barray[..num_bytes.len()].copy_from_slice(&num_bytes);
@@ -139,7 +139,7 @@ async fn main() -> Result<(), anyhow::Error> {
     // rollup processing
     let total = Instant::now();
     let mut apply_block_time = Duration::new(0, 0);
-    for height in start_height..end_height + 1 {
+    for height in start_height..=end_height {
         let filtered_block = &blocks[height as usize];
 
         let mut data_to_commit = SlotCommit::new(filtered_block.clone());
@@ -168,7 +168,13 @@ async fn main() -> Result<(), anyhow::Error> {
 
     let total = total.elapsed();
     if timer_output {
-        print_times(total, apply_block_time, end_height, num_txns_per_block,num_success_txns);
+        print_times(
+            total,
+            apply_block_time,
+            end_height,
+            num_txns_per_block,
+            num_success_txns,
+        );
     }
     if prometheus_output {
         println!("{:#?}", registry.gather());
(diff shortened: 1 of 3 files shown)


examples/demo-rollup/benches/rollup_coarse_measure.rs
```diff
@@ -107,8 +117,7 @@ async fn main() -> Result<(), anyhow::Error> {
     // data generation
     let mut blobs = vec![];
     let mut blocks = vec![];
-
-    for height in start_height..end_height {
+    for height in start_height..end_height + 1 {
```
preston-evans98: nit: This would be slightly clearer if we used `..=` instead of `..` and `+1`

examples/demo-rollup/benches/rollup_coarse_measure.rs
```diff
@@ -26,12 +23,24 @@ use tempfile::TempDir;
...
+    apply_block_time: Duration,
+    blocks: u64,
+    num_txns: u64,
+    num_success_txns: u64,
+) {
     let mut table = Table::new();
 
     table.add_row(row!["Blocks", format!("{:?}", blocks)]);
     table.add_row(row!["Txns per Block", format!("{:?}", num_txns)]);
+    // subtract 1 because the first block contains only 1 txn which is the token create txn
```
preston-evans98: Not sure I follow this comment. Why do we exclude the first block?

examples/demo-rollup/benches/rollup_coarse_measure.rs
```diff
@@ -130,7 +139,7 @@ async fn main() -> Result<(), anyhow::Error> {
     // rollup processing
     let total = Instant::now();
     let mut apply_block_time = Duration::new(0, 0);
-    for height in start_height..end_height {
+    for height in start_height..end_height + 1 {
```
preston-evans98: nit: This would be slightly clearer if we used `..=` as well

examples/demo-rollup/benches/rollup_coarse_measure.rs
dubbelosix: ah because the first block of the rng da is a block that includes only 1 transaction to create the token. the rest are as per the bench spec - 10k transfers

dubbelosix: As per discussion, I cleaned up this logic to completely exclude block 1 and the first token creation transaction, so we don't need to -1 in the counts and we also have more accurate timing

preston-evans98 (approved):

Issue #640, opened 2023-08-10
EVM implement transaction validation logic in the EVM module.


bkolad:
## Description
Implement transaction validation logic in the EVM module: relevant code:
`reth`:  [code](https://github.com/paradigmxyz/reth/blob/499ca3d15378d2aff13d299ff4967fcab7b009a3/crates/consensus/common/src/validation.rs#L399-L410)
`nethermind`:[code](https://github.com/NethermindEth/nethermind/blob/5ab59cc2a4694eefebc6b6d16007541353df52d0/src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs#L36-L43)

The signer recovery was already done in #639
Invalid transactions should be skipped. 

This is a low-priority task.

Issue #504, opened 2023-07-17
EVM: `getTransactionReceipt` should return fully initialized `TransactionReceipt` 


bkolad:
## Description
Some fields in the `TransactionReceipt` are hardcoded with constant values. We should update these fields so that they reflect the actual values obtained from the EVM execution.

## Acceptance Criteria
1. Relevant TODOs removed (unless they are related to [this issue](#737))
2. Code tested with unit tests.

bkolad:
All the block-related data will be added after https://github.com/Sovereign-Labs/sovereign-sdk/issues/737 is done.

Issue #503, opened 2023-07-17
EVM: Initialize `EvmTransaction`


bkolad:
## Description
Ensure that all the required fields are initialized in the `EvmTransaction` object. Currently, some fields still have `TODO` comments indicating that they have been left out.

## Acceptance Criteria
1. All relevant TODOS removed (unless they are related to [this issue](#737))
2. Code tested with unit tests

bkolad:
All the block-related data will be added after https://github.com/Sovereign-Labs/sovereign-sdk/issues/737 is done.

bkolad:
@LukaszRozmej? can we close it?

Issue #502, opened 2023-07-17
EVM: Implement basic Ethereum JSON-RPC endpoints.


bkolad:
## Description
Here is the comprehensive list of Ethereum JSON-RPC endpoints: [link to the documentation](https://ethereum.org/en/developers/docs/apis/json-rpc/). While it would be ideal to support all of them, we need to prioritize the most useful ones initially. Here is the prioritized list of endpoints (needed to deploy `Uniswap-v2`) that need to be implemented first:

- [x] eth_chainId #877
- [x] eth_getBlockByNumber  #878
- [x] eth_getTransactionByHash
- [x] eth_getTransactionReceipt
- [x] eth_call #879
- [x] eth_sendTransaction #880 (done for the `latest` block)
- [x] eth_blockNumber  #881
- [x] eth_getTransactionCount #882
- [ ] eth_gasPrice #883
- [ ] eth_feeHistory

## Acceptance Criteria
Unit test for each endpoint.

PR #830, opened 2023-09-09
EVM: Storage improvements , Blocks, Transactions, Receipts

### LINKED ISSUE #737: EVM: introduce tx-block that is compatible with the rollup batching meechanism.

LukaszRozmej:
# Description
- Implements `DoubleEndedIterator` and `last` on `AccessoryStateVec` and `StateVec`.
- Splits `end_slot_hook` into `end_slot_hook` (zk-context) and `finalize_slot_hook` (only non-zk-context)
- In `sov-evm` module:
  - `evm` state fields are properly split between `state` and `AccessoryState` types. Using Reth domain types where possible. Enclosing it in our types for `Block`, `Transaction` and `Receipt`. All serialized with binary serialization.
  - `execute_call` now properly creates pending transaction and receipt (for both successful and failed transactions)
  - `begin_slot_hook` properly creates `pending_block`
  - `end_slot_hook` creates final block and sets `head`, moves transactions from `pending_transactions` to `transactions` and `receipts`.
  - `finalize_slot_hook` - finalizes head block and moves it to `blocks` for RPC
  - Implements `eth_getTransactionByHash` and `eth_getTransactionReceipt` with new model

## Linked Issues
- Fixes #737, #504, #503 
- Related to #640, #502, #413

## Testing
Unit test + existing integration tests

### DIFF
--- a/examples/demo-rollup/tests/evm/mod.rs
+++ b/examples/demo-rollup/tests/evm/mod.rs
@@ -192,6 +192,7 @@ async fn send_tx_test_to_eth(rpc_address: SocketAddr) -> Result<(), Box<dyn std:
     test_client.execute().await
 }
 
+#[cfg(feature = "experimental")]
 #[tokio::test]
 async fn evm_tx_tests() -> Result<(), anyhow::Error> {
     let (port_tx, port_rx) = tokio::sync::oneshot::channel();
--- a/module-system/module-implementations/sov-evm/src/call.rs
+++ b/module-system/module-implementations/sov-evm/src/call.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use reth_primitives::TransactionSignedEcRecovered;
 use reth_revm::into_reth_log;
-use revm::primitives::{CfgEnv, SpecId};
+use revm::primitives::{CfgEnv, EVMError, SpecId};
 use sov_modules_api::CallResponse;
 use sov_state::WorkingSet;
 
@@ -40,39 +40,47 @@ impl<C: sov_modules_api::Context> Evm<C> {
 
         let evm_db: EvmDb<'_, C> = self.get_db(working_set);
         let result = executor::execute_tx(evm_db, &block_env, &evm_tx_recovered, cfg_env);
+        let previous_transaction = self.pending_transactions.last(working_set);
+        let previous_transaction_cumulative_gas_used = previous_transaction
+            .as_ref()
+            .map_or(0u64, |tx| tx.receipt.receipt.cumulative_gas_used);
+        let log_index_start = previous_transaction.as_ref().map_or(0u64, |tx| {
+            tx.receipt.log_index_start + tx.receipt.receipt.logs.len() as u64
+        });
 
         let receipt = match result {
             Ok(result) => {
                 let logs: Vec<_> = result.logs().into_iter().map(into_reth_log).collect();
-                let previous_transaction = self.pending_transactions.last(working_set);
+                let gas_used = result.gas_used();
 
                 Receipt {
                     receipt: reth_primitives::Receipt {
                         tx_type: evm_tx_recovered.tx_type(),
                         success: result.is_success(),
-                        cumulative_gas_used: previous_transaction.as_ref().map_or(0u64, |tx| {
-                            tx.receipt.receipt.cumulative_gas_used + result.gas_used()
-                        }),
+                        cumulative_gas_used: previous_transaction_cumulative_gas_used + gas_used,
                         logs,
                     },
-                    gas_used: result.gas_used(),
-                    log_index_start: match previous_transaction {
-                        Some(tx) => {
-                            tx.receipt.log_index_start + tx.receipt.receipt.logs.len() as u64
-                        }
-                        None => 0u64,
-                    },
+                    gas_used,
+                    log_index_start,
+                    error: None,
                 }
             }
-            Err(err) => todo!(
-                "{}",
-                match err {
-                    revm::primitives::EVMError::Transaction(error) =>
-                        serde_json::to_string(&error).unwrap(),
-                    revm::primitives::EVMError::PrevrandaoNotSet => "PrevrandaoNotSet".to_string(),
-                    revm::primitives::EVMError::Database(_) => "DB".to_string(),
-                }
-            ),
+            Err(err) => Receipt {
+                receipt: reth_primitives::Receipt {
+                    tx_type: evm_tx_recovered.tx_type(),
+                    success: false,
+                    cumulative_gas_used: previous_transaction_cumulative_gas_used,
+                    logs: vec![],
+                },
+                // TODO: Do we want failed transactions to use all gas?
+                gas_used: 0,
+                log_index_start: log_index_start,
+                error: Some(match err {
+                    EVMError::Transaction(err) => EVMError::Transaction(err),
+                    EVMError::PrevrandaoNotSet => EVMError::PrevrandaoNotSet,
+                    EVMError::Database(_) => EVMError::Database(0u8),
+                }),
+            },
         };
 
         let pending_transaction = PendingTransaction {
--- a/module-system/module-implementations/sov-evm/src/call.rs
+++ b/module-system/module-implementations/sov-evm/src/call.rs
@@ -1,7 +1,5 @@
 use anyhow::Result;
-use ethereum_types::U64;
-use reth_primitives::bloom::logs_bloom;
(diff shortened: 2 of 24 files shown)


module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -29,53 +33,93 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                    transaction_type: transaction
+                        .transaction_type
+                        .unwrap()
+                        .as_u64()
+                        .try_into()
+                        .unwrap(),
+                    logs_bloom: logs_bloom(logs.iter()),
+                    logs: {
+                        let mut log_index = 0;
+                        logs.into_iter()
```
bkolad: 1. Let's use `logs.into_iter().enumerate()` and remove explicit index tracking and move the code to a separate function.

bkolad (approved):

module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -29,53 +30,75 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                let mut accessory_state = working_set.accessory_state();
+                let tx_count = self.pending_transactions.len(&mut accessory_state);
 
-        self.pending_transactions
-            .push(&transaction, &mut working_set.accessory_state());
+                let previous_transaction: Option<PendingTransaction> = if tx_count == 0 {
+                    None
+                } else {
+                    Some(
+                        self.pending_transactions
```
bkolad: It looks like we should add `last ` method to `AccessoryStateVec & StateVec` and use it here instead of `tx_count`.  Can be done in another PR.

module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -29,53 +30,75 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                    log_index_start: match &previous_transaction {
+                        Some(tx) => {
+                            tx.receipt.log_index_start + tx.receipt.receipt.logs.len() as u64
+                        }
+                        None => 0u64,
+                    },
+                }
+            }
+            Err(err) => todo!(
+                "{}",
```
bkolad: Is this todo! needed?

module-system/module-implementations/sov-evm/src/evm/transaction.rs
```diff
@@ -29,11 +31,51 @@ impl Default for BlockEnv {
...
+}
+
+#[cfg_attr(
+    feature = "native",
+    derive(serde::Serialize),
+    derive(serde::Deserialize)
+)]
+#[derive(Debug, PartialEq, Clone)]
+pub struct Receipt {
+    pub receipt: reth_primitives::Receipt,
```
bkolad: Does `Receipt ` and all the fields need to be public? The same for `Block`? Can we make it private or `pub (crate)`?

module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -86,23 +91,67 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+
+        let tx_number = self.transaction_hashes.get(&hash, &mut accessory_state);
+
+        let transaction = tx_number
+            .map(|number| {
+                self.transactions
+                    .get(number as usize, &mut accessory_state)
+                    .expect("Transaction with known hash must be set")
+            })
+            .map(|tx| {
```
bkolad: The two `map`s can be collapsed into one.

module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -86,23 +91,67 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+
+        let tx_number = self.transaction_hashes.get(&hash, &mut accessory_state);
+
+        let receipt = tx_number
+            .map(|number| {
+                self.transactions
+                    .get(number as usize, &mut accessory_state)
+                    .expect("Transaction with known hash must be set")
+            })
+            .map(|tx| {
```
bkolad: we can collapse it into a single mapping.

module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -29,53 +30,75 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                            .expect("Pending transaction must be set"),
+                    )
+                };
+
+                Receipt {
+                    receipt: reth_primitives::Receipt {
+                        tx_type: evm_tx_recovered.tx_type(),
+                        success: result.is_success(),
+                        cumulative_gas_used: match &previous_transaction {
+                            Some(tx) => tx.receipt.receipt.cumulative_gas_used + result.gas_used(),
```
bkolad: What about:
```
previous_transaction.map_or(|tx| tx.receipt.receipt.cumulative_gas_used + result.gas_used(), 0u64);
```

module-system/module-implementations/sov-evm/src/call.rs
LukaszRozmej: this will be populated next, but wanted to output what was failing in tests

module-system/sov-modules-stf-template/src/lib.rs
```diff
@@ -94,17 +94,24 @@ where
...
 
         let (root_hash, authenticated_node_batch) = self
             .current_storage
             .compute_state_update(cache_log, &witness)
             .expect("jellyfish merkle tree update must succeed");
 
-        self.runtime.end_slot_hook(root_hash, &mut working_set);
+        let mut working_set = checkpoint.to_revertable();
+        self.runtime
+            .finalize_slot_hook([0; 32], &mut working_set.accessory_state());
```
LukaszRozmej, line 116: needs a root_hash

module-system/sov-modules-api/src/hooks.rs
```diff
@@ -64,7 +64,14 @@ pub trait SlotHooks<Da: DaSpec> {
 
     fn end_slot_hook(
         &self,
+        //TODO remove
```
LukaszRozmej: remove here and add previous hash in begin_slot

module-system/sov-modules-stf-template/src/lib.rs
bkolad: the `root_hash` is defined a few line above so we can just pass it.

examples/demo-rollup/tests/evm/mod.rs
```diff
@@ -203,6 +203,7 @@ async fn send_tx_test_to_eth(rpc_address: SocketAddr) -> Result<(), Box<dyn std:
     test_client.execute().await
 }
 
+#[cfg(feature = "experimental")]
```
bkolad, line 206: why do we need it now?

LukaszRozmej: we can remove it, but I think it hid some dependency problems in CI

LukaszRozmej: Otherwise this test can be run without that flag and fail in that context

module-system/module-implementations/sov-evm/src/tests/genesis_tests.rs
```diff
@@ -73,14 +80,57 @@ fn genesis_cfg_missing_specs() {
...
+                    transactions_root: EMPTY_TRANSACTIONS,
+                    receipts_root: EMPTY_RECEIPTS,
+                    logs_bloom: Bloom::default(),
+                    difficulty: U256::ZERO,
+                    number: 0,
+                    gas_limit: ETHEREUM_BLOCK_GAS_LIMIT,
+                    gas_used: 0,
+                    timestamp: 50,
+                    extra_data: Bytes::default(),
+                    mix_hash: H256::default(),
```
bkolad, line 112: Is this something for later. If yes can you please add //TODO EVM (so it is easy to grep) to all the places where we use default() temporarily.

examples/demo-rollup/tests/evm/mod.rs
LukaszRozmej: It helps tools to auto-add this flag when running this test only. As evm is also hidden by this flag, so this test doesn't test much without it. So you can see test green and have wrong idea that it works if run without this flag.

bkolad (approved):

PR #831, opened 2023-09-11
feat: downgrade WorkingSet to sov-modules-api

vlopes11:
`WorkingSet`, defined in `sov-state`, might have definitions specified under `Context`, that is defined in `sov-modules-api`.

Given `sov-state` is a dependency of `sov-modules-api`, this means `WorkingSet` must be moved to the latter; otherwise, such relationship will not be possible.

One example is the gas meter implementation that requires runtime gas price information that is best defined under `Context`. The gas will be deducted from the `WorkingSet`, but must be computed via the `Context`, making this relationship overly convoluted, unless `WorkingSet` contains a generic `Context`.

This commit solely moves the `WorkingSet` to the `sov-modules-api`, without making changes on its structures. A subsequent commit will introduce `Context` as generic attribute of `WorkingSet`.


examples/demo-nft-module/src/lib.rs
```diff
@@ -21,11 +20,11 @@ pub struct NonFungibleToken<C: Context> {
 
     #[state]
     /// Admin of the NonFungibleToken module.
-    admin: sov_state::StateValue<C::Address>,
+    admin: sov_modules_api::StateValue<C::Address>,
```
bkolad, line 23: I think `StateXX` should remain in `sov_state` if possible.

module-system/sov-modules-api/src/state/prover_storage.rs
```diff
@@ -7,13 +7,13 @@ use jmt::storage::{NodeBatch, TreeWriter};
 use jmt::{JellyfishMerkleTree, KeyHash, RootHash, Version};
```
bkolad: Is it possible to leave `storage.rs, prover_storage.rs, zk_storage.rs, tree_db.rs` in `sov-state`?  
It should be possible to import `sov-modules-api` as dev-dependency in `sov-state` if `WorkingSet` is needed for testing.

vlopes11: It's possible, yes, we just need to make a couple of private components public

vlopes11: I think the current change is on the minimal. I moved only the components that are required to live with the `WorkingSet`. The root of the problem is with the `StateMap`, that consumes the `WorkingSet`:

https://github.com/Sovereign-Labs/sovereign-sdk/blob/a1d6e7dac747288d478cdb4c390a7711cdf7c9b5/module-system/sov-state/src/containers/map.rs#L120-L128

And it is required by the `Storage` trait:

https://github.com/Sovereign-Labs/sovereign-sdk/blob/a1d6e7dac747288d478cdb4c390a7711cdf7c9b5/module-system/sov-state/src/storage.rs#L225-L232

And `Storage` is required by everything else that was moved.

vlopes11: We can open a tracking issue to revisit this in the future; wdyt? To be fair, it seems to me that `sov-state` is a module of `sov-modules-api`

bkolad: Thanks @vlopes11 yeah so we have circular dependency here.  To me the `sov-state` is a bit more lower level than the `sov-modules-api`  The good news is that `verify_proof` has fairly simple implementation and it is used only in one place. One solution would be to bring only the `verify_proof` to the `sov-module-api` but I agree it is out of scope of this PR.

bkolad (approved):
LGTM, thanks!

PR #842, opened 2023-09-12
Enable end-to-end integrations tests including prover->verifier workflow

preston-evans98:
# Description
This PR implements the `StateTransitionVerifier`, an a analogue to the `StateTransitionRunner` for use in zk. It uses the `StateTransitionVerifier` to implement an end-to-end integration test running both the prover and verifier workflows.


examples/demo-rollup/tests/all_tests.rs
```diff
@@ -1,3 +1,6 @@
+// For now, exclude the evm from the bank tests
+#[cfg(not(feature = "experimental"))]
 mod bank;
+#[cfg(feature = "experimental")]
```
bkolad, line 4: Why do we need to exclude the evm here?

bkolad (approved):

examples/demo-rollup/tests/all_tests.rs
preston-evans98: There was some odd interaction between EVM and non-evm code which caused genesis to fail when EVM was not feature gated here. I'll investigate this afternoon.

adapters/risc0/src/host.rs
```diff
@@ -9,50 +9,50 @@ use sov_rollup_interface::zk::{Zkvm, ZkvmHost};
...
-    #[cfg(not(feature = "bench"))]
-    pub fn new(elf: &'a [u8]) -> Self {
-        let default_env = ExecutorEnvBuilder::default();
-
-        Self {
-            env: RefCell::new(default_env),
-            elf,
-        }
-    }
+#[cfg(not(feature = "bench"))]
```
zvolin, line 22: it looks like a good candidate for `inline(always)`

examples/demo-rollup/tests/bank/mod.rs
```diff
@@ -67,14 +68,22 @@ async fn send_test_create_token_tx(rpc_address: SocketAddr) -> Result<(), anyhow
...
-        start_rollup(port_tx).await;
+        start_rollup(port_tx, Some(prover)).await;
     });
 
     // Wait for rollup task to start:
     let port = port_rx.await.unwrap();
 
-    send_test_create_token_tx(port).await?;
-    rollup_task.abort();
+    // If the rollup throws an error, stop trying to send the transaction and return it
```
zvolin, line 81:
```suggestion
    // If the rollup throws an error, return it and stop trying to send the transaction
```

A bit of a rewording there because I was confused where you 'return the transaction'

examples/demo-rollup/tests/bank/mod.rs
```diff
@@ -67,14 +68,22 @@ async fn send_test_create_token_tx(rpc_address: SocketAddr) -> Result<(), anyhow
...
+    tokio::select! {
+        err = rollup_task =>  {
+            err?;
+        },
+        res = send_test_create_token_tx(port) =>{res?},
```
zvolin, line 86:
```suggestion
        res = send_test_create_token_tx(port) => {
            res?;
        },
```
just for consistency with an arm above

preston-evans98:
Hey, thanks for the review @zvolin! Going to merge as-is because it's a big PR and getting CI to pass is a pain - but I'll include all of those suggestions in a fast follow 👍

PR #850, opened 2023-09-13
Create a more complete demo-nft-module with collections, metadata uri, freezing, updates, supply

dubbelosix:
## Overview
`demo-nft-module` has been enhanced to include the concept of Collections, NFTs and calls to create, modify, freeze collections and NFTs and to transfer NFTs. Existing `demo-nft-module` has been moved to `simple-nft-module` since it still serves as a very simply and easy to understand tutorial. `simple-nft-module` focuses more on the sovereign-sdk usage aspects, while `demo-nft-module` aims to be a more ready implementation to plug-n-play and customize if necessary

## Changes
- `demo-nft-module` has no genesis
- `README.md` added to explain the basic concepts, calls provided as well as example `sov-cli` commands to interact with the `demo-nft-module` on a running rollup
- `demo-nft-module` has also been added to runtime, so it would be a permanent module

## Further changes
The offchain component for indexing the NFTs into an offchain store will be part of a second PR.

## Test cases added
- Create Collection
- Mint NFT to created collection
- Mint NFT to non-existent collection
- Update a collection
- Freeze a non existent collection
- Freeze Collection
- Update collection uri for frozen collection
  - assert that the collection uri hasn't been changed
  - assert that supply hasn't been modified
- Supply change after minting an NFT to a Collection
- Minting to a Frozen Collection
  - ensure supply hasn't changed
- Transferring an NFT from a non-owner address
- Transferring NFT with non-existent token id
- Update NFT token_uri
- Freeze NFT
- Update NFT token uri for frozen NFT


Cargo.toml
```diff
@@ -9,6 +9,7 @@ members = [
     "examples/demo-simple-stf",
     "examples/demo-rollup",
     "examples/demo-nft-module",
+    "examples/simple-nft-module",
```
bkolad, line 11: Is the module  generic enough to be considered as a standard module? If yes then let's put it in `module-implementations`

examples/demo-nft-module/queries.md
```diff
@@ -0,0 +1,3 @@
+curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"nft_getCollectionAddress","params":["sov1l6n2cku82yfqld30lanm2nfw43n2auc8clw7r5u5m6s7p8jrm4zqrr8r94","Test Collection"],"id":1}' http://127.0.0.1:12345
```
bkolad: Let's explain what each of the command does. If we decide to make `demo-not-module` a standard module then we could integrate it in `demo-rollup`  the same way as for bank.

examples/demo-nft-module/src/call.rs
```diff
@@ -2,94 +2,314 @@ use anyhow::{bail, Result};
...
+
+        if let Some(mut c) = self.collections.get(&collection_address, working_set) {
+            let nft_identifier = NftIdentifier(token_id, collection_address.clone());
+            if c.frozen {
+                bail!(
+                    "Collection with name {} by sender {} is already frozen",
+                    collection_name,
+                    creator.to_string()
+                )
+            } else if self.nfts.get(&nft_identifier, working_set).is_some() {
```
bkolad: Nit I think we could avoid `3-branch if` in the following way:
1. The first ` if c.frozen` just early returns so we don'y need the second else_if
2. For the second and third else we use explicit pattern matching on Option.

examples/demo-nft-module/src/call.rs
```diff
@@ -2,94 +2,314 @@ use anyhow::{bail, Result};
...
+            .get(collection_address, working_set)
+            .is_some()
+        {
+            let nft_identifier = NftIdentifier(nft_id, collection_address.clone());
+            if let Some(mut n) = self.nfts.get(&nft_identifier, working_set) {
+                if owner.as_ref() == n.owner.as_ref() {
+                    n.owner = to.clone();
+                    self.nfts.set(&nft_identifier, &n, working_set);
+                } else {
+                    bail!("Transfer sent with owner {}, NFT id {} in Collection with address {} is owned by {}",
```
bkolad: This branch and the one in line 250 is repeated a few times.
What about adding methods like:
1. `check_owner(..)->Result`
2. `check_if_collection_exists()-> Result`
3. etc..

And the at the beginning of the methods that need validation we do:
```
check_owner()?;
check_if_collection_exists()?;
...
```
We use this pattern here: https://github.com/Sovereign-Labs/sovereign-sdk/blob/nightly/module-system/module-implementations/sov-accounts/src/call.rs#L42

examples/demo-nft-module/src/call.rs
```diff
@@ -2,94 +2,314 @@ use anyhow::{bail, Result};
...
+                        if Some(true) == frozen {
+                            n.frozen = true;
+                        }
+                        if let Some(murl) = token_uri {
+                            n.token_uri = murl;
+                        }
+                        self.nfts.set(&nft_identifier, &n, working_set);
+                    } else {
+                        bail!(
+                            "NFT id {} in Collection with address {} is frozen",
```
bkolad: Same: https://github.com/Sovereign-Labs/sovereign-sdk/pull/850/files#r1325662094

examples/demo-nft-module/src/call.rs
```diff
@@ -2,94 +2,314 @@ use anyhow::{bail, Result};
...
+        context: &C,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<CallResponse> {
+        let creator = context.sender();
+        let collection_address = get_collection_address::<C>(collection_name, creator.as_ref());
 
-        working_set.add_event("NFT mint", &format!("A token with id {id} was minted"));
+        if let Some(mut c) = self.collections.get(&collection_address, working_set) {
+            if c.frozen {
+                bail!(
```
bkolad: Same: https://github.com/Sovereign-Labs/sovereign-sdk/pull/850/files#r1325662094

examples/demo-nft-module/tests/nft_test.rs
```diff
@@ -1,159 +1,452 @@
-use demo_nft_module::{CallMessage, NonFungibleToken, NonFungibleTokenConfig, OwnerResponse};
```
bkolad, line 1: We should add integration tests in `demo-rollup/tests`. The tests there are higher level and use real rpc and mock DA. It can be done as a separate PR.

examples/demo-nft-module/queries.md
dubbelosix: ah sorry, this particular file was checked in accidentally and i was just saving the curl commands there - the explanation for the queries is here https://github.com/Sovereign-Labs/sovereign-sdk/blob/aa75d9ca57e533feb837a671d89eb6eba17c84bb/examples/demo-nft-module/README.md#create-collection I can add more detail there as well and remove this file

examples/demo-nft-module/src/call.rs
dubbelosix: yup that makes sense and it also keeps the core business logic separated and free from all the validation logic - avoids bugs too! will make this change

examples/demo-nft-module/tests/nft_test.rs
dubbelosix: https://github.com/Sovereign-Labs/sovereign-sdk/issues/871

Cargo.toml
dubbelosix: moved it to "module-implementations/sov-nft-module"

module-system/module-implementations/sov-nft-module/src/call.rs
```diff
@@ -0,0 +1,300 @@
...
+            ))
+        }
+    }
+
+    fn get_collection_by_name(
+        &self,
+        collection_name: &str,
+        context: &C,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Result<(CollectionAddress<C>, Collection<C>)> {
```
preston-evans98: It seems like it would be pretty easy to introduce a bug where someone fetches a collection by address and then mutates it - effectively skipping the `is_owner` check. Should we introduce an extra type so that `get_collection_by_address` returns a wrapped `Collection` which can't be stored back in the tree - effectively making it immutable?

module-system/module-implementations/sov-nft-module/src/call.rs
```diff
@@ -0,0 +1,300 @@
...
+        } else {
+            Err(anyhow!(
+                "Collection with name: {} does not exist for creator {}",
+                collection_name,
+                creator
+            ))
+        }
+    }
+
+    fn get_nft_by_id(
```
preston-evans98: Same question here - should we try to bake the `is_owner` check into this method using the type system. Maybe we have two methods: `get_owned_nft(context, collection_address, nft_id)` which returns an `OwnedfNft` that has a `pub owner` field and only allows `get` access to all other fields, and `get_nft_for_admin`, which has `pub uri` and `pub frozen` fields, but only provides `get` access to the rest of the fields. 

Then, we just need to implement `From<Nft>` and `into<Nft>` for each of these newtypes, and we get full type safety

preston-evans98: We also have functionality using the `EncodeLike` trait to let these kinds of newtypes be stored directly in the JMT instead of using `From` and `Into`

preston-evans98 (commented):
Looks really good overall. Left a couple of notes about type safety, but just one overall question - how would someone use this to create an NFT drop? It looks like in the current implementation the collection creator has to sign on every individual mint - maybe we should change the interface to have `mint` which can be called by other modules in addition to the current function  `mint_from_eoa`? The bank follows that pattern.

dubbelosix:
> Looks really good overall. Left a couple of notes about type safety, but just one overall question - how would someone use this to create an NFT drop? It looks like in the current implementation the collection creator has to sign on every individual mint - maybe we should change the interface to have `mint` which can be called by other modules in addition to the current function `mint_from_eoa`? The bank follows that pattern.

yeah, I agree. we can follow the same pattern, so that way a creator can mint to addresses directly, or the mint can also be created in conjunction with a token that exists in the Bank (will be doing this in a different PR to not make this one too large)

module-system/module-implementations/sov-nft-module/src/collection.rs
```diff
@@ -0,0 +1,134 @@
...
+impl<C: Context> Collection<C> {
+    pub fn new(
+        collection_name: &str,
+        collection_uri: &str,
+        collections: &StateMap<CollectionAddress<C>, Collection<C>>,
+        context: &C,
+        working_set: &mut WorkingSet<C>,
+    ) -> anyhow::Result<(CollectionAddress<C>, Collection<C>)> {
+        let creator = context.sender();
+        let ca = get_collection_address(collection_name, creator.as_ref());
```
preston-evans98: nit: let's use more expressive names here. maybe `collection_address`?

module-system/module-implementations/sov-nft-module/src/collection.rs
```diff
@@ -0,0 +1,134 @@
...
+        let c = collections.get(&ca, working_set);
```
preston-evans98: nit: Same here, let's use `collection`

module-system/module-implementations/sov-nft-module/src/call.rs
```diff
@@ -132,35 +152,46 @@ impl<C: Context> NonFungibleToken<C> {
...
-
-        Ok(CallResponse::default())
+        )?;
+        match collection_state {
+            CollectionState::Frozen(c) => bail!(
+                "Collection with name: {} , creator: {} is frozen",
+                c.get_name(),
+                c.get_creator()
+            ),
+            CollectionState::Mutable(mut collection) => {
```
preston-evans98: nit: It looks like this pattern of: 
- get collection
- match
- bail if frozen
... is repeated a few times. Maybe a helper function would be good?

module-system/module-implementations/sov-nft-module/src/collection.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    pub fn get_owned_collection(
+        collection_name: &str,
+        collections: &StateMap<CollectionAddress<C>, Collection<C>>,
+        context: &C,
+        working_set: &mut WorkingSet<C>,
+    ) -> anyhow::Result<(CollectionAddress<C>, CollectionState<C>)> {
+        let creator = context.sender();
+        let ca = get_collection_address(collection_name, creator.as_ref());
+        let c = collections.get(&ca, working_set);
```
preston-evans98: nit: let's use full variable names here. Even `addr` is clear enough, but one to two letters is a little short :)

module-system/module-implementations/sov-nft-module/src/collection.rs
```diff
@@ -0,0 +1,134 @@
...
+// the members of the struct to be mutable
+/// NewType representing a mutable (or unfrozen) collection
+pub struct MutableCollection<C: Context>(pub Collection<C>);
+
+/// Member Functions to allow controlled mutability for the Collection struct
+/// Can only freeze. Cannot unfreeze
+/// Can modify collection_uri
+/// Can increment supply. Cannot decrement
+/// Cannot modify creator address
+/// Cannto modify name
```
preston-evans98: nit: `cannto` => `cannot`

module-system/module-implementations/sov-nft-module/src/nft.rs
```diff
@@ -0,0 +1,177 @@
...
+                working_set,
+            )
+            .is_some()
+        {
+            bail!(
+                "NFT with id {} already exists for collection address {}",
+                token_id,
+                collection_address
+            )
+        }
```
preston-evans98, line 107: nit: the anyhow `Context` trait provides a nice method for this. `with_context` is like expect, but bails instead of panicking.

module-system/module-implementations/sov-nft-module/src/nft.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        let nft_identifier = NftIdentifier(token_id, collection_address.clone());
+        let nft = nfts.get(&nft_identifier, working_set).ok_or_else(|| {
+            anyhow!(
+                "Nft with token_id: {} in collection_address: {} does not exist",
+                token_id,
+                collection_address
+            )
+        })?;
+
```
preston-evans98: nit: anyhow::Context::with_context is meant for exactly this

module-system/module-implementations/sov-nft-module/src/nft.rs
```diff
@@ -0,0 +1,177 @@
...
+        if nft.owner == OwnerAddress::new(sender) {
+            Ok(OwnedNft(nft))
+        } else {
```
preston-evans98: nit: let's move this check inside the `OwnedNft` constructor for type safety. 

```
fn new(nft: Nft, claimed_owner: C::Address) -> Result<Self, anyhow::Error>{ 
        if nft.owner == OwnerAddress::new(sender) {
            Ok(OwnedNft(nft))
        } else { 
         ...
        }
}

File: module-system/sov-modules-macros/src/lib.rs [hunk 18]
```diff
@@ -196,3 +198,9 @@ pub fn custom_enum_clap(input: TokenStream) -> TokenStream {
     let input: syn::DeriveInput = parse_macro_input!(input);
     handle_macro_error(derive_cli_wallet_arg(input))
 }
+
+#[proc_macro_attribute]
+pub fn address_type(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
preston-evans98, line 262: Nice! Can we add a doc comment explaining what the output is?

module-system/module-implementations/sov-nft-module/src/collection.rs
preston-evans98: nit: let's use anyhow::Context::with_context here instead of the `if let`

preston-evans98 (approved):

PR #851, opened 2023-09-13
Test `demo-rollup` commands in CI

neysofu:
To generate the script:

`$ cargo run --bin bashtestmd -- --input examples/demo-rollup/README.md --output script.sh --tag test-ci`

Then, `chmod +x script.sh` and run with `./script.sh`.

I will remove `script.sh` and update CI workflows if approved, but I'm leaving the script committed to facilitate reviews for now.


script.sh
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env bash
+trap 'jobs -p | xargs -r kill' EXIT
+echo 'Running: '\''cd examples/demo-rollup/'\'''
+cd examples/demo-rollup/
```
preston-evans98: Nice! Looks like this it would be easy to get the `echo` and the actual command out of sync - maybe we should define a bash function which runs `echo` followed by `sh`?

preston-evans98: Oh, I see this is auto-generated. Nevermind! Did you intend to check it in?

preston-evans98 (approved):
LGTM! Should we add this into CI?

script.sh
neysofu: Yeah, but I'm removing it before merging. I committed it to facilitate reviews

neysofu:
Done! Merging.

PR #858, opened 2023-09-14
Evm/trim packages

LukaszRozmej:
# Description
Removes excessive packages
Fixes references as workspace references
rename `transactions` module to `primitive_types`

### DIFF
--- a/module-system/module-implementations/sov-evm/src/signer/mod.rs
+++ b/module-system/module-implementations/sov-evm/src/signer/mod.rs
@@ -1,0 +1,48 @@
+use std::collections::HashMap;
+
+use reth_primitives::{sign_message, Address, Transaction, TransactionSigned, H256};
+use reth_rpc::eth::error::SignError;
+use secp256k1::{PublicKey, SecretKey};
+
+/// Ethereum transaction signer.
+pub struct DevSigner {
+    signers: HashMap<Address, SecretKey>,
+}
+
+impl DevSigner {
+    /// Creates a new DevSigner.
+    pub fn new(secret_keys: Vec<SecretKey>) -> Self {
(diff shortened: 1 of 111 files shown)


bkolad (approved):

module-system/module-implementations/sov-evm/src/signer/mod.rs
```diff
@@ -1,14 +1,29 @@
...
 use secp256k1::{PublicKey, SecretKey};
 
 /// Ethereum transaction signer.
 pub struct DevSigner {
     signers: HashMap<Address, SecretKey>,
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum SignError {
+    /// Error occured while trying to sign data.
```
bkolad, line 13: We also have to add conversion to `jsonrpsee::core::Error` (eth_send_transaction will need it).

LukaszRozmej: I would prefer not to use `EthApiError` at all and this PR removes its usage.
We use `EthApiError` only to convert it into `jsonrpsee:error`, so if we need it lets convert this to `jsonrpsee:error` in the first place. The only thing `EthApiError` brings can be encapsulated like here: https://github.com/Sovereign-Labs/sovereign-sdk/pull/858/files#diff-f6d4618cb7e320bcd31b9c312498b000956f99c98fc0b0480ee666a9086d84fdR120-R127
The other problem with `EthApiError` is that it brings `reth-rpc` package, and that package has very big dependencies including `mdbx` and others, which we don't need and probably want to avoid.

LukaszRozmej: lets add when we need to

PR #859, opened 2023-09-14
Remove feature gate form `EVM` integration tests.

### LINKED ISSUE #844: Bank integration test in demo-rollup fails if EVM is not feature gated

bkolad:
# Description
Remove feature gate form `EVM` integration tests.

## Linked Issues
- Fixes #844

## Testing
CI passes.

## Docs
No changes.

### DIFF
--- a/examples/demo-rollup/Cargo.toml
+++ b/examples/demo-rollup/Cargo.toml
@@ -67,6 +67,7 @@ ethers-signers = { workspace = true }
 ethers = { workspace = true }
 revm = { workspace = true }
 
+sov-demo-rollup = { path = ".", features = ["experimental"] }
 
 [features]
 default = []
--- a/examples/demo-rollup/tests/all_tests.rs
+++ b/examples/demo-rollup/tests/all_tests.rs
@@ -1,6 +1,3 @@
-// For now, exclude the evm from the bank tests
-#[cfg(not(feature = "experimental"))]
 mod bank;
-#[cfg(feature = "experimental")]
 mod evm;
 mod test_helpers;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6698,6 +6698,7 @@ dependencies = [
  "sov-celestia-adapter",
  "sov-cli",
  "sov-db",
+ "sov-demo-rollup",
  "sov-ethereum",
  "sov-evm",
  "sov-modules-api",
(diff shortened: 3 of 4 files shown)


examples/demo-rollup/tests/all_tests.rs
```diff
@@ -1,6 +1,3 @@
-// For now, exclude the evm from the bank tests
-#[cfg(not(feature = "experimental"))]
 mod bank;
-#[cfg(feature = "experimental")]
```
LukaszRozmej, line 4: this flags should stay as they were, but CI should get `experimental` feature flag when invoking `cargo llvm-cov --locked --lcov --output-path lcov.info`

bkolad: We used to import `sov-demo-rollup = { path = ".", features = ["experimental"] }` in tests, but it was removed with #859. I am adding it back as the underlying issue is fixed. cc @preston-evans98

LukaszRozmej: going to approve, but I still think they should be feature-gated, maybe by `sov-demo-rollup/experimental` feature as they are dependent on that.

examples/demo-rollup/Cargo.toml
```diff
@@ -67,6 +67,7 @@ ethers-signers = { workspace = true }
 ethers = { workspace = true }
 revm = { workspace = true }
 
+sov-demo-rollup = { path = ".", features = ["experimental"] }
```
LukaszRozmej, line 70: was this removed only because of issue in evm?

LukaszRozmej (approved):

examples/demo-rollup/tests/all_tests.rs
bkolad: I believe it is more practical to have the tests for the `evm` enabled as the default. This  allows other developers to quickly detect any issues they may have caused without having to rely on the CI. Additionally, while it's acceptable to have incomplete documentation for `experimental` features, if someone has taken the time to write a test, we should make sure it functions correctly under all conditions.

preston-evans98 (approved):

examples/demo-rollup/Cargo.toml
preston-evans98: Yep, that's correct

PR #864, opened 2023-09-14
Fast-forward `stable` to v0.2.0

neysofu:
Let's bring `stable` up to date with `v0.2.0`.


neysofu:
I'll wait until https://github.com/Sovereign-Labs/sovereign-sdk/pull/817 is merged, so the commit history is cleaner and we won't have conflicts when we release v0.3.0.

PR #867, opened 2023-09-14
chore: create a new docker compose setup with local celestia network

### LINKED ISSUE #461: Sov-celestia-local container needs Dockerfile so that we can modify, update and build other versions of celestia binaries
## Summary

zvolin:
# Description

A new docker compose setup for tests / examples / local development. It's meant to replace both the `/docker` setup and `dubbelosix/sov-celestia-local:genesis-v0.7.1` image for running examples.

It uses latest celestia nodes so a few things are now done differently than in the `/docker` but the general idea is the same. Validator starts the chain and announces the hash of genesis block. Then it creates addresses for bridge (sequencer) nodes and transfers them coins. Each bridge node reads the genesis hash and trusts it. Then starts a node and syncs with validator.

Validator also saves each bridge address and keys in persistent volume, and if it finds it on startup, it just imports it. It allows for having constant address for the tests with just ungitignoring this file and commiting (I did this for the first sequencer).

It isn't particularly useful now before I update the `sov-celestia-adapter` but opening this PR early just fyi, as I saw that the original `/docker` setup was modified recently so to not double up work.

## Linked Issues
- Fixes #461 
- Related to # (issue) 

## Testing

## Docs
TODO!


zvolin:
@citizen-stig please check if it'll fit your workflow, as an author of `/docker`, I think about replacing it with this setup

preston-evans98:
FYI @zvolin, unfortunately @citizen-stig is out of the office for the rest of the week.

cc @dubbelosix, would you mind taking a look?

zvolin:
thanks @preston-evans98. I don't expect to even undraft this within a few days so no hurry. I've opened this as more of a preview for you to check how that affects your daily workflow. Also I saw some work done in `/docker` and didn't want to double up. Feel free to suggest potential use-cases that should be supported :+1:

 Ideally this change would be delivered together with the `sov-celestia-adapter` update, so I am thinking about closing this and opening a new PR which consists of all the changes or delivering changes one by one to some intermediate branch and merging later as a whole, as things can get quite big. No hard preferences there, whatever you'll like more

zvolin:
closing in favor of #948

PR #870, opened 2023-09-14
Add FinalizeSlotHook and invoke it after genesis

LukaszRozmej:
# Description
Separates `SlotHooks` and `FinalizeSlotHook` as they have different context. Invoke `FinalizeSlotHook` after normal slots as well as genesis.

## Linked Issues
- Fixes #874 

## Testing
Integration tests

PR #884, opened 2023-09-15
EVM: Implement getBlockByNumber query

### LINKED ISSUE #878: EVM: Implement `eth_getBlockByNumber` endpoint.
## Summary
Implement the endpoint according to the Ethereum spec.

orkunkilic:
# Description
Implements get_block_by_number method in sov-evm query. Implemented "earliest", "latest", and hex parameters. Implemented both details(full / hash) on transaction return type.

Block_number defaults to latest block if not present.

Looking forward for efficiency feedbacks!

## Linked Issues
- Fixes #878
- Related to #502 

## Testing
Tests WIP

## Docs
No docs added yet.


module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -25,42 +27,93 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
-            withdrawals_root: Default::default(),
+        // "safe" and "finalized" are not implemented
+        let block = match block_number {
+            Some(ref block_number) if block_number == "earliest" => self
+                .blocks
+                .get(0, &mut working_set.accessory_state())
+                .expect("Genesis block must be set"),
+            Some(ref block_number) if block_number == "latest" => self
+                .head
+                .get(working_set)
```
LukaszRozmej: `head` might not be available in this context, nor it will have `sealed` block -> please use `blocks.last()`

module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -25,42 +27,93 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                .expect("Head block must be set")
+                .seal(),
+            Some(ref block_number) if block_number == "pending" => self
+                .pending_head
+                .get(&mut working_set.accessory_state())
+                .expect("Pending head block must be set")
+                .seal(),
```
LukaszRozmej: I think we won't support `pending` as it doesn't make sense for us, for `pending` block we would have to get the next DA block.

module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -25,42 +27,93 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+            Some(ref block_number) => {
+                let block_number = usize::from_str_radix(block_number, 16).unwrap();
+                self.blocks
+                    .get(block_number, &mut working_set.accessory_state())
+                    .expect("Block must be set")
```
LukaszRozmej, line 38: this can fail and instead of panic, we need to return a proper error to rpc.

module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -25,42 +27,93 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+            }
+            None => self
+                .head
```
LukaszRozmej: `blocks.last` like above

module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -25,42 +27,93 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                .iter()
+                .map(|hash| {
+                    (
+                        hash,
+                        self.transaction_hashes
+                            .get(hash, &mut working_set.accessory_state())
+                            .unwrap(),
+                    )
+                })
+                .collect::<HashMap<_, _>>();
```
LukaszRozmej: we shouldn't iterate over all transactions_hashes, that will load the entire database!

module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -25,42 +27,93 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+
+                            reth_rpc_types::Transaction::from_recovered_with_block_context(
+                                tx.into(),
+                                block.header.hash,
+                                block.header.number,
+                                block.header.base_fee_per_gas,
+                                U256::from(tx_number - block.transactions.start),
+                            )
+                        })
+                        .collect::<Vec<_>>(),
```
LukaszRozmej: same here we cannot load entire database.

module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -25,42 +27,93 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                self.blocks
+                    .get(block_number, &mut working_set.accessory_state())
+                    .expect("Block must be set")
+            }
+            None => self
+                .head
+                .get(working_set)
+                .expect("Head block must be set")
+                .seal(),
+        };
```
LukaszRozmej: This code should be extracted to helper function it will be used in many RPC calls.

LukaszRozmej (dismissed):

module-system/module-implementations/sov-evm/src/query.rs
orkunkilic: I am not sure about that. There will be some rollups that use single centralized sequencer to give soft confirmations, such as ours. In that case, there will be a pending block on the sequencer level which is not broadcasted or sent to the DA layer.

module-system/module-implementations/sov-evm/src/query.rs
orkunkilic: That was my mistake, I confused block.transactions with db transactions. Fixed.

module-system/module-implementations/sov-evm/src/query.rs
LukaszRozmej: ok but then it should live on the sequencer and not on the module. Sequencer can extend the api, it will be a separate component that can us the module.. I don't see this doable inside the module.

LukaszRozmej (commented):
needs a test

module-system/module-implementations/sov-evm/src/query.rs
bkolad: I think we need to merge [#870](https://github.com/Sovereign-Labs/sovereign-sdk/pull/870) before writing a test. [@orkunkilic](https://github.com/orkunkilic) let's discuss the use case you have on our sync-call. I am sure we will accommodate it. In the meantime, I'm ok with two options: we can either merge this PR now and make a comment on #878 that we need to further work on the `test`  (once #870 lands) and the `pending block` case, or we can wait to discuss it during our upcoming sync-up call, whichever you prefer, [@orkunkilic](https://github.com/orkunkilic). Thank you for your excellent contribution!

LukaszRozmej: I think you can write test without it, I've written receipts and transactions tests without it. This should be similar In full node it should work. 

But also depends if it should be unit test on evm module or integration test on full rollup.

orkunkilic:
Tests are added, I think good to go after final review.

bkolad (approved):

module-system/module-implementations/sov-evm/src/query.rs
```diff
@@ -32,42 +61,63 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+            .transactions
+            .clone()
+            .map(|id| {
+                let tx = self
+                    .transactions
+                    .get(id as usize, &mut working_set.accessory_state())
+                    .expect("Transaction must be set");
+                (id, tx)
+            })
+            .collect::<Vec<_>>();
```
LukaszRozmej, line 89: You shouldn't need to collect it here, this causes unnecessary allocation on heap.

PR #893, opened 2023-09-18
Integrate "demo-prover" into workspace. Replace demo-prover impl with `StateTransitionRunner`.

preston-evans98:
# Description
This PR updates the demo-prover package to use the `StateTransitionRunner`, eliminating nearly all of the custom code from the host. A follow-up will do the same for the guest. It also integrates the demo-prover into the main workspace, and updates CI accordingly.


bkolad (approved):
LGTM, thx.

PR #895, opened 2023-09-19
feat: add wrapper struct for macros manifest

vlopes11:
This commit introduces a wrapper struct to encapsulate the macros manifest behavior.

It is intended to make it easier to change resolve strategies in the future (i.e. add full qualified module path for the appropriate manifest section, etc)


module-system/sov-modules-macros/src/manifest.rs
```diff
@@ -1,61 +1,180 @@
+#![allow(dead_code)]
```
preston-evans98, line 2: Should this be removed in future? If so, let's add a TODO so we don't forget

module-system/sov-modules-macros/src/manifest.rs
```diff
@@ -1,61 +1,180 @@
...
+    type Target = Table;
+
+    fn deref(&self) -> &Self::Target {
+        &self.table
+    }
+}
+
+impl Manifest {
+    /// The file name of the manifest.
+    pub const MANIFEST_NAME: &str = "sovereign.toml";
```
preston-evans98: nit: Let's rename this to something a bit more generic (`constants.toml`?) so that end-users aren't stuck with sovereign-specific branding

module-system/sov-modules-macros/src/manifest.rs
```diff
@@ -1,61 +1,180 @@
...
+
+    /// Reads a `sovereign.toml` manifest file, recursing from the target directory that builds the
+    /// current implementation.
+    ///
+    /// If the environment variable `SOVEREIGN_MANIFEST` is set, it will use that instead.
+    pub fn read() -> anyhow::Result<Self> {
+        let initial_path = match env::var("SOVEREIGN_MANIFEST") {
```
preston-evans98: nit: Same here - let's use something other than SOVEREIGN_MANIFEST

preston-evans98 (approved):

module-system/sov-modules-macros/src/manifest.rs
vlopes11: Added! It's there only until the module isn't consumed anywhere

module-system/sov-modules-macros/src/manifest.rs
vlopes11: Agreed; using `constants.toml`

module-system/sov-modules-macros/src/manifest.rs
vlopes11: Agreed; to maintain consistency with the file name, the name `CONSTANTS_MANIFEST` was used

PR #901, opened 2023-09-20
`ValueSetter` module: read `genesis` config from a file.

### LINKED ISSUE #872: Remove hardcoded genesis values and instead read the genesis configuration of a rollup from files.
## Summary
Currently, the rollup's `genesis` initialization relies on hard-coded values within the code rather than being sourced from configuration files. Moreover, the initialization logic for the rollup is duplicated in multiple places, including the `demo-rollup, demo-prover, tests, and benchmarks`

To tackle these issues, we have two primary tasks:
1. Standardize the rollup initialization logic to ensure consistency across different parts of the codebase.
2. Retrieve the initial genesis configuration from external files.

bkolad:
# Description
Read `genesis` configuration for `ValueSetter` module from a file rather than having these values hard-coded in the code.
Once this approach has been reviewed and approved, we have two tasks to complete:
1. Make the same change for all the modules
2. Pass all the required paths (for all modules) as parameters to the `get_genesis_config`

## Linked Issues
- Related to #872

## Testing
CI passes.

## Docs
No doc changes

### DIFF
--- a/examples/demo-stf/src/genesis_config.rs
+++ b/examples/demo-stf/src/genesis_config.rs
@@ -1,3 +1,4 @@
+use anyhow::Context as AnyhowContext;
 #[cfg(feature = "experimental")]
 use reth_primitives::Bytes;
 use sov_chain_state::ChainStateConfig;
@@ -41,6 +42,7 @@ pub fn get_genesis_config<C: Context, Da: DaSpec>(
         #[cfg(feature = "experimental")]
         evm_genesis_addresses,
     )
+    .expect("Unable to read genesis configuration")
 }
 
 fn create_genesis_config<C: Context, Da: DaSpec>(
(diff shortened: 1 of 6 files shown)


examples/demo-stf/src/genesis_config.rs
```diff
@@ -81,10 +81,13 @@ fn create_genesis_config<C: Context, Da: DaSpec>(
         is_preferred_sequencer: true,
     };
 
-    // This will be read from a file: #872
-    let value_setter_config = ValueSetterConfig {
-        admin: value_setter_admin_private_key.pub_key().to_address(),
-    };
+    // This path will be injected as a parameter: #872
+    let value_setter_genesis_path = "../test-data/genesis/value_setter.json";
```
bkolad, line 85: In the following PR we will inject paths as parameters.

citizen-stig (approved):

PR #903, opened 2023-09-20
dont merge: gas experiment discussion

vlopes11:


module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,160 @@
+pub mod sov_modules_api {
+    use anyhow::Result;
+
+    use core::marker::PhantomData;
+
+    // We add `remainder_gas` to `WorkingSet` as a simple `u64`; the gas will already be converted
+    // to scalar value when charged from `WorkingSet`.
+
+    pub struct WorkingSet<C> {
+        remainder_gas: u64,
```
bkolad: I think it should be reminding_funds: u64

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    impl<C> WorkingSet<C> {
+        pub fn new(remainder_gas: u64) -> Self {
+            Self {
+                remainder_gas,
+                _context: PhantomData,
+            }
+        }
+
+        pub fn charge_gas(&mut self, gas: u64) -> Result<()> {
```
bkolad, line 22: And here `  pub fn charge_gas(&mut self, gas: C:GasUnit) -> Result<()> `

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    use core::marker::PhantomData;
+
+    // We add `remainder_gas` to `WorkingSet` as a simple `u64`; the gas will already be converted
+    // to scalar value when charged from `WorkingSet`.
+
+    pub struct WorkingSet<C> {
+        remainder_gas: u64,
+        _context: PhantomData<C>,
+    }
```
bkolad, line 12: we can add also gas_price: GasUnit and together with reminding_funds put it in GasoMeter

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,160 @@
...
+        pub fn new(remainder_gas: u64) -> Self {
+            Self {
+                remainder_gas,
+                _context: PhantomData,
+            }
+        }
+
+        pub fn charge_gas(&mut self, gas: u64) -> Result<()> {
+            self.remainder_gas = self
+                .remainder_gas
```
bkolad: here we could of `self.remining_funds = self.self.remining_funds - gas.value(self.price)`

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        fn charge_gas(
+            ws: &mut WorkingSet<Self::Context>,
+            key: &str,
+            price: &<<Self::GasConfig as GasConfig>::GasUnit as GasUnit>::Price,
+        ) -> Result<()> {
+            Self::GasConfig::get_gas_unit(key)
+                .ok_or(anyhow::anyhow!("No gas unit found"))
+                .map(|gas_unit| gas_unit.value(price))
+                .and_then(|gas| ws.charge_gas(gas))
```
bkolad, line 55: now we we do the calculation inside ws so we can remove `map(|gas_unit| gas_unit.value(price))`

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+}
+
+#[test]
+fn it_works() {
+    use foo::FooModule;
+    use sov_modules_api::WorkingSet;
+
+    let gas = 1_000_000;
```
bkolad, line 139: here we crate a GasoMeter with `gas_price, remining_funds` and pass it to WS. In the RPC queries we can set the GasoMeter to ignore gas counting.

module-system/sov-modules-api/src/gas.rs
vlopes11: This would imply we need to move `C::GasUnit` definition to context, that could increase the complexity of the solution. Since it is not strictly required as `GasUnit` lives only inside of the module, and we have everything there to resolve a `u64` value, should we expect in the future to extend the functionality of the gas unit definition in working set?

I think having it as a simple `u64` will greatly simplify things

vlopes11: I'll push a commit with the proposed change so we check which one we prefer

module-system/sov-modules-api/src/gas.rs
vlopes11: Done!

module-system/sov-modules-api/src/gas.rs
vlopes11: Added here as a simplified version. Let me know wdyt. Attention point is the need to create a dedicated context for `Foo`

https://github.com/Sovereign-Labs/sovereign-sdk/pull/903/commits/7efa44ef695fa79c4ee0f60bfc8e18f73c32de3c

bkolad: Yeah the GasUnit is the same for all the modules in the rollup.

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,190 @@
...
+    // the dynamic price calculation
+
+    pub trait Module {
+        type Context: Context;
+        type GasConfig: GasConfig<GasUnit = <Self::Context as Context>::GasUnit>;
+
+        fn charge_gas(
+            ws: &mut WorkingSet<Self::Context>,
+            key: &str,
+            price: &<<Self::GasConfig as GasConfig>::GasUnit as GasUnit>::Price,
```
bkolad: I think price is already in GasoMeter

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,190 @@
...
+    // On the `Module`, we simply receive a `GasUnit` static type, and everything is unblocked by
+    // the dynamic price calculation
+
+    pub trait Module {
+        type Context: Context;
+        type GasConfig: GasConfig<GasUnit = <Self::Context as Context>::GasUnit>;
+
+        fn charge_gas(
+            ws: &mut WorkingSet<Self::Context>,
+            key: &str,
```
bkolad: what is the `key`?

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,190 @@
...
+    // Here we declare a custom gas unit
+
+    pub struct FooGasUnit {
+        pub a: u64,
+        pub b: u64,
+        pub c: u64,
+    }
+
+    impl GasUnit for FooGasUnit {
+        type Price = String;
```
bkolad: this should be in  `sov_modules_api`?

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,190 @@
...
+    // The context will be defined as concrete type for execution
+
+    pub type FooContext = super::sov_modules_api::DefaultContext<FooGasUnit>;
+
+    // The code below will be generated by a macro that will look like
+    // gas_config!(FooGasConfig, FooGasUnit);
+
+    pub struct FooGasConfig {
+        pub keys: [&'static str; 2],
+        pub values: [FooGasUnit; 2],
```
bkolad, line 87: can have something like: 
```
FooGasConfig<C: Context> {
 some_martix_multiplication: C: GasUnit
}
```

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,190 @@
...
+        _context: PhantomData<C>,
+    }
+
+    impl<C> WorkingSet<C>
+    where
+        C: Context,
+    {
+        pub fn new(remainder_funds: u64) -> Self {
+            Self {
+                remainder_funds,
```
preston-evans98, line 17: nit: let's call this `remaining_funds`. Remainder funds sounds like we're doing division

PR #904, opened 2023-09-20
Generic sequencer address in `SequencerRegistry` module

### LINKED ISSUE #622: use `AddressTrait` as generic param for input/output value in Sequencer Registry and BlobStorage
Currently `Vec<u8>` is used as actual representation of addresses. 

Idea is to keep it as storage attributes, but for call/query and other parameters switch to generic parameter, so we have type safety

citizen-stig:
# Description

Using generic `Da::Address` instead of raw bytes. That allows seamleasly interact with STF and other parts of SDK which rely on concrete type of BasicAddress. 

Makes module `SequencerRegistry` generic over `DaSpec`. SequencerRegistry module switches to BcsCodec for storage, but still uses Borsh for CallMessage

This PR removes bunch of `.as_ref().to_vec()`

Also fixes bug in `MockAddress` string representation

## Linked Issues
- Fixes #622

## Testing
Tests have been updated

## Docs
Docs have been checked.


preston-evans98 (approved):
Looks great!

PR #907, opened 2023-09-21
WIP Blaze/GAS

bkolad:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,158 @@
...
+    //#[derive(serde::Serialize, serde::Deserialize)]
+    pub struct FooGasConfig<GS> {
+        complex_math_operation: GS,
+        another_operation: GS,
+    }
+
+    impl<GS: GasUnit> GasConfig for FooGasConfig<GS> {
+        const MANIFEST: Self = {
+            Self {
+                complex_math_operation: todo!(),
```
bkolad: I'm unsure about this part; how we should read the manifest? I assume this will be done by the `build.rs`?

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,158 @@
...
+    impl<C: Context> FooModule<C> {
+        pub fn new() -> Self {
+            Self {
+                _context: PhantomData::<C>,
+                gas_config: GasConfig::MANIFEST,
+            }
+        }
+
+        pub fn some_cool_function(&self, ws: &mut WorkingSet<C>) -> Result<()> {
+            ws.charge_gas(&self.gas_config.complex_math_operation)?;
```
bkolad: This is how the module charges gas.

module-system/sov-modules-api/src/gas.rs
vlopes11: It will be a macro expansion executed at compilation

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,157 @@
...
+        _context: PhantomData<C>,
+    }
+
+    impl<C: Context> WorkingSet<C> {
+        pub fn new(gaso_meter: GasoMeter<C::GasUnit>) -> Self {
+            Self {
+                gaso_meter,
+                _context: PhantomData,
+            }
+        }
```
vlopes11: I was thinking its best to keep the GasoMeter fully encapsulated so we don't have an additional type. Wdyt?

```rust
        pub fn new(remaining_funds: u64, gas_price: C::GasUnit) -> Self {
            Self {
                gaso_meter: GasoMeter::new(remaining_funds, gas_price),
                _context: PhantomData,
            }
        }
```

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,157 @@
...
+        fn value(&self, price: &Self) -> u64 {
+            self.a.saturating_mul(price.a).saturating_add(
+                self.b
+                    .saturating_mul(price.b)
+                    .saturating_add(self.c.saturating_mul(price.c)),
+            )
+        }
+    }
+
+    pub struct DefaultContext {}
```
vlopes11: What if we use a generic gas unit for the default context, so the user won't have to declare his own context every time he creates a gas unit?

```rust
    pub struct DefaultContext<GU>
    where
        GU: GasUnit,
    {
        _gas_unit: PhantomData<GU>,
    }
```

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,149 @@
...
+    use anyhow::Result;
+
+    use core::marker::PhantomData;
+
+    pub trait GasUnit {
+        fn value(&self, price: &Self) -> u64;
+    }
+
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct TupleGasUnit<const N: usize>(pub [u64; N]);
```
vlopes11, line 11: This is not a concrete struct representing a tuple. The manifest file will expect an array, and will use this struct as hardcoded for now. Example

```json
{ some_complex_math: [1, 2, 3], other: [4, 5, 6] }
```

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    pub struct FooGasConfig<GU>
+    where
+        GU: GasUnit,
+    {
+        pub complex_math_operation: GU,
+        pub some_other_operation: GU,
+    }
+
+    pub struct FooModule;
```
vlopes11, line 119: Module is stateless (for the gas config)

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    impl Module for FooModule {
+        type Context = DefaultContext<TupleGasUnit<3>>;
+        type GasConfig = FooGasConfig<TupleGasUnit<3>>;
+        const GAS_CONFIG: Self::GasConfig = FooGasConfig {
+            complex_math_operation: TupleGasUnit([1, 2, 3]),
+            some_other_operation: TupleGasUnit([4, 5, 6]),
+        };
```
vlopes11, line 127: The expanded macro will look like this

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,149 @@
...
+            ws.charge_gas(price)
+        }
+    }
+
+    pub struct DefaultContext<GU>
+    where
+        GU: GasUnit,
+    {
+        _gas_unit: PhantomData<GU>,
+    }
```
vlopes11, line 96: Added complexity/point of alert: the default context now takes a generic gas unit so it can be resolved at compilation

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub struct WorkingSet<C>
+    where
+        C: Context,
+    {
+        gaso_meter: GasoMeter<C::GasUnit>,
+        _context: PhantomData<C>,
+    }
+
+    impl<C: Context> WorkingSet<C> {
+        pub fn new(remaining_funds: u64, gas_price: C::GasUnit) -> Self {
```
vlopes11, line 66: Here I opted to keep `GasoMeter` fully encapsulated in the module instead of exposing it, since it is an implementation details of the working set. Thoughts?

vlopes11 (commented):
@bkolad I pushed a commit with some changes to your proposal that reflects the code to be deployed. Let me know your thoughts

module-system/sov-modules-api/src/gas.rs
vlopes11: N/A; static type declared under `Context`

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,149 @@
...
+
+            Ok(())
+        }
+    }
+
+    pub struct WorkingSet<C>
+    where
+        C: Context,
+    {
+        gaso_meter: GasoMeter<C::GasUnit>,
```
preston-evans98, line 61: nit: `gas_o_meter` or `gas_meter`

module-system/sov-modules-api/src/gas.rs
bkolad: lgtm

module-system/sov-modules-api/src/gas.rs
bkolad: The problem is that this generic would propagate to many places. Can we go with associated type and revisit it later?  If we decide to add a Generic it should be for something bigger than just `GasUnit` maybe we should include DA as well but this is a separate discussion.

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    pub trait Module {
+        type Context: Context;
+        type GasConfig;
+        const GAS_CONFIG: Self::GasConfig;
+
+        fn charge_gas(
+            ws: &mut WorkingSet<Self::Context>,
+            price: &<Self::Context as Context>::GasUnit,
```
bkolad, line 85: here we don't need price it is already in ws

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,149 @@
...
+        GU: GasUnit,
+    {
+        pub complex_math_operation: GU,
+        pub some_other_operation: GU,
+    }
+
+    pub struct FooModule;
+
+    impl Module for FooModule {
+        type Context = DefaultContext<TupleGasUnit<3>>;
```
bkolad, line 122: It is the rollup which sets the GU not the Module

module-system/sov-modules-api/src/gas.rs
bkolad: This is great, but how the macro will know this is `TupleGasUnit`?

module-system/sov-modules-api/src/gas.rs
bkolad: lgtm

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,149 @@
...
+    where
+        GU: GasUnit,
+    {
+        pub complex_math_operation: GU,
+        pub some_other_operation: GU,
+    }
+
+    pub struct FooModule;
+
+    impl Module for FooModule {
```
bkolad, line 121: In the previous version `FooModule` was parmetrized by a <C: Context> and the `TupleGasUnti` didn't leak anywhere in the module implemntetaion. Will the macro expansion still work if the only information the module see is "GU" and not `TupleGasUnit`?

module-system/sov-modules-api/src/gas.rs
bkolad: lgtm

module-system/sov-modules-api/src/gas.rs
```diff
@@ -0,0 +1,149 @@
...
+        type GasConfig = FooGasConfig<TupleGasUnit<3>>;
+        const GAS_CONFIG: Self::GasConfig = FooGasConfig {
+            complex_math_operation: TupleGasUnit([1, 2, 3]),
+            some_other_operation: TupleGasUnit([4, 5, 6]),
+        };
+    }
+
+    impl FooModule {
+        pub fn some_cool_function(
+            ws: &mut WorkingSet<DefaultContext<TupleGasUnit<3>>>,
```
bkolad, line 132: this will be `<C>` not `DefaultContext<TupleGasUnit<3>>` right?

module-system/sov-modules-api/src/gas.rs
vlopes11: True! The problem is actually the name of the variable; should be `gas`.

module-system/sov-modules-api/src/gas.rs
vlopes11: This problem is only applicable for this example as all modules will have a Context (hence, we take the unit from there)

module-system/sov-modules-api/src/gas.rs
vlopes11: That's correct, I'll push a commit with the fixed nits so we can better visualize the expected result

PR #908, opened 2023-09-21
Use `jmt` update proofs. Add `Storage::Root` generic.

### LINKED ISSUE #246: Update jmt to remove honest treereader assumption.

preston-evans98:
# Description
This PR switches from our temporary (insecure) method of proving updates to the new jmt update proofs that @theochap implemented while he was here. In order to make this change, it was convenient to clean up many of the hard-coded `[u8;32]`s scattered throughout the codebase and replace them with a generic `Storage::Root` type. 

## Linked Issues
- Fixes #246 
- Related to # (issue) 

## Testing
This change is covered by existing tests.


module-system/sov-state/src/storage.rs
```diff
@@ -160,6 +160,17 @@ pub trait Storage: Clone {
         + BorshSerialize
         + BorshDeserialize;
 
+    /// A cryptographic commitment to the contents of this storage
+    type Root: Serialize
```
citizen-stig, line 164: Nit: just note on naming. For me it looks like this is more "Hash", because root_hash is hash in the end, but maybe hash is too broad as well.

full-node/sov-stf-runner/src/runner.rs
```diff
@@ -75,28 +75,30 @@ where
...
-            } else {
-                debug!("Chain is already initialized. Skipping initialization.");
-                app.get_current_state_root()?
-            }
+            debug!("Chain is already initialized. Skipping initialization.");
+            prev_state_root
+        } else {
+            info!("No history detected. Initializing chain...");
+            let genesis_root = app.init_chain(genesis_config);
+            info!("Chain initialization is done.");
```
citizen-stig: Could you please log genesis root here?

citizen-stig (approved):

module-system/sov-state/src/storage.rs
preston-evans98: Hmm, Hash isn't quite accurate here because the root can be something other than a Hash (I.e. a curve point). Also, it's a bit overloaded - it might be confused with the "Hasher" generic

full-node/sov-stf-runner/src/runner.rs
preston-evans98: Done!

PR #913, opened 2023-09-22
Use `Strorage` generic instead of `Context` for `Delta` and `AccessoryDelta`

citizen-stig:
# Description

Refactoring, that might be used for re-orgs

## Testing
Existing tests are passing

## Docs
Documentation has not been updated

### DIFF
--- a/module-system/sov-modules-api/src/state/scratchpad.rs
+++ b/module-system/sov-modules-api/src/state/scratchpad.rs
@@ -268,10 +268,6 @@ where
 }
 
 impl<T: StateReaderAndWriter> StateReaderAndWriter for RevertableWriter<T> {
-    fn delete(&mut self, key: &StorageKey) {
-        self.writes.insert(key.to_cache_key(), None);
-    }
-
     fn get(&mut self, key: &StorageKey) -> Option<StorageValue> {
         if let Some(value) = self.writes.get(&key.to_cache_key()) {
             value.as_ref().cloned().map(Into::into)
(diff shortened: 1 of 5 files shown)


neysofu (approved):
Much cleaner!

module-system/sov-modules-api/src/state/scratchpad.rs
```diff
@@ -11,48 +11,40 @@ use crate::{Context, Spec};
...
-    witness: <<C as Spec>::Storage as Storage>::Witness,
+pub struct Delta<S: Storage> {
+    inner: S,
+    witness: S::Witness,
     cache: StorageInternalCache,
 }
 
-impl<C: Context> Delta<C> {
-    fn new(inner: <C as Spec>::Storage) -> Self {
+impl<S: Storage> Delta<S> {
```
bkolad, line 20: This was changed recently by @vlopes11 from` Storage` to `Context`.

citizen-stig:
Requesting review from @vlopes11 , as there was this change: https://github.com/Sovereign-Labs/sovereign-sdk/pull/832

@vlopes11 , can you please provide more details, if this PR will break something needed for Gas estimation?

vlopes11 (approved):
@citizen-stig nothing will break! The only reason it was changed is to keep the bounds with a higher level trait so changes can be centralized into a single point (ie spec)

preston-evans98 (approved):

PR #914, opened 2023-09-24
`EVM`: Add missing docs

bkolad:
# Description
Documentation for the EVM module.

## Testing
CI passes.

## Docs
Docs added


module-system/sov-modules-api/src/hooks.rs
```diff
@@ -69,7 +69,7 @@ pub trait SlotHooks<Da: DaSpec> {
 pub trait FinalizeHook<Da: DaSpec> {
     type Context: Context;
 
-    fn finalize_slot_hook(
+    fn finalize_hook(
```
citizen-stig, line 72: Nit: What is the reason for renaming? 

I personally find `finalize_slot_hook` to be clearer, as it describes what it actually finalizes.

Ideally I would prefer it to be this way:

```rust
pub trait FinalizeSlotHook<Da: DaSpec> {
    fn finalize_slot()
}
```

citizen-stig (approved):

module-system/sov-modules-api/src/hooks.rs
bkolad: This method is also called just after genesis, so the `slot` was misleading.

preston-evans98 (approved):

PR #922, opened 2023-09-25
feat: use json as manifest constants instead of toml

vlopes11:
Currently, we parse the manifest file as TOML. However, we will use JSON
for the genesis format. For improved consistency, it is desirable to
have the constants manifest file with the same format.

### DIFF
--- a/module-system/sov-modules-macros/src/manifest.rs
+++ b/module-system/sov-modules-macros/src/manifest.rs
@@ -4,35 +4,43 @@
 use std::path::PathBuf;
 use std::{env, fmt, fs, ops};
 
-use proc_macro2::TokenStream;
-use toml::{Table, Value};
-
-use crate::common::StructDef;
-use crate::module_info::parsing::ModuleField;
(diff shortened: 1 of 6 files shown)


preston-evans98 (approved):

module-system/sov-modules-macros/src/manifest.rs
```diff
@@ -186,13 +226,43 @@ fn fetch_manifest_works() {
...
+    let input = r#"{
+        "comment": "Sovereign SDK constants",
+        "gas": {
+            "complex_math_operation": [1, 2, 3],
+            "some_other_operation": [4, 5, 6]
+        }
+    }"#;
+
+    let parent = Ident::new("foo", proc_macro2::Span::call_site());
+    let gas_config = Manifest::read_str(input, PathBuf::from("foo.toml"), &parent)
```
bkolad: Should be `foo.json`?

vlopes11: True, fixed!

PR #932, opened 2023-09-26
Run tests with prover enabled. 

bkolad:
# Description
1. Moves `AppVerifier` to `demo-stf` which breaks the `demo-prover -> demo-rollup` dependency.
2. Adds `methods/guest-mock` a prover that works with `MockDa`
3. Makes uses of the `guest-mock` prover in `demo-rollup` tests with the `DemoProverConfig::Execute` setting. 
4. Updates CI accordingly.
5. Removes `ide_steup` readme, is it still relevant?
6. Removes the `local` feature as it is relevant only for the `EthSequencer`
## Testing
Ci passes.

## Docs
No changes.


examples/demo-prover/methods/guest-mock/riscv32im-risc0-zkvm-elf.json
```diff
@@ -0,0 +1,21 @@
+{
```
preston-evans98: I think we can delete this file too. That's also a holdover from the old Risc0 version

preston-evans98 (commented):
Looks good to me, except that it feels like we might be breaking an abstraction by adding a dependency on `sov-stf-runner` in `demo-stf`. Maybe we need to revisit something there?

bkolad:
> Looks good to me, except that it feels like we might be breaking an abstraction by adding a dependency on `sov-stf-runner` in `demo-stf`. Maybe we need to revisit something there?

The underlying issue is that at some point we have to connect `sov_stf_runner::StateTransitionVerifier` with `demo_stf::Runtime`. The above approach adds the `sov-stf-runner -> demo-stf`  and assembles the types in the `demo-stf`. Alternativly we can import `sov_stf_runner & demo_stf` in both `demo-rollup` & `demo-prover` and create a 
```
pub type AppVerifier<DA, Zk> = StateTransitionVerifier<
    AppTemplate<
        ZkDefaultContext,
        <DA as DaVerifier>::Spec,
        Zk,
        Runtime<ZkDefaultContext, <DA as DaVerifier>::Spec>,
    >,
    DA,
    Zk,
>;
```
type in each crate but the downside is that we  will repeat the same type alias 2x.

preston-evans98 (approved):
Synced with @bkolad. We're going to merge this as-is and fix the abstraction in a follow-up

examples/demo-prover/methods/guest-mock/riscv32im-risc0-zkvm-elf.json
bkolad: fixed

PR #937, opened 2023-09-26
fix: require `native` when applicable for all targets

vlopes11:
Prior to this commit, some checks of all-targets without the feature `native` would break.

This commit introduces a fix for every workspace member to be consistent with the feature set.


citizen-stig:
I am curious, how this solution differs from adding "native" dev-dependency to itself, we have in some cases, for example for `sov-sequencer-registry`:

```toml
[dev-dependencies]
sov-sequencer-registry = { path = ".", features = ["native"] }
```

preston-evans98 (approved):

PR #938, opened 2023-09-26
feat: add addressable trait

vlopes11:
Prior to this commit, the module info was the sole responsible to
provide a module address with its dependencies.

If the module info is ever to increase in complexity, such complexity
will be leaked to functionality that expects to handle only its
addresses, such as sort modules per deps/check cyclic dependencies.

This commit introduces `Addressable`, a trait that will be implemented
for every `ModuleInfo` and will provide the addresses, when required.

### DIFF
--- a/module-system/sov-modules-api/src/lib.rs
+++ b/module-system/sov-modules-api/src/lib.rs
@@ -321,6 +321,7 @@ pub trait ModuleCallJsonSchema: Module {
 
 /// Every module has to implement this trait.
 pub trait ModuleInfo {
+    /// Execution context.
     type Context: Context;
 
     /// Returns address of the module.
@@ -333,10 +334,37 @@ pub trait ModuleInfo {
     fn dependencies(&self) -> Vec<&<Self::Context as Spec>::Address>;
 }
 
+/// A type that is addressable.
+pub trait Addressable {
+    /// Execution context.
+    type Context: Context;
+
+    /// Returns address of the instance.
+    fn address(&self) -> &<Self::Context as Spec>::Address;
+
+    /// Returns the addresses of all dependencies of this instance.
+    fn dependencies(&self) -> Vec<&<Self::Context as Spec>::Address>;
+}
+
+impl<T, C: Context> Addressable for T
+where
+    T: ModuleInfo<Context = C>,
+{
+    type Context = C;
+
+    fn address(&self) -> &<Self::Context as Spec>::Address {
+        self.address()
+    }
+
+    fn dependencies(&self) -> Vec<&<Self::Context as Spec>::Address> {
+        self.dependencies()
+    }
+}
+
 struct ModuleVisitor<'a, C: Context> {
     visited: HashSet<&'a <C as Spec>::Address>,
     visited_on_this_path: Vec<&'a <C as Spec>::Address>,
-    sorted_modules: std::vec::Vec<&'a dyn ModuleInfo<Context = C>>,
+    sorted_modules: std::vec::Vec<&'a dyn Addressable<Context = C>>,
 }
 
 impl<'a, C: Context> ModuleVisitor<'a, C> {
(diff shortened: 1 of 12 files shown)


vlopes11:
Blocked by #937

module-system/sov-modules-api/src/lib.rs
```diff
@@ -333,10 +334,37 @@ pub trait ModuleInfo {
...
+    type Context: Context;
+
+    /// Returns address of the instance.
+    fn address(&self) -> &<Self::Context as Spec>::Address;
+
+    /// Returns the addresses of all dependencies of this instance.
+    fn dependencies(&self) -> Vec<&<Self::Context as Spec>::Address>;
+}
+
+impl<T, C: Context> Addressable for T
```
bkolad, line 349: @vlopes11 can you please remind me why do we need `Addressable `. `ModuleInfo` has only one method more which shouldn't be problematic ` fn prefix(&self) -> Prefix;`? And `Addressable` requires `ModuleInfo`

bkolad: Ahh this is because `ModuleInfo` will have a GasConfig? But then why `GasConfig`  goes to `ModuleInfo`? It should be available via the `sov-modules-api::Module` trait.

vlopes11: It's mainly because we sometimes take a module as a dynamic dispatch of `ModuleInfo`

https://github.com/Sovereign-Labs/sovereign-sdk/blob/98bdc3b52414967c964ead063ab1784168d2b76e/module-system/sov-modules-api/src/lib.rs#L424-L430

While the `Context` will always be the same for all dependencies in a module, their `GasConfig` will vary. It means this code would break when we introduce an associated type that is not the same for the whole set.

This is a restriction of the language. It will not allow us to accept `Any` generic argument there, even if we don't interact at all with the `GasConfig` on this implementation. It would be a problem to set the size of the type on other contexts, but not for us here.

vlopes11: @bkolad we currently don't have a mechanism of implementing `sov-modules-api::Module` automatically via macro, as the `call` will be always customized for the module.

From my understanding, we should put all "metadata" into `ModuleInfo`, and have it implemented automatically (i.e. derive macro). The implication of moving `GasConfig` to `Module` is the user will have to define manually his `GasConfig`.

bkolad: I have the following reasons for putting it in the `Module`
1.  I don't think a `GasConfig.` is metadata. It is crucial for module development and interfaces directly with the developers.
2. `CallMessage` is different for each module, and it is the same for `GasConfig`, so it is more uniform to keep all the module-specific types in the `Module`. In various places of the macro system we do `<SomeModule as Module>::CallMessage` so if we ever need a`GC` we can uniformly do the same `<SomeModule as Module>::GasConfig` and this can be done inside `ModuleInfo` macro as well. 

What does it mean for the user to define `GC`  manually? I'm still not sure why `ModuleInfo` needs to interact with `GasConfig`. Can we discuss this further to clarify the mechanism involved?

citizen-stig: I second this about having `GasConfig` being its own trait, similar to `CallMessage`

vlopes11: As discussed on call, a couple of changes:

- move the `GasConfig` to `Module`
- move the `const` to `Default` implementation of `Module`
- create a derive macro helper attribute `gas` that will bind the type to the constant

vlopes11: This is how the bank module will look like. Let me know your thoughts:

https://gist.github.com/vlopes11/c6413473b6c38d456ac07beba7512775

vlopes11:
N/A, gas config moved to the `Module` trait

PR #939, opened 2023-09-27
offchain processing macro and example for sov-nft-module

dubbelosix:
# Description
This PR contains the code for `offchain` macro which enables creating functions that can update and manage offchain state/computation that doesn't impact on-chain state or consensus. 
Decorating a function with the offchain macro causes it to no-op unless the "offchain" feature is explicitly enabled. When the offchain feature is enabled while starting the rollup as described in `offchain_readme.md`, the code is executed.

Components added
* offchain macro in sov-module-macros
* offchain.rs inside the sov-nft-module for updating postgres tables in sync with transactions
* init_db.sql which has the scripts to create the necessary tables for offchain processing
* utils/nft-utils - a binary that can be used to submit transactions to the sequencer
* simple_client exists with `send_transaction`, which sends a single transaction in a batch. added `send_transactions` which takes a vector of transactions, serializes them into a batch and submits to the sequencer. optionally, send_transactions can also take a parameter that controls the batch size. For example `Some(3)` would mean that multiple batches of up to 3 transactions each are submitted, and `None` would mean that there is micro batching, and all transactions are submitted.

Outstanding Issues
* There is a timing/concurrency issue with the sequencer and the way it reads from state to accept/reject transactions, so the nft-utils script contains a sleep of 500ms between sends
* sequencer also performs an execute so offchain processing is being done twice. This is not an issue for any of the tables except "top_owners" since the operation is not idempotent (increments are done twice, once in batch_builder and once more in sov_modules_stf_template. investigating options to avoid this double processing

## Testing
The script added to create collections and mint NFTs at `utils/nft-utils` can also be used to test

## Docs
Docmentation specific to using and testing the offchain macro is at `module-system/module-implementations/sov-nft-module/offchain_readme.md`


full-node/sov-sequencer/src/utils.rs
```diff
@@ -38,6 +38,39 @@ impl SimpleClient {
         Ok(())
     }
 
+    /// Sends multiple transactions to the sequencer for immediate publication.
+    pub async fn send_transactions<Tx: BorshSerialize>(
+        &self,
+        txs: Vec<Tx>,
+        micro_batch: Option<usize>,
```
preston-evans98: nit: The name `micro_batch` doesn't convey that this is the chunk size very clearly. Maybe it would be clearer to call it `chunk_size`? 

We could also look at changing the API to `fn send_batches(&self, batches: Vec<Vec<Tx>>` and then we wouldn't need the extra param and logic here

module-system/module-implementations/sov-nft-module/Cargo.toml
```diff
@@ -21,6 +21,9 @@ sov-modules-api = { path = "../../sov-modules-api" }
 sov-modules-macros = {path = "../../sov-modules-macros"}
 sov-state = { path = "../../sov-state" }
 
+postgres = {version = "0.19.7", optional = true}
+tokio = {version = "1.32.0", features=["full"],optional = true}
```
preston-evans98: nit: formatting
```
postgres = { version = "0.19.7", optional = true }
tokio = { version = "1.32.0", features=["full"], optional = true }
```

module-system/module-implementations/sov-nft-module/offchain_readme.md
```diff
@@ -0,0 +1,56 @@
+## Offchain testing
```
preston-evans98, line 1: This readme could use some context on what the user is supposed to be accomplishing by following these steps. Can we add an intro to orient the user?

module-system/module-implementations/sov-nft-module/src/call.rs
```diff
@@ -87,6 +88,8 @@ impl<C: Context> NonFungibleToken<C> {
         )?;
         self.collections
             .set(&collection_address, &collection, working_set);
+        std::thread::sleep(std::time::Duration::from_millis(300));
```
preston-evans98: Two comments here: 
1. If we're going to introduce a sleep into code, we should definitely add a comment explaining it! 
2. I don't think we can safely add a sleep into the critical path of processing a call message. I don't want to block this tutorial on something unrelated to its functionality (adding synchronization to the sequencer), but I don't feel good about this. Users running the NFT demo will see 3 tps, and it won't be obvious why

module-system/module-implementations/sov-nft-module/src/init_db.sql
```diff
@@ -0,0 +1,47 @@
+-- Drop existing tables if they exist
```
preston-evans98, line 1: nit: The indentation of create table is all over the place. Is that fixable, or does some linter do this automatically?

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,168 @@
...
+use crate::{Collection, CollectionAddress, Nft};
+
+// CREATE TABLE collection (
+//     collection_address TEXT PRIMARY KEY,
+//     collection_name TEXT NOT NULL,
+//     creator_address TEXT NOT NULL,
+//     frozen BOOLEAN NOT NULL,
+//     metadata_url TEXT,
+//     supply BIGINT NOT NULL
+// );
```
preston-evans98: nit: Let's remove commented code or add some explanation of what it's doing here.

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,168 @@
...
+
+#[offchain]
+pub fn track_collection<C: sov_modules_api::Context>(collection: &Collection<C>) {
+    // data extraction
```
preston-evans98: nit: `// data extraction` -> `// Extract the necessary metadata from the collection`

module-system/module-implementations/sov-nft-module/src/offchain.rs
preston-evans98: nit: This comment is a little unclear. Does it go with the `track_collection function?` If so, why isn't it a doc comment. Also, let's add some explanation explaining what it means - is this a description of the logic that the track collection function implements?

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,168 @@
...
+//     collection_address TEXT PRIMARY KEY,
+//     collection_name TEXT NOT NULL,
+//     creator_address TEXT NOT NULL,
+//     frozen BOOLEAN NOT NULL,
+//     metadata_url TEXT,
+//     supply BIGINT NOT NULL
+// );
+
+#[offchain]
+pub fn track_collection<C: sov_modules_api::Context>(collection: &Collection<C>) {
```
preston-evans98: nit: `track_collection` implies something ongoing. How about `put_collection` or `insert_collection`?

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// CREATE TABLE nft (
+//     collection_address TEXT NOT NULL,
+//     nft_id BIGINT NOT NULL,
+//     metadata_url TEXT,
+//     owner TEXT NOT NULL,
+//     frozen BOOLEAN NOT NULL,
+//     PRIMARY KEY (collection_address, nft_id)
+// );
```
preston-evans98: nit: comment is unclear

module-system/module-implementations/sov-nft-module/src/offchain.rs
preston-evans98: Also, this function is missing a doc comment

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,168 @@
...
+
+#[offchain]
+pub fn track_nft<C: sov_modules_api::Context>(nft: &Nft<C>) {
```
preston-evans98: nit: Same comment as above about funciton name (maybe use "insert" or "put"?). Also, missing doc comment

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,168 @@
...
+                    if let Err(e) = result {
+                        println!("Failed to execute query: {}", e);
+                    }
+                }
+                Err(e) => {
+                    println!("Failed to connect to the database: {}", e);
+                }
+            }
+        } else {
+            println!("Environment variable POSTGRES_CONNECTION_STRING is not set");
```
preston-evans98: nit: Can we `tracing::error!` instead of `print`ing in all this cases, or do we want to avoid a `tracing` dependency here?

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+    })
+}
```
preston-evans98: Both of these `track_` functions are long and mostly duplicated. I think we can break them down and then factor some code into a shared helper. 

It looks to me like the two components are: 
1. Build a sql query from the data (depends on the shape of the item we're tracking)
2. Try to execute that query (can take query as a simple argument, identical in both cases)

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,168 @@
...
+
+// CREATE TABLE top_owners (
```
preston-evans98: nit: Same issue with comment

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,168 @@
...
+//     owner TEXT NOT NULL,
+//     collection_address TEXT NOT NULL,
+//     count BIGINT NOT NULL,
+//     PRIMARY KEY (owner, collection_address)
+// );
+
+#[offchain]
+pub fn update_top_owners<C: sov_modules_api::Context>(
```
preston-evans98: nit: missing doc comment

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,168 @@
...
+    collection_address: &CollectionAddress<C>,
+    owner_count_incr: Option<&[(String, u64)]>,
+    owner_count_decr: Option<&[(String, u64)]>,
```
preston-evans98: Question: why do we use two vecs instead of using a signed integer?

module-system/sov-modules-macros/src/lib.rs
```diff
@@ -263,3 +265,9 @@ pub fn address_type(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let input = parse_macro_input!(item as DeriveInput);
     handle_macro_error(address_type_helper(input))
 }
+
+#[proc_macro_attribute]
+pub fn offchain(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
preston-evans98, line 299: nit: Missing doc comment

utils/nft-utils/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+use sov_modules_api::transaction::Transaction;
+use sov_modules_api::{Address, PrivateKey};
+use sov_nft_module::utils::get_collection_address;
+use sov_nft_module::{CallMessage, CollectionAddress, UserAddress};
+use sov_rollup_interface::mocks::MockDaSpec;
+use sov_sequencer::utils::SimpleClient;
+
+const COLLECTION_1: &str = "Sovereign Squirrel Syndicate";
+const COLLECTION_2: &str = "Celestial Dolphins";
+const COLLECTION_3: &str = "Risky Rhinos";
```
preston-evans98, line 14: nice :)

utils/nft-utils/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
```
preston-evans98, line 1: We have a lot of packages in the workspace already. Did we look at options that don't add this at the top level? Maybe we can make this a binary in the other nft package?

utils/nft-utils/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+    161, 208, 245, 116, 93, 84, 37, 87, 171, 44, 30, 239,
+];
+
+fn get_collection_metadata_url(collection_address: &str) -> String {
+    format!("{}/collection/{}", DUMMY_URL, collection_address)
+}
+
+fn get_nft_metadata_url(collection_address: &str, nft_id: u64) -> String {
+    format!("{}/nft/{}/{}", DUMMY_URL, collection_address, nft_id)
+}
```
preston-evans98: This package mixes real functionality and mocks. Can we separate it cleanly into module with real helpers ("create collection", etc.), and then a binary with tge dummy data and main function? That way, someone could depend on the helpers and get a working implementation by just passing some actual data instead of mocks

utils/nft-utils/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+
+async fn create_collection(
```
preston-evans98: nit: missing doc comment

utils/nft-utils/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+        },
+    );
+    Transaction::<DefaultContext>::new_signed_tx(
+        signer,
+        create_collection_message.try_to_vec().unwrap(),
+        nonce,
+    )
+}
+
+async fn mint_nft(
```
preston-evans98: nit: missing doc comment

utils/nft-utils/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+            frozen: false,
+        });
+    Transaction::<DefaultContext>::new_signed_tx(
+        signer,
+        mint_nft_message.try_to_vec().unwrap(),
+        nonce,
+    )
+}
+
+async fn transfer_nft(
```
preston-evans98: nit: missing doc comment

utils/nft-utils/src/main.rs
preston-evans98: I don't think this function needs to be async. Am I missing something?

utils/nft-utils/src/main.rs
preston-evans98: Also, Same question about async as above

utils/nft-utils/src/main.rs
preston-evans98: Same question about async as above

utils/nft-utils/src/main.rs
preston-evans98: Same comment about naming as above

utils/nft-utils/src/main.rs
preston-evans98: Also, this name is a bit misleading. Can we rename to something which conveys that this is just generating a call message, not actually creating the collection `build_create_collection_msg` or something?

utils/nft-utils/src/main.rs
preston-evans98: Same comment about naming as above

utils/nft-utils/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+    transactions.push(create_collection(&creator_pk, nonce, COLLECTION_1).await);
+    nonce += 1;
+    transactions.push(create_collection(&creator_pk, nonce, COLLECTION_2).await);
+    nonce += 1;
+    transactions.push(create_collection(&creator_pk, nonce, COLLECTION_3).await);
+
+    client.send_transactions(transactions, None).await.unwrap();
+    // sleep is necessary because of how the sequencer currently works
+    // without the sleep, there is a concurrency issue and some transactions would be ignored
+    thread::sleep(Duration::from_millis(1000));
```
preston-evans98: This comment is great. Can we add an issue to track the problem and link it here as well?

full-node/sov-sequencer/src/utils.rs
preston-evans98: Update: We never use `micro_batch`. Can we just remove it?

utils/nft-utils/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+    let mut transactions = vec![];
+
+    let mut owner_1_nonce = 0;
+    let mut nft_id = 1;
+    let collection_1_address = get_collection_address::<DefaultContext>(
+        COLLECTION_1,
+        creator_pk.default_address().as_ref(),
+    );
+
+    #[allow(clippy::explicit_counter_loop)]
```
preston-evans98: Can we use the loop counter instead of using separate `owner_1_nonce` and `nft_id` variables here? Will make clippy happy and remove a little code.

utils/nft-utils/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+                &creator_pk,
+                nonce,
+                COLLECTION_1,
+                nft_id,
+                &owner_1_pk.default_address(),
+            )
+            .await,
+        );
+        nft_id += 1;
+    }
```
preston-evans98: nit: We have several of these loops with mutable nonces etc. it's not the end of the world, but feels a bit messy. Maybe we can simplify with an iterator, and/or make a helper function?

module-system/module-implementations/sov-nft-module/src/call.rs
dubbelosix: oh yeah, this was just to illustrate the issue but it got checked in accidentally. removed now!

full-node/sov-sequencer/src/utils.rs
dubbelosix: I use the second parameter primarily while switching between MockDA and celestia.
switched the logic to use Vec<Vec<Tx>> so that the chunking can be handled by the caller

module-system/module-implementations/sov-nft-module/src/offchain.rs
dubbelosix: changed the name to update_collection!

utils/nft-utils/src/main.rs
dubbelosix: ah yeah, the asyncs were part of a prior version where create_collection also ended up sending the transaction directly using simple_client. originally tested with celestia docker and it was too slow, so i added the logic to fetch the transactions and send batches of multiple. both the name and the async ended up remaining. fixed now!

module-system/module-implementations/sov-nft-module/src/offchain.rs
dubbelosix: ah yeah, this logic is changed now with merging update_nft and update_top_owners (original reason was to decerement counts of owner1 and increment counts of owner2 for a transfer, but that logic doesnt matter right now)

module-system/module-implementations/sov-nft-module/src/offchain.rs
dubbelosix: dependency should be fine! the more offchain or convenience libs and methods we can use, the more it shows the benefits of using the offchain component (in fact they can even add sentry logging etc and would not need to rely on emitting events or using on-chain compute). 
will add tracing in

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,142 @@
...
+        } else {
+            tracing::error!("Environment variable POSTGRES_CONNECTION_STRING is not set");
+        }
+    })
+}
+
+/// Syncs an NFT to the corresponding table "nfts" in postgres
+/// Additionally, this function also has logic to track the counts of NFTs held by each user
+/// in each collection.
+#[offchain]
```
bkolad, line 59: Can this macro be used with `Module` attributes. I am thinking that the `AccesorryXX` should be always behind offchain. @neysofu wdt?

module-system/module-implementations/sov-nft-module/src/offchain.rs
```diff
@@ -0,0 +1,142 @@
...
+                    "INSERT INTO top_owners (owner, collection_address, count) VALUES ($1, $2, 1) \
+                    ON CONFLICT (owner, collection_address) \
+                    DO UPDATE SET count = top_owners.count + 1",
+                    &[&new_owner_str, &collection_address],
+                );
+            }
+
+            // Update NFT information after handling top_owners logic
+            let _ = client.execute(
+                "INSERT INTO nfts (\
```
bkolad: What about extracting all the sql strings as constants in a separate `sql.rs` file?

module-system/module-implementations/sov-nft-module/src/offchain.rs
neysofu: I very much agree, I think accessory state should be zero-cost and only written to during native execution. The current model works but we need to integrate it more tightly with `offchain`. Tracked in https://github.com/Sovereign-Labs/sovereign-sdk/issues/944.

bkolad: Cool, @dubbelosix feel free to resolve this comment. We will do it is as #944.

utils/nft-utils/Cargo.toml
dubbelosix: had a discussion with team and it would be good to come up with a more generalized structure to place util libs and util bins for each module. created an issue to track this https://github.com/Sovereign-Labs/sovereign-sdk/issues/957 for now, resolving the comment and keeping the structure at `sovereign/utils/nft-utils`

bkolad (approved):

PR #947, opened 2023-09-28
EVM: Implement estimate_gas

### LINKED ISSUE #906: EVM: Implement gas limit estimation
# Summary
Implement gas limit estimation so that we can use it in eth_sendTransaction endpoint. It needs to take a transaction, replay it in evm module, and return the estimated gas consumption for the transaction.

orkunkilic:
# Description
This PR implements estimate_gas endpoint which estimates the gas used by transaction. It is a copy of reth's code but with small differences to be compatible with our module system.

This also adds gas estimation feature to eth_sendTransaction endpoint.

## Linked Issues
- Fixes #906
- Related to #502

## Testing
Integration tests are added.

## Docs

### DIFF
--- a/full-node/sov-ethereum/src/lib.rs
+++ b/full-node/sov-ethereum/src/lib.rs
@@ -18,9 +18,9 @@ pub mod experimental {
     use jsonrpsee::types::ErrorObjectOwned;
     use jsonrpsee::RpcModule;
     use reth_primitives::{
-        Address as RethAddress, TransactionSignedNoHash as RethTransactionSignedNoHash,
+        Address as RethAddress, TransactionSignedNoHash as RethTransactionSignedNoHash, U128, U256,
     };
-    use reth_rpc_types::{TransactionRequest, TypedTransactionRequest};
+    use reth_rpc_types::{CallRequest, TransactionRequest, TypedTransactionRequest};
     use sov_evm::{CallMessage, Evm, RlpEvmTransaction};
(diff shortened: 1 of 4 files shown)


orkunkilic:
My main concern on this PR is that it copies evm_db a lot because it performs several evm inspect runs. This will happen outside the zk context and it is also the same with Reth so I guess it is not a big problem but please share your thoughts.

full-node/sov-ethereum/src/lib.rs
```diff
@@ -225,21 +225,74 @@ pub mod experimental {
                     .map(|id| id.as_u64())
                     .unwrap_or(1);
 
-                // TODO: implement gas logic after gas estimation (#906) is implemented
-                // https://github.com/Sovereign-Labs/sovereign-sdk/issues/906
+                // TODO: we need an oracle to fetch the gas price of the current chain
```
bkolad: `eth_sendTransaction` is quite long let' split it into several methods? We can also crate a file `eth_sendTransaction .rs` and move the code there.

bkolad:
> My main concern on this PR is that it copies evm_db a lot because it performs several evm inspect runs. This will happen outside the zk context and it is also the same with Reth so I guess it is not a big problem but please share your thoughts.

Yes this is fine, we have a `cache` in the `WrokingSet` so it won't be that have on the db.

bkolad (approved):

PR #948, opened 2023-09-28
feat(celestia-adapter): update to celestia-node v0.11 and celestia app v1.0

### LINKED ISSUE #461: Sov-celestia-local container needs Dockerfile so that we can modify, update and build other versions of celestia binaries

### LINKED ISSUE #480: sov-celestia-local is opaque to users since nothing is displayed after "make start"
## Summary

With sov-celestia-local, the goal is to provide users a local celestia environment to develop against using the Sovereign SDK. Currently the docker container setup and start is all abstracted under one command `make start`. The problem here is that once it completes, the logs from celestia are hidden from the end user so they have no idea what's happening inside the celestia container.

The fix for this should be relatively easy since the logs are already there - just not exposed to the developer.
1. Add instruction in README to run `docker logs -f sov-celestia-local`
2. Let make start automatically run the above command so that they don't need to manually run it and can also see the container logs in another tab

zvolin:
# Description

This updates the `sov-celestia-adapter` to the latest (at the time of submitting) versions of celestia node (0.11.0-rc14) and app (1.0.0-rc17).

Sorry for such a big PR, but it was hard to introduce changes separately there

changes includes:
- a complete rewrite of celestia docker, this includes both `/docker` compose setup and the `Dockerfile` used in `demo-rollup/Makefile`. Now there is a single source of truth inside the `docker`. I tried to support all the existing use-cases, but please check if I haven't broken your workflows :pray: Also there seems to currently be no easy way to stop the running network in a way to resume it later, so it's recreated each time.
- usage of `celestia-types` in `sov-celestia-adapter`. `celestia-types` is a crate we maintain at @eigerco that consists of types compatible with celestia's protobuf and json API's. That way you shouldn't worry that you'll fall incompatible with the celestia network as we run integration tests for all those types with the real celestia-node. `celestia-types` have very little logic itself (yet) and is mostly about conforming to correct representations, so some types stayed as they were, but I tried to replace everything that was possible currently and we can then try to upstream the logic bit by bit.
- removal of `borsh` wherever `celestia-types` were included. We currently don't support borsh, but we can introduce the support for it in `celestia-types` wherever we don't include the `tendermint` types. All those places were currently marked as `TODO`
- recreation and addition of test data, updating all the tests and introducing a few new ones

## Linked Issues
- Fixes #461, fixes #480, 
- Related to # (issue) 

## Testing

Ran tests for sov-celestia-adapter
ran  demo-prover (with SKIP_PROVER=1) and demo-rollup + make test-create-token

## Docs

Updated the docs were relevant, also included some inline documentation


zvolin:
I gotta check that https://github.com/Sovereign-Labs/sovereign-sdk/actions/runs/6338399051/job/17219162315

zvolin:
@preston-evans98 so I allowed the [`Zlib`](https://www.zlib.net/zlib_license.html) licenses in the `deny.toml` as it seems minimal and very permissive. If you don't want to allow it for the sovereign sdk tho, I can just hardcode the constants that we build with `const_format` (the crate with zlib license) directly in `celestia-types`

.github/workflows/rust.yml
```diff
@@ -145,6 +149,10 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: true
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v2
+        with:
+          version: "23.2"
```
preston-evans98, line 170: Where does the protoc compiler get introduced? Is this from `celestia-types`?

zvolin: It's a dependency of our other pkg, `celestia-proto`.

zvolin: It's used to build MsgPayForBlobs and other celestia and cosmos relevant protos

adapters/celestia/src/verifier/mod.rs
```diff
@@ -182,20 +184,19 @@ impl da::DaVerifier for CelestiaVerifier {
...
             return Err(ValidationError::MissingTx);
         }
-        let namespace = NamespaceGroup::from_shares_unchecked(rollup_shares_u8);
 
         // Check the e-tx proofs...
         // TODO(@preston-evans98): Remove this logic if Celestia adds blob.sender metadata directly into blob
         let mut tx_iter = txs.iter();
         let square_size = block_header.dah.row_roots.len();
-        for (blob, tx_proof) in namespace.blobs().zip(inclusion_proof.into_iter()) {
+        for (blob, tx_proof) in verified_shares.blobs().zip(inclusion_proof.into_iter()) {
```
preston-evans98: This doesn't look like a new bug - but I think this check is broken. If `inclusion_proof` is an empty iterator, this loop will never execute. Looks like we need to check that `inclusion_proof.len()` == `verified_shares.blobs().len()`

.github/workflows/rust.yml
preston-evans98: Makes sense, thanks!

preston-evans98:
Thanks for the hard work here @zvolin! Generally the diff looks fine - I feel pretty confident that the changes here didn't introduce any new bugs. Unfortunately, I'm not at all confident that there aren't important bugs in the verifier (in fact, I think I spotted one fatal one during this review). 

Any updates from the Celestia side about incorporating `blob.sender` information into the data namespace directly? That would let us drastically simplify the logic and remove nearly all of the surface for bugs. If so, let's keep pushing! If not, we may need to consider a significant refactor or even a total rewrite of the verifier logic

adapters/celestia/src/verifier/mod.rs
zvolin: this case should be covered now

zvolin:
> Any updates from the Celestia side about incorporating blob.sender information into the data namespace directly

Yes.

So I was initially searching for a way to just get relevant PFB's with proofs, so that the adapter doesn't need to fetch the entire EDS and recreate the proofs for only the PFB's it's interested in. The celestia team is aware of the issue that it's currently not possible to get the metadata relevant only for the blobs of the given namespace and will likely introduce some way to do it gently in a future, perhaps with an addition of a new node type, partial node (or namespaced node).

Quoting @Wondertan:

> FN and PN can theoretically index the reserved namespace to store that info and provide over API. If that becomes a very common usecase and we decide that core protocol should not handle that, then the node will likely provide this functionality.
>
> The node will need to optimize its BlobModule proofs. Proving the inclusion of a blob is equivalent to the inclusion of its PFB and the PFB inclusion proof is much smaller than the blob’s.
To achieve this FNs and PNs would need to parse the reserved namespace which contains all PFBs and metadata such as signer info. 
>
> Besides just described optimization, there is your use case which would benefit from node parsing and indexing PFBs for blobs. And in general, allowing  devs using our APIs to inspect PFBs is a huge enabler for various dashboards and explorers

I asked about the inclusion of the `sender` (or `signer`) metadata directly into blob shares and the chances are rather slim, again quoting @Wondertan:

> That's a simpler but consensus breaking solution and why I involved app team. My problem is that does not scale for other fields inside of PFB tx and duplicates data. We don't wanna break consensus everytime someone needs to map blob to some metadata in PFB and abstracting away this (necessary) separation between Blobs and PFBs on the node's API seems like a better solution to me.

adapters/celestia/src/verifier/mod.rs
zvolin: https://github.com/Sovereign-Labs/sovereign-sdk/pull/948/commits/d755a2d3a2a6b15da77ea82abdb553da554ffcc0

preston-evans98: Fanstastic, thanks @zvolin!

preston-evans98:
> Quoting @Wondertan:

Thanks for the update! Unfortunately, I don't @Wondertan has the full picture on the use case here (changing node APIs doesn't help in this case, since we need to cryptographic verification). I'll ping the Celestia folks and see if we can get discussions started again

zvolin:
> changing node APIs doesn't help in this case, since we need to cryptographic verification

I was asking for a way to obtain the relevant pfb's with inclusion proofs just for them. Currently when we fetch the whole PFB namespace we get all the pfbs for block with proof for all the pfbs even tho we care only about the few. The adapter currently fetches the entire EDS and recreates the proofs only for relevant pfbs and that was a problem I was asking to address, thus the node's api change.

zvolin:
Including the sender directly in blob's shares is kind of related but also separated topic and only the last quote is in regards to that

preston-evans98:
Got it, thanks for the additional context

zvolin:
rebased and fixed stuff from previous CI run

zvolin:
rebased one more time. aligned new method name in tests from #977 and reverted changes to celestia addresses from #972. Could you retrigger one more time please @preston-evans98 ?

preston-evans98 (approved):

zvolin:
rebased

zvolin:
Closing in favor of #997

PR #950, opened 2023-09-28
EVM: Fix hex prefix on block number

orkunkilic:
# Description
EVM applications generally use hex strings with 0x prefix so block_number call was failing for example. This pr supports both 0x and plain string version.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


bkolad (approved):

orkunkilic:
Fixed in #964

PR #960, opened 2023-09-29
Use build script to install cargo-risczero

preston-evans98:
# Description
Try to install cargo-risczero during demo-prover builds if absent.


cemozerr (approved):
LGTM.

Issue #915, opened 2023-09-25
`module-system`: read genesis data from files.


bkolad:
## Summary
The genesis data is hardcoded in the code but should be read from a file instead.
List of modules:
   - [ ] module-template
   - [x] sov-blob-storage
   - [x] sov-chain-state 
   - [x] sov-nft
   - [x] sov-simple-nft
   - [x] sov-accounts
   - [x] sov-bank 
   - [ ] sov-prover-incentives
   - [ ] sov-attester-incentives
   - [x] sov-sequencer-registry 
   - [x] sov-evm 
   - [x] value-setter
   - [ ] vec-setter
   - [ ] accessory-state

PR #972, opened 2023-10-02
Initialize genesis state form genesis files.

bkolad:
# Description
This PR introduces the following chnages:
1. Added `genesis` file for `sov-sequencer-registry`
2. Implemented custom serde logic for CelestiaAddress and MockAddress.
3. The genesis state is now read from files.

## Linked Issues
- Related to #915

## Testing
CI passes.

## Docs
No changes.


adapters/celestia/src/verifier/address.rs
```diff
@@ -15,11 +14,41 @@ const VARIANT: bech32::Variant = bech32::Variant::Bech32;
...
 /// Spec says: "Addresses have a length of 32 bytes.", but in reality it is 32 `u5` elements, which can be compressed as 20 bytes.
 /// TODO: Switch to bech32::u5 when it has repr transparent: <https://github.com/Sovereign-Labs/sovereign-sdk/issues/646>
-#[derive(
-    Debug, PartialEq, Clone, Eq, Serialize, Deserialize, BorshDeserialize, BorshSerialize, Hash,
-)]
+#[derive(Debug, PartialEq, Clone, Eq, BorshDeserialize, BorshSerialize, Hash)]
 pub struct CelestiaAddress([u8; 32]);
 
+impl serde::Serialize for CelestiaAddress {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
```
bkolad, line 21: Taken from: https://github.com/Sovereign-Labs/sovereign-sdk/pull/979/files#diff-1f70d28d7a9cb24577e66718a40d689fbc35945816d11ae912d1cd68e9f81300R21

rollup-interface/src/state_machine/mocks/da.rs
```diff
@@ -19,26 +20,47 @@ use crate::{BasicAddress, RollupAddress};
...
     pub addr: [u8; 32],
 }
 
-impl core::str::FromStr for MockAddress {
+impl serde::Serialize for MockAddress {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        if serializer.is_human_readable() {
```
neysofu, line 40: TIL! Pretty cool.

examples/demo-stf/src/genesis_config.rs
```diff
@@ -32,63 +43,58 @@ pub const DEMO_TOKEN_NAME: &str = "sov-demo-token";
...
+        let token_address = sov_bank::get_genesis_token_address::<C>(
+            &bank_config.tokens[0].token_name,
+            bank_config.tokens[0].salt,
+        );
 
-    let nft_config: NonFungibleTokenConfig = NonFungibleTokenConfig {};
+        assert_eq!(
+            sequencer_registry_config.coins_to_lock.token_address,
+            token_address
+        );
```
neysofu: This crash results from user-provided configuration values instead of a logic error in the code, so it should probably be an `anyhow::Error`.

module-system/module-implementations/sov-sequencer-registry/src/tests.rs
neysofu, line 1: It looks like this file can be put into `sov-sequencer-registry/tests/` as an integration test instead, which might be preferable if we're not relying on any internal details of the module. WDYT?

neysofu (commented):
Looks great! I left two comments below.

The `GenesisPaths` pattern is a bit verbose, but can can remove it either with the discussed env. var. approach or with default path values, so it's not a big deal.

module-system/module-implementations/sov-sequencer-registry/src/tests.rs
bkolad: I thought about it more as an unit test that integration test that's way I put it inside src. We could move this to /tests but we will need to do it for all other modules to be uniform. So if we decide to do it let's make it in a separate PR.

examples/demo-stf/src/genesis_config.rs
bkolad: fixed

module-system/module-implementations/sov-sequencer-registry/src/tests.rs
neysofu: Consistency is a good point; let's leave it as it is.

examples/demo-stf/src/genesis_config.rs
neysofu: Fixed by https://github.com/Sovereign-Labs/sovereign-sdk/pull/972/commits/cb8932ccdb122f2e71ea97a5229148b7835753c7.

neysofu (approved):

PR #978, opened 2023-10-03
Removing `Storage::backing` method

citizen-stig:
# Description
As it heavily leaks abstraction.

Replace it with `fn get_with_proof(&self, key: StorageKey) -> Option<StorageProof<Self::Proof>> ` which at least does not expose underlying storage.

## Testing
Current tests are passing

## Docs
No updates in documentation

### DIFF
--- a/module-system/module-implementations/sov-attester-incentives/src/query.rs
+++ b/module-system/module-implementations/sov-attester-incentives/src/query.rs
@@ -62,9 +62,7 @@ where
     where
         C::Storage: NativeStorage,
     {
-        let prefix = self.bonded_attesters.prefix();
-        let codec = self.bonded_attesters.codec();
-        working_set.get_with_proof(StorageKey::new(prefix, &address, codec))
+        working_set.get_with_proof(self.get_attester_storage_key(address))
     }
 
     /// TODO: Make the unbonding amount queryable:
--- a/module-system/module-implementations/sov-attester-incentives/src/query.rs
+++ b/module-system/module-implementations/sov-attester-incentives/src/query.rs
@@ -1,7 +1,7 @@
 //! Defines the query methods for the attester incentives module
 use serde::{Deserialize, Serialize};
 use sov_modules_api::{Spec, ValidityConditionChecker, WorkingSet};
-use sov_state::storage::{Storage, StorageKey, StorageProof};
+use sov_state::storage::{NativeStorage, Storage, StorageKey, StorageProof};
 
 use super::AttesterIncentives;
 use crate::call::Role;
@@ -54,12 +54,14 @@ where
     /// Used by attesters to get a proof that they were bonded before starting to produce attestations.
     /// A bonding proof is valid for `max_finality_period` blocks, the attester can only produce transition
     /// attestations for this specific amount of time.
-    #[cfg(feature = "native")]
     pub fn get_bond_proof(
         &self,
         address: C::Address,
         working_set: &mut WorkingSet<C>,
-    ) -> StorageProof<<C::Storage as Storage>::Proof> {
+    ) -> StorageProof<<C::Storage as Storage>::Proof>
+    where
+        C::Storage: NativeStorage,
+    {
         let prefix = self.bonded_attesters.prefix();
         let codec = self.bonded_attesters.codec();
         working_set.get_with_proof(StorageKey::new(prefix, &address, codec))
--- a/module-system/module-implementations/sov-attester-incentives/src/query.rs
+++ b/module-system/module-implementations/sov-attester-incentives/src/query.rs
@@ -1,7 +1,7 @@
 //! Defines the query methods for the attester incentives module
 use serde::{Deserialize, Serialize};
 use sov_modules_api::{Spec, ValidityConditionChecker, WorkingSet};
-use sov_state::storage::{NativeStorage, Storage, StorageKey, StorageProof};
+use sov_state::storage::{Storage, StorageKey};
 
 use super::AttesterIncentives;
 use crate::call::Role;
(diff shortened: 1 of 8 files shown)


module-system/module-implementations/sov-attester-incentives/src/query.rs
```diff
@@ -44,22 +44,11 @@ where
         }
     }
 
-    /// Used by attesters to get a proof that they were bonded before starting to produce attestations.
-    /// A bonding proof is valid for `max_finality_period` blocks, the attester can only produce transition
-    /// attestations for this specific amount of time.
-    pub fn get_bond_proof(
```
citizen-stig, line 50: @preston-evans98 how important is it for actual users?

Could we provide this functionality just from witness?

citizen-stig: Another solution would be to generalize it full node somehow, so you can take proof for any prefix, then clients just have to query storage prefix for address (or build it themself).

Or it should be part of WorkingSet API

preston-evans98: Hmm, we do definitely need to expose this method via RPC. Otherwise, there's no way to generate attestations because you can't prove that you're bonded. I think you're right that it should eventually be generalized to allow creating a merkle proof for any storage key, but I don't have a strong opinion about whether we should do that now or later

preston-evans98 (commented):
We do need the `get_bond_proof` method for optimistic rollups (and in other places), so I would definitely prefer not to remove it. If we don't remove it, what options do we have to unblock you?

module-system/module-implementations/sov-attester-incentives/src/query.rs
citizen-stig: I need to investigate if I making this part of WorkingSet will solve the problem.

citizen-stig: Did it differently

module-system/sov-modules-api/src/state/scratchpad.rs
```diff
@@ -268,6 +262,15 @@ impl<C: Context> WorkingSet<C> {
...
         self.gas_meter.charge_gas(gas)
     }
+
+    #[cfg(feature = "native")]
+    pub fn get_with_proof(
+        &mut self,
+        key: StorageKey,
+    ) -> sov_state::storage::StorageProof<<C::Storage as Storage>::Proof> {
+        // First inner is `RevertableWriter` and second inner is actually a `Storage` instance
+        self.delta.inner.inner.get_with_proof(key).unwrap()
```
bkolad: Can this unwrap panic?

module-system/sov-state/src/prover_storage.rs
```diff
@@ -191,25 +185,25 @@ impl<S: MerkleProofSpec> Storage for ProverStorage<S> {
...
-        &self,
-        key: StorageKey,
-        _witness: &Self::Witness,
-    ) -> StorageProof<Self::Proof> {
+    // Based on assumption `validate_and_commit` increments version.
+    fn is_empty(&self) -> bool {
+        self.db.get_next_version() <= 1
+    }
+
+    #[cfg(feature = "native")]
```
bkolad: I believe ` #[cfg(feature = "native")]` is not needed here.

module-system/module-implementations/sov-attester-incentives/src/query.rs
```diff
@@ -44,22 +44,25 @@ where
...
+    pub fn get_attester_storage_key(&self, address: C::Address) -> StorageKey {
+        let prefix = self.bonded_attesters.prefix();
+        let codec = self.bonded_attesters.codec();
+        StorageKey::new(prefix, &address, codec)
+    }
+
     /// Used by attesters to get a proof that they were bonded before starting to produce attestations.
     /// A bonding proof is valid for `max_finality_period` blocks, the attester can only produce transition
     /// attestations for this specific amount of time.
+    #[cfg(feature = "native")]
```
bkolad: The `query.rs` is behind `native` already.

module-system/sov-state/src/storage.rs
```diff
@@ -237,14 +237,20 @@ pub trait Storage: Clone {
...
         proof: StorageProof<Self::Proof>,
     ) -> Result<(StorageKey, Option<StorageValue>), anyhow::Error>;
 
     /// Indicates if storage is empty or not.
     /// Useful during initialization.
     fn is_empty(&self) -> bool;
+
+    /// Returns the value corresponding to the key or None if key is absent and a proof to
+    /// get the value.
+    /// Only supposed to be used in native context
```
bkolad: We can't assume this here as  the `Storage` trait has to work with any context. We already have this: 
https://github.com/Sovereign-Labs/sovereign-sdk/blob/nightly/module-system/sov-state/src/storage.rs#L273

module-system/sov-modules-api/src/state/scratchpad.rs
citizen-stig: Yes, unfortunately. It will panic if underlying Storage implementation does not have implementation. For example if it called in Zk context by accident. I think native flag should protect against it.

Another option would be to have a trait definition to be different in `native` and `zk`, then we can get rid of the option

module-system/module-implementations/sov-attester-incentives/src/query.rs
```diff
@@ -44,22 +44,25 @@ where
...
     pub fn get_bond_proof(
         &self,
         address: C::Address,
-        witness: &<<C as Spec>::Storage as Storage>::Witness,
         working_set: &mut WorkingSet<C>,
-    ) -> StorageProof<<C::Storage as Storage>::Proof>
-    where
-        C::Storage: NativeStorage,
```
bkolad, line 57: Isn't  ` C::Storage: NativeStorage` needed to avoid: https://github.com/Sovereign-Labs/sovereign-sdk/pull/978/files#r1360613601

module-system/sov-state/src/prover_storage.rs
citizen-stig: I followed approach of

```
    #[cfg(feature = "native")]
    fn get_accessory(&self, key: &StorageKey) -> Option<StorageValue> {
    ```

File: module-system/module-implementations/sov-attester-incentives/src/query.rs (hunk 4 above)
Comment (citizen-stig) [reply module-system/module-implementations/sov-attester-incentives/src/query.rs]: That's a good point!

File: module-system/sov-modules-api/src/state/scratchpad.rs [hunk 7]
```diff
@@ -16,6 +17,7 @@ pub struct Delta<S: Storage> {
     inner: S,
     witness: S::Witness,
     cache: StorageInternalCache,
+    parent: PhantomData<S>,
```
bkolad: I think this was added by accident.

citizen-stig: Yes, it was!

bkolad (approved):

PR #979, opened 2023-10-03
Create rollup template

preston-evans98:
# Description
This PR implements a basic rollup template, which combines an STF and the "full-node" implementation into a single package. A follow-up will add the following missing functionality:
- Working CLI
- Prover/verifier logic

## Testing
One end-to-end test has been added. Additional test coverage will be added in follow-ups.


templates/rollup-template/src/bin/node.rs
```diff
@@ -0,0 +1,56 @@
...
+    let da_service = start_da_service(&rollup_config).await;
+
+    // Load genesis data from file
+    let genesis_path = env::args()
+        .nth(1)
+        .map(PathBuf::from)
+        .unwrap_or_else(|| PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("genesis.json"));
+    info!("Reading genesis configuration from {genesis_path:?}");
+    let genesis_config =
+        std::fs::read_to_string(genesis_path).context("Failed to read genesis configuration")?;
```
bkolad: I made a similar change in `demo-rollup, demo-prover, benches, tests` but instead of reading the whole genesis I read each module file separatly: #972.

Let's coordinate on daily.

templates/rollup-template/rollup_config.toml
```diff
@@ -0,0 +1,23 @@
+[da]
+# The JWT used to authenticate with the celestia light client. Instructions for generating this token can be found in the README
+celestia_rpc_auth_token = "MY.SECRET.TOKEN"
```
bkolad: What are your thoughts on using `MockDa`. This way our unit test also test that if the `main` works.

templates/rollup-template/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+#[cfg(feature = "native")]
```
bkolad: Let's add
```
#![deny(missing_docs)]
```

templates/rollup-template/tests/test_genesis.json
```diff
@@ -0,0 +1,64 @@
...
+					"sov1l6n2cku82yfqld30lanm2nfw43n2auc8clw7r5u5m6s7p8jrm4zqrr8r94"
+				],
+				"salt": 0
+			}
+		]
+	},
+	"sequencer_registry": {
+		"seq_rollup_address": "sov1l6n2cku82yfqld30lanm2nfw43n2auc8clw7r5u5m6s7p8jrm4zqrr8r94",
+		"seq_da_address": [
+			1,
```
bkolad, line 25: With the custom serde change it should be a human readable address.

templates/rollup-template/src/rollup.rs
```diff
@@ -0,0 +1,100 @@
...
+
+use crate::builder::StfWithBuilder;
+use crate::rpc::{register_ledger, register_sequencer};
+use crate::runtime::{get_rpc_methods, GenesisConfig, Runtime};
+
+type ZkStf<Da, Vm> = AppTemplate<ZkDefaultContext, Da, Vm, Runtime<ZkDefaultContext, Da>>;
+
+/// Dependencies needed to run the rollup.
+// This is duplicated exactly from demo-rollup. Should go to stf-runner crate?
+pub struct Rollup<Vm: ZkvmHost, Da: DaService + Clone> {
```
bkolad: It looks like the `new`, `run_and_report_rpc_port`, and the `jsonrpsee::RpcModule` registration logic is common for any rollup.

Would it make sense to introduce:
```rust
struct RollupTemplate <GenesiConfig, R: Runtime>{
     pub runtime: R,
     pub da_service: Da,
     pub ledger_db: LedgerDB,
     pub runner_config: RunnerConfig,
     pub genesis_config: GenesisConfig,
     #[allow(clippy::type_complexity)]
     pub prover: Option<Prover<ZkStf<Da::Spec, Vm::Guest>, Da, Vm>>,
     ///======  This is new
     methods: Vet<jsonrpsee::RpcModule<()>
} 
```
And implement `run_and_report_rpc_port ` generically (we could merge all the `methods` in a loop).
This structure would serve a similar role to `AppTemplate`.
 
`AppTemplate` provides implementation of STF for free
`RollupTemplate` provides implementation of a Rollup for free 
And they could leave in the same place. 

The `demo-rollup` & `this package (we would need to change the name)` would be an instantiation of `RollupTemplate ` with different dependencies.

examples/demo-stf/src/hooks_impl.rs
```diff
@@ -41,6 +44,7 @@ impl<C: Context, Da: DaSpec> ApplyBlobHooks<Da::BlobTransaction> for Runtime<C,
         blob: &mut Da::BlobTransaction,
         working_set: &mut WorkingSet<C>,
     ) -> anyhow::Result<()> {
+        // Before executing each batch, check that the sender is regsitered as a sequencer
```
cemozerr:
```suggestion
        // Before executing each batch, check that the sender is registered as a sequencer
```

templates/rollup-template/rollup_config.toml
cemozerr: +1

templates/rollup-template/src/builder.rs
```diff
@@ -0,0 +1,45 @@
+//! This module implements the batch builder for the rollup.
+//! To swap out the batch builder, simply replace the
+//! FifoStructBatchBuilder in `StfWithBuilder` with a type of your choosing.
```
cemozerr:
```suggestion
//! FifoStrictBatchBuilder in `StfWithBuilder` with a type of your choosing.
```

cemozerr: For customizability, this is probably good for now, but in the future I hope we can get even simpler abstractions, where you basically set this from the rollup config file or something.

cemozerr (commented):
The PR looks great. I'll leave it to someone more day-to-day in the codebase to leave a complete review with an approve.

templates/rollup-template/src/lib.rs
preston-evans98: Done

templates/rollup-template/rollup_config.toml
preston-evans98: I think it's probably preferable to use a local Celestia devnet like we do for demo-rollup. I was planning that for a follow-up - @cemozerr would you prefer a Mock instead? It's easy to switch, but I feel like it's nice for users to see the real thing instead of a mock

cemozerr: I actually think it'd be better to have mock DA for demo-rollup too. Easier to spin up the better IMO.

templates/rollup-template/src/rollup.rs
preston-evans98: Yeah, this definitely makes sense. Planning to do this in the next PR

templates/rollup-template/rollup_config.toml
preston-evans98: Ok, switched to mocks

bkolad (approved):

bkolad (approved):

bkolad (approved):

PR #984, opened 2023-10-04
fix: broken fuzz lib

vlopes11:
Prior to this commit, there was no CI check for the fuzz library.

Since it is isolated from the workspace (so it won't be affected by fuzz specific build configuration), it could be silently broken by a breaking change.

This commit introduces a fix to `AccountConfig`, and adds the fuzz check to the link job of the `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,10 +24,11 @@ install-dev-tools:  ## Installs all necessary cargo helpers
 	cargo install cargo-risczero
 	cargo risczero install
 
-lint: check-fuzz  ## cargo check and clippy. Skip clippy on guest code since it's not supported by risc0
+lint:  ## cargo check and clippy. Skip clippy on guest code since it's not supported by risc0
 	## fmt first, because it's the cheapest
 	cargo +nightly fmt --all --check
 	cargo check --all-targets --all-features
+	$(MAKE) check-fuzz
 	CI_SKIP_GUEST_BUILD=1 cargo clippy --all-targets --all-features
 
 lint-fix:  ## cargo fmt, fix and clippy. Skip clippy on guest code since it's not supported by risc0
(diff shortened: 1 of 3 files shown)


citizen-stig (approved):

neysofu (changes requested):
Since `fuzz` is in a separate workspace, we need to change the caching configuration, otherwise it will compile from scratch for every CI run (our caching configuration is workspace-based). I'll propose a change below:

Makefile
```diff
@@ -24,7 +24,7 @@ install-dev-tools:  ## Installs all necessary cargo helpers
 	cargo install cargo-risczero
 	cargo risczero install
 
-lint:  ## cargo check and clippy. Skip clippy on guest code since it's not supported by risc0
+lint: check-fuzz  ## cargo check and clippy. Skip clippy on guest code since it's not supported by risc0
 	## fmt first, because it's the cheapest
 	cargo +nightly fmt --all --check
 	cargo check --all-targets --all-features
```
neysofu: This would maintain the ordering from cheapest to most expensive check:

```suggestion
lint:  ## cargo check and clippy. Skip clippy on guest code since it's not supported by risc0
	## fmt first, because it's the cheapest
	cargo +nightly fmt --all --check
	cargo check --all-targets --all-features
	$(MAKE) check-fuzz
```

neysofu (commented):
I pushed a commit to a separate branch as a suggestion on how to fix the CI caching:

https://github.com/Sovereign-Labs/sovereign-sdk/commit/aadfefe0fae3ae6c7e364a0f8d9c179c954c39b3

vlopes11:
@neysofu amazing suggestion, pushed it into the branch!

neysofu (approved):

Issue #957, opened 2023-09-29
Have a more generalized structure and process to create utils for each module


dubbelosix:
We ran into this issue when trying to create nft-utils. The `sov-nft-module` provides calls to create collections, mint nfts in collections. However creating a collection and minting a large number of NFTs to it would be too complicated to do using the sov-cli. To do so programmatically, it made sense to import the runtime from demo-stf and then create a serialized transaction as follows

```
    let mint_nft_message =
        RuntimeCall::<DefaultContext, MockDaSpec>::nft(CallMessage::<DefaultContext>::MintNft {
            collection_name: collection_name.to_string(),
            token_uri,
            token_id,
            owner: UserAddress::new(owner),
            frozen: false,
        });
    Transaction::<DefaultContext>::new_signed_tx(
        signer,
        mint_nft_message.try_to_vec().unwrap(),
        nonce,
    )
```

We foresee a general pattern where each module might require utils to programmatically interact with the Call Messages to create transactions and perhaps other utlities.

For now `nft-utils` has been placed at the root, but it would be good to design a generalized structure for utils so that developers of modules also have a process to follow and some templates they can use as a guideline. Eg: bank utils, sequencer utils as well.

Note: placing this in the module directly causes a circular dependency since the runtime imports the module but the util needs the runtime to generate the call message

bkolad:
The utils for a specific module should not depend on the `Runtime` but only on the individual module. The reason is that if someone changes the `Runtime` (by removing the module) it will break the util specific code.

PR #990, opened 2023-10-06
Fixes for `simple-nft-module`

bkolad:
# Description
1. Adds `nft_path` to `GenesisPaths` so the nft genesis is read from a file.
2. Adds cli support in the `simple-nft-module`
3. Imports `simple-nft-module` in `demo-stf/Cargo.toml` (our customer is using it) 

## Linked Issues
- Related to #957

## Testing
Ci passes.

## Docs
No doc changes.


examples/simple-nft-module/src/tests.rs
```diff
@@ -0,0 +1,27 @@
+use sov_modules_api::default_context::DefaultContext;
```
citizen-stig, line 1: Could we please have tests to be not inside `src`, but in the module root?

module-system/module-implementations/sov-nft-module/src/tests.rs
```diff
@@ -0,0 +1,14 @@
+use crate::NonFungibleTokenConfig;
```
citizen-stig, line 1: Same for tests location

vlopes11 (approved):
LGTM!

examples/simple-nft-module/src/tests.rs
bkolad: We had a long discussion offline. In my opinion this should be solved as a separate issue. @citizen-stig  Let me know if you are ok with that and I will create the issue.

bkolad: https://github.com/Sovereign-Labs/sovereign-sdk/issues/993

PR #991, opened 2023-10-06
`sov-stf-runner` should not depend on modules

citizen-stig:
# Description
`sov-stf-runner` suppose to be independent from module system.

This PR solves this by doing following things:

* Replace re-exports from `sov-modules-api` to direct import from `sov-rollup-interface`
* Moving concrete implementation of batch builder `FiFoStrictBatchBuilder` to `sov-sequencer`
* `RollupConfig` now do not depend on concrete `StorageConfig`, but it is generic instead.
* `sov-modules-api` was dependent on `sov-sequencer` only for `to_jsonrpsee_error_object` function. So it got moved to `sov-modules-api`, since `sov-sequencer` already depends on it. And for ledger_rpc it is copied, to avoid bringing whole `sov-modules-api` dependency.

## Testing
Existing tests are passing

## Docs
No updates in documentation


full-node/sov-stf-runner/src/config.rs
```diff
@@ -26,12 +25,12 @@ pub struct RpcConfig {
 
 /// Rollup Configuration
 #[derive(Debug, Clone, PartialEq, Deserialize)]
-pub struct RollupConfig<DaServiceConfig> {
-    /// Runner configuration.
+pub struct RollupConfig<DaServiceConfig, StorageConfig> {
+    /// Path to serialized storage configuration.
     pub storage: StorageConfig,
```
bkolad, line 37: Would it be possible to use e `PathBuffer` maybe in a netype and remove the `StorageConfig` generic?

citizen-stig: Yes, it is possible, if we assume that all potential storage implementations will need only path parameter and nothing else.

bkolad: My inclination would be to make it generic only when we have the use case otherwise we have the generic bubbling up prematurely.

citizen-stig: Done. This new type effectively duplicates `sov_state::config::Config`, but I think this is cheap tradeoff.

bkolad (approved):

PR #996, opened 2023-10-06
feat(celestia-adapter): update to celestia-node v0.11 and celestia app v1.

### LINKED ISSUE #461: Sov-celestia-local container needs Dockerfile so that we can modify, update and build other versions of celestia binaries

### LINKED ISSUE #480: sov-celestia-local is opaque to users since nothing is displayed after "make start"

preston-evans98:
# Description

This copy is an exact copy of #948, rebased with the latest nightly. This is a separate PR due to a permissions issue, but @zvolin deserves all credit for this work. The original PR description is posted below.

This updates the `sov-celestia-adapter` to the latest (at the time of submitting) versions of celestia node (0.11.0-rc14) and app (1.0.0-rc17).

Sorry for such a big PR, but it was hard to introduce changes separately there

changes includes:
- a complete rewrite of celestia docker, this includes both `/docker` compose setup and the `Dockerfile` used in `demo-rollup/Makefile`. Now there is a single source of truth inside the `docker`. I tried to support all the existing use-cases, but please check if I haven't broken your workflows :pray: Also there seems to currently be no easy way to stop the running network in a way to resume it later, so it's recreated each time.
- usage of `celestia-types` in `sov-celestia-adapter`. `celestia-types` is a crate we maintain at @eigerco that consists of types compatible with celestia's protobuf and json API's. That way you shouldn't worry that you'll fall incompatible with the celestia network as we run integration tests for all those types with the real celestia-node. `celestia-types` have very little logic itself (yet) and is mostly about conforming to correct representations, so some types stayed as they were, but I tried to replace everything that was possible currently and we can then try to upstream the logic bit by bit.
- removal of `borsh` wherever `celestia-types` were included. We currently don't support borsh, but we can introduce the support for it in `celestia-types` wherever we don't include the `tendermint` types. All those places were currently marked as `TODO`
- recreation and addition of test data, updating all the tests and introducing a few new ones

## Linked Issues
- Fixes #461, fixes #480, 
- Related to # (issue) 

## Testing

Ran tests for sov-celestia-adapter
ran  demo-prover (with SKIP_PROVER=1) and demo-rollup + make test-create-token

## Docs

Updated the docs were relevant, also included some inline documentation


cemozerr (approved):

zvolin:
Closing in favor of #997

zvolin:
Hey, if that's not a problem for you I'd prefer the #997 but appreciate the credits :+1: I found out that the issue to the lack of write access was due to it being from an organisation-forked repository

preston-evans98:
Ah, makes sense - thanks @zvolin. Closing in favor of #997

PR #1000, opened 2023-10-09
feat: add arbitrary to sov-accounts

vlopes11:
This commit introduces arbitrary implementations for sov-accounts. It will generate primitives such as `Account` transparently, while will use a `Mutex` for `WorkingSet` mutation while the accounts are being generated.


module-system/module-implementations/sov-accounts/src/lib.rs
```diff
@@ -82,54 +83,141 @@ impl<C: Context> sov_modules_api::Module for Accounts<C> {
...
+        C: Context,
+        C::Address: proptest::arbitrary::Arbitrary,
+        C::PrivateKey: proptest::arbitrary::Arbitrary,
+    {
+        /// Creates an arbitrary set of accounts and stores it under `working_set`.
+        ///
+        /// We take the `WorkingSet` as `Arc<Mutex<_>>` so the strategy can freely implement
+        /// parallelism while preserving interior mutability safety. The return is a reference to a
+        /// promise of a strategy; hence, the `WorkingSet` lifetime must be locked to its lifetime.
+        pub fn arbitrary_proptest_workset(
```
bkolad: Where is this function called?

module-system/module-implementations/sov-accounts/src/lib.rs
```diff
@@ -82,54 +83,141 @@ impl<C: Context> sov_modules_api::Module for Accounts<C> {
...
+
+                            let addr = C::Address::from(address);
+                            let nonce = rng.next_u64();
+                            let account: Account<C> = Account { addr, nonce };
+
+                            public_keys.set(&account.addr, &public, working_set);
+                            accounts.set(&public, &account, working_set);
+                        }
+
+                        Accounts {
```
bkolad: The ` Accounst.accounts, Accounts.public_keys` prefixes are set by a macro and are always the same.  Why do we fuzz the `Accounts` structure? I thought we will only fuzz the CallMessages.

module-system/module-implementations/sov-accounts/src/lib.rs
vlopes11: This is equivalent (but for proptest) to the example here:

https://github.com/Sovereign-Labs/sovereign-sdk/blob/1c3f77f812e7f0e36e96db5066baa20fa225752f/fuzz/fuzz_targets/accounts_call_random.rs#L11-L25

It is a convenience function to allow us to fuzzy things faster, without worrying about the internals of the struct

module-system/module-implementations/sov-accounts/src/lib.rs
vlopes11: Makes sense, updated to `Accounts::default`. As the previous comment, this is just a convenience function available. Do you think its best to remove it?

bkolad: Ahh ok now I got it, but is it going to be used somwhere? If it is not used the let's remove it. Btw should  we move the fuzz tests into the modules (so accounts fuzz tests to `sov-accounts/fuzz` etc)?

vlopes11: Makes sense; moved to `fuzz`

module-system/module-implementations/sov-accounts/src/lib.rs
vlopes11: Since it's currently not used, the last commit removes it. If we need it in the future, we can just add

bkolad (approved):

PR #1014, opened 2023-10-10
EVM: Fix Sequencer Nonce

orkunkilic:
# Description
Sequencer only uses its own private key to wrap & sign EVM transactions. However current implementation tries to use original sender's nonce while signing, which fails on stateful verification.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/full-node/sov-ethereum/src/lib.rs
+++ b/full-node/sov-ethereum/src/lib.rs
@@ -8,7 +8,6 @@ pub use sov_evm::DevSigner;
 #[cfg(feature = "experimental")]
 pub mod experimental {
     use std::array::TryFromSliceError;
-    use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


full-node/sov-ethereum/src/lib.rs
```diff
@@ -58,7 +55,7 @@ pub mod experimental {
     }
 
     pub struct Ethereum<C: sov_modules_api::Context, Da: DaService> {
-        nonces: Mutex<HashMap<RethAddress, u64>>,
+        nonce: Mutex<u64>,
```
bkolad: Let's use https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicU64.html. It is more lightweight than mutex

bkolad (approved):

PR #1021, opened 2023-10-11
fix: add `risc0` feature to celestia adapter

vlopes11:
The tests are currently breaking as they require an optional `risc0-zkvm` dependency, assuming it is always available. A new feature is required to shield such case.

This commit also introduces a check on `cargo hack` to inspect all targets. This will prevent the same problem from happening in the future. Finally, it removes the unused `default` argument from `cargo hack`, as it conflicts with `all-targets`.


cemozerr (approved):
LGTM.

bkolad (approved):

PR #1030, opened 2023-10-12
Allow setting constants via manifest

preston-evans98:
# Description
This PR adds the `#[config_constant]` attribute, which sets a constant from the manifest file. This functionality will be used in several places, but is blocking for the soft-confirmations workflow (where we need to set the `inbox_delay` for forced transactions per-rollup instead of hard-coding it). 

```rust
#[config_constant]
pub const TEST_ARRAY_OF_U8: [u8; 32];

fn main() {
 // The value is filled from `constants.json`, allowing this assertion to pass
 assert_eq!(TEST_ARRAY_OF_U8, [255;32])
}
```

This macro is exported from `sov_modules_api::macros`.

**N.B.**: This PR only implements basic types (numeric types, strings, booleans, and arrays). Support for more complex types is left unimplemented for now.

## Testing
A new test has been added using `trybuild` to ensure that valid values are produced for the covered types. 

## Docs
A doc comment has been added


module-system/sov-modules-macros/src/make_constants.rs
```diff
@@ -0,0 +1,12 @@
...
+
+use crate::manifest::Manifest;
+
+pub(crate) fn make_const(
+    field_ident: &Ident,
+    ty: &Type,
+    vis: syn::Visibility,
+    attrs: &[syn::Attribute],
+) -> Result<proc_macro2::TokenStream, syn::Error> {
+    Manifest::read_constants(field_ident)?.parse_constant(ty, field_ident, vis, attrs)
```
vlopes11, line 37: nit: should we consider making it a function of an instance of `Manifest`? This way will work, but we will parse the `constants.json` file every time the macro is called, as opposed to having it parsed only once.

This is optional as making the macro expansion stateful isn't trivial. One alternative would be to create a `static ref MANIFEST: Option<Manifest>` that would read the file once and serve it many times.

module-system/sov-modules-macros/src/manifest.rs
```diff
@@ -48,12 +49,16 @@ impl<'a> Manifest<'a> {
     /// The `parent` is used to report the errors to the correct span location.
     pub fn read_constants(parent: &'a Ident) -> Result<Self, syn::Error> {
         let manifest = "constants.json";
-        let initial_path = match env::var("CONSTANTS_MANIFEST") {
+        let initial_path = match env::var("CONSTANTS_MANIFEST").and_then(|v| if v.is_empty() {
+            Err(env::VarError::NotPresent)
```
vlopes11: The previous implementation intended to suppress env var errors as it might not be set.

The proposed change introduces an improvement that is to check a specific variant where the key/value pair of the environment has the value empty.

I think we will still fall into the line `64` `Err(_)` pattern, as we will correctly raise the error, but then fallback to reading from manifest dir. The most consistent behavior would be to fail if the user `export CONSTANTS_MANIFEST=""`.

I think the best way to achieve that is to add a variant of the `match` pattern, before line `57`, with `Ok(p) p.is_empty() => `, and throw an error

module-system/sov-modules-macros/src/manifest.rs
```diff
@@ -236,6 +241,104 @@ impl<'a> Manifest<'a> {
...
     }
 
+    pub fn parse_constant(
+        &self,
+        ty: &Type,
+        field: &Ident,
+        vis: syn::Visibility,
+        attrs: &[syn::Attribute],
+    ) -> Result<TokenStream, syn::Error> {
+        let map = self
```
vlopes11: nit: this code is repeating with the `parse_gas_config`. Might be worthy to create a specialized, private scope function

```rust
fn get_object(&self, field: &Ident, key: &str) -> Result<TokenStream, syn::Error> {
  self
    .value
    .as_object()
    .ok_or_else(|| Self::err(&self.path, field, "manifest is not an object"))?
    .get(key)
    .ok_or_else(|| {
      Self::err(
        &self.path,
        field,
        format!("manifest does not contain a `{key}` attribute"),
      )
    })?
    .as_object()
    .ok_or_else(|| {
      Self::err(
        &self.path,
        field,
        format!("`{key}` attribute of `{field}` is not an object"),
      )
    })
}

File: module-system/sov-modules-macros/src/lib.rs [hunk 4]
```diff
@@ -82,6 +84,15 @@ pub fn codec(input: TokenStream) -> TokenStream {
     handle_macro_error(codec_macro.derive_message_codec(input))
 }
 
+/// Specifies that a constant should be configured from the manifest file.
+#[proc_macro_attribute]
```
vlopes11, line 88: One alternative to consider is to, instead of making it an attribute macro (i.e. `#[config_constant]`), a simple macro, as it might be easier to use. Example:

from:
```rust
#[config_constant]
pub const TEST_U32: u32 = 0;
```

to:
```rust
config_constant!(TEST_U32, u32);
```

vlopes11 (approved):
Looks good! It's a nice improvement to have tests over the functionality.

A couple of non-blocking comments & suggestions

module-system/sov-modules-macros/src/lib.rs
vlopes11: I think it would be possible to have something like

```rust
#[config_constant]
pub const TEST_U32: u32;
```

And the macro expansion will append the `= 42`. Having it declared with a default value might give the user the wrong impression the constant will fallback to the provided value if the parsing fails (or the constants.json file is absent)

module-system/sov-modules-macros/src/make_constants.rs
preston-evans98: Going to hold off on this for now to keep the PR small. I'm not too worried about the performance, since the JSON files we're parsing here are small

module-system/sov-modules-macros/src/lib.rs
preston-evans98: TBH, I think the first syntax is a little bit more readable.

vlopes11: Wdyt of introducing that default value as fallback to when `constants.json` is absent, or has no `constants.TEST_U32`? We could ofc add this as future functionality by creating an issue

module-system/sov-modules-macros/src/manifest.rs
preston-evans98: Good idea. Done!

module-system/sov-modules-macros/src/manifest.rs
preston-evans98: Yeah, I think this makes sense. Adjusting.

module-system/sov-modules-macros/src/lib.rs
preston-evans98: Oops, I hadn't refreshed the page, so my comment missed your second one. 

I like this syntax a lot
```rust
#[config_constant]
pub const TEST_U32: u32;
```

I'm afraid that using the default value as a fallback will introduce a footgun where the rollup just silently accepts not being configured , so I think it's preferable to fail explicitly when the constants aren't set.

preston-evans98: Let's plan to convert to this new syntax in a follow-up. Don't want to block this work, and it will be a bit painful because we can't rely on syn's default parser since the declaration is incomplete.

preston-evans98: It actually turned out not to be too complicated, so I went ahead!

PR #1032, opened 2023-10-13
Dedicated ledger JSON-RPC crate

neysofu:
# Description
With this PR, I'm introducing a new crate `full-node/sov-ledger-rpc` which has two components:
- A `jsonrpsee::RpcModule` builder for the `LedgerDB` JSON-RPC API.
- A type-safe client trait `RpcClient`.

This results in some code deduplication across rollup crates, which can just import the server implementation to serve ledger data. The provided server implementation is also more extensive compared to what we had in `demo-rollup`, and exposes all getters that `LedgerRpcProvider` offers.

`rollup-template` has been modified accordingly.

## Testing
Existing `test_rpc.rs` in `demo-rollup`, combined with new doc tests and two integration tests.


full-node/sov-ledger-rpc/src/lib.rs
```diff
@@ -0,0 +1,79 @@
+#![forbid(unsafe_code)]
```
bkolad, line 1: let's add `#![deny(missing_docs)]`

neysofu: Yeah, initially I didn't want to document all the `Rpc` trait methods because it would just be a duplication of the `LedgerRpcProvider` documentation, but it makes sense. Will address

neysofu: Fixed in https://github.com/Sovereign-Labs/sovereign-sdk/pull/1032/commits/fd55cbdd85cabc950db3a3cc20226af0705ced33.

bkolad (approved):
LGTM!

PR #1042, opened 2023-10-16
Split `rng_xfers` into its own crate

neysofu:
# Description
I'm splitting off `rng_xfers` into its own utility crate, to facilitate code reuse with the block explorer. 

## Testing
I'm mostly relying on existing tests, which covert pretty much all of this logic.

### DIFF
--- a/utils/rng-da-service/src/lib.rs
+++ b/utils/rng-da-service/src/lib.rs
@@ -16,6 +16,8 @@ use sov_rollup_interface::mocks::{
 use sov_rollup_interface::services::da::DaService;
 
 pub fn sender_address_with_pkey() -> (Address, DefaultPrivateKey) {
+    // TODO: maybe generate address and private key randomly, instead of
+    // hardcoding them?
     let addr_bytes = "sov15vspj48hpttzyvxu8kzq5klhvaczcpyxn6z6k0hwpwtzs4a6wkvqmlyjd6".to_string();
     let addr = Address::from(
         AddressBech32::try_from(addr_bytes)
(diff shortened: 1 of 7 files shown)


utils/rng-da-service/src/lib.rs
```diff
@@ -15,81 +15,29 @@ use sov_rollup_interface::mocks::{
...
-
-fn generate_transfers(n: usize, start_nonce: u64) -> Vec<u8> {
-    let sender_address =
-        "sov15vspj48hpttzyvxu8kzq5klhvaczcpyxn6z6k0hwpwtzs4a6wkvqmlyjd6".to_string();
-    let token_name = "sov-test-token";
-    let sa = Address::from(
-        AddressBech32::try_from(sender_address)
-            .unwrap_or_else(|_e| panic!("Failed generating transfers")),
+pub fn sender_address_with_pkey() -> (Address, DefaultPrivateKey) {
+    let addr_bytes = "sov15vspj48hpttzyvxu8kzq5klhvaczcpyxn6z6k0hwpwtzs4a6wkvqmlyjd6".to_string();
```
vlopes11, line 21: Any reason why we don't generate the address with entropy (i.e. from a RNG)?

utils/rng-da-service/src/lib.rs
```diff
@@ -15,81 +15,29 @@ use sov_rollup_interface::mocks::{
...
+    let addr = Address::from(
+        AddressBech32::try_from(addr_bytes)
```
vlopes11, line 23: If the origin string is constant, then the address could be

vlopes11 (approved):
LGTM; I pointed one question. But since it isn't introduced in this PR, we can merge as it is and maybe address the question @ future

citizen-stig (approved):

utils/rng-da-service/src/lib.rs
neysofu: I don't think there's any particular reason; arguably we should. I'll make a note in the code.

neysofu: Left a comment! Thank you.

PR #1050, opened 2023-10-16
Small tweaks to make CI faster

neysofu:
- Reintroduces `--exclude-features default` in `make check-features`, which was accidentally removed in https://github.com/Sovereign-Labs/sovereign-sdk/pull/1021. I chatted with @vlopes11 at we are confident we can reintroduce it without issues. This will significantly speed up the `features` job.
- Install `cargo-risczero` *after* `Swatinem/rust-cache`, which I suspect might make it cacheable. I'm not sure about this; at best we'll cut 1-2 minutes from every job, at worse nothing will happen because of the reordering.


neysofu:
We'll be unable to evaluate the perf penalty/improvement until merged, as caches are only populated from `nightly`.

bkolad (approved):

PR #1053, opened 2023-10-16
Bug fix; re-run const resolution on change

preston-evans98:
# Description
Currently, our `#[const_config]` and `#[gas]` macros read from a config file which is resolved during macro processing based on either the `CONSTANTS_MANIFEST` environment variable or the `CARGO_MANIFEST_DIR`. However, the macro only runs if `cargo` decides to recompile the crate - so changes which only affect the `constants.json` file are not propagated to the crate by default. Changes only propagate if the user changes both `constants.json` (or the environment variable) *and* a `.rs` file.

This PR fixes the issue by:
1. Moving path resolution logic to a `build.rs` script in `sov-modules-macros`
2. Invoking `cargo:rerun-if-changed` on the manifest
3. Writing a special `target-path` file into `CARGO_MANIFEST_DIR` telling the macros where to look for the `constants.json` file.


module-system/sov-modules-macros/build.rs
```diff
@@ -1,14 +1,59 @@
...
+        Ok(p) if p.is_empty() => {
+            Err(anyhow!(
+                "failed to read target path for sovereign manifest file: env var `CONSTANTS_MANIFEST` was set to the empty string"
+            ))
+        },
+        Ok(p) => PathBuf::from(&p).canonicalize().map_err(|e| {
+                anyhow!("failed to canonicalize path for sovereign manifest file from env var `{p}`: {e}")
+        }),
+        Err(_) => {
+            let mut current_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
```
vlopes11: It looks like we should start from the `OUT_DIR`. If the `sov-*` is downloaded as crate, we will backtrack from `$HOME/.cargo/**/*`, and will never reach the workspace root of the current project.

module-system/sov-modules-macros/src/manifest.rs
```diff
@@ -48,76 +48,40 @@ impl<'a> Manifest<'a> {
...
-                    "failed to read target path for sovereign manifest file: env var `CONSTANTS_MANIFEST` was set to the empty string".to_string(),
-                ))
-            },
-            Ok(p) => PathBuf::from(&p).canonicalize().map_err(|e| {
+        let target_path_pointer = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .canonicalize()
+            .map_err(|e| {
                 Self::err(
-                    &p,
+                    "constants.json",
```
vlopes11: Consider renaming it to `target-path`. The previous implementation intended to display the faulty path to the user; in this case, we are trying to get the `OUT_DIR`

module-system/sov-modules-macros/build.rs
```diff
@@ -1,14 +1,59 @@
...
+    // Write the path "OUT_DIR" to a file in the manifest directory so that it's available to macros
+    fs::write(target_path_record, out_dir.display().to_string())?;
+
+    let manifest_path = resolve_manifest_path()?;
+    let output_manifest_path = out_dir.join("constants.json");
+
+    // Copy the constants.json file into out_dir
+    fs::copy(&manifest_path, &output_manifest_path).with_context(|| anyhow!("could not copy manifest from {manifest:?} to output directory `{output_manifest_path:?}`"))?;
+    // Tell cargo to rebuild if constants.json changes
+    println!("cargo:rerun-if-changed={}", output_manifest_path.display());
```
vlopes11, line 64: Consider not tracking the out dir artifact with the build system. This file is not the source of truth and shouldn't be edited by the user - the source of truth must be only the single `constants.json` at his root.

module-system/sov-modules-macros/build.rs
```diff
@@ -1,14 +1,59 @@
...
+                "failed to read target path for sovereign manifest file: env var `CONSTANTS_MANIFEST` was set to the empty string"
+            ))
+        },
+        Ok(p) => PathBuf::from(&p).canonicalize().map_err(|e| {
+                anyhow!("failed to canonicalize path for sovereign manifest file from env var `{p}`: {e}")
+        }),
+        Err(_) => {
+            let mut current_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+            loop {
+                if current_path.join(manifest).exists() {
```
vlopes11: Should we check if [is_file](https://doc.rust-lang.org/std/path/struct.Path.html#method.is_file) ? Also, we should probably canonicalize the `current_path`, so the `parent` backtrack will be unambiguous.

vlopes11 (changes requested):
The change looks good overall! It is a good idea to re-compile if the constants file changes.

Added a couple of potential issues to be discussed

module-system/sov-modules-macros/src/manifest.rs
preston-evans98: Done.

module-system/sov-modules-macros/build.rs
preston-evans98: Per the [docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script), `rerun-if-changed` works by looking at the timestamp - so if we remove this line, I think we create a corner case where:
1. The user has two different `constants.json` files
2. They update `CONSTANTS_MANIFEST` from `a/constants.json` to `b/constants.json`

In this case, the build script up to this point will behave correctly - copying `b/constants.json` to the `out_dir`. However, since the timestamp on `b/constants.json` might be older than the previous build, this script won't trigger a re-run of the macros unless we watch the output file

module-system/sov-modules-macros/build.rs
preston-evans98: Sure, both good ideas. Done!

module-system/sov-modules-macros/build.rs
preston-evans98: That makes sense, yeah.

module-system/sov-modules-macros/build.rs
vlopes11: Good point!

module-system/sov-modules-macros/build.rs
vlopes11: As side note, if we start from `OUT_DIR`, we can just copy the `constants.json` directly into `CARGO_MANIFEST_DIR` so we just read it during macro expansion, as our `CARGO_MANIFEST_DIR` for the context of `sov-module-macros` will be guaranteed to contain the correct `constants.json`.

In short; this hack of setting the `OUT_DIR` in `target-path` is deprecated with this new approach, we can just copy the desired file

preston-evans98: Done.

vlopes11 (approved):
LGTM after CI

PR #1059, opened 2023-10-17
CI workflow to manually delete caches

neysofu:
# Description
The GH Actions UI offers a way to list and manually delete cache entries. There's no UI solution for BuildJet, but we can use a custom workflow to manually delete entries. I'm not yet sure if listing cache entries if possible, for now the cache entry key to delete must be copy-pasted from logs.


neysofu:
Must be rebased once https://github.com/Sovereign-Labs/sovereign-sdk/pull/1061 is merged.

bkolad (approved):

PR #1072, opened 2023-10-18
Give preferred sequencer the ability to process "deferred blobs" early

preston-evans98:
# Description
This PR adds functionality to the `BlobStorage` module allowing the `preffered_sequencer` to request that some number of `deferred_blobs` be processed early. 

As a reminder, blobs which are sent by someone other than the preferred sequencer are usually stored in state for some number of slots  (`DEFERRED_SLOTS_COUNT`) before being processed. This is to ensure that they don't "front-run" the preferred sequencer, invalidating its soft confirmations.

This PR allows the preferred sequencer to specify a minimum number of such "deferred blobs" for the rollup to process during the next block. Note that this change *does not* allow the preferred sequencer to censor, since he can only set a lower bound on the number of deferred blobs to be processed. Deferred blobs will always be processed no later than `DEFERRED_SLOTS_COUNT` after appearing on the DA layer. If the preferred sequencer requests more blobs than are in the current deferred queue, the queue will be emptied.

## Linked Issues
- Related to #408 

## Testing
Tests have been added


module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -16,36 +20,88 @@ impl<C: Context, Da: DaSpec> BlobSelector<Da> for BlobStorage<C, Da> {
...
         let mut to_defer: Vec<&mut Da::BlobTransaction> = Vec::new();
-
         for blob in current_blobs {
+            // Blobs from the preferred sequencer get priority
             if blob.sender() == preferred_sequencer {
                 priority_blobs.push(blob);
             } else {
-                to_defer.push(blob);
+                // Other blobs get deferred unless the sequencer has requested otherwise
+                if remaining_blobs_requested != 0 {
```
bkolad: Nit: Let's change it to `remaining_blobs_requested > 0`. I know it is impossible to go negative here but I think `>` is more defensive.

bkolad (approved):

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -16,36 +20,88 @@ impl<C: Context, Da: DaSpec> BlobSelector<Da> for BlobStorage<C, Da> {
...
 
+        // If we reach this point, there is a preferred sequencer, so we need to handle cases 1 and 3.
+
+        // First, compute any "bonus blobs" requested
+        // to be processed early.
+        let num_bonus_blobs_requested = self
+            .deferred_blobs_requested_for_execution_next_slot
+            .get(working_set)
+            .unwrap_or_default();
+        self.deferred_blobs_requested_for_execution_next_slot
```
citizen-stig, line 95: Should we hide this `.set` behind condition, so it is only set if was non zero? To reduce amount of state accesses?

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -16,36 +20,88 @@ impl<C: Context, Da: DaSpec> BlobSelector<Da> for BlobStorage<C, Da> {
...
+            .set(&0, working_set);
+
+        let mut remaining_blobs_requested =
+            (num_bonus_blobs_requested as usize).saturating_sub(expiring_deferred_blobs.len());
+        let mut bonus_blobs: Vec<BlobRefOrOwned<Da::BlobTransaction>> =
+            Vec::with_capacity(remaining_blobs_requested);
+        let mut next_slot_to_check = slot_for_expiring_blobs + 1;
+        while remaining_blobs_requested > 0 && next_slot_to_check < current_slot {
```
citizen-stig, line 105: I was initially thought why not to get up to current blob including, but then later realizaed why.
Maybe let's put this comment. Feel free to edit the wording.

```suggestion
        // Blobs from the current slot are excluded, ensuring that priority blobs are not deprioritized.
        while remaining_blobs_requested > 0 && next_slot_to_check < current_slot {
```

module-system/module-implementations/sov-sequencer-registry/src/lib.rs
```diff
@@ -136,6 +136,21 @@ impl<C: sov_modules_api::Context, Da: sov_modules_api::DaSpec> SequencerRegistry
         self.preferred_sequencer.get(working_set)
     }
 
+    /// Returns the preferred sequencer, or [`None`] it wasn't set.
```
citizen-stig:
```suggestion
    /// Returns a rollup address of the preferred sequencer, or [`None`] it wasn't set.
```

module-system/module-implementations/sov-blob-storage/tests/capability_tests.rs
```diff
@@ -120,16 +151,251 @@ fn priority_sequencer_flow_general() {
...
         }
     }
     // Ensure that all blobs have been processed
     assert!(expected_blobs.next().is_none());
 }
 
+#[test]
+fn bonus_blobs_are_delivered_on_request() {
+    // If blobs are deferred for less than two slots, "early processing" is not possible
+    if DEFERRED_SLOTS_COUNT < 2 {
```
citizen-stig, line 190: Do you think it makes sense to have a test with this condition, to show how rollup suppose to behave?

My understanding is, that if `DEFERRED_SLOTS_COUNT < 2` logic suppose to be the same as with flushing request, as without it. 

It might sound stupid, but having test prove that would be nice, in case if there are some subtle bugs.

citizen-stig (approved):

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
preston-evans98: No need! No-op writes are already optimized away by our `CacheLog`

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
preston-evans98: Blobs from the current slot are excluded because they haven't been stored yet, not because that would cause priority blobs to be deprioritized. Adding a comment to clarify.

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
preston-evans98: Good idea. Done.

module-system/module-implementations/sov-sequencer-registry/src/lib.rs
preston-evans98: 👍

module-system/module-implementations/sov-blob-storage/tests/capability_tests.rs
preston-evans98: Good idea. Added tests and had to special case some things when `DEFERRED_SLOTS_COUNT == 0`. I don't think that's a case we'll likely run into, but this way it's covered.

module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -107,22 +146,7 @@ impl<C: Context, Da: DaSpec> BlobSelector<Da> for BlobStorage<C, Da> {
...
-                    if !is_allowed {
-                        info!(
-                            "Blob hash=0x{} from sender {} is going to be discarded",
-                            hex::encode(b.hash()),
-                            b.sender()
-                        );
-                    }
-                    is_allowed
-                })
+                .filter(|b| self.filter_by_sender(b, working_set))
```
citizen-stig: Nit: Maybe call it `filter_by_allowed_sender`

preston-evans98: Done.

PR #1073, opened 2023-10-18
feat: add `no-std` to `sov-modules-api`

vlopes11:


neysofu:
Should we close this one in favor of https://github.com/Sovereign-Labs/sovereign-sdk/pull/1077?

neysofu:
Nvm, I'm dumb. I just realized this is for `sov-modules-api`.

vlopes11:
wontfix for now; there will be impact on the users experience if we introduce no-std, as the feature gate will increase in complexity. We can compile the tx signature using a permissive wasm backend

PR #1076, opened 2023-10-19
Remove `NativeDB::set_value` in favour of `set_values`

citizen-stig:
As it is only used in test.

Explicit use of `anyhow::Result` instead of importing it, I believe it improves readability.

This is ongoing effort of simplifying public interface of `sov-db` and `sov-schema-db`

### DIFF
--- a/full-node/db/sov-schema-db/src/lib.rs
+++ b/full-node/db/sov-schema-db/src/lib.rs
@@ -55,7 +55,7 @@ impl DB {
         column_families: impl IntoIterator<Item = impl Into<String>>,
         db_opts: &rocksdb::Options,
     ) -> anyhow::Result<Self> {
-        let db = DB::open_cf(
+        let db = DB::open_with_cfds(
             db_opts,
             path,
             name,
@@ -69,7 +69,8 @@ impl DB {
     }
 
     /// Open RocksDB with the provided column family descriptors.
-    fn open_cf(
+    /// This allows to configure options for each column family.
+    pub fn open_with_cfds(
         db_opts: &rocksdb::Options,
         path: impl AsRef<Path>,
         name: &'static str,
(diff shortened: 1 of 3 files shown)


full-node/db/sov-schema-db/src/lib.rs
```diff
@@ -69,12 +69,12 @@ impl DB {
     }
 
     /// Open RocksDB with the provided column family descriptors.
-    pub fn open_cf(
+    fn open_cf(
```
neysofu: I think we should to keep this public. If private, there's no way for a downstream user to open a `DB` with custom `rocksdb::Options` and we're forcing everybody to use the options used by `DB::open`.

citizen-stig: public `open` also has ` db_opts: &rocksdb::Options,`, this method only provides opportunity to specify column options. 

I understand that it adds another level of configurability, but on the other hand it makes public interface more complicated. I don't have a strong evidence, that somebody might need custom options for column families. But it also does not mean, that this use case does not exist. As they say, absence of evidence is not evidence of absence.

Overall I am fine to keep it public. I will probably open follow up PR, with different approach regarding many public methods

neysofu: Well, apparently we do have a need for custom options (LZ4), so I think it's likely someone else will too, and it'd be nice for them to be able to use the crate. That said, I don't feel too strongly about this!

neysofu (approved):
LGTM! I think `open_cf` should be public but I don't want to block the PR based on that, so it's your call 😄 the rest of it seems great, especially the iterator approach to `set_values`.

full-node/db/sov-schema-db/src/lib.rs
citizen-stig: Ok, let's keep it public

citizen-stig:
> LGTM! I think `open_cf` should be public but I don't want to block the PR based on that, so it's your call 😄 the rest of it seems great, especially the iterator approach to `set_values`.

I decided to bring it back, because I want to make this preparation work less intrusive.

PR #1077, opened 2023-10-19
feat: add `no-std` to `rollup-interface`

vlopes11:
This commit introduces a `no-std` build for `rollup-interface`. This is required to create WASM modules for module implementations, having one use-case the web wallet.

We follow a similar approach to `borsh` and create a `maybestd` module so dependencies can rely on that for fallback implementations of common stdlib components that are available either on `alloc` or commonly used crates.

Most of the requirements exists on `alloc`, but we do have a couple of exceptions. The first is `HashMap`, that we fallback for `hashbrown` when `std` isn't available. We shouldn't expect issues from risc0 integration as `risc0` code will be compiled with `std`.

The second is `Mutex`; for that, we fallback to `spin`, another common choice on the Rust ecosystem.

Finally, this commit introduces a couple of changes on the root `Cargo.toml` to not use default-features for crates that offers `no-std` support.


full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -168,12 +176,14 @@ macro_rules! define_table_with_default_codec {
 /// little-endian, but RocksDB uses lexicographic ordering which is only
 /// compatible with big-endian, so we use [`bincode`] with the big-endian option
 /// here.
+// TODO `bincode` is expected to have `no-std` soon; should be usable under `no-std`
+#[cfg(feature = "std")]
```
bkolad: From what I see all the `cfg(feature = "std")]` pragmas in this file are on functions signatures. Would it be possible to move all the `std` code to another file and reimport it as:
```
 `cfg(feature = "std")]`
 mod table_moule;
  `cfg(feature = "std")]`
  pub use table_moule::*;
```
And inside `tables_std` we won't need   `cfg(feature = "std")]`

full-node/db/sov-schema-db/src/iterator.rs
```diff
@@ -28,19 +28,23 @@ pub trait SeekKeyEncoder<S: Schema + ?Sized>: Sized {
...
+#[cfg(feature = "std")]
 pub(crate) enum ScanDirection {
     Forward,
     Backward,
 }
 
 /// DB Iterator parameterized on [`Schema`] that seeks with [`Schema::Key`] and yields
 /// [`Schema::Key`] and [`Schema::Value`] pairs.
+#[cfg(feature = "std")]
 pub struct SchemaIterator<'a, S> {
```
bkolad: I think the ` SchemaIterator`and all its dependencies  can be moved to separate file and then only the file could be imported with `#[cfg(feature = "std")]`

full-node/db/sov-schema-db/src/lib.rs
```diff
@@ -292,58 +62,101 @@ impl SchemaBatch {
...
+
+        #[cfg(feature = "std")]
         self.rows
             .lock()
             .expect("Lock must not be poisoned")
             .entry(S::COLUMN_FAMILY_NAME)
             .or_default()
             .push(WriteOp::Value { key, value });
 
+        #[cfg(not(feature = "std"))]
```
bkolad, line 81: we have `cfg_attr(not(feature = "std"), allow(dead_code))]` if it is not used in `std` why do we have the flag here?

full-node/db/sov-schema-db/src/lib.rs
```diff
@@ -292,58 +62,101 @@ impl SchemaBatch {
...
-        let _timer = SCHEMADB_BATCH_PUT_LATENCY_SECONDS
+        #[cfg(feature = "std")]
+        let _timer = metrics::SCHEMADB_BATCH_PUT_LATENCY_SECONDS
             .with_label_values(&["unknown"])
             .start_timer();
+
         let key = key.encode_key()?;
         let value = value.encode_value()?;
+
+        #[cfg(feature = "std")]
```
bkolad: We have `cfg_attr(not(feature = "std"), allow(dead_code))]` in line 39.  If the code is not used in `std` why do we have the flag here?

full-node/db/sov-schema-db/src/lib.rs
```diff
@@ -292,58 +62,101 @@ impl SchemaBatch {
...
 }
 
-/// For now we always use synchronous writes. This makes sure that once the operation returns
-/// `Ok(())` the data is persisted even if the machine crashes. In the future we might consider
-/// selectively turning this off for some non-critical writes to improve performance.
-fn default_write_options() -> rocksdb::WriteOptions {
-    let mut opts = rocksdb::WriteOptions::default();
-    opts.set_sync(true);
-    opts
+#[cfg(not(feature = "std"))]
```
bkolad: Let's move all the conversions to a separate module and feature gate only the module.

full-node/db/sov-schema-db/src/lib.rs
bkolad: I have just noticed that the `row` in `SchemaBatch` is always `Mutex`  from std. We should feature gate the row in `SchemaBatch` instead in the caller.

full-node/db/sov-db/Cargo.toml
```diff
@@ -17,30 +17,44 @@ resolver = "2"
...
-tokio = { workspace = true }
+tokio = { workspace = true, optional = true }
 
 
 [dev-dependencies]
 tempfile = { workspace = true }
 
 
 [features]
+default = ["std", "tokio"]
```
neysofu: Why do we need `std` for `sov-db` and `sov-schema-db`? They seem to be almost completely useless without it. Are the `no_std` variants of these crates used/expected to be used anywhere?

rollup-interface/src/state_machine/mocks/zk_vm.rs
```diff
@@ -79,17 +78,30 @@ impl Zkvm for MockZkvm {
...
+        Add: crate::RollupAddress,
+        Da: crate::da::DaSpec,
+        Root: Serialize + serde::de::DeserializeOwned,
+    >(
+        _serialized_proof: &[u8],
+        _code_commitment: &Self::CodeCommitment,
+    ) -> Result<crate::zk::StateTransition<Da, Add, Root>, Self::Error> {
+        todo!("the current version of bincode doesn't support no-std; however, the next version is scheduled to")
+    }
 }
```
neysofu, line 105: A unified function body seems more compact and uses less feature gating magic:

```suggestion
    fn verify_and_extract_output<
        Add: crate::RollupAddress,
        Da: crate::da::DaSpec,
        Root: Serialize + serde::de::DeserializeOwned,
    >(
        serialized_proof: &[u8],
        code_commitment: &Self::CodeCommitment,
    ) -> Result<crate::zk::StateTransition<Da, Add, Root>, Self::Error> {
        if cfg!(feature = "std") {
            let output = Self::verify(serialized_proof, code_commitment)?;
            Ok(bincode::deserialize(output)?)
        } else {
            todo!("the current version of bincode doesn't support no-std; however, the next version is scheduled to")
        }
    }
```

rollup-interface/src/state_machine/mod.rs
```diff
@@ -23,10 +23,10 @@ pub trait BasicAddress:
     + Send
     + Sync
     + Clone
-    + std::hash::Hash
+    + core::hash::Hash
```
neysofu, line 27: This is exactly the same pattern I'm suggesting we also use for `Vec`, `Box`, `String`, etc.. These types are equivalent and more portable.

rollup-interface/src/state_machine/mocks/da.rs
```diff
@@ -12,8 +10,12 @@ use serde::{Deserialize, Serialize};
 use crate::da::{
     BlobReaderTrait, BlockHashTrait, BlockHeaderTrait, CountedBufReader, DaSpec, DaVerifier, Time,
 };
+use crate::maybestd::string::String;
+#[cfg(all(feature = "native", feature = "tokio"))]
+use crate::maybestd::sync::Arc;
```
neysofu: Similarly, `alloc::sync::Arc` would probably work just fine here.

rollup-interface/src/state_machine/mocks/da.rs
```diff
@@ -302,12 +304,12 @@ impl DaSpec for MockDaSpec {
...
 use tokio::sync::mpsc::{self, Receiver, Sender};
-#[cfg(feature = "native")]
+#[cfg(all(feature = "native", feature = "tokio"))]
 use tokio::sync::Mutex;
 
-#[cfg(feature = "native")]
+#[cfg(all(feature = "native", feature = "tokio"))]
 #[derive(Clone)]
 /// DaService used in tests.
 pub struct MockDaService {
```
neysofu: It seems that `MockDaService` requires a ton of feature gating. Can we maybe put it (and its required imports) in a module and re-export it with just 2 feature gates?

rollup-interface/src/state_machine/da.rs
```diff
@@ -212,8 +212,12 @@ pub struct Time {
...
 }
 
-#[derive(Debug, Error)]
-#[error("Only intervals less than one second may be represented as nanoseconds")]
+#[derive(Debug)]
+#[cfg_attr(
+    feature = "std",
+    derive(thiserror::Error),
+    error("Only intervals less than one second may be represented as nanoseconds")
+)]
```
neysofu, line 219: Since implementing `core::error::Error` for this type seems very simple, we can easily do this even without `std` and `thiserror` so that the `no_std`/`std` APIs are more similar.

rollup-interface/src/lib.rs
```diff
@@ -2,12 +2,29 @@
...
+    /// A facade for the `sync` std module.
+    pub mod sync {
+        #[cfg(feature = "std")]
+        pub use std::sync::Mutex;
+
+        pub use borsh::maybestd::sync::*;
+        #[cfg(not(feature = "std"))]
+        pub use spin::Mutex;
+    }
+}
```
neysofu: I really like this pattern, and it seems to be commonplace among crates that need to expose 3rd-party types in their API with optional `std`. Very clean!

rollup-interface/src/lib.rs
```diff
@@ -2,12 +2,29 @@
...
+    pub use borsh::maybestd::{borrow, boxed, collections, format, io, rc, string, vec};
+
+    /// A facade for the `sync` std module.
+    pub mod sync {
+        #[cfg(feature = "std")]
+        pub use std::sync::Mutex;
+
+        pub use borsh::maybestd::sync::*;
+        #[cfg(not(feature = "std"))]
+        pub use spin::Mutex;
```
neysofu: Nit: let's put the two re-exports for `Mutex` next to each other.

neysofu (changes requested):
Looks great! I have some reservations mostly around three items:

1. `sov_db` and `sov_state_db`. Why is `no_std` needed there?
2. We need an effective way to test that our dependencies are `no_std` as well. Running `cargo check` in CI for `sov-rollup-interface` with a target that has no `std` support would be a good start. I see that there's a few `std` dependencies that slipped in, it'll be way too easy for that to happen again even if we fix it in this PR without adding a CI check.
3. I would suggest we rely more on `alloc` for the types that it exposes, without `maybestd`. Namely, `Arc`, `Vec`, `String`, `Box`, etc..

full-node/db/sov-db/src/schema/tables.rs
vlopes11: I did a slightly different approach to create `use_std` and `no_std` sub-modules on the same file. Otherwise, a bigger change to move `tables.rs` to `tables/mod.rs` and add `tables/tables_std.rs` would be required.

Let me know wdyt

full-node/db/sov-schema-db/src/iterator.rs
vlopes11: Ditto; let's centralize on the previous comment

full-node/db/sov-schema-db/src/lib.rs
vlopes11: We do have a `Mutex` available for no-std; `SchemaBatch` type definition is available for such environment.

Regarding the `dead_code`, we make the type definition available, but the concrete implementation is done only in `std`, where we interact with a rocksdb backend. I think it's best to make types available for `no-std` so we can represent them, even without certain functionality. Wdyt?

full-node/db/sov-schema-db/src/lib.rs
vlopes11: Ditto; let's centralize on the previous comment

full-node/db/sov-db/Cargo.toml
vlopes11: The rationale behind that is to make as many type representations available to `no-std` as possible, having their concrete implementation as added feature.

The trade-off is that we increase code bloat - especially feature shielding. Wdyt?

neysofu: I think it's a noble intent, but I don't see any possible usage of `sov-db` and `sov-schema-db` in a `no_std` environment at the moment. I may be wrong – if we can think of anything worthwhile, or if my assumption is incorrect, let's do it. As long as `sov_schema_db` and `sov_db` are only ever used in native execution environments with full access to `std`, I don't see a need for `no_std`.

rollup-interface/src/state_machine/mocks/da.rs
vlopes11: The problem of requiring that is we introduce the need to `extern crate alloc` on every library. They all compile to the same code, it's just a matter of what is more convenient. Wdyt?

rollup-interface/src/state_machine/mod.rs
vlopes11: Let's centralize the discussion on the other comment

rollup-interface/src/state_machine/da.rs
vlopes11: To double check: do you mean removing `thiserror` in favor of `core::error::Error`?

rollup-interface/src/lib.rs
vlopes11: Yes, it's a good approach from borsh that I think it's worthy to extend here.

rollup-interface/src/lib.rs
vlopes11: This is more of a lint limitation. The first is imported from `std`, and the second from the crate `spin`. Unfortunately, the lint will enforce the two are separated

rollup-interface/src/state_machine/mocks/da.rs
neysofu: I personally think `extern crate alloc` is ideal, it simplifies the API surface area at the cost of just 1 extra line of code.

rollup-interface/src/state_machine/da.rs
neysofu: Yes, indeed!

rollup-interface/src/state_machine/mocks/zk_vm.rs
vlopes11: Tested; the `bincode` will be compiled on non-optimized builds for `no-default-features`. Since it's not present under `no_std`, it will cause a compiler error

rollup-interface/src/state_machine/da.rs
vlopes11: Wdyt if we leave this change to this issue? https://github.com/Sovereign-Labs/sovereign-sdk/issues/1105

That way, we keep the changeset concise

rollup-interface/src/state_machine/mocks/da.rs
vlopes11: Moved to `src/state_machine/mocks/service.rs`

vlopes11:
@bkolad @neysofu review re-requested; `sov-schema-db` and `sov-db` removed in favor of a more minimal approach

full-node/db/sov-schema-db/src/db.rs
neysofu, line 1: Unused file, I guess it's a leftover from the `sov-schema-db` refactor.

rollup-interface/src/state_machine/da.rs
neysofu: Works for me!

neysofu (commented):
`sov-rollup-interface` is not `no_std`-compatible yet because of its dependency tree, unfortunately. We'll also need some checks in CI for it. @vlopes11 how do you want to handle this? It doesn't feel right to merge `no_std` support without actual `no_std` support, but if that's what it takes to reduce merge conflicts I'm okay with it, assuming we'll have the bandwidth to work on these fixes ASAP.

full-node/db/sov-schema-db/src/db.rs
vlopes11: It's odd that this file was preserved on the previous restore. Good catch; removed

neysofu (changes requested):
As per our discussion on Slack, let's address the dependency tree problem (and CI) before merging

bkolad (approved):

.github/workflows/rust.yml
```diff
@@ -122,6 +122,28 @@ jobs:
...
+          target: thumbv6m-none-eabi
+          override: true
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-provider: "buildjet"
+          shared-key: cargo-check-cache
+          save-if: ${{ github.ref == 'refs/heads/nightly' }}
+          workspaces: |
+            .
+            fuzz
```
neysofu: `cache-provider`, `shared-key` and `workspaces` are not necessary in this case. We can remove them

vlopes11: Done!

neysofu (approved):

PR #1078, opened 2023-10-20
Remove rust-version from cargo files

preston-evans98:
# Description
Currently, all of our `cargo.toml`s specify a `rust-version` even though we have ` rust-toolchain.toml` file. This removes that redundancy.


cemozerr (approved):

preston-evans98:
cc @neysofu @vlopes11, do either of you see any downsides here? I think these are just redundant, but don't remember why they were added

bkolad (approved):

neysofu:
`rust-toolchain.toml` specifies the intended Rust toolchain to develop the workspace in, whereas `rust-version` lets downstream users know the minimum required `rustc` version to be able to use the package.

It's a good thing to have both, but only if checked in CI with a tool like https://github.com/foresterre/cargo-msrv, otherwise `rust-version` will inevitably get out of sync, just like in our case. I'm in favor of removing it. It would be ideal to reintroduce it once we're confident we can provide correct values: https://github.com/Sovereign-Labs/sovereign-sdk/issues/1081.

neysofu (approved):

adapters/risc0/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = { workspace = true }
 homepage = { workspace = true }
 license = { workspace = true }
 repository = { workspace = true }
-rust-version = { workspace = true }
```
vlopes11, line 9: nit: consider not adding a blank line here so the header is concise

vlopes11 (approved):
Agreed with @neysofu . We do enforce `stable`, and that should suffice for us, as `stable` versions will upgrade passively over time (i.e. release train).

Unless there is an explicit reason to bind ourselves to an older `stable` (i.e. `1.66`), we shouldn't do that

citizen-stig (approved):

citizen-stig:
> `rust-toolchain.toml` specifies the intended Rust toolchain to develop the workspace in, whereas `rust-version` lets downstream users know the minimum required `rustc` version to be able to use the package.
> 
> It's a good thing to have both, but only if checked in CI with a tool like https://github.com/foresterre/cargo-msrv, otherwise `rust-version` will inevitably get out of sync, just like in our case. I'm in favor of removing it. It would be ideal to reintroduce it once we're confident we can provide correct values: #1081.

I agree that checking MSRV is good thing, but I believe that at this moment of rapid changes of the repo it might slow as down. So I am voting to remove rust_version now and re-untroduce it later. Because with MSRV we will also have to respect MSRV of core dependencies.

adapters/risc0/Cargo.toml
preston-evans98: Fixed

preston-evans98:
Ah, I'd forgotten this was used for MSRV checks. Yeah, we definitely need at least 1.70 already and will probably keep bumping that for a while. But agreed, let's add this back as things stabilize.

PR #1085, opened 2023-10-20
Implement additional Event macro

cemozerr:
# Description
Adds an Event macro that creates a <runtime_name>Event enum, almost identical to how the DispatchCall macro generates a <runtime_name>Call enum.


examples/simple-nft-module/src/lib.rs
```diff
@@ -36,6 +36,8 @@ impl<C: Context> Module for NonFungibleToken<C> {
 
     type CallMessage = CallMessage<C>;
 
+    type Event = ();
```
bkolad, line 39: Question:
How are we going to treat `EventKey`? Will it be always `Vec<u8>`, or should we allow any type to be the key?
cc @preston-evans98

bkolad (approved):
LGTM.

PR #1091, opened 2023-10-23
fix(sequencer): reject txs bigger than batch size

rkrasiuk:
# Description

Reject transactions bigger than the batch size from the mempool. Eliminates a misconfiguration issue where the sequencer would "choke" on a valid transaction bigger than batch size and never return a blob.

## Testing
Added a test case for rejecting tx bigger than batch size


bkolad:
Grate work @rkrasiuk. Thx!

citizen-stig (approved):

PR #1099, opened 2023-10-23
Limit CPU usage by `docker-compose` services

neysofu:
# Description
It might help with CI times of the `check-demo-rollup-bash-commands` job, which runs in parallel with `rustc` thus making it very CPU-starved.


neysofu:
Doesn't seem to make much of a difference; there must be another reason why `rustc` is so slow in this specific job.

PR #1107, opened 2023-10-25
Add get_key_string method to RuntimeEvent

cemozerr:
# Description
Added a `get_key_string` method to the `<runtime>Event` enum generated by the `Event` macro.

This method will eventually be used to replace the user-defined event keys currently utilized for accessing events in the database.

Additionally, added the `Event` macro to the `demo_stf` runtime, and an `Event` enum to the `ValueSetter` module for testing purposes.

## Testing
Tested the new `get_key_string` method in `derive_event.rs`


module-system/sov-modules-macros/src/event.rs
```diff
@@ -26,6 +26,44 @@ impl<'a> StructDef<'a> {
...
+        let enum_ident = self.enum_ident(EVENT);
+
+        let match_legs: Vec<proc_macro2::TokenStream> = self
+            .fields
+            .iter()
+            .map(|field| {
+                let module_name = &field.ident;
+                let module_name_str = &field.ident.to_string();
+                quote::quote!(
+                    #enum_ident::#module_name(inner)=>{
```
bkolad, line 40: Will this code still behave as expected if the `event` is `struct/tuple/array` instead of `enum`?

module-system/sov-modules-macros/tests/dispatch/derive_event.rs
```diff
@@ -12,12 +12,35 @@ struct Runtime<C: Context>
...
 
 fn main() {
     // Check to see if the runtime events are getting initialized correctly
-    let _event = RuntimeEvent::<DefaultContext>::first(first_test_module::Event::FirstModuleEnum1(10));
-    let _event = RuntimeEvent::<DefaultContext>::first(first_test_module::Event::FirstModuleEnum2);
-    let _event = RuntimeEvent::<DefaultContext>::first(first_test_module::Event::FirstModuleEnum3(vec![1;3]));
-    let _event = RuntimeEvent::<DefaultContext>::second(second_test_module::Event::SecondModuleEnum);
+    {
+        let event = RuntimeEvent::<DefaultContext>::first(first_test_module::Event::FirstModuleEnum1(10));
+        assert_eq!(event.get_key_string(), "first-FirstModuleEnum1");
```
bkolad, line 22: Will this generate an expected string even if  `Event` is `struct/tuple/array/nested enum` instead of an enum type?

module-system/sov-modules-macros/tests/dispatch/derive_event.rs
```diff
@@ -12,12 +12,35 @@ struct Runtime<C: Context>
...
+    }
+
+    {
+        let event = RuntimeEvent::<DefaultContext>::first(first_test_module::Event::FirstModuleEnum2);
+        assert_eq!(event.get_key_string(), "first-FirstModuleEnum2");
+    }
+
+    {
+        let event = RuntimeEvent::<DefaultContext>::first(first_test_module::Event::FirstModuleEnum3(vec![1; 3]));
+        assert_eq!(event.get_key_string(), "first-FirstModuleEnum3");
```
bkolad, line 32: To create the key string using the `get_key_string` function, we require a value of `Event::FirstModuleEnum3(..)`. However, obtaining this value necessitates providing a `Vec` this is fine when we save to the db.  

Now, let's consider a scenario where a user wishes to query the database using the `first-FirstModuleEnum3` key. What will be the method for the user to calculate the `first-FirstModuleEnum3`? Since he doesn't have a value of the `Event::FirstModuleEnum3(...)` he can't call `get_key_string(&self)`. We have to think about the cases where the enums contain arbitrary complex structures and possibly other nested enums.

cemozerr:
Closing in favor of: https://github.com/Sovereign-Labs/sovereign-sdk/pull/1116

Issue #1023, opened 2023-10-12
Data Availability re-orgs handling tracking issue


citizen-stig:
At present, the Sovereign SDK processes only finalized blocks. To enhance response times,  the SDK should be equipped to manage re-organizations of the blockchain structure due to forks in underlying Data Availability(DA) layers.

## Steps

- [x] Update `rollup-interface` with notion of external storage: https://github.com/Sovereign-Labs/sovereign-sdk/pull/1103
- [x] https://github.com/Sovereign-Labs/sovereign-sdk/issues/1024
- [x] https://github.com/Sovereign-Labs/sovereign-sdk/issues/1026
    - [x] Add snapshot support to `sov-schema-db` https://github.com/Sovereign-Labs/sovereign-sdk/pull/1112
    - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1177
    - [x] https://github.com/Sovereign-Labs/sovereign-sdk/issues/1027
    - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1197 
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/issues/1217
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/issues/1218

## Prerequisites:
  - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/978 as it leaks abstraction
 
 
 ## Finalization and Refactoring
   - [x] Bug: Handle state root change in runner in case of re-org
   - [x] ProverStorageManager: wrapper struct for 3 RwLocks so they all written together.
   - [x] ChangeSet from default Stf should be a different type from PreState
   - [ ] Remove explicit Arc<RwLock<Storage>> from Blueprint and Runner and unify it with RpcStorage.
   - [x] https://github.com/Sovereign-Labs/sovereign-sdk/issues/1249
   - [x] https://github.com/Sovereign-Labs/sovereign-sdk/issues/643 as less types will simplify the change
   - [ ] https://github.com/Sovereign-Labs/sovereign-sdk/issues/1185
   - [ ] Refactor `SchemaIterator` so it utilizes already tested logic of `RawDbIter`
   - [ ] Explore removing Mutex<ChangeSet> inside CacheDb and change API to have &mut self
   - [ ] Optimize CacheDb for multithreaded execution.

Issue #791, opened 2023-09-04
Make writes of `SchemaBatch`es atomic across `StateDb`'s `JmtNodes`, `JmtValues`, and `NativeDb`'s accessory state


neysofu:
## Summary

Our writes to `StateDb` are not fully atomic as our `SchemaBatch` writes tend to be isolated to a single column family, even when the write operation should arguably be atomic across multiple column families. The same holds for `NativeDb`. We should change the `SchemaBatch` API to allow atomic writes across several column families.

Context: #773.

PR #1112, opened 2023-10-26
Add `DbSnapshot` to `sov-db-schema`

### LINKED ISSUE #1026: `Storage` should support nested snapshots
So re-orgs can be handled from data perspective

citizen-stig:
# Description

This PR adds: `DbSnapshot`, a wrapper around `sov_schema_db::DB`, which has it's local cache in form of adapted `SchemaBatch` and link to `QueryManager`. This allows `DbSnapshot` read/write value to local cache, fetch them from parent snapshots or directly from DB.

Please note, that this functionality is added, but it is not integrated. Next step would be modifying `StateDB`, `NativeDB` and `LedgerDB` to work with `DbSnapshot` instead of `DB` directly.

Main modifications:

- *`sov_schema_db::SchemaBatch`: vector of operations replaced with HashMap, which allows efficient reads. 
- internal `Mutex` is removed from `SchemaBatch` and now it is responsibility of the caller to handle. Current usage of SchemaBatch is not concurrent. `DbSnapshot` uses `Mutex` over whole internal cache. Potentially related to https://github.com/Sovereign-Labs/sovereign-sdk/issues/791

## Linked Issues
- Related to #1023 

## Testing
New integration test has been added to test snapshot

## Docs
Documentation to the new modules have been added

### DIFF
--- a/full-node/db/sov-schema-db/src/snapshot.rs
+++ b/full-node/db/sov-schema-db/src/snapshot.rs
@@ -116,11 +116,7 @@ pub struct FrozenDbSnapshot {
 impl FrozenDbSnapshot {
     /// Get value from its own cache
     pub fn get<S: Schema>(&self, key: &impl KeyCodec<S>) -> anyhow::Result<Option<Operation>> {
-        if let Some(operation) = self.cache.read(key)? {
-            return Ok(Some(operation));
-        }
-
-        Ok(None)
+        Ok(self.cache.read(key)?)
     }
 
     /// Get id of this Snapshot
--- a/full-node/db/sov-schema-db/src/snapshot.rs
+++ b/full-node/db/sov-schema-db/src/snapshot.rs
@@ -24,7 +24,6 @@ pub struct ReadOnlyLock<T> {
 }
 
 impl<T> ReadOnlyLock<T> {
-    #[allow(dead_code)]
     /// Create new [`ReadOnlyLock`] from [`Arc<RwLock<T>>`].
     pub fn new(lock: Arc<RwLock<T>>) -> Self {
         Self { lock }
@@ -37,7 +36,6 @@ impl<T> ReadOnlyLock<T> {
 }
 
 /// Wrapper around [`DB`] that allows to read from snapshots
-#[allow(dead_code)]
 pub struct DbSnapshot<Q> {
     id: SnapshotId,
     cache: Mutex<SchemaBatch>,
@@ -45,7 +43,6 @@ pub struct DbSnapshot<Q> {
     db_reader: Arc<DB>,
 }
 
-#[allow(dead_code)]
 impl<Q: QueryManager> DbSnapshot<Q> {
     /// Create new [`DbSnapshot`]
     pub fn new(id: SnapshotId, manager: ReadOnlyLock<Q>, db_reader: Arc<DB>) -> Self {
--- a/full-node/db/sov-schema-db/src/snapshot.rs
+++ b/full-node/db/sov-schema-db/src/snapshot.rs
@@ -80,7 +80,7 @@ impl<Q: QueryManager> DbSnapshot<Q> {
                 .manager
                 .read()
                 .expect("Parent lock must not be poisoned");
-            if let Some(operation) = parent.get(self.id, key)? {
+            if let Some(operation) = parent.get::<S>(self.id, key)? {
                 return decode_operation::<S>(operation);
             }
         }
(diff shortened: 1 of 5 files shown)


full-node/db/sov-schema-db/src/snapshot.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub fn read<S: Schema>(&self, key: &impl KeyCodec<S>) -> anyhow::Result<Option<S::Value>> {
+        // Some(Operation) means that key was touched,
+        // but in case of deletion we early return None
+        // Only in case of not finding operation for key,
+        // we go deeper
+
+        // 1. Check in cache
+        if let Some(operation) = self
+            .cache
+            .lock()
```
bkolad, line 65: I am not sure what is the scope for `if condition`. Is it the entire function of just the `if statement`? In order to avoid confusion let's read the variable outside of the `if` and drop it explicitly?

full-node/db/sov-schema-db/src/snapshot.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+impl FrozenDbSnapshot {
+    /// Get value from its own cache
+    pub fn get<S: Schema>(&self, key: &impl KeyCodec<S>) -> anyhow::Result<Option<Operation>> {
+        if let Some(operation) = self.cache.read(key)? {
+            return Ok(Some(operation));
+        }
+
+        Ok(None)
```
bkolad: Would ` Ok(self.cache.read(key)?)` work

full-node/db/sov-schema-db/tests/snapshot_test.rs
```diff
@@ -0,0 +1,142 @@
...
+
+impl QueryManager for LinearSnapshotManager {
+    fn get<S: Schema>(
+        &self,
+        snapshot_id: SnapshotId,
+        key: &impl KeyCodec<S>,
+    ) -> anyhow::Result<Option<Operation>> {
+        for snapshot in self.snapshots[..snapshot_id as usize].iter().rev() {
+            if let Some(operation) = snapshot.get(key)? {
+                return Ok(Some(operation));
```
bkolad, line 73: Same `Ok(self.cache.read(key)?)`

full-node/db/sov-schema-db/tests/snapshot_test.rs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+impl QueryManager for LinearSnapshotManager {
+    fn get<S: Schema>(
+        &self,
+        snapshot_id: SnapshotId,
+        key: &impl KeyCodec<S>,
+    ) -> anyhow::Result<Option<Operation>> {
+        for snapshot in self.snapshots[..snapshot_id as usize].iter().rev() {
+            if let Some(operation) = snapshot.get(key)? {
```
bkolad, line 72: Same: `Ok(self.cache.read(key)?)`

full-node/db/sov-schema-db/src/snapshot.rs
bkolad: Alternatively we could add another method `self.read` similar to: `self.put` and `self.delete` bellow and do the locking there.

bkolad (approved):

full-node/db/sov-schema-db/src/snapshot.rs
```diff
@@ -0,0 +1,158 @@
...
+pub type SnapshotId = u64;
+
+/// A trait to make nested calls to several [`Schema`]
+pub trait QueryManager {
+    /// Get a value from snapshot or its parents
+    fn get<S: Schema>(
+        &self,
+        snapshot_id: SnapshotId,
+        key: &impl KeyCodec<S>,
+    ) -> anyhow::Result<Option<Operation>>;
```
neysofu: I wonder if `Option<Operation>` is a leaky abstraction in this case. Do we really need to distinguish between an absent value and a deleted value?

citizen-stig: Yes, the distinction is important, because if value is absent, then snapshot needs to query database, but if it was deleted, "None" is the correct response, even if value is still in database.

There might be different approach tho, when snapshot only reads from local and parent, but manager have to read from the database if value wasn't found in snapshots.

neysofu: Ah, yes! I totally missed that.

neysofu (approved):
LGTM!

full-node/db/sov-schema-db/src/snapshot.rs
citizen-stig: Pros for `QueryManager` to read from database:

 * There's single instance of QueryManager, which means only single handle to the database.
 * It is ReadLock, so many readers should not block each other
 
 Cons:
 
  * When QueryManager will be flushing snapshot to rocksdb, it will block all readers. It will block them in current design, but for less time. 
  
I wonder if benchmark is needed.

Issue #219, opened 2023-05-01
separating sov-state into two crates (an API crate and an implementation crate)


citizen-stig:
That would signficantly speed up CI, since we could stop importing RocksDB everywhere

citizen-stig:
I think it is already happend with introduction of `sov-modules-core`

PR #1113, opened 2023-10-26
feat: add `no-std` to `sov-state`

vlopes11:
This commit introduces `no-std` to `sov-state`. It adds a `sync` feature that will shield targets that contains atomic synchronization implementations.

A completely bare target will not have such functionality available, and could benefit from some of the primitives that will not rely on components such as `Arc`.


module-system/sov-state/Cargo.toml
```diff
@@ -37,11 +36,37 @@ tempfile = { workspace = true }
...
+    "sov-rollup-interface/default",
+    "sov-first-read-last-write-cache/default",
+    "sync",
+]
+sync = [
+    "borsh/rc",
+    "jmt",
+    "serde/rc",
+    "sov-first-read-last-write-cache",
 ]
```
neysofu, line 72: Do we absolutely need all of these 3 new features? Seems like a lot :/

vlopes11: We need to decouple `sov-db` from native as it is std-only.

The `sync` is for bare environments without atomic synchronization (might be the case for some hardware wallets). Basically it would require this state without `sync` capabilities, but still have critical components available such as the `Storage` trait definition

module-system/utils/sov-first-read-last-write-cache/src/cache.rs
```diff
@@ -1,14 +1,18 @@
-use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+use alloc::vec::Vec;
 
-use thiserror::Error;
+use sov_rollup_interface::maybestd::collections::hash_map::Entry;
+use sov_rollup_interface::maybestd::collections::HashMap;
```
neysofu, line 4: These two are the only imports for which we use `sov_rollup_interface`. I think it's best if we create a new facade instead of depending on it, maybe a minimal one with just `HashMap`. WDYT?

module-system/sov-state/src/lib.rs
neysofu, line 1: While this kind of extensive feature gating is the most obvious way to do the job, it's also quite the code quality smell and I think we ought to find a better solution. If all we need is the `Storage` trait and a handful of types/utils, then we ought to find a better place for them. Either another crate (an existing one or, if we have no good candidate, a new one) or a deeper module hierarchy that clearly separates crate modules between native-only and `no_std`.

@preston-evans98 do you have any ideas here?

neysofu (changes requested):
As discussed during standup. Let me know when you need another round of review!

module-system/sov-state/Cargo.toml
citizen-stig: Just to add to the discussion, we stumbled upon similar issue before:

https://github.com/Sovereign-Labs/sovereign-sdk/issues/219

So maybe we can reconsider split of `sov-state` create into 2

module-system/sov-state/src/lib.rs
citizen-stig: Just to chime in: https://github.com/Sovereign-Labs/sovereign-sdk/issues/219

vlopes11:
Superseded by https://github.com/Sovereign-Labs/sovereign-sdk/pull/1126

PR #1115, opened 2023-10-27
Add 'kernel' to AppTemplate

preston-evans98:
# Description
This PR adds the notion of a `kernel` to the SDK. This is a very small code diff, but represents a significant departure from our former philosophy that "everything is a module". 

The motivation for this change is to support soft-confirmations more effectively. When only based sequencing was enabled, there was no need for any "kernel" logic because no entity had any real power over the rollup's state transition. Blobs were processed as they appeared on the DA layer, and that was that. 

Once we introduce a preferred sequencer, things get more complicated. We want to give the sequencer enough power over the rollup to give out reliable soft confirmations. At the same time, we need to ensure that a malicious sequencer can't halt the rollup or censor transactions. These two goals are in tension with each other. On the one hand, the goal of giving out reliable soft confirmations implies that we want the rollup state to be independent of things that happen on the DA layer. If the preferred sequencer incurs a one or two block delay while trying submit a batch to L1, that should not change the results of the batch. On the other hand, we need the rollup to be "aware of" the underlying DA layer so that it can prevent censorship (by forcing the inclusion of certain transactions). 

The role of the kernel is to reconcile these different goals. Functionality which depends directly on the DA layer (things like processing forced transactions, watching for attestations, etc.) can be moved into the kernel, while ordinary "user-space" transactions remain independent of the DA layer. 

In this PR, the Kernel's only responsibility is to select which blobs will be executed in the current slot. The next several PRs will:
1. Introduce the concept of a KernelModule, which has "protected" state that is not always readable from user space
2. Ensure that user-space transactions are independent of DaLayer information
3. Move the ChainState and BlobStorage modules into the kernel
4. Introduce a constrained user-space mechanism for reading kernel values. This will be kernel-dependent - the simple kernel with no preferred sequencer will allow immediate access to all kernel state, while the preferred sequencer variant will gate access to recent updates in order to avoid introducing non-determinism  from the DA layer


module-system/sov-modules-stf-template/src/app_template.rs
```diff
@@ -23,10 +24,11 @@ use sov_zk_cycle_macros::cycle_tracker;
 /// An implementation of the
 /// [`StateTransitionFunction`](sov_rollup_interface::stf::StateTransitionFunction)
 /// that is specifically designed to work with the module-system.
-pub struct AppTemplate<C: Context, Da: DaSpec, Vm, RT: Runtime<C, Da>> {
+pub struct AppTemplate<C: Context, Da: DaSpec, Vm, RT: Runtime<C, Da>, K: Kernel<C, Da>> {
     /// State storage used by the rollup.
     /// The runtime includes all the modules that the rollup supports.
     pub(crate) runtime: RT,
+    pub(crate) kernel: K,
```
bkolad, line 31: Question: Will we have a finite set of `Kernel` implementations or do we expect the users co create their own `Kernels`? If we have a finite set we could have something like:
```
enum Kernels{
 BasicKernel(BC),
 SomeOtherKernel(..),
...
}
```
And choose the correct one from the runtime config, this way we could avoid introducing the additional generic.

bkolad (approved):

bkolad:
LGTM. After merging to nightly, could you please update  `https://github.com/Sovereign-Labs/sov-rollup-starter` with the relevant changes as well? (we should see nightly CI failing)

module-system/sov-modules-stf-template/src/app_template.rs
preston-evans98: I think we will probably be able to reduce this to a finite set, but I'm not quite sure yet

PR #1116, opened 2023-10-27
Add auto-generated key methods for Events

cemozerr:
# Description
We want to use typed objects instead of the handwritten events in the codebase so that the event processes are less error-prone, and its easer/ more obvious for users to index these events.

For this purpose, in this PR, we:
- Added an `Event` trait with methods an `event_key()`, which returns a given Event object's key, and `get_all_event_keys()`, which returns all possible event keys for a module, and put this trait as a trait bound on the `Event` associated type on the `Module` trait.
- Added an `Event` macro that derives the `Event` trait methods for a given enum.
- Renamed the original `Event` macro that creates an enum that represents all possible events from all the modules to `RuntimeEvent`.
- Added a `get_key_string` method to the `<runtime>Event` enum generated by the (now called) `RuntimeEvent` macro, which returns a given `<Runtime>Event`'s key.
- Added a `get_all_key_strings` method to the `<runtime>Event` enum, which returns the key strings for all events in the runtime, spanning across all modules.
- Renamed the legacy `Event` type from `rollup_interface` to `LegacyEvent` to fix name clashing.  The `LegacyEvent` will be completely removed in an upcoming PR as we move to a simpler `Event` flow.

The diffs for the macro files are unfortunately bigger than necessary as the name `event.rs` was more appropriate for the new macro than the previous `RuntimeEvent` macro.


module-system/sov-modules-api/src/lib.rs
```diff
@@ -340,6 +340,23 @@ pub trait Module {
     }
 }
 
+impl Event for () {
+    fn event_key(&self) -> &'static str {
+        unimplemented!("Event type for this module does not exist.")
+    }
+}
```
preston-evans98: I know I told you to go ahead and do this, but I'm having second thoughts now 😅. 

With the current setup, there might a possibility that someone who meant to have no events could accidentally leave an `event.emit()` in their module and cause a `panic` if a user ever triggered that code path. Two potential options are:
1. Switch to a `NonInstantiable` event type which has no constructor - so the compiler will prevent them from ever emitting the event
2. Change this to emit a dummy key like `"()"` or `"nil"`

cemozerr: I think it makes sense to prevent user from emitting the event, so I like switching to `NonInstantiable`. 

Can you explain how this prevents that though?

cemozerr: TBF, the downside of `NonInstantiable` is that a module that does not have call messages are implemented as 
```type CallMessage = ()```, so having ```type Event = NonInstantiable``` is a bit of a surprise for the user.

Comment (preston-evans98) [reply module-system/sov-modules-api/src/lib.rs]: Yeah. The reason for this is that `Callmessage`s have to be serializable and deserializable - which by gives users a way to instantiate them and force the rollup to deal with it.

Comment (preston-evans98) [reply module-system/sov-modules-api/src/lib.rs]: We used to have `CallMessage = NonInstantiable`, but it was insecure for that reason. If a user made a transaction for (say) BlobStorage (which had a `NonInstantiable` CallMessage, the rollup would be forced to process it. That processing *should* just be a no-op - but because the type claimed to be non-instantiable devs would sometimes assume they could panic instead

File: module-system/sov-modules-macros/src/runtime_event.rs [hunk 2]
```diff
@@ -0,0 +1,115 @@
...
+impl<'a> StructDef<'a> {
+    fn create_event_enum_legs(&self) -> Vec<proc_macro2::TokenStream> {
+        self.fields
+            .iter()
+            .map(|field| {
+                let name = &field.ident;
+                let ty = &field.ty;
+
+                quote::quote!(
+                    #[doc = "Module event."]
```
preston-evans98: nit: This doc is a little terse. Maybe we could say something like, "an event emitted by the {} module"?

module-system/sov-modules-api/src/lib.rs
cemozerr: Switched to emitting dummy key with warning after our offline discussion.

module-system/sov-modules-macros/src/runtime_event.rs
cemozerr: Great catch, this was from previous PR, but will improve it with your suggestion.

module-system/sov-modules-macros/src/runtime_event.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        let match_legs: Vec<proc_macro2::TokenStream> = self
+            .fields
+            .iter()
+            .map(|field| {
+                let module_name = &field.ident;
+                let module_name_str = &field.ident.to_string();
+                quote::quote!(
+                    #enum_ident::#module_name(inner)=>{
+                        let enum_name: String = format!("{:?}", inner)
```
preston-evans98: Looks like we forgot to replace this with a call to `inner.event_key()`

preston-evans98 (changes requested):
Looks good except for the `create_get_key_string_impl` body. Just need to replace the `format!("{:?}", inner)` call with `inner.event_key()` and we should be good to go

module-system/sov-modules-macros/src/runtime_event.rs
cemozerr: Done.

module-system/sov-modules-api/src/lib.rs
```diff
@@ -340,6 +340,24 @@ pub trait Module {
...
+
+/// Every module `Event` has to implement this trait.
+///
+/// You can use `#[derive(Event)]` to automatically implement the trait. If your `Event` is not an enum,
+/// or if your `Event` type is a nested enum and you want to index events based on their nested fields,
+/// you will need to implement the trait manually.
+pub trait Event: Debug + BorshSerialize + BorshDeserialize {
+    /// Returns a static string slice that identifies the Event variant.
+    /// This string slice is primarily used to index Events from the DB.
+    fn event_key(&self) -> &'static str;
```
bkolad, line 368: Is the `event_key(&self)` going to be used also for quering the database (to get all events related to a particlualar keys)?  If yes then a user will need to generate an `Event` value  to use this method.  However, it may be unclear how to achieve this when dealing with a complex structure or enum as the event.

cemozerr: Yeah, great point. 

Shall we have some cli tooling to generate this? I haven't looked into sov-cli much but I can imagine that being able to printing event_key's for event variants.

@preston-evans98

cemozerr: Added a method to retrieve all event_key's of a runtime. Next step will be to output them to a file on startup.

PR #1117, opened 2023-10-27
feat: add `tokio` as part of `native` feature

vlopes11:
Prior to this commit, `tokio` was split from the `rollup-interface` native feature so the ledger functionality would be available for no-std.

This commit simplifies the structure and will require `std/tokio` when `native` is active.


preston-evans98 (approved):
LGTM, thanks

PR #1118, opened 2023-10-30
Update `DaService` interface to support re-orgs

### LINKED ISSUE #1024: Make `DaService` trait aware about re-ogranizations and finalization
`DaService` should be able to do following things:

* Provide blocks as they arrive, even if their height is below latest height.
* Provide finalization events for blocks.

So current methods related to getting blocks should be changed:

```rust

pub trait DaService: Send + Sync + 'static {
    type FilteredBlock: SlotData<
        BlockHeader = <Self::Spec as DaSpec>::BlockHeader,
        Cond = <Self::Spec as DaSpec>::ValidityCondition,
    >;


    type Error: fmt::Debug + Send + Sync + Display;

    async fn get_finalized_at(&self, height: u64) -> Result<Self::FilteredBlock, Self::Error>;
    async fn get_block_at(&self, height: u64) -> Result<Self::FilteredBlock, Self::Error>;
    
```

Details of the implementation to be defined.

citizen-stig:
# Description

* Add `get_last_finalized_block_header` and subscription based version `subscribe_finalized_header`
* Add `get_head_block_header` for better UX and other optimizations
* Remove `get_finalized_at` in favour of `get_block_at` + `get_latest_finalized_block_header` 

## Difference between  `get_last_finalized_block_header` and `subscribe_finalized_header`

* `get_last_finaliezed_block_header` is simple request to get finalized block header at current moment. It can be used in simple for of rollup runner, or for syncronization
* `subscribe_finalized_header` returns a stream, that can be consumed continously and used for persisting non-finalized state to the database, publish proofs, update RPC storage.

Currently it seems that both methods are needed, but maybe in the future one of them will be removed

## Note on subscription implementation

Interface defines pretty basic constraint of [`futures::Stream`](https://docs.rs/futures/0.3.29/futures/stream/trait.Stream.html) trait.

Most implementation just implement a wrapper that does result type conversion. I've added `pin-project` dependency to handle Pin projections inside `Stream` implementation.

## Linked Issues
- Fixes #1024

## Testing
`MockDaService` have new unit tests added. Celestia and Avail tested manually.

## Docs
Doc strings in relevant structs and traits have been updated


rollup-interface/src/node/services/da.rs
```diff
@@ -30,17 +32,33 @@ pub trait DaService: Send + Sync + 'static {
...
 
+    /// Fetch the [`DaSpec::BlockHeader`] of the last finalized block.
+    async fn get_last_finalized_block_header(
+        &self,
+    ) -> Result<<Self::Spec as DaSpec>::BlockHeader, Self::Error>;
+
+    /// Subscribe to finalized headers as they are finalized.
+    /// Optimized version of `get_last_finalized_block_header`.
+    /// For the environments that support [`tokio::sync::broadcast`].
+    fn subscribe_finalized_header(
```
bkolad: Wheat environment doesn't support `tokio::sync::broadcast`?

adapters/avail/src/service.rs
```diff
@@ -252,3 +300,20 @@ impl DaService for DaProvider {
         Ok(())
     }
 }
+
+/// Used to keep channel open even if there's no other consumers
+fn run_finalized_header_consumer(mut receiver: Receiver<AvailHeader>) {
```
bkolad: Can you please expand why do we ned this method?

rollup-interface/src/node/services/da.rs
```diff
@@ -30,17 +32,33 @@ pub trait DaService: Send + Sync + 'static {
...
+        &self,
+    ) -> Result<tokio::sync::broadcast::Receiver<<Self::Spec as DaSpec>::BlockHeader>, Self::Error>;
```
preston-evans98: Why do we return a `Receiver` instead of an associated type that implements [`Stream`](https://docs.rs/futures/latest/futures/stream/trait.Stream.html)?

rollup-interface/src/node/services/da.rs
```diff
@@ -30,17 +32,33 @@ pub trait DaService: Send + Sync + 'static {
...
-    /// The returned block may not be final, and can be reverted without a consensus violation
+    /// The returned block may not be final, and can be reverted without a consensus violation.
+    /// Call it for the same height are allowed to return different results.
+    /// Should always returns the block at that height on the best fork.
     async fn get_block_at(&self, height: u64) -> Result<Self::FilteredBlock, Self::Error>;
 
+    /// Fetch the [`DaSpec::BlockHeader`] of the last finalized block.
+    async fn get_last_finalized_block_header(
+        &self,
+    ) -> Result<<Self::Spec as DaSpec>::BlockHeader, Self::Error>;
```
preston-evans98, line 51: Does this need to be an `Option<BlockHeader>`? I suppose we can consider the genesis block of all chains to be final and expect clients to return that block even if no others have been finalized - but we may want to document that?

adapters/avail/src/service.rs
preston-evans98: +1

rollup-interface/src/node/services/da.rs
citizen-stig: In mock implementation, I return "error" if chain hasn't reached finality from the genesis.

That's a good point.

rollup-interface/src/node/services/da.rs
citizen-stig: I think `Stream` is a better option.

Originally I followed [`LedgerRpcProvider::subscribe_slots`](https://github.com/Sovereign-Labs/sovereign-sdk/blob/nightly/rollup-interface/src/node/rpc/mod.rs#L305)

adapters/avail/src/service.rs
citizen-stig: TLDR: It can be removed, as discussed with Blazej yesterday.

To clarify. In the implementation, subcription to finalized headers starts at the initialization of Client, so there should be background task that will do 2 things:

1. Keep channel open for new subsribers
2. Keep channel not filled to the end, until actual subcribers appear.


But if `subscribe_finalized_header` takes `&mut self`, then it can have an option and only subcribe when first `subscribe_finalized_header` is called.

bkolad (approved):

PR #1125, opened 2023-10-31
Prefer `sov_modules_api` over direct `sov_rollup_interface` for modules

citizen-stig:
# Description

Following approach for modules to use `sov_modules_api` instead of sov_rollup_interface. 

it also simplifies potential feature matrix

## Testing
No updates

## Docs
No updates


module-system/sov-modules-api/Cargo.toml
```diff
@@ -68,5 +68,6 @@ native = [
     "macros",
     "sov-modules-macros/native",
     "sov-state/native",
+    "sov-rollup-interface/native",
```
citizen-stig: I can remove this, but in reality it is imported as "native" from `sov-state/native`, as `sov-db` imports rollup interface with native

bkolad (approved):

PR #1126, opened 2023-10-31
feat: add sov-modules-core

vlopes11:
This commit introduces `sov-modules-core`, a new workspace member that will contain the required declarations for `no-std` `DispatchCall` runtime message declarations.

The new module is re-exported via `sov-state` and `sov-modules-api` to minimize breaking changes downstream.

Finally, it introduces a `RefCount` in `sov-rollup-interface` that will contain atomic synchronization, if the compile target contains atomic pointers. It safely fallback to `Rc`, that is a reference count without atomic synchronization, but equivalent API, that will be the used type on targets without synchronization. This can replace all the `Arc` instances in the future, but for now, it is required only for the previously `sov-state` storage/cache keys and values, so the `Storage` trait can be properly defined.


Cargo.toml
```diff
@@ -25,10 +25,10 @@ members = [
     "module-system/sov-modules-stf-template",
     "module-system/sov-modules-rollup-template",
     "module-system/sov-modules-macros",
+    "module-system/sov-modules-core",
     "module-system/sov-state",
     "module-system/sov-modules-api",
     "module-system/module-schemas",
-    "module-system/utils/sov-first-read-last-write-cache",
```
citizen-stig, line 34: Do I understand corrrectly, that this crate merged into `sov-modules-core`

adapters/celestia/Cargo.toml
```diff
@@ -9,7 +9,7 @@ publish = false
 
 [dependencies]
 borsh = { workspace = true, features = ["bytes"] }
-bech32 = { workspace = true }
+bech32 = { workspace = true, default-features = true }
```
citizen-stig, line 12: I honestly don't know, but what is the point of having `default-features = true` explicitly

module-system/sov-modules-core/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+name = "sov-modules-core"
+description = "Defines the core components of the Sovereign SDK module system"
+authors = { workspace = true }
+edition = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+repository = { workspace = true }
+
+version = { workspace = true }
+readme = "README.md"
```
citizen-stig, line 11: It looks like REDME file is missing

module-system/sov-modules-core/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
citizen-stig, line 1: Can we also have `#![deny(missing_docs)]` here, so we don't have to do it again.

module-system/sov-state/src/codec/borsh_codec.rs
```diff
@@ -7,7 +10,7 @@ pub struct BorshCodec;
 
 impl<K> StateKeyCodec<K> for BorshCodec
 where
-    K: borsh::BorshSerialize + borsh::BorshDeserialize,
+    K: BorshSerialize + borsh::BorshDeserialize,
```
citizen-stig: Nit: could we stick to one approach for both?

citizen-stig (approved):
This is big change, hard to spot if something is missing.

But LGTM, only couple small comments.

module-system/sov-modules-core/src/storage.rs
```diff
@@ -1,24 +1,26 @@
-use std::fmt::Display;
```
bkolad: The file hierarchy is too flat, and we should split it logically into several rust modules:
- `storage, witness, cache, scratchpad` 
- `key, codec`
- everything else

module-system/sov-modules-api/src/dispatch.rs
```diff
@@ -43,10 +23,27 @@ pub trait EncodeCall<M: Module> {
...
-pub trait CliWallet: DispatchCall {
+pub trait CliWallet: sov_modules_core::DispatchCall {
     /// The type that is used to represent this type in the CLI. Typically,
     /// this type implements the clap::Subcommand trait. This type is generic to
     /// allow for different representations of the same type in the interface; a
     /// typical end-usage will impl traits only in the case where `CliStringRepr<T>: Into::RuntimeCall`
     type CliStringRepr<T>;
 }
+
+impl<T> Genesis for T
```
bkolad: What is the methodology behind moving code from `sov-modules-api` to `core`? 
If we moved the `Module and DispatchCall` traits there, we should also move `Genesis`, so they are next to each other.

Cargo.toml
vlopes11: Yes, that is correct. All the types there are required by `sov-modules-core`

adapters/celestia/Cargo.toml
vlopes11: `bech32` is a requirement of `sov-modules-core`, and it must be `default-features = false`. This way, the bare minimum we declare on the root workspace is with such state, and if the library is declared elsewhere, it is expected to have the default features as these crates won't split std/no-std by feature

module-system/sov-modules-core/Cargo.toml
vlopes11: Good catch; fixed

module-system/sov-modules-core/src/lib.rs
vlopes11: Good suggestion; updated

module-system/sov-state/src/codec/borsh_codec.rs
vlopes11: Done!

module-system/sov-modules-api/src/dispatch.rs
vlopes11: The approach was to move only the minimum possible set of traits/structs to core so we can declare a runtime as no-std, but it makes sense. Done!

module-system/sov-modules-core/src/storage.rs
vlopes11: Good suggestion; divided into 2 levels

bkolad (approved):

PR #1128, opened 2023-11-01
Move mock adapters to separate crates

citizen-stig:
# Description

* Remove `mocks` feature in `sov-rollup-interface`
* Add 2 new crates: `sov-mock-da` and `sov-mock-zkvm`
* Remove `verifier` feature from **all** DA adapters
* `demo-simple-stf` was calling "native" method from `BlobReader`. There is some work needs to be done: https://github.com/Sovereign-Labs/sovereign-sdk/issues/1129

(!) Note. `sov-mock-da` should be imported with "native" flag if any module is imported with native flag, because modules indirectly import `rollup-interface` with native, which means that `sov-mock-da` can have not fully implement `BlobReader` if it also does not have "native"

## Testing
Existing tests are passing

## Docs
Documentation has been updated


bkolad (approved):
LGTM, please also update `sov-seq-starter` with the relevant changes.

PR #1133, opened 2023-11-02
Prover Storage Manager 

### LINKED ISSUE #1027: Crate for managing Storage snapshots relationship to DA forks
Crate that manages relations between snapshot and underlying forks of Data availability service

citizen-stig:
# Description

* Adds `HierarchicalStorageManager` trait to rollup interface, which will replace `StorageManager` after these changes are integrated
* Adds crate `sov-prover-storage-manager`, which have implmenetations for `HierarchicalStorageManager` and `sov-schema-db::snapshot::QueryManager` traits.
* Adds `std::hash::Hash` bound to `rollup_interface::BlockHashTrait`. 
* Minor interface improvement for `DbSnapshot` to use references instead of value.

Note, this change will require performance benchmark after being integrated.

## Linked Issues
- Related to #1023 

## Testing
Unit tests have been added to new crate

## Docs
Rust documentation have been updated

### DIFF
--- a/full-node/sov-prover-storage-manager/src/dummy_storage.rs
+++ b/full-node/sov-prover-storage-manager/src/dummy_storage.rs
@@ -1,3 +1,5 @@
+// This module is used as a temporary filler for `ProverStorage`, until this module is integrated.
+// It is going to be deleted after integration has been completed
 use std::marker::PhantomData;
 
 use byteorder::{BigEndian, ReadBytesExt};
@@ -6,7 +8,7 @@ use sov_schema_db::snapshot::{DbSnapshot, QueryManager};
 use sov_schema_db::{define_schema, CodecError};
 use sov_state::MerkleProofSpec;
 
-/// Oversimplified representation of [`sov_state::ProverStorage`]
+// Oversimplified representation of [`sov_state::ProverStorage`]
 pub struct NewProverStorage<Mps: MerkleProofSpec, Q> {
     state_db: DbSnapshot<Q>,
     native_db: DbSnapshot<Q>,
--- a/full-node/sov-prover-storage-manager/src/dummy_storage.rs
+++ b/full-node/sov-prover-storage-manager/src/dummy_storage.rs
@@ -1,0 +1,145 @@
+use std::marker::PhantomData;
+
+use byteorder::{BigEndian, ReadBytesExt};
+use sov_schema_db::schema::{KeyDecoder, KeyEncoder, Result as CodecResult, ValueCodec};
+use sov_schema_db::snapshot::{DbSnapshot, QueryManager};
+use sov_schema_db::{define_schema, CodecError};
+use sov_state::MerkleProofSpec;
+
+/// Oversimplified representation of [`sov_state::ProverStorage`]
+pub struct NewProverStorage<Mps: MerkleProofSpec, Q> {
+    state_db: DbSnapshot<Q>,
+    native_db: DbSnapshot<Q>,
+    p: PhantomData<Mps>,
+}
+
+impl<Mps: MerkleProofSpec, Q: QueryManager> NewProverStorage<Mps, Q> {
+    pub(crate) fn with_db_handlers(
+        state_db_snapshot: DbSnapshot<Q>,
+        native_db_snapshot: DbSnapshot<Q>,
+    ) -> Self {
+        NewProverStorage {
+            state_db: state_db_snapshot,
+            native_db: native_db_snapshot,
+            p: Default::default(),
+        }
+    }
+
+    pub(crate) fn freeze(self) -> (DbSnapshot<Q>, DbSnapshot<Q>) {
+        let NewProverStorage {
+            state_db,
+            native_db,
+            ..
+        } = self;
+        (state_db, native_db)
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn read_state(&self, key: u64) -> anyhow::Result<Option<u64>> {
+        let key = DummyField(key);
+        Ok(self
+            .state_db
+            .read::<DummyStateSchema>(&key)?
+            .map(Into::into))
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn write_state(&self, key: u64, value: u64) -> anyhow::Result<()> {
+        let key = DummyField(key);
+        let value = DummyField(value);
+        self.state_db.put::<DummyStateSchema>(&key, &value)
+    }
+
+    #[allow(dead_code)]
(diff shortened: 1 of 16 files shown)


rollup-interface/src/state_machine/da.rs
```diff
@@ -179,6 +179,8 @@ pub trait BlobReaderTrait: Serialize + DeserializeOwned + Send + Sync + 'static
 
 /// Trait with collection of trait bounds for a block hash.
 pub trait BlockHashTrait:
+    // TODO: Can we replace `Into<[u8; 32]>` with std::hash::Hash
```
citizen-stig: Or, I can use `[u8; 32]` as Key in hashmap, so BlockHash does not need to implement Hash

full-node/sov-prover-storage-manager/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+            }
+            // Storage requested first time
+            None => {
+                let new_snapshot_id = self.latest_snapshot_id + 1;
+                if let Some(parent_snapshot_id) =
+                    self.block_hash_to_snapshot_id.get(&prev_block_hash)
+                {
+                    let mut snapshot_id_to_parent = self.snapshot_id_to_parent.write().unwrap();
+                    snapshot_id_to_parent.insert(new_snapshot_id, *parent_snapshot_id);
+                }
```
preston-evans98, line 106: Should we enforcing consistency between this check and the check on line 107? In other words, is it possible for the `prev_block` to have a snapshot but not exist in the `chain_forks` map?

full-node/sov-prover-storage-manager/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+    // Reverse: child_block -> parent
+    blocks_to_parent: HashMap<Da::SlotHash, Da::SlotHash>,
+
+    latest_snapshot_id: SnapshotId,
+    block_hash_to_snapshot_id: HashMap<Da::SlotHash, SnapshotId>,
+    // Same reference for individual managers
+    snapshot_id_to_parent: Arc<RwLock<HashMap<SnapshotId, SnapshotId>>>,
+
+    state_snapshot_manager: Arc<RwLock<SnapshotManager>>,
+    native_snapshot_manager: Arc<RwLock<SnapshotManager>>,
```
bkolad: what is the difference between `state_snapshot_manager ` vs `native_snapshot_manager `?

full-node/sov-prover-storage-manager/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+
+        while let Some(block_hash) = to_discard.pop() {
+            let child_block_hashes = self.chain_forks.remove(&block_hash).unwrap_or_default();
+            self.blocks_to_parent.remove(&block_hash).unwrap();
+            let snapshot_id = self.block_hash_to_snapshot_id.remove(&block_hash).unwrap();
+
+            state_manager.discard_snapshot(&snapshot_id);
+            native_manager.discard_snapshot(&snapshot_id);
+
+            to_discard.extend(child_block_hashes);
```
bkolad, line 231: Does this loop ends? It looks like we remove one item from `to_discard` and add one on every step of the loop.

full-node/sov-prover-storage-manager/src/lib.rs
citizen-stig: one is for `state_db` and another one is for `native_db` (accessory db)

I suppose ledger_db will be added here as well. 

Technically underlying maps are the same in shape, but values are different. This makes each type smaller and easier to reason.

full-node/sov-prover-storage-manager/src/lib.rs
citizen-stig: Yes, it ends, when all "orphans" are removed. Basically it ends, when all heads from non-canonical forks are removed.

full-node/sov-prover-storage-manager/src/lib.rs
citizen-stig: Yes, there should be enforcing. It is technically possible to create snapshot id incorrect snapshot id, I plan to have test for it.

StorageManager should protect its own consistency

full-node/sov-prover-storage-manager/src/lib.rs
citizen-stig: Renamed to accessory

full-node/sov-prover-storage-manager/src/lib.rs
citizen-stig: I've tests all of the cases I think of, it looks this should be covered

full-node/sov-prover-storage-manager/src/dummy_storage.rs
```diff
@@ -0,0 +1,157 @@
+use std::marker::PhantomData;
+
+use byteorder::{BigEndian, ReadBytesExt};
+use sov_schema_db::schema::{KeyDecoder, KeyEncoder, Result as CodecResult, ValueCodec};
+use sov_schema_db::snapshot::{DbSnapshot, QueryManager};
+use sov_schema_db::{define_schema, CodecError};
+use sov_state::MerkleProofSpec;
+
+/// Oversimplified representation of [`sov_state::ProverStorage`]
```
bkolad: WHy do we put it as  public doc? Is it going to be changed?

full-node/sov-prover-storage-manager/src/dummy_storage.rs
```diff
@@ -0,0 +1,157 @@
...
+            state_db,
+            native_db,
+            ..
+        } = self;
+        (state_db, native_db)
+    }
+
+    #[cfg(test)]
+    pub(crate) fn read_state(&self, key: u64) -> anyhow::Result<Option<u64>> {
+        let key = DummyField(key);
```
bkolad, line 41: Would be nice not to mix `Dummy..` and prod code in one file. Would it be possible to make `state_db` `pub(crate)` and implemnthe these methods in test file?

bkolad (approved):

full-node/sov-prover-storage-manager/src/dummy_storage.rs
citizen-stig: Accidentally did it, replaced it with internal.

Also, added comment on the top of the file, that it will be removed.

full-node/sov-prover-storage-manager/src/dummy_storage.rs
citizen-stig: I will keep it this way for now, as it is going to be removed in the next PR.

PR #1134, opened 2023-11-02
Fixing broken prover benchmark

### LINKED ISSUE #981: Benchmark in `demo-prover` fails.
## Summary
Steps to reproduce:
1. `cd /examples/demo-prover/benches`
2. `cargo bench --features bench`

Output:
```
Skipping prover at block 2 to capture cycle counts

thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `252`: Batch deserialization failed and some data was not provided. The prover might be malicious', /Users/blaze/programming/sovereign-labs/sovereign/module-system/sov-modules-stf-template/src/app_template.rs:271:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'Prover should run successfully: Guest panicked: assertion failed: `(left == right)`
  left: `0`,
 right: `252`: Batch deserialization failed and some data was not provided. The prover might be malicious', examples/demo-prover/benches/prover_bench.rs:234:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### LINKED ISSUE #1131: Check feasibility to remove .cargo/config.toml in provers/risc0
There was a bug  on 08/01/2023 which caused risc0 to break. This was a bug related to an incompatibility with risc0, it's usage of the -Zstd flag in cargo and the latest version of `cc` pushed. 
This caused risc0 to break and the prover to stop working. Because of how cargo builds std when the -Zstd flags are passed, the only way to ensure that we could pin `cc` to an older version was to use .cargo/config.toml. Cargo creates a new virtual working space and ignores any dependency overrides in Cargo.toml itself, but .cargo/config.toml still works. risc0 recommended using that approach as well 
risc0issue

this issue seems to have been fixed currently, but we still have the .cargo/config.toml checked in and it's generating warnings. We want to test and remove the override if its not needed so that we can keep warnings clean
```
sov-demo-prover-guest-mock: Starting build for riscv32im-risc0-zkvm-elf
sov-demo-prover-guest-mock: warning: Patch `cc v1.0.79 (https://github.com/rust-lang/cc-rs?rev=e5bbdfa#e5bbdfa1)` was not used in the crate graph.
```

dubbelosix:
# Changes
prover-bench was broken due to multiple reasons.
- The accumulator being passed to the prover was empty - root caused this the advance function being called when running in native before the prover is executed. This was fixed by moving the hint to add the blob transactions for the prover after apply_blob (earlier it was before apply_blob, right after extract_relevant_blobs_with_proof, but without apply_blob, the accumulator is empty since advance is not called)
- The prover benchmark relied on blocks.hex which was a static file that was checked in with a few pre-generated blocks. If any of the types involved in the deserialization process or the DA adapters changed, this was causing it to break. Modified prover-bench to use mock-da and added a datagen.rs which generates the necessary transactions on the fly
- Feature flags along the dependency chain were broken for prover-bench (which requires the bench flag to be passed through build.rs and to multiple modules to enable the "cycle-tracker". In some places "native" dependency was added
- consistently use risc0 zkvm Bytes and remove additional dependencies added to guest
- .cargo/config.toml was removed since the `cc` bug is fixed (linked in issues)
- The original code to track was inside celestia rollup.rs, this was removed during refactoring it looks like - but added it now to mock-da's rollup.rs

## Linked Issues
- Fixes #1131 
- Fixes #981


bkolad (approved):

bkolad:
LGTM, it would be good to think about how we can make the benchmark generic over DA/ZkVM. Maybe we could reuse the `RollupTemplate`

PR #1146, opened 2023-11-13
Introduce the `ProverService` in `RollupBlueprint`.

bkolad:
# Description
In this PR, we abstract the `ZKP` proving logic in the `ProverService` trait. The current implementation, `BlockingProver`, is just a wrapper over the existing logic. In the next PR, we will replace `BlockingProver` with `MultiThreadedProver`.

## Testing
Ci passes.

## Docs
Relevant docs added.


full-node/sov-stf-runner/src/prover_service/blocking_prover.rs
```diff
@@ -0,0 +1,126 @@
...
+use sov_rollup_interface::services::da::DaService;
+use sov_rollup_interface::stf::StateTransitionFunction;
+use sov_rollup_interface::zk::ZkvmHost;
+
+use super::{Hash, ProverService, ProverServiceError};
+use crate::verifier::StateTransitionVerifier;
+use crate::{ProofGenConfig, Prover, RollupProverConfig, StateTransitionData};
+
+/// Prover that blocks the current thread and creates a ZKP proof.
+pub struct BlockingProver<StateRoot, Witness, Da, Vm, V>
```
bkolad, line 17: This part is a wrapper over existing proving logic and will be replaced by a multithreaded prover.

full-node/sov-stf-runner/src/prover_service/blocking_prover.rs
```diff
@@ -0,0 +1,126 @@
...
+        >,
+    ) {
+        let header_hash = state_transition_data.da_block_header.hash().into();
+        self.witness
+            .lock()
+            .expect("Lock was poisoned")
+            .insert(header_hash, state_transition_data);
+    }
+
+    async fn prove(&self, block_header_hash: Hash) -> Result<(), ProverServiceError> {
```
citizen-stig, line 100: I see that `ZkStorage` is an attribute of prover.

It makes an assumption, that ZkStorage is stateless, so it is free to clone it. But this is implicit assumption.

What are the cons of passing zk_storage as parameter to `prove` method? And RollupBluePrint creates it as before?

full-node/sov-stf-runner/src/runner.rs
```diff
@@ -11,19 +11,18 @@ use tokio::sync::oneshot;
...
 /// Combines `DaService` with `StateTransitionFunction` and "runs" the rollup.
-pub struct StateTransitionRunner<Stf, Sm, Da, Vm, V>
+pub struct StateTransitionRunner<Stf, Sm, Da, Vm, PS>
 where
     Da: DaService,
     Vm: ZkvmHost,
     Sm: StorageManager,
     Stf: StateTransitionFunction<Vm, Da::Spec, Condition = <Da::Spec as DaSpec>::ValidityCondition>,
-    V: StateTransitionFunction<Vm::Guest, Da::Spec>,
+    PS: ProverService,
```
citizen-stig: Nit: Our naming convention for abbreviations is title case, could you please make `Ps`

full-node/sov-stf-runner/src/prover_service/blocking_prover.rs
citizen-stig: Ok, it was already like that in rollup runner , so no need to resolve it this PR.

But I still would like to have your comment. If you think that this implicit assumption might bring some problems along the way, we need to address it in separate issue.

citizen-stig (approved):

full-node/sov-stf-runner/src/prover_service/blocking_prover.rs
bkolad: There are several reasons for this design:

1. In the above code, the `ProverService` owns the storage and (currently). By passing it explicitly, we leak it, and the `ZkStorage` would be accessible in more places unnecessarily.
2. Once we implement `ProverService` for bonsai, we won't need it at all, so it should not be exposed to the rest of the system.
3. In the next "multithreaded" version of the `ProverService`, every thread will need to have its own clone of the `ZkStorage`. If there ever will be a version of stateful `ZkStorage`, then the implementer needs to make sure that the `Clone` is implemented correctly and can be used this way.

Also: The assumption about being stateless is kind of explicit, as the `set/get` methods accept a `witness` and don't have `&mut self`, the implementer opts-out via interior mutability mechanism.

full-node/sov-stf-runner/src/runner.rs
bkolad: fixed

full-node/sov-stf-runner/src/prover_service/blocking_prover.rs
citizen-stig: 1. I agree that leaking one more associated type is not nice.
2. I don't fully get it, I thought that Bonsai will be just running guest VM, which still needs ZkStorage
3. This part, where I don't feel comfortable. Have correct clone behavior is not the same as being stateless. For instance, having `Arc<RwLock<Vec>>` will have correct clone, but underlying data still be shared. Of course type system gives a hint here. Same goes with "not having `&mut self`". ProverStorage is mutable, and follow the same pattern.

I agree that it is not a problem at the moment, as ZkStorage indeed stateless, but as this assumption leaks into other parts of the system, I would prefer not to pull it under the rug.

If you are sure, that this approach won't back fire in the future, I trust you.

PR #1154, opened 2023-11-15
Fix intermittent failures related to config_constant macro

dubbelosix:
# Description

The primary issue here seems to be that the `sov-modules-macros/target-path` file which holds the directory containing `constants.json` (which is required by the macro to function) is created by the `build.rs` script inside sov-modules-macros. This file `sov-modules-macros/target-path` is behaving like a global variable and there are two `constants.json` files (one inside sov-modules-macros for test cases, and the second is the one in the root folder that contains the gas constants etc) that keep being written to it.

In short trybuild compiles `sov-modules-macros` multiple times and stores the test constants.json at  `sov-modules-macros/target-path`  while the non trybuild tests require the default one. The intermittent nature of the failures depends on which tests get run first (and is complicated by the fact that nextest runs in parallel and cargo test runs for docs)

This doesn't happen in CI because 
1. nextest doesn't run doc tests (which are the ones that break locally)
2. nextest and cargo test --doc are run in different jobs
3. in local, we frequently run individual tests which ends up setting the global variable (target-path) to one of the two constants files which breaks for 1 test case, but works for the other. This is why cargo clean fixes the issue

## Fix chosen
The fix for now is to have a second environment variable `TARGET_PATH_OVERRIDE` which is set to `target-path-trybuild` in sov-modules-macros/all_tests.rs and the default is `target-path`. This ensures that both the processes don't step on each other and break each other.

## Changes
* New environment variable `TARGET_PATH_OVERRIDE` added to build.rs to change the file that stores the constants.json folder
* `TARGET_PATH_OVERRIDE` support added to manifest.rs as well since the macro needs to read the folder containing constants.json from a different folder if present, if not falls back to `target-path`
* `target-path*` added to gitignore so both the global var files are ignored

## Alternative approaches tried
One of the approaches tried was to use some kind of a signal to force run build.rs (timestamp file), but this seemed complicated and not always working (since sov-modules-macros is re-exported from sov-modules-api, and it isn't being rebuilt) additionally sov-modules-macros isn't always being rebuilt. Tried using `cargo:rerun-if-env-changed=CONSTANTS_MANIFEST` in build.rs as well, but it was being ignored. I think the core problem there is that trybuild is modifying the file AFTER it is first built, but when the module is being re-imported later (after the trybuild tests complete) it is not being re-created. Was monitoring the `target-path` file with fswatch, and each trybuild run repopulates the file, but once the trybuild is completed and sov-modules-macros is needed again by sov-blob-storage, the package is not being rebuilt and build.rs is not being run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@ demo-rollup-readme.sh
 /target
 
 .idea/
-target-path
+target-path*
 
 target/
 fuzz/Cargo.lock
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,7 +44,7 @@ dependencies = [
  "memmap2",
  "object 0.31.1",
  "rustc-demangle",
- "smallvec 1.11.1",
+ "smallvec 1.11.2",
 ]
 
 [[package]]
@@ -115,7 +115,7 @@ version = "0.7.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a824f2aa7e75a0c98c5a504fceb80649e9c35265d44525b5f94de4771a395cd"
 dependencies = [
- "getrandom 0.2.10",
+ "getrandom 0.2.11",
  "once_cell",
  "version_check",
 ]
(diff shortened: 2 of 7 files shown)


bkolad (approved):
Thx a lot for this fix!

vlopes11 (approved):
Looks good, thanks!

We should probably add doc tests to the CI as well, to avoid this situation in the future.

As side note, it is in our backlog to remove the target-path entirely, in favor of a home dir path for the constants file.

citizen-stig (approved):

PR #1158, opened 2023-11-15
refactor: Deduplicate implementations of State and AccessoryState

preston-evans98:
# Description
This PR consolidates the implementations of `State{Value|Map|Vec}` and `AccessoryState{Value|Map|Vec}` into three traits (one each for `Value`, `Map`, and `Vec`). This reduces code duplication, and makes it much easier to add additional state types in future. The primary drawback to this approach is that now users have to import `sov_modules_api::StateValueAccessor` in order to call any of the standard accessors on their `State`/`Accessory` values. To help reduce the friction, I've added a `sov_modules_api::prelude` which exports all three accessor traits. In future, this prelude should be expanded with other useful types.

## Testing
These changes are covered by existing tests


neysofu (approved):
LGTM, I have no concerns. Some of the type signatures are a bit more complex now – which, I now remember, was the reason why we tried the other approach first – but it's all neatly hidden away from the user, and I think the tradeoff is definitely worth it considering the future additions and new state wrappers!

bkolad (approved):

PR #1160, opened 2023-11-16
multi-threaded prover service

bkolad:
# Description
This PR replaces the `BlockingProver` with `ParallelProver` running on a separate `rayon` thread pool.
In subsequent pull requests, we plan to incorporate the following enhancements:
1.  Implementation of handling for prover `busy` status.
2.  Addition of unit tests for the `ProverService.` This will be a separate PR, as it involves mocking STF and ZkVm.

Related issues: #1165

## Testing
CI passes.

## Docs
Relevant docs added.


full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
```diff
@@ -1,21 +1,20 @@
...
 
 use super::{Hash, ProverService, ProverServiceError};
 use crate::verifier::StateTransitionVerifier;
-use crate::{ProofGenConfig, RollupProverConfig, StateTransitionData};
+use crate::{ProofGenConfig, ProofSubmissionStatus, RollupProverConfig, StateTransitionData};
 
-/// Prover that blocks the current thread and creates a ZKP proof.
-pub struct BlockingProver<StateRoot, Witness, Da, Vm, V>
+/// Prover service that generates proofs in parallel.
+pub struct ParallelProverService<StateRoot, Witness, Da, Vm, V>
```
vlopes11, line 17: Should we leave the previous blocking implementation so the user can pick it in case he wants to manage the threads himself?

full-node/sov-stf-runner/src/prover_service/mod.rs
```diff
@@ -19,6 +19,16 @@ pub enum RollupProverConfig {
     Prove,
 }
 
+/// Indicates the status of the DA proof submission.
+pub enum ProofSubmissionStatus {
+    /// Proof was submitted to the DA.
+    Success,
+    /// Proof generation is still in progress.
+    ProvingInProgress,
```
vlopes11, line 29: Should this variants be handled by the stdlib prelude async? Meaning: We always wait for a future `anyhow::Result<()>`

full-node/sov-stf-runner/src/prover_service/mod.rs
```diff
@@ -55,4 +65,7 @@ pub trait ProverService {
 
     /// Creates ZKP prove for a block corresponding to `block_header_hash`.
     async fn prove(&self, block_header_hash: Hash) -> Result<(), ProverServiceError>;
+
+    /// Sends the ZK proof to the DA.
```
vlopes11: nit: consider adding some comments on why the block header has suffices to accept the proof (i.e. the proofs already exists on component `ProverState`, that lives on `X`)

full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
```diff
@@ -24,17 +23,15 @@ where
     V: StateTransitionFunction<Vm::Guest, Da::Spec> + Send + Sync,
 {
     vm: Vm,
-    prover_config: Arc<Option<ProofGenConfig<V, Da, Vm>>>,
+    prover_config: Option<Arc<ProofGenConfig<V, Da, Vm>>>,
```
vlopes11: Instead of making the config optional, consider making its values default to something. With that, it will be easier to access its properties - instead of having to check for the `Option::Some` variant

full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
```diff
@@ -94,44 +92,26 @@ where
...
-            })?;
+            self.prover_state
+                .start_proving(block_header_hash, config, vm, zk_storage)?;
         }
-
         Ok(())
     }
+
+    async fn send_proof_to_da(&self, block_header_hash: Hash) -> ProofSubmissionStatus {
+        if self.prover_config.is_some() {
```
vlopes11: nit: this looks like a `match` pattern or a

```rust
self.prover_config
  .map(|_| self.prover_state
    .get_proof_submission_status_and_remove_on_success(
      block_header_hash
    ))
  .unwrap_or(ProofSubmissionStatus::Success)
```

full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
```diff
@@ -94,44 +92,26 @@ where
...
+            self.prover_state
+                .get_proof_submission_status_and_remove_on_success(block_header_hash)
+        } else {
+            ProofSubmissionStatus::Success
```
vlopes11: It's not clear to me why we return success when the config is absent. Let's add an explanatory comment to this code?

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
```diff
@@ -0,0 +1,178 @@
...
+
+use super::{Hash, ProverServiceError};
+use crate::{ProofGenConfig, ProofSubmissionStatus, StateTransitionData};
+
+enum ProverStatus<StateRoot, Witness, Da: DaSpec> {
+    WitnessSubmitted(StateTransitionData<StateRoot, Witness, Da>),
+    Proving,
+    Proved(Proof),
+    Err(anyhow::Error),
+}
```
vlopes11, line 20: Consider making this a `anyhow::Result<ProverStatus>` so it's more easily interoperable with the stdlib

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
```diff
@@ -0,0 +1,178 @@
...
+
+#[derive(Default)]
+struct ProverState<StateRoot, Witness, Da: DaSpec> {
+    prover_status: HashMap<Hash, ProverStatus<StateRoot, Witness, Da>>,
```
vlopes11, line 24: Won't we lose the witness state if the job ever moves to `proving` and the backend prover postpones it due to busy resources? Meaning: consider making the `Proving(StateTransitionData<StateRoot, Witness, Da>)` variant retain the stf while the proof is not produced.

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+impl<StateRoot, Witness, Da: DaSpec> ProverState<StateRoot, Witness, Da> {
+    fn remove(&mut self, hash: &Hash) -> Option<ProverStatus<StateRoot, Witness, Da>> {
```
vlopes11, line 28: Consider making this `async` and using a thread synchronization implementation such as `tokio::RwLock` so we don't need to take `self` as `&mut self`

This might help if the prover status queue gets large

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
```diff
@@ -0,0 +1,178 @@
...
+use sov_rollup_interface::stf::StateTransitionFunction;
+use sov_rollup_interface::zk::{Proof, ZkvmHost};
+
+use super::{Hash, ProverServiceError};
+use crate::{ProofGenConfig, ProofSubmissionStatus, StateTransitionData};
+
+enum ProverStatus<StateRoot, Witness, Da: DaSpec> {
+    WitnessSubmitted(StateTransitionData<StateRoot, Witness, Da>),
+    Proving,
+    Proved(Proof),
```
vlopes11, line 18: Consider moving the `proved` variant to a different component so it can be served once, and won't lock the wait queue to inquire if other proofs are available. Meaning:

- Alice submits stf `A`
- Prover yields proof `A`
- Bob submits stf `B` shortly after Alice asks for proof `A`

Bob will have to wait for alice in order to submit his proof. Not only that, but if Alice never asks for her proof, the map state will grow.

It seems to me that `WitnessSubmitted` and `Proving` states are of different nature than `Proved`. The former will never expire (i.e. we might have to ping a prover backend for the proof or error), while `Proved` might have to implement a time to live for its elements. If we have such split, it will be easier to do so.

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
```diff
@@ -0,0 +1,178 @@
...
+            Err(e) => self.prover_status.insert(hash, ProverStatus::Err(e)),
+        }
+    }
+
+    fn get_prover_status(&self, hash: Hash) -> Option<&ProverStatus<StateRoot, Witness, Da>> {
+        self.prover_status.get(&hash)
+    }
+}
+
+pub(crate) struct Prover<StateRoot, Witness, Da: DaService> {
```
vlopes11, line 52: Consider simplifying `Prover(ProverState)` into `Prover` as the `ProverState` functionality is a wrapper for a map structure and shouldn't be expected to have different behavior and/or dedicated test logic

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub(crate) fn submit_witness(
+        &self,
+        state_transition_data: StateTransitionData<StateRoot, Witness, Da::Spec>,
+    ) {
+        let header_hash = state_transition_data.da_block_header.hash().into();
+        let data = ProverStatus::WitnessSubmitted(state_transition_data);
+
+        self.prover_state
+            .lock()
+            .expect("Lock was poisoned")
```
vlopes11: Consider using async `RwLock` as it is extremely resistant to poisoning. `RwLock` might be better over `Mutex` as we should expect more state read than write (i.e. users will query about their proof status more often than submitting a new stf)

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
```diff
@@ -0,0 +1,178 @@
...
+            .insert(header_hash, data);
+    }
+
+    pub(crate) fn start_proving<Vm, V>(
+        &self,
+        block_header_hash: Hash,
+        config: Arc<ProofGenConfig<V, Da, Vm>>,
+        mut vm: Vm,
+        zk_storage: V::PreState,
+    ) -> Result<(), ProverServiceError>
```
vlopes11, line 90: IIUC, we perform a lazy load of proving, meaning we have a "reactor" or "work management unit" that lives somewhere. One alternative is to have such work management performed here, and always prove everything that is submitted - and serve when requested. Otherwise, it might be analogous to the work management unit proving the STF himself.

In other words: the same actor who call `ParallelProverService::prove` might as well just prove the transaction directly, instead

One alternative is to have a `config` that will set the mode of the proving service to `lazy` (i.e. we just start a proving thread when claled) or `eager (i.e. we have a fixed number of worker threads that will keep taking proofs and will submit them to a channel/queue once done)

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
```diff
@@ -0,0 +1,178 @@
...
+        let prover_status = prover_state
+            .remove(&block_header_hash)
+            .ok_or_else(|| anyhow::anyhow!("Missing status for {:?}", block_header_hash))?;
+
+        match prover_status {
+            ProverStatus::WitnessSubmitted(state_tranistion_data) => {
+                prover_state.set_to_proving(block_header_hash);
+                vm.add_hint(state_tranistion_data);
+
+                rayon::spawn(move || {
```
vlopes11, line 109: Consider having a limited number of working threads so the behavior of the proving service is predictable by the system administrator (i.e. he knows exactly how many resources are needed for such service)

It might be hard to set a cloud service if we spawn on request, as the number of concurrent threads might become large

full-node/sov-stf-runner/src/runner.rs
```diff
@@ -201,6 +201,19 @@ where
...
                     .expect("The proof creation should succeed");
+
+                loop {
+                    let status = self.prover_service.send_proof_to_da(header_hash).await;
+                    match status {
+                        crate::ProofSubmissionStatus::Success => {
+                            break;
+                        }
+                        crate::ProofSubmissionStatus::ProvingInProgress => {
+                            tokio::time::sleep(tokio::time::Duration::from_millis(100)).await
```
vlopes11, line 234: instead of managing the interval, consider delegating this task to [tokio::select](https://docs.rs/tokio/1.34.0/tokio/macro.select.html)

Tokio has a great mechanism to set a fair interval & task pick based on very complicated criteria - we can delegate them this responsibility

vlopes11 (commented):
Looks good overall! I added a couple of questions & suggestions as a first round of review

full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
bkolad: We can't block during the proof creation as it will be too slow in production.  Parallel prover blocks when we send to DA so the workflow is different.

full-node/sov-stf-runner/src/prover_service/mod.rs
bkolad: Yes, this is a good point. Currently, the caller waits in a loop with some delay until the proof is ready. I think we will need something like a `future` or channel subscription, but I would like to defer the implementation after we have the fork management mechanism from @citizen-stig ready. This is because we will send proofs to DA only when the corresponding block is finalized, so we have to design with the `ForkMenager` in mind.

full-node/sov-stf-runner/src/prover_service/mod.rs
bkolad: Can you please extend the comment?

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
```diff
@@ -0,0 +1,178 @@
...
+        let prover_state_clone = self.prover_state.clone();
+
+        let mut prover_state = self.prover_state.lock().expect("Lock was poisoned");
+
+        let prover_status = prover_state
+            .remove(&block_header_hash)
+            .ok_or_else(|| anyhow::anyhow!("Missing status for {:?}", block_header_hash))?;
+
+        match prover_status {
+            ProverStatus::WitnessSubmitted(state_tranistion_data) => {
```
cemozerr:
```suggestion
            ProverStatus::WitnessSubmitted(state_transition_data) => {
```

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
bkolad: Yes, this is point `1` in the PR description. I left it as a follow-up because the code became a bit complicated, and currently, there is no good way to test it, so I don't want to add even more complexity before #1165  is fixed. 
The following PR will be #1165, and after that, I will add handling for the busy status.

full-node/sov-stf-runner/src/runner.rs
```diff
@@ -201,6 +201,19 @@ where
                     .prove(header_hash)
                     .await
                     .expect("The proof creation should succeed");
+
```
cemozerr, line 226: Since the `run_in_process` method is a for loop, does this new proof submission & check loop mean that we block on proof generation before we move on to executing the next block? 

Since we want to provide soft confirmations and generally stay on par with the DA layer natively, we likely don't want this to happen. Are we doing this to decrease PR size for now?

full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
bkolad: Done

full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
bkolad: FIxed with https://github.com/Sovereign-Labs/sovereign-sdk/pull/1160#discussion_r1399070829

full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
bkolad: Fixed with: https://github.com/Sovereign-Labs/sovereign-sdk/pull/1160#discussion_r1399070829

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
bkolad: This will be added together with the `busy` status implementation.

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
bkolad: The lock is taken over `ProverState` in the `Prover`. It is done this way because in the following PR, I will add a flag related to the `busy` state management and this will need to be protected by the same lock.

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
bkolad: It is a separate type because we will add more data to the `ProverState` in the next PR. We will manage the `busy` flag there as well.

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
bkolad: I was going this way initially, but instead, I opted for a single map. Here is the reasoning:
1. Quering the map behind lock or rw lock is fast and done relatively rearly. We do it when we submit the witness and ask for proof status. 
2. The proving process is a slow operation but we don't hold the lock there so nothing will be blocked.

Given 1 & 2 just locking the map will never be a buttlneck and adding additional structures would add more complexity to the code.

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
bkolad: Done

full-node/sov-stf-runner/src/runner.rs
bkolad: Certainly, we're currently implementing this as a temporary measure. The plan is to share the proof only after the relevant block has been finalized. However,  we haven't implemented the logic for `block-finalization` yet. Once Nikolai completes the `fork-management` feature, we'll integrate this logic in the appropriate location.

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
bkolad: Do you mean that the caller should spawn a separate thread or wait for the prove method to complete?

full-node/sov-stf-runner/src/runner.rs
bkolad: I would postpone the task for two reasons.:

1.  https://github.com/Sovereign-Labs/sovereign-sdk/pull/1160#discussion_r1399107880 (the `fork-menager` is ready) so we have a better sense of how exactly this method is used
2. #1165 We have a means to test the code better

full-node/sov-stf-runner/src/prover_service/mod.rs
vlopes11: Sounds good! I'd suggest something like a [oneshot](https://docs.rs/tokio/1.34.0/tokio/sync/oneshot/index.html) channel (or some other library that implements a similar functionality) as it does perform well and is quite robust, but you're right that we should prioritize consistency with other parts of the code

full-node/sov-stf-runner/src/prover_service/mod.rs
vlopes11: If we check the code, we know the `ProverState` keeps an internal map `block_header_hash: Hash |-> WitnessStf`, and that's why only the block header hash suffices to sbumit a proof to the DA - as compared to a scenario where you need to send the state transition witness.

It means also that the user must submit, prior to this call, the state transition witness elsewhere - and would be nice to have documented also where he is expected to submit it

full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
vlopes11: I meant, we don't replace the `BlockerProverService`; instead, we add a new `ParallelProverService`.

Indeed we shouldn't expect the user to run single-threaded, but he might want to implement the parallelism & work distribution himself. In that case, he can spawn several `BlockerProverService` on his worker threads

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
vlopes11: Sounds good! If we need to implement a time to live for `Proved(Proof)` in the future, we can revisit this point and make the split then

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
vlopes11: No, I think we should indeed delegate the thread management to the prover service (as we are doing now).

However, spawning one thread per user request is analogous to having the user spawning his own threads. We can name this as `lazy` approach. The main problem of this approach is that the number of concurrent threads will be defined by the number of simultaneous requests, and this might deplete the resources of a server in case it faces a spike of requests.

What I'm suggesting is the future possibility of implementation of a `eager` load approach. Consider the following sketch:

image

Here the user submits a STF to a prover service, that will generate him a oneshot channel. Such prover will then forward the work to be done (i.e. prove the STF) to a "work management unit" (in other words, a load balancer), that will pick an available worker thread and submit the `(stf, oneshot)` pair.

Provided the user can configure a "number of maximum workers", we end with a couple of advantages:

- The resource consumption of the architecture is predictable and is just the required resources for a single STF function, multiplied by the number of workers
- We don't build a queue of proved STFs waiting to be fetched. On the `lazy` approach, if the user never fetches such proofs, they will just stay in memory, growing indefinitely. Instead, if the user never requests for the proof, he will have at some point to drop the ownership of his oneshot channel - and that would free the resources allocated for the unused proof
- We preload the worker threads, removing the overhead of initiating a new thread per request
- We will end up having lower latency times as no CPU cycle is wasted on waiting for the user to request his proof. They are always computed when submitted, as the `eager` approach

We might not need to make this entire change now, but we might profit from some infrastructure unrelated to the STF itself such as a configuration argument with the maximum number of threads (to avoid denial of service). As next step, we can introduce this `eager` work management

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
vlopes11: Sounds good, we can also take this topic on the other item

https://github.com/Sovereign-Labs/sovereign-sdk/pull/1160/files#r1399167455

full-node/sov-stf-runner/src/runner.rs
vlopes11: It would however cover a potential edge case where the service is locked with such STF, and an unreasonable timeout is reached. In short, the suggestion would look like

```rust
/// Select will trigger whatever event happens first, returning it
let response = tokio::select! {
  v = self.prover_service.send_proof_to_da(header_hash) => v,
  /// This would halt the entire request if the timeout is reached
  v = tokio::time::sleep(tokio::time::Duration::from_millis(100)) => panic!("timeout"),

};
if let crate::ProofSubmissionStatus::Err(e) = response {
  // report error
}
```

full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
bkolad: discusse offline

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
bkolad: discussed offline

full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
vlopes11: As discussed, the prover service locks the witness computation, and this is overall undesirable. Even tho it is simpler to write unit tests with the blocking interface, we want all the tests to reflect a production-like interface, making the blocker prover service entirely obsolete

full-node/sov-stf-runner/src/prover_service/parallel/prover.rs
vlopes11: As discuss, we might embed the parallelism arguments `ProofGenConfig` for now, and maybe change that once the fork management implementation is completed

vlopes11 (approved):
Looks good! Thanks

PR #1163, opened 2023-11-17
Create a versioned `KernelStateValue` which is only partially accessible from user space

preston-evans98:
# Description
This PR makes two related changes:
1. Add a `slot_height` method to `Context`. This simple change accounts for almost all of the lines changed
2. Introduce a new `KernelStateValue` type which appears as a `map` from `slot_number` to `value` when accessed from the `kernel`, but appears as a regular state value (which may be absent) when accessed from user space. To support this value, two new variants of `WorkingSet` are introduced - a `KernelWorkingSet` which can only be constructed by a `Kernel`, and a `VersionedWorkingSet`, which is constructed from the context and allows access to only the current version of a `Kernel` Value. 

## Testing
Tests have been added to cover the new `KernelStateValue`'s  behavior


module-system/sov-modules-api/src/containers/kernel_value.rs
```diff
@@ -0,0 +1,396 @@
...
+};
+use sov_state::codec::BorshCodec;
+
+/// A `versioned` value stored in kernel state. The semantics of this type are different
+/// depending on the priveleges of the accessor. For a standard ("user space") interaction
+/// via a `VersionedWorkingSet`, only one version of this value is accessible. Inside the kernel,
+/// (where access is mediated by a `KernelWorkingSet`), all versions of this value are accessible.
+///
+/// Under the hood, a versioned value is implemented as a map from a slot number to a value. From the kernel, any
+/// value can be accessed using the `StateMapAcessor` trait with the slot number as the key. For convenience,
```
cemozerr:
```suggestion
/// value can be accessed using the `StateMapAccessor` trait with the slot number as the key. For convenience,
```

module-system/sov-modules-core/src/runtime/capabilities.rs
```diff
@@ -50,26 +79,67 @@ impl<'a, B: BlobReaderTrait> From<&'a mut B> for BlobRefOrOwned<'a, B> {
...
-        I: IntoIterator<Item = &'a mut Da::BlobTransaction>;
+    use super::{BlobRefOrOwned, BlobSelector, Kernel};
+    use crate::Context;
+
+    /// A mock kernel for use in tests
+    #[derive(Debug, Clone)]
+    pub struct MockKernel<C, Da> {
+        /// The current slot height
+        pub true_height: u64,
+        /// The height at which transactions appear to be eecuting
```
cemozerr:
```suggestion
        /// The height at which transactions appear to be executing
```

module-system/sov-modules-core/src/runtime/capabilities.rs
```diff
@@ -1,15 +1,44 @@
...
 //! you can bypass the Sovereign module-system completely
 //! and write a state transition function from scratch.
 //! [See here for docs](https://github.com/Sovereign-Labs/sovereign-sdk/blob/nightly/examples/demo-stf/README.md)
 
-use sov_modules_core::{Context, WorkingSet};
 use sov_rollup_interface::da::{BlobReaderTrait, DaSpec};
 
+use crate::{Context, WorkingSet};
+
+/// The kernel is responsible for managing the inputs to the `apply_blob` method.
```
cemozerr, line 13: Are these comments outdated? It seems like the Kernel and BlobSelector functionality is separated and the power of "reorder blobs" is more appropriate for BlobSelector?

module-system/sov-modules-api/src/containers/kernel_value.rs
```diff
@@ -0,0 +1,396 @@
...
+            let mut kernel_state = KernelWorkingSet::from_kernel(&kernel, &mut working_set);
+            value.set(&100, &mut kernel_state);
+            assert_eq!(value.get(&mut kernel_state), Some(100));
+        }
+
+        {
+            {
+                let mut versioned_state =
+                    working_set.versioned_state(&DefaultContext::new(Address::from([1; 32]), 1));
+                // Try to read the value from user space with the slot number set to 1. Should fail.
```
cemozerr, line 352: Question:

Where does the check of `VersionedWorkingSet` not returning values until they're in the past happen? Couldn't find it.

module-system/sov-modules-core/src/storage/scratchpad.rs
```diff
@@ -449,6 +460,88 @@ impl<'a, C: Context> StateReaderAndWriter for AccessoryWorkingSet<'a, C> {
     }
 }
 
+/// Provides specialized working set wrappers for dealing with protected state.
+pub mod kernel_state {
+    use sov_rollup_interface::da::DaSpec;
```
bkolad, line 465: I added this issue #1166 and made this PR #1164 in order to bring more tests to `sov-modules-core`. Let;s add some tests for  `VersionedWorkingSet` there?

module-system/sov-modules-api/src/containers/kernel_value.rs
```diff
@@ -0,0 +1,396 @@
...
+        fn codec(&self) -> &Codec {
+            &self.codec
+        }
+
+        fn set(&self, value: &V, working_set: &mut VersionedWorkingSet<'a, C>) {
+            working_set.set_value(
+                self.prefix(),
+                &working_set.slot_num(),
+                value,
+                StateValueAccessor::<V, Codec, VersionedWorkingSet<'a, C>>::codec(self),
```
bkolad, line 83: Why can't we use` self.codec()` here?

bkolad (approved):

module-system/sov-modules-api/src/containers/kernel_value.rs
preston-evans98: Since this type implements two `Accessor` traits which each have a `codec` method, we have to disambiguate for rustc. Both impls are the same, but rustc doesn't know that :)

module-system/sov-modules-core/src/storage/scratchpad.rs
preston-evans98: Good idea. Will add.

module-system/sov-modules-api/src/containers/kernel_value.rs
preston-evans98: Chatted offline. The "check" is internal to the prefix generation logic

module-system/sov-modules-core/src/runtime/capabilities.rs
preston-evans98: These are accurate, although we'll want to extend them with the new functionality as we add it. But kernel is responsible to provide the `BlobSelector` in addition to managing the kernel state.

module-system/sov-modules-api/src/containers/kernel_value.rs
preston-evans98: Fixed, thanks!

module-system/sov-modules-core/src/runtime/capabilities.rs
preston-evans98: Fixed!

module-system/sov-modules-core/src/storage/scratchpad.rs
preston-evans98: Done

PR #1177, opened 2023-11-23
K, V Iterator for `DbSnapshot`

### LINKED ISSUE #1023: Data Availability re-orgs handling tracking issue

### LINKED ISSUE #1026: `Storage` should support nested snapshots

citizen-stig:
# Description

`DbSnapshot` struct suppose to swap `DB` in `state_db` and `native_db`, so they can work on top of snapshots.
This PR adds 2 methods to `DbSnapshot` struct:

* `get_largest` - returns the value of the largest key in the schema
* `get_prev` - returns the value of the key, that is equal or lower that the given

## Secondary changes

* Move `SchemaBatch` into its own file
* Replace HashMap in for `BTreeMap` inside `ScemaBatch`, so keys are sorted inside single snapshot
* Add new method to `QueryManager` trait, that allows to iterate over all key value pairs in reverse order.

## Linked Issues
 - Related to #1023 

## Testing
Unit and integration tests have been added.

## Docs
Docs on public methods have been added.

### DIFF
--- a/full-node/db/sov-schema-db/src/snapshot.rs
+++ b/full-node/db/sov-schema-db/src/snapshot.rs
@@ -316,3 +316,116 @@ impl QueryManager for NoopQueryManager {
         Ok(std::iter::empty())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::define_schema;
+    use crate::schema::KeyEncoder;
+    use crate::test::{TestCompositeField, TestField};
+
+    define_schema!(TestSchema, TestCompositeField, TestField, "TestCF");
+
+    struct SingleSnapshotQueryManager {
+        cache: SchemaBatch,
+    }
+
+    impl QueryManager for SingleSnapshotQueryManager {
+        type Iter<'a, S: Schema> = std::vec::IntoIter<(SchemaKey, SchemaValue)>;
+
+        fn get<S: Schema>(
+            &self,
+            _snapshot_id: SnapshotId,
+            key: &impl KeyCodec<S>,
+        ) -> anyhow::Result<Option<S::Value>> {
+            if let Some(Operation::Put { value }) = self.cache.read(key)? {
+                let value = S::Value::decode_value(value)?;
+                return Ok(Some(value));
+            }
+            Ok(None)
+        }
+
+        fn iter<S: Schema>(&self, _snapshot_id: SnapshotId) -> anyhow::Result<Self::Iter<'_, S>> {
+            let collected: Vec<(SchemaKey, SchemaValue)> = self
+                .cache
+                .iter::<S>()
+                .filter_map(|(k, op)| match op {
+                    Operation::Put { value } => Some((k.to_vec(), value.to_vec())),
+                    Operation::Delete => None,
+                })
+                .collect();
+
+            Ok(collected.into_iter())
+        }
+    }
+
+    fn encode_key(key: &TestCompositeField) -> Vec<u8> {
+        <TestCompositeField as KeyEncoder<TestSchema>>::encode_key(key).unwrap()
+    }
+
+    fn encode_value(value: &TestField) -> Vec<u8> {
+        <TestField as ValueCodec<TestSchema>>::encode_value(value).unwrap()
+    }
+
+    #[test]
+    fn test_db_snapshot_iterator_empty() {
+        let local_cache = SchemaBatch::new();
+        let parent_values = SchemaBatch::new();
+
+        let manager = SingleSnapshotQueryManager {
+            cache: parent_values,
+        };
+
+        let local_cache_iter = local_cache.iter::<TestSchema>().peekable();
+        let manager_iter = manager.iter::<TestSchema>(0).unwrap().peekable();
+
+        let snapshot_iter = SnapshotIter::<'_, SingleSnapshotQueryManager, TestSchema> {
+            local_cache_iter,
+            parent_iter: manager_iter,
+        };
+
+        let values: Vec<(SchemaKey, SchemaValue)> = snapshot_iter.collect();
+
+        assert!(values.is_empty());
+    }
+
+    #[test]
+    fn test_db_snapshot_iterator_values() {
+        let k1 = TestCompositeField(0, 1, 0);
+        let k2 = TestCompositeField(0, 1, 2);
+        let k3 = TestCompositeField(3, 1, 0);
+        let k4 = TestCompositeField(3, 2, 0);
+
+        let mut parent_values = SchemaBatch::new();
+        parent_values.put::<TestSchema>(&k2, &TestField(2)).unwrap();
+        parent_values.put::<TestSchema>(&k1, &TestField(1)).unwrap();
+        parent_values.put::<TestSchema>(&k4, &TestField(4)).unwrap();
+        parent_values.put::<TestSchema>(&k3, &TestField(3)).unwrap();
+
+        let mut local_cache = SchemaBatch::new();
+        local_cache.delete::<TestSchema>(&k3).unwrap();
+        local_cache.put::<TestSchema>(&k1, &TestField(10)).unwrap();
+        local_cache.put::<TestSchema>(&k2, &TestField(20)).unwrap();
+
+        let manager = SingleSnapshotQueryManager {
+            cache: parent_values,
+        };
+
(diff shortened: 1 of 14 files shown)


full-node/db/sov-schema-db/src/snapshot.rs
```diff
@@ -104,6 +115,129 @@ impl<Q: QueryManager> DbSnapshot<Q> {
...
+            .expect("Parent snapshots lock must not be poisoned");
+
+        let parent_iter = parent.iter::<S>(self.id)?;
+
+        let combined_iter: SnapshotIter<'_, Q, S> = SnapshotIter {
+            local_cache_iter: local_cache_iter.peekable(),
+            parent_iter: parent_iter.peekable(),
+        };
+
+        if let Some((_, value)) = combined_iter
```
bkolad, line 173: Here we keep `local_cache` & `parent` guards in scope so the method is locked during the filtering, is it intended?

full-node/db/sov-schema-db/src/snapshot.rs
```diff
@@ -104,6 +115,129 @@ impl<Q: QueryManager> DbSnapshot<Q> {
...
+
+struct SnapshotIter<'a, Q: QueryManager + 'a, S: Schema> {
+    local_cache_iter: std::iter::Peekable<SchemaBatchIterator<'a, S>>,
+    parent_iter: std::iter::Peekable<Q::Iter<'a, S>>,
+}
+
+impl<'a, Q: QueryManager + 'a, S: Schema> Iterator for SnapshotIter<'a, Q, S> {
+    type Item = (SchemaKey, SchemaValue);
+
+    fn next(&mut self) -> Option<Self::Item> {
```
bkolad, line 197: This look squite complex, do we have a unit test for it?

full-node/sov-prover-storage-manager/src/snapshot_manager.rs
```diff
@@ -55,9 +61,140 @@ impl SnapshotManager {
...
+            }
+
+            // We don't need key anymore
+            let mut max_values: Vec<DataLocation> = max_values
+                .into_iter()
+                .map(|(location, _)| location)
+                .collect();
+
+            // Save location of max value to be probably returned
+            let last_max_location = max_values.pop().unwrap();
```
bkolad: It looks like we convert a vec into iter (line 155) thne back into a vec just to get firsy element? Woul `max_values[0].location` work?

full-node/db/sov-schema-db/src/snapshot.rs
citizen-stig: Yes. Because we need total ordering, so it is possible that snapshot holds update the smallest key, so we will reach to it last.

This method can be optimized for the case when seek_key is present, then we can just query it. 
But if we need "previous key", then we need iterate.

I personally think, this whole iterator is not entrile used.

If you see how `state_db` uses it, it just need state_key at current or previous version.
So maybe it will be more efficient, if state_db just tries to do N queries where N = 0..=last_version. 
So instead of holding all locks.

full-node/db/sov-schema-db/src/snapshot.rs
citizen-stig: Not directly. Intergration tests `tests/iterator_test` cover that via `get_last` and `get_prev`

But I can add additional iterator test just for it

full-node/sov-prover-storage-manager/src/snapshot_manager.rs
citizen-stig: We need to do conversion, so it can drop &SchemaKey it got from iterator, only to keep index.

If we do just by index, then we have life time issue. Let's consider this change:

```rust
// // We don't need key anymore
            // let mut max_values: Vec<DataLocation> = max_values
            //     .into_iter()
            //     .map(|(location, _)| location)
            //     .collect();

            // Save location of max value to be probably returned
            let (last_max_location, _) = max_values.last().unwrap();

            // Move all iterators to next value
            for (location, _) in max_values.iter() {
                match location {
                    DataLocation::Db => {
                        let _ = self.db_iter.next().unwrap();
                    }
                    Snapshot(idx) => {
                        let _ = self.snapshot_iterators[*idx].next().unwrap();
                    }
                }
            }
```

We will get this error:

```
   --> full-node/sov-prover-storage-manager/src/snapshot_manager.rs:167:33
    |
125 |             let max_db_value = self.db_iter.peek();
    |                                ------------ first mutable borrow occurs here
...
164 |             for (location, _) in max_values.iter() {
    |                                  ----------------- first borrow later used here
...
167 |                         let _ = self.db_iter.next().unwrap();
    |                                 ^^^^^^^^^^^^ second mutable borrow occurs here

```

So we use original max_values vector. 

Also, we need "last" element, not first.

full-node/db/sov-schema-db/src/snapshot.rs
bkolad: ok

full-node/db/sov-schema-db/src/snapshot.rs
bkolad: Let's add them

full-node/sov-prover-storage-manager/src/snapshot_manager.rs
```diff
@@ -121,47 +121,44 @@ impl<'a, S: Schema> Iterator for SnapshotManagerIter<'a, S> {
...
                             Ordering::Less => {}
-                        }
-                    }
+                        },
+                    };
                 }
             }
 
-            if max_values.is_empty() {
+            if max_value_locations.is_empty() {
```
bkolad: we cam remove the unwraps bellow with `if let Some(xx) =max_value_locations.pop() ` and use xx

bkolad (approved):

full-node/sov-prover-storage-manager/src/snapshot_manager.rs
citizen-stig: Fixed

full-node/db/sov-schema-db/src/snapshot.rs
citizen-stig: Done!

PR #1179, opened 2023-11-23
feat: add `sov-modules-wallet-blueprint`

vlopes11:
This commit introduces `sov-modules-wallet-blueprint`, a blueprint unlocks wallet implementations for signed runtime calls.

The `WalletBlueprint` trait exposes the required types to declare `RuntimeCall` so these can be embeded into wallet modules.

### DIFF
--- a/module-system/sov-modules-wallet-blueprint/Cargo.toml
+++ b/module-system/sov-modules-wallet-blueprint/Cargo.toml
@@ -1,0 +1,42 @@
+[package]
+name = "sov-modules-wallet-blueprint"
+description = "Defines the interface of the Sovereign SDK wallet"
+authors = { workspace = true }
+edition = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+repository = { workspace = true }
+
+version = { workspace = true }
+readme = "README.md"
+resolver = "2"
+
+[dependencies]
+anyhow = { workspace = true, optional = true }
+async-trait = { workspace = true, optional = true }
+borsh = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
+serde_json = { workspace = true, optional = true }
+sov-cli = { path = "../sov-cli", optional = true }
+sov-mock-da = { path = "../../adapters/mock-da", version = "0.3", optional = true }
+sov-modules-api = { path = "../sov-modules-api", version = "0.3" }
+sov-modules-rollup-blueprint = { path = "../sov-modules-rollup-blueprint", optional = true }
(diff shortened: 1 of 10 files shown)


module-system/sov-modules-wallet-blueprint/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+name = "sov-modules-wallet-blueprint"
```
bkolad, line 2: Why do we need a new crate? Can we just put the `snap-wallet` in `sov-modules-rollup-blueprint` so everything is in one place?

module-system/sov-modules-wallet-blueprint/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+pub trait WalletBlueprint {
+    /// Context used to define the asymetric cryptography for keys generation and signing.
+    type Context: Context;
+    /// DA specification used to declare runtime call message of the module, that is signed by the
+    /// wallet.
+    type DaSpec: DaSpec;
+}
+
+#[cfg(feature = "default-impl")]
+pub struct DefaultWalletBlueprint;
```
bkolad, line 23: We shouldn't need this. If we move the code to `sov-modules-rollup-blueprint` like suggested in another comment we will just implemnt the `WalletBlueprint` for `CelestiaDemoRollup` & `MockDemoRollup` the same way as it was done for `CliWallet`.

vlopes11: The goal of this struct is to make a `(Context, DaSpec)` available for crates outside our demos. Meaning: if we create a new runtime that is not the one of celestia or mock, we can use `DefaultWalletBlueprint` to generate the WASM module with `ed25519` keys.

If we implement `WalletBlueprint` for celestia/mock, we will be able to generate WASM only for these runtimes

module-system/sov-modules-wallet-blueprint/Cargo.toml
vlopes11: That would bring more feature complexity to `sov-modules-rollup-blueprint` as I need to differentiate when we generate for the CLI (so we use things like `clap` as dependency), and for the WASM (that is as minimal as possible to keep the binary size small)

If we put it under `sov-modules-rollup-blueprint`, we will have to shield anything non-wallet behind feature flags

vlopes11:
wontfix; the blueprint will remain as simple as possible, while the process to generate a snap will be done entirely by the sov-snap-generator

PR #1170, opened 2023-11-21
Archival state support for state_db and native_db

dubbelosix:
# Description
Basic support at the code level to read archival state for
1. state_db (jmt)
2. native_db (accessory)

## Changes
1. Added archival_version to ProverStorage
2. Support for NativeDB and  StateDB for versioning
3. Test cases for querying archival state using ProverStorage and WorkingSet for accessory and jmt

## Pending
Expose archival state via RPC macros

PR #1184, opened 2023-11-27
Support archival queries by modifying the WorkingSet

dubbelosix:
# Description
This accomplishes support for archival / versioned queries against state-db and native-db. This is similar to #1170 but instead of modifying the Storage level, this modification is at the WorkingSet level

## Linked Issues
- https://github.com/Sovereign-Labs/sovereign-sdk/pull/1170


module-system/sov-modules-core/src/storage/scratchpad.rs
```diff
@@ -460,6 +479,69 @@ impl<'a, C: Context> StateReaderAndWriter for AccessoryWorkingSet<'a, C> {
...
+        Jmt(S),
+        /// Accessory store
+        Accessory(S),
+    }
+
+    /// Archival working set
+    pub struct ArchivalWorkingSet<C: Context> {
+        delta: ArchivalStore<<C as Spec>::Storage>,
+        archival_version: Version,
+        witness: <<C as Spec>::Storage as Storage>::Witness,
```
bkolad: Why do we need witness in `ArchivalWorkingSet`?

module-system/module-implementations/sov-bank/src/call.rs
```diff
@@ -263,7 +263,7 @@ impl<C: sov_modules_api::Context> Bank<C> {
     pub fn get_token_name(
         &self,
         token_address: &C::Address,
-        working_set: &mut WorkingSet<C>,
+        working_set: &mut impl StateMapWorkingSet,
```
bkolad: Is the goal to replace `&mut WorkingSet<C>` with `&mut impl StateMapWorkingSet`? If that is the case, then the `StateMapWorkingSet` should also support the `charge_gas` functionality. And what about the other modules? Will they all be moved to work with `StateMapWorkingSet`?

module-system/sov-modules-core/src/storage/scratchpad.rs
```diff
@@ -460,6 +479,69 @@ impl<'a, C: Context> StateReaderAndWriter for AccessoryWorkingSet<'a, C> {
...
+    }
+
+    impl<C: Context> StateReaderAndWriter for ArchivalWorkingSet<C> {
+        fn get(&mut self, key: &StorageKey) -> Option<StorageValue> {
+            match &self.delta {
+                ArchivalStore::Jmt(jmt_store) => {
+                    jmt_store.get(key, Some(self.archival_version), &self.witness)
+                }
+                ArchivalStore::Accessory(accessory_store) => {
+                    if !cfg!(feature = "native") {
```
bkolad: I think this should guard the entire function and we should panic in this case.

module-system/sov-modules-core/src/storage/scratchpad.rs
```diff
@@ -460,6 +479,69 @@ impl<'a, C: Context> StateReaderAndWriter for AccessoryWorkingSet<'a, C> {
...
+                        None
+                    } else {
+                        accessory_store.get_accessory(key, Some(self.archival_version))
+                    }
+                }
+            }
+        }
+
+        fn set(&mut self, _key: &StorageKey, _value: StorageValue) {}
```
bkolad: I think `set` also should be supported. So the user can run arbitrary logic against the archival state.

module-system/sov-modules-api/src/containers/map.rs
```diff
@@ -1,10 +1,11 @@
 use std::marker::PhantomData;
 
-use sov_modules_core::{Context, Prefix, StateCodec, StateKeyCodec, StateValueCodec, WorkingSet};
+#[cfg(feature = "arbitrary")]
+use sov_modules_core::{Context, WorkingSet};
```
neysofu: Instead of gated imports, I'd suggest using the full type path inside `arbitrary` code. The less feature gates we have, the better.

module-system/sov-modules-core/src/storage/scratchpad.rs
```diff
@@ -7,12 +7,21 @@ pub use kernel_state::{KernelWorkingSet, VersionedWorkingSet};
...
     CacheKey, CacheValue, EncodeKeyLike, NativeStorage, OrderedReadsAndWrites, StateCodec,
     StateValueCodec, Storage, StorageInternalCache, StorageKey, StorageProof, StorageValue,
 };
+use crate::Version;
+
+/// Super trait for convenience in macro parsing and identifying Working sets that need StateMap implementation of StateMapAccessor
+pub trait StateMapWorkingSet: StateReaderAndWriter {}
+
+impl<C: Context> StateMapWorkingSet for WorkingSet<C> {}
+impl<C: Context> StateMapWorkingSet for ArchivalWorkingSet<C> {}
```
neysofu: This idea makes sense to me, i.e. we only accessing `StateMap` values with certain `StateReaderAndWriter` implementations because by doing so we're protecting the user against misuse and mixing accessory and JMT state together. This retains the protection that `StateMapAccessor` gave us before the change, which only worked for `WorkingSet`.

We can make this more generic though, so that it's not limited to `StateMap`. I think these four traits may be enough:

- `StateReader`
- `StateWriter`
- `AccessoryStateReader`
- `AccessoryStateWriter`

With this system, we won't need new traits for every state container, and we can have e.g.

```rust
    fn check_balance(
        &self,
        from: &C::Address,
        amount: Amount,
        working_set: &mut impl StateReader,
    ) -> Result<Amount> {
        // ...
    }
```

...which doesn't require a `StateMap`-specific working set, but retains type safety.

module-system/sov-modules-core/src/storage/scratchpad.rs
neysofu: How would that work, without breaking state and having ripple effects on all future versions? We'd need branching versions, right?

module-system/sov-modules-core/src/storage/scratchpad.rs
neysofu: I agree getters on accessory state in non-native environments should panic, that's something we've wanted to change for a while. But IMHO it's best to tackle it separately and change the behavior all at once for all accessory getters in a separate PR than to start with just one and leave it inconsistent.

module-system/sov-modules-core/src/storage/scratchpad.rs
neysofu: As a compromise not to block this PR, can we rename `StateMapWorkingSet` to something like `StateBackend` and change all state containers to use it? Maybe we can discuss this option at standup.

neysofu (commented):
The PR looks good – I'd just like to make sure that we're not accumulating too much tech debt with this change, since it doesn't replace all instances of `WorkingSet` with generics so it leaves things in a somewhat messy state. With the addition of `ArchivalWorkingSet`, we also have too many `WorkingSet`-like structs – although that's definitely not this PR's fault. Let's brainstorm at standup!

module-system/sov-modules-core/src/storage/scratchpad.rs
bkolad: This will work only per http request and won't be committed. The idea is that if you have a query method that writes to a state and then reads the same state the changes should be visible inside this single query and discarded after.

module-system/module-implementations/sov-bank/src/call.rs
dubbelosix: ah, so not entirely (and not in all places, primarily query.rs).
but that's a good point since gas should be charged for cross-module calls, otherwise it will prevent module1 from calling module2. will add a gas trait for that an impl it for regular working set - will check if that is feasible

module-system/sov-modules-core/src/storage/scratchpad.rs
dubbelosix: yup, set can be supported as well - one of the reasons for making it at workingset level was to let the archival state be used to "simulate" transactions by using set, but for the immediate purpose, wanted to just support it at the RPC level

module-system/sov-modules-core/src/storage/scratchpad.rs
dubbelosix: that makes sense actually - I'll rename it

module-system/sov-modules-core/src/storage/scratchpad.rs
citizen-stig: I think we don't, but my guess is to keep it compatible with `Storage` methods, which require `Witness` to be passed explicitly via parameters.

I personally would prefer, that witness would be kept internally to the ProverStorage and ZkStorage, and created new one together with Storage instance itself. But this change is defenetly out of the scope of this PR.

Now we have case of tight coupling, so all users of "storage" trait should maintain witness for it.

@preston-evans98 , what do you think? I vaguely recall several months ago I've asked this question already.

module-system/sov-modules-core/src/storage/scratchpad.rs
citizen-stig: > I think set also should be supported. So the user can run arbitrary logic against the archival state.

I agree, that it is nice to have feature, but in reality, how many queries there are, which actually write to state? 
I just want to prevent us to adding support for something, that won't be actually used, but we will keep burden of maintaining it. Sometimes limiting functionality can be benefitial.

I also see, how simulating feature definitely will be needed, but I am not sure, that it makes sense to simulate something on archival state.

citizen-stig (approved):
Look good to me. I don't see any blockers to integrating fork management into it.

Changes in `sov-db` are not entierly new, native db just gets same functionality as state_db already has.

module-system/sov-modules-core/src/storage/scratchpad.rs
dubbelosix: will add the set functionality in another PR for simulating module calls on the WorkingSet (without committing).

bkolad: Changing state inside a query is a standard use-case.

module-system/module-implementations/sov-bank/src/call.rs
bkolad: Q: What was the reason for `impl StateMapWorkingSet` instead a concrete type (`ArchivalWorkingSet`)?

module-system/sov-modules-core/src/storage/scratchpad.rs
bkolad: Here is the other place that needs to be changed: https://github.com/Sovereign-Labs/sovereign-sdk/blob/nightly/module-system/sov-modules-core/src/storage/mod.rs#L209

module-system/sov-modules-core/src/storage/scratchpad.rs
```diff
@@ -338,6 +362,28 @@ impl<C: Context> WorkingSet<C> {
         AccessoryWorkingSet { ws: self }
     }
 
+    /// Returns a handler for the archival state (JMT state).
+    pub fn archival_state(&mut self, version: Version) -> ArchivalJmtWorkingSet<C> {
```
bkolad: it looks like `archival_accessory_state & archival_state` methods cane be private.

bkolad (approved):
LGTM, would it be possible to add more tests in `sov-modules-code/tests` ? (In a separate PR)

PR #1188, opened 2023-11-28
feat: add transaction id to DaService

vlopes11:
This commit introduces an associated type `TransactionId` to `DaService`.

It will unlock the ability to track a submitted transaction after when it is sent to the DA, and before it is included on a block.


preston-evans98 (approved):
Approving to unblock customers, with https://github.com/Sovereign-Labs/sovereign-sdk/issues/1189 as a TODO.

PR #1192, opened 2023-11-29
feat: add chain-id to transaction

vlopes11:
This commit introduces `chain-id` to `Transaction`, a constant that will prevent transactions from being replayed on different chains.

It also includes `gas_tip`, a scalar value to be added to the runtime gas amount so sequencers can prioritize a broadcasted transaction.


vlopes11:
Superseded by https://github.com/Sovereign-Labs/sovereign-sdk/pull/1213

PR #1193, opened 2023-11-29
MockDa support for `demo-rollup`

bkolad:
# Description
Updates `demo-rollup` README with `mock-da` setup.

## Testing
CI tests added.

## Docs
Added readme.


examples/demo-rollup/README.md
```diff
@@ -60,87 +43,46 @@ understand how to build your own state transition function, check out at the doc
...
 [sov-rollup-starter](https://github.com/Sovereign-Labs/sov-rollup-starter/)
 
-### Run a local DA layer instance
-
-1. Install Docker: <https://www.docker.com>.
 
-2. Follow [this guide](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic)
-to authorize yourself in github's container registry. (we use original celestia images which they publish in ghcr)
+### Run a local DA layer instance
+This setup works with an in-memory DA that is easy to set up for testing purposes.
```
dubbelosix, line 59: should we include a link here to README_CELESTIA.md saying "check here if you want to run with dockerized local celestia) ?
might be good, since how to customize is also moved to that README

dubbelosix (approved):

examples/demo-rollup/README.md
```diff
@@ -23,30 +21,15 @@ This is a demo full node running a simple Sovereign SDK rollup on [Celestia](htt
...
     - [2. Generate the Transaction](#2-generate-the-transaction)
-    - [3. Submit the Transaction(s)](#3-submit-the-transactions)
-    - [4. Verify the Token Supply](#4-verify-the-token-supply)
-  - [Makefile](#makefile)
-  - [Remote setup](#remote-setup)
-- [How to Customize This Example](#how-to-customize-this-example)
-  - [1. Initialize the DA Service](#1-initialize-the-da-service)
-  - [2. Run the Main Loop](#2-run-the-main-loop)
-- [Disclaimer](#disclaimer)
-- [Interacting with your Node via RPC](#interacting-with-your-node-via-rpc)
```
cemozerr, line 36: Is there a reason we removed the "Interacting with your Node via RPC" section? Happy to be convinced otherwise, but that section seems more appropriate to be in the main README.md

examples/demo-rollup/README.md
cemozerr: I think a link here could be distracting, too early on. We'd rather have users to see the thing running simply before we introduce complexity like DA's and Celestia's docker stuff. But I think it we should have a section in the end such as "Testing on specific DA layers" and then we'd have the Celestia Readme.

bkolad: Added the Celestia link at the end.

examples/demo-rollup/README.md
bkolad: I agree the main readme is a better place. I have just updated this part.

examples/demo-rollup/README.md
```diff
@@ -240,4 +249,135 @@ $ cargo run --bin sov-cli rpc submit-batch by-address sov1l6n2cku82yfqld30lanm2n
...
+```shell
+$ curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"ledger_getEvents","params":[1],"id":1}' http://127.0.0.1:12345
+
+{"jsonrpc":"2.0","result":[null],"id":1}
+```
+
+This response indicates that event `1` has not been emitted yet.
+
+## Testing with specific DA layers
+Check [here](./README_CELESTIA.md) if you want to run with dockerized local celestia.
```
cemozerr:
```suggestion
Check [here](./README_CELESTIA.md) if you want to run with dockerized local Celestia instance.
```

cemozerr (approved):

PR #1205, opened 2023-12-06
refactor: simplify CliWalletArg decodable message

vlopes11:
This commit removes the `CliWallet` and `CliFrontEnd` traits. Additionally, it refactors the `CliWalletArg` to encode runtime call messages directly into a readable format, thereby eradicating an intermediary representation.

The prior implementation employed an intermediary representation that was adjustable by users but necessitated manual conversion into a decodable message format. As of now, this step is deemed superfluous because the readable transaction can embody the new `CliWalletArg` and offer decoding functionality instantaneously. Typically, the input will be a string that can be translated from a common serialization format like JSON or YAML. By default, JSON is utilized; however, other common serialization formats can also be supported.

The changeset aims to make frontend portrayal of superstructures possible for runtime call messages. It simplifies the decoding process so that it can be seamlessly plugged into wrapper superstructures containing additional metadata. One potential instance is `UnsignedTransaction` as a wrapping superstructure for runtime call data, with relevant protocol-level information pertaining to the rollup.

Moreover, this commit renames `sov_modules_api::cli::FileNameArg` to `JsonFileNameArg`. This type inherently represents the path to a JSON file; hence, the name must reflect that. This allows for effortless addition of paths symbolizing other formats in the future (e.g., `YamlFileNameArg`).

### DIFF
--- a/examples/demo-rollup/src/sov-cli/main.rs
+++ b/examples/demo-rollup/src/sov-cli/main.rs
@@ -1,13 +1,15 @@
 use demo_stf::runtime::RuntimeSubcommand;
 use sov_demo_rollup::CelestiaDemoRollup;
-use sov_modules_api::cli::{FileNameArg, JsonStringArg};
+use sov_modules_api::cli::{JsonFileNameArg, JsonStringArg};
 use sov_modules_rollup_blueprint::WalletBlueprint;
 
 #[tokio::main]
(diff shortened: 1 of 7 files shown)


bkolad (approved):

vlopes11:
Superseded by https://github.com/Sovereign-Labs/sovereign-sdk/pull/1213

PR #1213, opened 2023-12-12
feat: add `UnsignedTransaction`

### LINKED ISSUE #624: Transactions can be replayed
## Background
Currently, the SDK does not include a chain-id. So, any message sent on one Sovereign SDK chain can potentially be replayed on any other chain. To fix this, we should add a chain-id field to transactions.

vlopes11:
This commit introduces `UnsignedTransaction`, an intermediary type that will contain transaction metadata before it is signed.

To maintain consistency with the JSON runtime call messages serialized format, we preserve this format because the serialization occurs prior to the unsigned transaction serialization. The runtime will define a module name for its implementation; therefore, the runtime call will have a custom field name, so it must be serialized beforehand.

In order to maintain simplicity, two CLI arguments are added to sov-cli to contain the two new scalar attributes of the transaction: chain id and gas tip.

The chain ID is introduced to prevent cross-chain transaction reproduction. The signed transaction will also take the chain id field as part of the signature; therefore, a signature is valid only for a specified chain.

The gas tip is an incentive for the sequencer to pick the given transaction from the pool.

## Linked Issues
- Fixes #624

### DIFF
--- a/examples/demo-rollup/README.md
+++ b/examples/demo-rollup/README.md
@@ -223,7 +223,7 @@ Let's go ahead and import the transaction into the wallet
 
 
 ```bash,test-ci,bashtestmd:compare-output
-$ cargo run --bin sov-cli -- transactions import from-file bank --path ../test-data/requests/transfer.json
+$ cargo run --bin sov-cli -- transactions import from-file bank --chain-id 0 --path ../test-data/requests/transfer.json
 Adding the following transaction to batch:
 {
   "tx": {
--- a/examples/demo-rollup/README.md
+++ b/examples/demo-rollup/README.md
@@ -226,15 +226,19 @@ Let's go ahead and import the transaction into the wallet
 $ cargo run --bin sov-cli -- transactions import from-file bank --path ../test-data/requests/transfer.json
 Adding the following transaction to batch:
 {
-  "bank": {
-    "Transfer": {
-      "to": "sov1l6n2cku82yfqld30lanm2nfw43n2auc8clw7r5u5m6s7p8jrm4zqrr8r94",
-      "coins": {
-        "amount": 200,
-        "token_address": "sov1zdwj8thgev2u3yyrrlekmvtsz4av4tp3m7dm5mx5peejnesga27svq9m72"
+  "tx": {
+    "bank": {
+      "Transfer": {
+        "to": "sov1l6n2cku82yfqld30lanm2nfw43n2auc8clw7r5u5m6s7p8jrm4zqrr8r94",
+        "coins": {
+          "amount": 200,
+          "token_address": "sov1zdwj8thgev2u3yyrrlekmvtsz4av4tp3m7dm5mx5peejnesga27svq9m72"
+        }
       }
     }
-  }
+  },
+  "chain_id": 0,
+  "gas_tip": 0
 }
 ```
 
--- a/module-system/sov-modules-api/src/transaction.rs
+++ b/module-system/sov-modules-api/src/transaction.rs
@@ -1,6 +1,6 @@
-use std::{fmt, io, marker};
-
-use serde::ser::SerializeStruct;
+use borsh::{BorshDeserialize, BorshSerialize};
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
 #[cfg(feature = "native")]
 use sov_modules_core::PrivateKey;
 use sov_modules_core::{Context, Signature};
@@ -23,8 +23,12 @@ pub struct Transaction<C: Context> {
 }
 
 /// An unsent transaction with the required data to be submitted to the DA layer
-#[derive(Debug)]
-pub struct UnsignedTransaction<Tx> {
+#[derive(Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+#[serde(bound = "Tx: serde::Serialize + serde::de::DeserializeOwned")]
+pub struct UnsignedTransaction<Tx>
+where
+    Tx: BorshSerialize + BorshDeserialize,
+{
     /// The underlying transaction
     pub tx: Tx,
     /// The ID of the target chain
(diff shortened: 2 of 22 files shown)


module-system/sov-modules-api/src/transaction.rs
```diff
@@ -63,23 +96,181 @@ impl<C: Context> Transaction<C> {
...
+    fn serialize<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
+        self.tx.serialize(writer)?;
+        self.chain_id.serialize(writer)?;
+        self.gas_tip.serialize(writer)?;
+
+        Ok(())
+    }
+}
+
+impl<Tx> borsh::BorshDeserialize for UnsignedTransaction<Tx>
```
bkolad: Why do we need this implementation? Can's this be auto derived?

module-system/sov-modules-api/src/transaction.rs
```diff
@@ -63,23 +96,181 @@ impl<C: Context> Transaction<C> {
...
+
+        Ok(Self {
+            tx,
+            chain_id,
+            gas_tip,
+        })
+    }
+}
+
+impl<Tx> serde::Serialize for UnsignedTransaction<Tx>
```
bkolad: The same question for auto derive.

examples/demo-rollup/README.md
```diff
@@ -226,15 +226,19 @@ Let's go ahead and import the transaction into the wallet
 $ cargo run --bin sov-cli -- transactions import from-file bank --path ../test-data/requests/transfer.json
```
bkolad: The `chains_id` and `gas_tip` parameters are not provided to the CLI. Where are these values obtained?

vlopes11: This is the stdout printed here

https://github.com/Sovereign-Labs/sovereign-sdk/blob/3ae9b7b92c8123864bf7471bfb4c086da93fe3af/module-system/sov-cli/src/workflows/transactions.rs#L136

I opted to add the full `UnsignedTransaction` because `(chain_id, gas_tip)` are part of the user input, so it makes sense to integrate these into the feedback

module-system/sov-modules-api/src/transaction.rs
vlopes11: `UnsignedTransaction` might live without a borsh serialization implementation, as in

https://github.com/Sovereign-Labs/sovereign-sdk/blob/3ae9b7b92c8123864bf7471bfb4c086da93fe3af/module-system/sov-cli/src/wallet_state.rs#L10-L20

module-system/sov-modules-api/src/transaction.rs
vlopes11: In this case, we have one scenario where only `Serialize` is required

https://github.com/Sovereign-Labs/sovereign-sdk/blob/3ae9b7b92c8123864bf7471bfb4c086da93fe3af/module-system/sov-cli/src/workflows/transactions.rs#L108

I opted to avoid introducing new unnecessary bounds, as we can implement serde/borsh serialization conditionally. I can naturally just put the bounds downstream and get rid of the manual implementations, if they come to be a problem

module-system/sov-modules-api/src/transaction.rs
bkolad: Not sue I understand. What would be the problem for deriving borsh for `UnsignedTransaction ` by a macro?

examples/demo-rollup/README.md
bkolad: My question is how these parameters are inserted as  there are no `--chain_id` & `--chain_id` parameters in the above commands.

module-system/sov-modules-api/src/transaction.rs
bkolad: The manual implementation [here](https://github.com/Sovereign-Labs/sovereign-sdk/pull/1213/files/3ae9b7b92c8123864bf7471bfb4c086da93fe3af#diff-710780d3f43bf298801069e657ff34727958335421a0642a72a99c1d8b019e12R190) is quite complex.

The main issue lies in the hardcoded field names. If someone modifies or adds a new field, there's a risk of forgetting to update the code accordingly, which could lead to unnoticed errors. What are the downsides of adding additional bounds apart that they are not strictly needed?

examples/demo-rollup/README.md
vlopes11: They have a default value of `0`, as of the clap implementation

https://github.com/Sovereign-Labs/sovereign-sdk/blob/3ae9b7b92c8123864bf7471bfb4c086da93fe3af/module-system/sov-modules-api/src/cli.rs#L27-L29

bkolad: Ahh, I think we should not have default values for this parameter, as someone could, by accident, send a tx to the wrong chain.

vlopes11: Done on https://github.com/Sovereign-Labs/sovereign-sdk/pull/1213/commits/32c6a8f33c234fcdc428f56036fef8d13e804680

module-system/sov-modules-api/src/transaction.rs
vlopes11: Unfortunately that is a restriction of serde. They have no means to implement easily the serialization for such case.

We could require all the `Tx` instances to implement `{serde, borsh}::{Serialize, Deserialize}`, if they intend to use the `sov-cli`. I don't see much of a trouble by doing that, as it should be ok to have a over-restrictive CLI interface.

We can't, however, leak this to service level: as RPC, we might want only `borsh`. I'll make a shot to see if it can be done without leaking

vlopes11: Done on https://github.com/Sovereign-Labs/sovereign-sdk/pull/1213/commits/aa3e742e72b17e54f3bb8c3d95e34324ebaaec09

module-system/sov-modules-api/src/transaction.rs
vlopes11: Resolved on https://github.com/Sovereign-Labs/sovereign-sdk/pull/1213/commits/aa3e742e72b17e54f3bb8c3d95e34324ebaaec09

bkolad (approved):
LGTM, can you please update sov-starter as well.

PR #1221, opened 2023-12-14
Implement ZK proof submissin for `MockDa`

bkolad:
# Description
This PR adds implementation of  `send_aggregated_zk_proof & get_aggregated_proofs_at` for `MockDa`

## Testing
Added a unit test.

## Docs
No changes.

### DIFF
--- a/adapters/mock-da/src/service.rs
+++ b/adapters/mock-da/src/service.rs
@@ -79,6 +79,62 @@ impl MockDaService {
         }
         anyhow::bail!("No blob at height={height} has been sent in time")
     }
+
+    async fn add_blob(&self, blob: &[u8], zkp_proof: Vec<u8>) -> Result<u64, anyhow::Error> {
+        let mut blocks = self.blocks.write().await;
+
+        let (previous_block_hash, height) = match blocks.iter().last().map(|b| b.header().clone()) {
+            None => (MockHash::from([0; 32]), 0),
+            Some(block_header) => (block_header.hash(), block_header.height + 1),
+        };
+        let data_hash = hash_to_array(blob);
+        let proof_hash = hash_to_array(&zkp_proof);
+        // Hash only from single blob
+        let block_hash = block_hash(height, data_hash, proof_hash, previous_block_hash.into());
+
+        let blob = MockBlob::new_with_zkp_proof(
+            blob.to_vec(),
+            vec![zkp_proof],
+            self.sequencer_da_address,
+            data_hash,
+        );
+        let header = MockBlockHeader {
+            prev_hash: previous_block_hash,
+            hash: block_hash,
+            height,
+            time: Time::now(),
+        };
+        let block = MockBlock {
+            header,
+            validity_cond: Default::default(),
+            blobs: vec![blob],
+        };
+        blocks.push_back(block);
+
+        // Enough blocks to finalize block
+        if blocks.len() > self.blocks_to_finality as usize {
+            let oldest_available_height = blocks[0].header().height();
+            let last_finalized_height = self.last_finalized_height.load(Ordering::Acquire);
+            let last_finalized_index = last_finalized_height
+                .checked_sub(oldest_available_height)
+                .unwrap();
+            let next_index_to_finalize = blocks.len() - self.blocks_to_finality as usize - 1;
+
+            if last_finalized_index > 0 {
+                assert_eq!(next_index_to_finalize as u64, last_finalized_index + 1);
+            }
+
+            self.finalized_header_sender
+                .send(blocks[next_index_to_finalize].header().clone())
+                .unwrap();
+
+            let this_finalized_height = oldest_available_height + next_index_to_finalize as u64;
+            self.last_finalized_height
+                .store(this_finalized_height, Ordering::Release);
+        }
+
+        Ok(height)
+    }
 }
 
 #[pin_project]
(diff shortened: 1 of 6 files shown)


adapters/mock-da/src/service.rs
```diff
@@ -218,61 +274,17 @@ impl DaService for MockDaService {
...
-            self.finalized_header_sender
-                .send(blocks[next_index_to_finalize].header().clone())
-                .unwrap();
-
-            let this_finalized_height = oldest_available_height + next_index_to_finalize as u64;
-            self.last_finalized_height
-                .store(this_finalized_height, Ordering::Release);
-        }
-
+        let _ = self.add_blob(blob, Default::default()).await;
```
vlopes11: We should probably evaluate the `Result` of `add_blob` via `self.add_blob(blob, Default::default()).await?;`

adapters/mock-da/src/types/mod.rs
```diff
@@ -140,6 +140,7 @@ pub struct MockBlob {
     pub(crate) hash: [u8; 32],
     /// Actual data from the blob. Public for testing purposes.
     pub data: CountedBufReader<Bytes>,
+    pub(crate) zk_proofs_data: Vec<u8>,
```
vlopes11, line 144: Let's add a small doc to describe the attribute? Also, as optional nit, if this is the concrete proof serialized, we should probably name it `zk_proofs`, as `zk_proofs_data` might give the impression we are storing proof metadata, such as witnesses only

adapters/mock-da/src/service.rs
```diff
@@ -79,6 +79,62 @@ impl MockDaService {
         }
         anyhow::bail!("No blob at height={height} has been sent in time")
     }
+
+    async fn add_blob(&self, blob: &[u8], zkp_proof: Vec<u8>) -> Result<u64, anyhow::Error> {
+        let mut blocks = self.blocks.write().await;
+
+        let (previous_block_hash, height) = match blocks.iter().last().map(|b| b.header().clone()) {
+            None => (MockHash::from([0; 32]), 0),
+            Some(block_header) => (block_header.hash(), block_header.height + 1),
```
vlopes11, line 88: nit: this looks like `unwrap_or_default`

adapters/mock-da/src/service.rs
```diff
@@ -79,6 +79,62 @@ impl MockDaService {
...
+            header,
+            validity_cond: Default::default(),
+            blobs: vec![blob],
+        };
+        blocks.push_back(block);
+
+        // Enough blocks to finalize block
+        if blocks.len() > self.blocks_to_finality as usize {
+            let oldest_available_height = blocks[0].header().height();
+            let last_finalized_height = self.last_finalized_height.load(Ordering::Acquire);
```
vlopes11, line 118: The operations we perform between the `Acquire` and `Release` are relatively complex, including a `send.finalized_header_sender.send` that will involve external resources.

Consider using a lock guard (RwLock/Mutex) instead of atomic synchronization, as the latter is expected to have very cheap ordered operations. This way, you will also unlock the ability to return logical errors instead of panic (i.e. `checked_sub`, `last_finalized_index`, `send.unwrap()`)

adapters/mock-da/src/service.rs
```diff
@@ -79,6 +79,62 @@ impl MockDaService {
...
+            blob.to_vec(),
+            zkp_proof,
+            self.sequencer_da_address,
+            data_hash,
+        );
+        let header = MockBlockHeader {
+            prev_hash: previous_block_hash,
+            hash: block_hash,
+            height,
+            time: Time::now(),
```
vlopes11, line 106: Dunno if we want to worry about that on the mock implementation, but using the local systime might be unreliable in terms of consensus. The topic is further discussed [here](https://ethresear.ch/t/time-as-a-public-service-in-byzantine-context/6895)

vlopes11 (approved):
Looks good! Added a couple of optional nits & discussions

adapters/mock-da/src/service.rs
bkolad: Good point fixed.

adapters/mock-da/src/types/mod.rs
bkolad: Thx, added a comment it. The `zk_proofs_data = proof + public input`

adapters/mock-da/src/service.rs
bkolad: This was changed yesterday: https://github.com/Sovereign-Labs/sovereign-sdk/commit/6e22b4c91bce64b70cb1624770ce10a773c36a08 . We use the `MockDa` only for tests and in the demo (it is always a single node)

adapters/mock-da/src/service.rs
bkolad: I have just tried, but because we modifiy the touple in ` Some(block_header)` the version with `unwrap_or_default` was longer;)

adapters/mock-da/src/service.rs
bkolad: I moved the code from another place and didn't change the functionality. In line `87`, we are holding a lock for the entire function so I am not sure why we need the `Ordering` here. I will create a separate issue to investigate it. 

Issue added here: https://github.com/Sovereign-Labs/sovereign-sdk/issues/1222

vlopes11: Sounds good! I think the lock does make sense as we manipulate the value that is outside the `self.blocks` lock. But OTOH we are facing this issue where the operation inside the atomic lock is just too complex, so maybe we want to lock a more complex structure than a single `VecDeque` for the blocks? A wrapper `struct BlockState { blocks: VecDeque<Block>, last_finalized_height: u64 }` might make more sense

PR #1223, opened 2023-12-15
feat: add gas limit to tx

vlopes11:
With this commit, we implement a gas limit for transactions. The introduced value signifies the available gas resources for a transaction prior to its execution by the runtime. Upon transaction dispatch finalization, the bank module will consume the excess gas and distribute the tip from the sender's account.

To identify the sender address, an event `accounts::sender` has been established between accounts and the bank. This is essential because the bank cannot determine or access context while executing hooks.

The definition of the dynamic gas price elasticity remains undetermined with this commit, as it relies on chain state and execution history for calculation. Furthermore, this change does not impact the accrued gas rewards held by the sequencer, which will be transferred to his bank account upon block execution completion.


module-system/module-implementations/sov-bank/src/hooks.rs
```diff
@@ -0,0 +1,67 @@
...
+        &self,
+        tx: &Transaction<C>,
+        working_set: &mut WorkingSet<C>,
+        arg: BankTxHook<C>,
+    ) -> anyhow::Result<()> {
+        let BankTxHook { sender, sequencer } = arg;
+
+        let amount = tx.gas_limit().saturating_add(tx.gas_tip());
+        if amount > 0 {
+            let token_address = self.address.clone();
```
bkolad: This is the address of the module, but we have to transfer from the address of the gas `token`.

examples/demo-rollup/stf/src/hooks_impl.rs
```diff
@@ -24,9 +27,27 @@ impl<C: Context, Da: DaSpec> TxHooks for Runtime<C, Da> {
...
+            tx,
+            working_set,
+            BankTxHook {
+                sender: sender.clone(),
+                sequencer: sequencer.clone(),
+            },
+        )?;
+
+        let sender =
+            Arc::try_unwrap(sender).map_err(|e| anyhow::anyhow!("synchronization failure: {e}"))?;
```
bkolad: It would be better if `Context` also keeps `Arc` as well so there will be no need for `try_unwrap` here

module-system/module-implementations/sov-bank/src/hooks.rs
vlopes11: Done!

examples/demo-rollup/stf/src/hooks_impl.rs
vlopes11: I moved the `Arc` to `Rc` as we don't perform the hooks under `async`. Is it ok if we move this change to a dedicated PR? Any changes to the `Context` will imply a lot of new LOC to this one

vlopes11: One easy way to achieve that is to make the constructor of the context receive anything that can be an arc:

```rust
/// Constructor for the Context.
fn new<S, Q>(sender: S, sequencer: Q, height: u64) -> Self
where
    S: Into<Arc<Self::Address>>,
    Q: Into<Arc<Self::Address>>;
```

Wdyt?

vlopes11: Regarding the problem above, I was able to get away with that since `Arc/Rc` employs a hack that returns the rc itself instead of the error.

This allows us to clone the value for the worst case scenario - that is if the value still lives elsewhere. To be fair, this would happen only if the module is trying to do something very weird, such as sending the rc into a bounded channel, but then we have to shield against this possible outcome.

bkolad: Do we need the `Rc` at all? What is the original reason for introducing it?

vlopes11: The reason is to avoid unnecessary cloning here

https://github.com/Sovereign-Labs/sovereign-sdk/blob/8472d1d76dc8bb25dcc98ae88e0e87ed7c08c655/examples/demo-rollup/stf/src/hooks_impl.rs#L23-L47

While avoiding lifetime complexity when declaring the types. We can pass them around as owned without bounding lifetimes to the runner and/or context

vlopes11: (we do clone there, but we clone the `RC` and not the data itself)

bkolad: Ok, but why is the ownership/cloning needed? I understand we `Rc::clone` cheaply, but why is there a reason for cloning? Even with `Rc(sender )`, the `pre_dispatch_tx_hook` doesn't have access to the owned sender if it doesn't `Rc::unwrap` in the module code.

vlopes11: We pass the ownership to `BankTxHook`. The alternative to that would be creating `BankTxHook<'a>`, that introduces lifetime complexity. Its easier to read the code without lifetime mess.

If we really want the lifetime there, we would have to introduce a lifetime to the bank hook declaration too.

bkolad: What about passing `hook: &BankTxHook` instead and taking the reference via `hook.sender` inside the `pre_dispatch_tx_hook `.

vlopes11: You mean changing the signature of the trait like this?

```diff
 /// Runs just before a transaction is dispatched to an appropriate module.
 fn pre_dispatch_tx_hook(
     &self,
     tx: &Transaction<Self::Context>,
     working_set: &mut WorkingSet<Self::Context>,
-    arg: Self::PreArg,
+    arg: &Self::PreArg,
 ) -> anyhow::Result<Self::PreResult>;
```

The downturn is that we always have to clone to pass the ownership to the context, if we do that. We receive the reference, we necessarily need to transform it to owned so we can create the context

bkolad: Why wouldn't it work if we do the following:
1. Create `BankTxHook` ` with owned values.
2. Pass  `&BankTxHook` to the `bank.pre_dispatch_tx_hook` so we only borrow.
3. We use `sender & sequencer` to construct the Context, they are still in the original `BankTxHook` which was only borrowed and the borrow ended so now can continue to use it?

vlopes11: We must use either owned everywhere or ref everywhere as we share the trait signature. There is an alternative to using Rc if we consume the type on bank and return the argument. I added it on https://github.com/Sovereign-Labs/sovereign-sdk/pull/1223/commits/4ab9d782c534a8b0f8a6adf78bee0c40efed73c0 ; let me know wdyt

vlopes11: Fixed on https://github.com/Sovereign-Labs/sovereign-sdk/pull/1223/commits/1ee482aebef655beb365669c00b018318f163129

module-system/module-implementations/sov-bank/tests/gas_test.rs
```diff
@@ -0,0 +1,223 @@
...
+            tmpdir,
+            gas_limit: sender_balance,
+            native_price: 0,
+            zk_price: 0,
+        }
+    }
+
+    pub fn override_gas_config(mut self) -> Self {
+        self.bank.override_gas_config(BankGasConfig {
+            create_token: [CREATE_TOKEN_NATIVE_PRICE, CREATE_TOKEN_ZK_PRICE],
```
bkolad: `CREATE_TOKEN_XX_PRICE` is a wrong name here. This is the cost of the operation not the price.

module-system/module-implementations/sov-bank/tests/gas_test.rs
```diff
@@ -0,0 +1,223 @@
...
+    let zk_price = 3;
+    let remaining_funds = BankGasTestCase::init(sender_balance)
+        .with_native_price(native_price)
+        .with_zk_price(zk_price)
+        .override_gas_config()
+        .execute()
+        .unwrap();
+
+    // compute the expected gas cost, based on the test constants
+    let gas_cost = native_price * CREATE_TOKEN_NATIVE_PRICE + zk_price * CREATE_TOKEN_ZK_PRICE;
```
bkolad: Same as below  let's rename`CREATE_TOKEN_X_PRICE` to something like `CREATE_TOKEN_X_COST `

`gas_cost => gas_used`

module-system/module-implementations/sov-bank/tests/gas_test.rs
vlopes11: Fixed!

module-system/module-implementations/sov-bank/tests/gas_test.rs
vlopes11: Fixed!

bkolad (approved):

PR #1225, opened 2023-12-15
feat: add associated types as tx hooks args

vlopes11:
This commit introduces associated types to the `TxHooks` trait, enhancing its ability to define custom interactions between the trait and its implementing modules. The associated types allow for greater flexibility in the implementation of `TxHooks`, enabling diverse use cases beyond the current `Accounts` module.

The current `Accounts` implementation of `TxHooks` serves the purpose of extracting a public key from a transaction and mapping it to a rollup address, creating an execution context. This functionality is achieved through the `pre_dispatch_tx_hook` signature which returns a `C::Address`. However, this tight coupling with the `accounts` module restricts future expansions of `TxHooks`.

With this modification, hooks gain increased versatility and modules are free to implement them as per their requirements. Additionally, runtimes can consume these implementations, utilizing them as essential building blocks, while the banks module is empowered to enforce a gas cap on the pre-dispatch hook.

### DIFF
--- a/examples/demo-rollup/stf/src/hooks_impl.rs
+++ b/examples/demo-rollup/stf/src/hooks_impl.rs
@@ -13,24 +13,34 @@ use crate::runtime::Runtime;
 
 impl<C: Context, Da: DaSpec> TxHooks for Runtime<C, Da> {
     type Context = C;
+    type PreArg = u64;
+    type PreResult = C;
+    type PostArg = ();
+    type PostResult = ();
 
     fn pre_dispatch_tx_hook(
         &self,
         tx: &Transaction<Self::Context>,
         working_set: &mut WorkingSet<C>,
-    ) -> anyhow::Result<<Self::Context as Spec>::Address> {
+        height: u64,
+    ) -> anyhow::Result<C> {
         // Before executing a transaction, retrieve the sender's address from the accounts module
         // and check the nonce
-        self.accounts.pre_dispatch_tx_hook(tx, working_set)
+        let sender_address = self.accounts.pre_dispatch_tx_hook(tx, working_set, ())?;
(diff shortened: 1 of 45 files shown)


examples/demo-rollup/stf/src/hooks_impl.rs
```diff
@@ -13,24 +13,34 @@ use crate::runtime::Runtime;
 
 impl<C: Context, Da: DaSpec> TxHooks for Runtime<C, Da> {
     type Context = C;
+    type PreArg = u64;
+    type PreResult = C;
```
bkolad, line 18: Let's construct Context outside the `pre_dispatch_tx_hook` as before. I will need to update the code and it will be easier this way.

vlopes11: Done!

examples/demo-rollup/stf/src/hooks_impl.rs
```diff
@@ -13,24 +13,31 @@ use crate::runtime::Runtime;
...
+    type PreResult = C::Address;
+    type PostArg = ();
+    type PostResult = ();
 
     fn pre_dispatch_tx_hook(
         &self,
         tx: &Transaction<Self::Context>,
         working_set: &mut WorkingSet<C>,
-    ) -> anyhow::Result<<Self::Context as Spec>::Address> {
+        _arg: (),
```
bkolad: Now we have unused parameters `()` lying around. 
And the `PostArg & PostResult` will always be `()`. What about going the other way we discussed and removing the `TxHooks` for `Accounts.`?

vlopes11: Removed these to simplify the api

bkolad (approved):

PR #1251, opened 2023-12-21
Move all DA layer information into Kernel. Prevent user-space access to Kernel State

preston-evans98:
# Description
This PR moves the chain-state module and related functionality into the kernel, and makes a handful of tweaks to ensure that the kernel values are inaccessible:
- Add a new `KernelModule` trait analagous to `Module`, but with no `CallMessage`
- Enforce that `#[derive(ModuleInfo)]` can only be called on structs that implement `Module` or `KernelModule`
- Add a new `#[derive(KernelModuleInfo)]` macro that  allows structs to use the `#[kernel_module]` attribute *in addition* to the existing attributes like `#[module]`, `#[state]`. This macro implements `ModuleInfo` for `KernelModules`.  Modules referenced by `#[kernel_module]` are required to implement the `KernelModule` trait. 

After this PR, the Kernel is now...
- Inaccessible to regular modules, without the need for complicated `KernelWorkingSet`s. The `KernelWorkingSet` and related functionality will be removed in a follow up PR.
- Fully responsible for chain-state management.
- Able to access DA information via `KernelSlotHooks`, which are invoked by the `StfBlueprint`
- Fully responsible to implement `BlobSelector`. The runtime no longer has this functionality.

Remaining work:
- [ ] Completely re-write the integration tests for chain-state (`module-system/module-implementations/integration-tests/src/chain_state/tests.rs` and `module-system/module-implementations/sov-chain-state/tests/all_tests.rs`)
- [ ] Completely re-write the integration tests for blob-selection
- [ ] Remove DA-related information from the normal SlotHooks
- [ ] Remove `KernelWorkingSet` and related types/functionality
- [ ] Implement kernel logic to manage the user-space view of DA layer information
- [ ] Simplify initialization logic for kernel and runtime
- [ ] Re-enable chain-state RPC 
- [ ] Document `Kernel` and it's responsibilities
- [ ] Document the method for soft-confirmations

## Linked Issues
- Related to #408 

## Testing
These changes are largely covered by existing tests. Improved integration tests will be added in a follow up.


module-system/sov-modules-macros/src/module_info.rs
```diff
@@ -208,11 +229,20 @@ fn make_init_state(
...
+) -> Result<proc_macro2::TokenStream, syn::Error> {
     let field_ident = &field.ident;
     let ty = &field.ty;
+    let trait_to_assert = match variant {
+        ModuleType::Standard => quote::quote! { ::sov_modules_api::Module },
+        ModuleType::Kernel => quote::quote! { ::sov_modules_api::KernelModule },
+    };
 
     Ok(quote::quote! {
+        // Ensure that the type implements "Module" or "KernelModule" at compile time
```
bkolad, line 244: Is it possible to check with try build if we get a suitable error message here?

bkolad (approved):

vlopes11 (commented):
Non-blocking; Consider implementing a "KernelContext" without the mutable reference to a "WorkingSet." Instead, make this context accessible under the "Context" umbrella. When initiating a kernel call, pass the desired WorkingSet to the context, which will subsequently forward it to the KernelContext. This design can also be scaled for new hook implementations.

The proposed KernelContext would solely store height data internally and be functional regarding working set state modifications (accepting its mutable state as an argument). By implementing this abstraction, module developers will gain the ability to invoke specific kernel functions while being confined by the safeguards built into the context itself. Consequently, we establish a clear separation between user space and kernel space.

module-system/sov-modules-macros/src/module_info.rs
preston-evans98: Yeah, that should be possible. I'll add this to the checklist 👍

preston-evans98:
>  Consider implementing a "KernelContext" without the mutable reference to a "WorkingSet." Instead, make this context accessible under the "Context" umbrella. When initiating a kernel call, pass the desired WorkingSet to the context, which will subsequently forward it to the KernelContext. This design can also be scaled for new hook implementations.

This is an interesting idea. It's a bit out of scope for this PR, but it might solve some problems in the next one where we try to remove KernelWorkingSet entirely.

vlopes11 (approved):

PR #1270, opened 2024-01-02
Fixes for Rust 1.75 release

citizen-stig:
# Description
Clippy fixes for Rust 1.75 and macro stderr update.

Publish it as separate PR, so currently active PRs have less conflicts.

Errors fixed:

* `warning: items after a test module`
* `warning: unused import` for some `hooks::*` and `call::*`. When there are no public types of functions, but just impl blocks.
* `first()` and `.front()` instead of `get(0)`
* `warning: use of a fallible conversion when an infallible one could be used`

### DIFF
--- a/full-node/sov-ethereum/src/gas_price/gas_oracle.rs
+++ b/full-node/sov-ethereum/src/gas_price/gas_oracle.rs
@@ -3,8 +3,6 @@
 
 // Adopted from: https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc/src/eth/gas_oracle.rs
 
-use std::array::TryFromSliceError;
-
 use reth_primitives::constants::GWEI_TO_WEI;
 use reth_primitives::{H256, U256, U64};
 use reth_rpc_types::BlockTransactions;
@@ -141,7 +139,7 @@ impl<C: sov_modules_api::Context> GasPriceOracle<C> {
         let mut results = Vec::new();
         let mut populated_blocks = 0;
 
-        let header_number = convert_u256_to_u64(header.number.unwrap()).unwrap();
+        let header_number = convert_u256_to_u64(header.number.unwrap());
 
         // we only check a maximum of 2 * max_block_history, or the number of blocks in the chain
         let max_blocks = if self.oracle_config.max_block_history * 2 > header_number {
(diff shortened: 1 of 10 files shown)


module-system/sov-modules-api/src/pub_key_hex.rs
```diff
@@ -80,6 +80,14 @@ impl TryFrom<&PublicKeyHex> for DefaultPublicKey {
     }
 }
 
+#[cfg(feature = "arbitrary")]
+impl<'a> arbitrary::Arbitrary<'a> for PublicKeyHex {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
```
vlopes11, line 86: Since this wasn't introduced on this PR, I'll leave it as optional.

Shouldn't we generate a secret key, then public, then finally its string representation? If we end with random strings, they will almost never encode to a key that can be used

module-system/module-implementations/sov-chain-state/src/lib.rs
```diff
@@ -3,7 +3,6 @@
 
 /// Contains the call methods used by the module
 mod call;
-pub use call::*;
```
vlopes11, line 6: I think the call implementations of the module will be required downstream. Any particular reason we are removing this re-export? Note that the `call` module is private.

module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -9,7 +9,7 @@ mod genesis;
 #[cfg(feature = "experimental")]
 mod hooks;
 #[cfg(feature = "experimental")]
-pub use {call::*, error::rpc::*, evm::*, genesis::*, hooks::*};
```
vlopes11, line 12: Ditto; why remove these exports? Hooks are also meant to be consumed downstream

full-node/sov-ethereum/src/gas_price/gas_oracle.rs
```diff
@@ -303,6 +303,12 @@ fn effective_gas_tip(
     }
 }
 
+fn convert_u256_to_u64(u256: U256) -> Result<u64, TryFromSliceError> {
+    let bytes: [u8; 32] = u256.to_be_bytes();
+    let bytes: [u8; 8] = bytes[24..].try_into()?;
```
vlopes11: Unrelated to this PR, but this looks infallible. If the array is `32` bytes, then it will always have `24 + 8` bytes, making this conversion check unnecessary.

vlopes11 (commented):
Initial round to be followed by approval

module-system/module-implementations/sov-chain-state/src/lib.rs
citizen-stig: Call module in this crate only has `impl` block and does not export any public types or standalone functions.
impl block is "attached", so this statement does not export anything. If my understanding is correct.

module-system/module-implementations/sov-evm/src/lib.rs
citizen-stig: Here is the same situation as we `pub use call::*`, only impl block inside hooks module.

I've also added this to PR description.

vlopes11 (approved):
LGTM; two optional, unrelated topics

full-node/sov-ethereum/src/gas_price/gas_oracle.rs
citizen-stig: Fixed + added proptest

module-system/sov-modules-api/src/pub_key_hex.rs
citizen-stig: Agree. 

And if we want to test "ArbitraryString" as PublicKeyHex, we will do exactly that: generate string manually and contruct PubliKeyHex from it

PR #1276, opened 2024-01-03
feat: add gas total to tx receipt

vlopes11:
This commit introduces the "gas total" as a new attribute in the transaction receipt. With this addition, the gas price elasticity can be calculated based on the demand of previous blocks.

This feature also allows setting a base fee for every transaction, which is paid by the sender to cover processing costs such as deserialization and signature verification. Insufficient funds from the sender will cause the fee to be charged to the sequencer, leading to a failed transaction and thwarting potential attacks. The sequencer should have sufficient funds available from their staked amount to cover these expenses; the implementation of the sequencer registry is responsible for ensuring the sequencer possesses adequate funds for selecting transactions to assemble into a blob.

To define the base cost constant, an adjustment was made to make this feature available under the `sov-modules-macros`. Due to the need for defining slices of unknown size for the constants file, this functionality had to be provided there as well.

### DIFF
--- a/module-system/module-implementations/sov-bank/src/hooks.rs
+++ b/module-system/module-implementations/sov-bank/src/hooks.rs
@@ -21,12 +21,6 @@ use crate::{Bank, Coins};
 // https://github.com/Sovereign-Labs/sovereign-sdk/issues/1234
 const GAS_TOKEN_ADDRESS: &'static str;
 
-#[config_constant]
-const GAS_TX_FIXED_COST: &[u64];
-
-#[config_constant]
-const GAS_TX_COST_PER_BYTE: &[u64];
-
 /// The computed addresses of a pre-dispatch tx hook.
 pub struct BankTxHook<C: Context> {
     /// The tx sender address
--- a/module-system/module-implementations/sov-bank/src/hooks.rs
+++ b/module-system/module-implementations/sov-bank/src/hooks.rs
@@ -3,7 +3,7 @@ use core::str::FromStr;
 use sov_modules_api::hooks::TxHooks;
 use sov_modules_api::macros::config_constant;
 use sov_modules_api::transaction::Transaction;
-use sov_modules_api::{Context, WorkingSet};
+use sov_modules_api::{Context, GasUnit, WorkingSet};
 
 use crate::{Bank, Coins};
 
@@ -20,6 +20,12 @@ use crate::{Bank, Coins};
 // TODO: fetch address as constant
 // https://github.com/Sovereign-Labs/sovereign-sdk/issues/1234
 const GAS_TOKEN_ADDRESS: &'static str;
+
+#[config_constant]
+const GAS_TX_FIXED_COST: &[u64];
+
+#[config_constant]
+const GAS_TX_COST_PER_BYTE: &[u64];
 
 /// The computed addresses of a pre-dispatch tx hook.
 pub struct BankTxHook<C: Context> {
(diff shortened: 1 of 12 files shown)


module-system/module-implementations/sov-bank/src/hooks.rs
```diff
@@ -21,6 +21,12 @@ use crate::{Bank, Coins};
 // https://github.com/Sovereign-Labs/sovereign-sdk/issues/1234
 const GAS_TOKEN_ADDRESS: &'static str;
 
+#[config_constant]
+const GAS_TX_FIXED_COST: &[u64];
+
+#[config_constant]
+const GAS_TX_COST_PER_BYTE: &[u64];
```
preston-evans98: It looks like these constants are declared twice? (Once here, and once in transaction.rs). Is that intentional?

vlopes11: Unintentional! Was removed a couple of seconds before the comment

module-system/sov-modules-core/src/common/gas.rs
```diff
@@ -80,9 +108,21 @@ where
         self.remaining_funds
     }
 
+    /// Returns the total gas incurred.
+    pub const fn gas_total(&self) -> &GU {
```
preston-evans98: nit: The name "gas_used" might be slightly more clear. WDYT?

module-system/sov-modules-core/src/common/gas.rs
```diff
@@ -34,12 +43,28 @@ impl<const N: usize> GasUnit for TupleGasUnit<N> {
...
         self.iter()
             .zip(price.iter().copied())
             .map(|(a, b)| a.saturating_mul(b))
             .fold(0, |a, b| a.saturating_add(b))
     }
+
+    fn combine(&mut self, rhs: &Self) {
+        for i in 0..N {
+            self[i] = self[i].saturating_add(rhs[i]);
+        }
```
preston-evans98: nit: This might insert unneeded bounds checks, and I think it's slight more idiomatic to use an iterator anyway. Is there any problem with doing that?

module-system/sov-modules-core/src/common/gas.rs
```diff
@@ -34,12 +43,28 @@ impl<const N: usize> GasUnit for TupleGasUnit<N> {
...
+    }
+
+    fn scalar_product(&mut self, scalar: u64) {
+        for i in 0..N {
+            self[i] = self[i].saturating_mul(scalar);
+        }
```
preston-evans98: nit: Same comment re: iterators

module-system/sov-modules-core/src/common/gas.rs
vlopes11: Makes sense; fixed!

module-system/sov-modules-core/src/common/gas.rs
vlopes11: It's mostly a case of style since we use a const `N`, so the compiler will end up optimizing both versions to the maximum. But indeed, we use iterators elsewhere on the same source, so it's best to use them here to keep consistency. Fixed!

module-system/sov-modules-core/src/common/gas.rs
vlopes11: Fixed above!

bkolad (approved):

PR #1279, opened 2024-01-04
Add prover_bench to CI to catch potential breakages resulting from format/serialization changes

dubbelosix:
# Description

Adding the prover bench to CI to ensure bugs in it are caught. 
We've had a couple of issues in the past where serialization changes etc caused the benchmark to break and this has only been caught when prover_bench was manually run. To avoid that, we want to run the bench in CI as well.
prover_bench.rs runs without actually proving i.e. just performs the riscv compilation and execution of the ELF to capture the cycles (without actually proving the riscv instructions) so it should be relatively fast. As of recent test it was 6-8 minutes, and since it runs in parallel with cargo hack, it has room to get 4-5x worse before it blocks CI.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -167,7 +167,11 @@ jobs:
           version: "23.2"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Install Rust
-        run: rustup show
+        run: rustup show && rustup install nightly && rustup component add rustfmt --toolchain nightly-x86_64-unknown-linux-gnu
+      - name: Install cargo-risczero
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-risczero@0.19
       - uses: Swatinem/rust-cache@v2
         with:
           cache-provider: "buildjet"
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -152,6 +152,41 @@ jobs:
...
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-provider: "buildjet"
+          shared-key: cargo-check-cache
+          save-if: ${{ github.ref == 'refs/heads/nightly' }}
+          workspaces: |
+            .
+            fuzz
+      - name: cargo bench check
+        run: cargo bench --bench prover_bench --features bench
```
citizen-stig: Have you tried to run all benches that we have? How much time would it take?

dubbelosix: that's a good question - prover_bench is unique though because it doesn't really use a bench framework like glassbench or criterion, it mainly uses bench for the build profile and dev-dependencies, but prover_bench.rs is just a regular binary that performs the benchmarks in its own way (counting instruction cycles). 
but i will give the rest of them a try too

dubbelosix: `cargo bench` at root takes 6.5 mins on local. i'll try adding it as another parallel job in CI

theochap:
Lgtm

theochap (approved):

.github/workflows/rust.yml
dubbelosix: prover_bench_check is for the prover cycles and bench_check is for the root folder level `cargo bench`
still lesser than cargo hack.
wdyt @citizen-stig ? should we keep regular bench as well ?
Screen Shot 2024-01-04 at 10 41 44 PM

citizen-stig (approved):

PR #1296, opened 2024-01-10
Add shell wrappers for checking prover and rollup benchmark thresholds

dubbelosix:
## Description
`prover_bench` is already being checked for sanity. This PR adds `rollup_bench` as well - specifically rollup_coarse_measure.rs which runs a small workload to get the TPS. This PR adds 2 shell scripts which cover the sanity check (to catch script breakage) as well as checking relaxed thresholds to catch issues such as accelerated patches not being applied. 
* `ci_prover_count_check.sh`: this script checks for the number of riscv cycles taken by the `verify` function and exits with code 1 if the number cycles are > 1,000,000. The number of cycles with acceleration should be  700k or better
```
+----------------------+----------------+-----------+
| Function             | Average Cycles | Num Calls |
+----------------------+----------------+-----------+
| Cycles per block     | 8101853        | 2         |
+----------------------+----------------+-----------+
| apply_blob           | 4986497        | 2         |
+----------------------+----------------+-----------+
| pre_process_batch    | 4465276        | 2         |
+----------------------+----------------+-----------+
| verify_txs_stateless | 4423303        | 2         |
+----------------------+----------------+-----------+
| end_slot             | 2056091        | 2         |
+----------------------+----------------+-----------+
| jmt_verify_update    | 1159849        | 2         |
+----------------------+----------------+-----------+
| jmt_verify_existence | 817044         | 2         |
+----------------------+----------------+-----------+
| verify               | 745055         | 11        |
+----------------------+----------------+-----------+
| apply_txs            | 487083         | 2         |
+----------------------+----------------+-----------+
| begin_slot           | 89594          | 2         |
+----------------------+----------------+-----------+
| deserialize          | 28928          | 11        |
+----------------------+----------------+-----------+
| decode_txs           | 27021          | 2         |
+----------------------+----------------+-----------+
| deserialize_batch    | 8720           | 2         |
+----------------------+----------------+-----------+
| hash                 | 6026           | 11        |
+----------------------+----------------+-----------+
| commit               | 7              | 2         |
+----------------------+----------------+-----------+

Total cycles consumed for test: 16203706

The value for 'verify' is less than 1000000. Passing the check. Value: 745055
```

* `ci_rollup_count_check.sh`: this script checks for the TPS count for a small workload.. 10 blocks, 10k transactions per block. if the TPS < 1000 it fails
```
+----------------------------------------+----------------------+
| Blocks                                 | 10                   |
+----------------------------------------+----------------------+
| Transactions per block                 | 10000                |
+----------------------------------------+----------------------+
| Processed transactions (success/total) | 100000/100000        |
+----------------------------------------+----------------------+
| Total                                  | 10s 153ms 198us 6ns  |
+----------------------------------------+----------------------+
| Apply block                            | 9s 417ms 695us 790ns |
+----------------------------------------+----------------------+
| Transactions per sec (TPS)             | 9849.1               |
+----------------------------------------+----------------------+
The value for TPS is greater than 1000. Passing the check. Value:  9849.1 
```

### DIFF
--- a/examples/demo-rollup/benches/node/rollup_config.toml
+++ b/examples/demo-rollup/benches/node/rollup_config.toml
@@ -21,3 +21,6 @@ start_height = 1
 # the host and port to bind the rpc server for
 bind_host = "127.0.0.1"
 bind_port = 12345
+
+[prover_service]
+aggregated_proof_block_jump = 0
--- a/examples/demo-rollup/benches/node/rollup_config.toml
+++ b/examples/demo-rollup/benches/node/rollup_config.toml
@@ -23,4 +23,4 @@ bind_host = "127.0.0.1"
 bind_port = 12345
 
 [prover_service]
-aggregated_proof_block_jump = 0
+aggregated_proof_block_jump = 1
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -188,7 +188,7 @@ jobs:
           BLOCKS: 1
           TXNS_PER_BLOCK: 10
           NUM_PUB_KEYS: 100
-        run: cargo bench --bench prover_bench --features bench
+        run: sh ci_prover_count_check.sh
 
   bench_check:
     name: bench_check
@@ -264,7 +264,7 @@ jobs:
           BLOCKS: 10
           TXNS_PER_BLOCK: 10000
           NUM_PUB_KEYS: 100
-        run: cd examples/demo-rollup make basic
+        run: sh ci_rollup_count_check.sh
 
   # Check that every combination of features is working properly.
   hack:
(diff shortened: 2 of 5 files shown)


examples/demo-rollup/benches/node/rollup_config.toml
```diff
@@ -21,3 +21,6 @@ start_height = 1
 # the host and port to bind the rpc server for
 bind_host = "127.0.0.1"
 bind_port = 12345
+
+[prover_service]
+aggregated_proof_block_jump = 0
```
bkolad: I think we should set it to 1 in the bench

bkolad (approved):

PR #1430, opened 2025-07-29
Switch to cacheable DB backend for NOMT

preston-evans98:
# Description

This PR drastically improves NOMT performance by switching to a version of Rockbound with an application-aware cache. It also improves our pruning logic for NOMT historical values, removing the need to scan the entire `historical_values` table for each run of the pruner. 

Along the way, this PR also fixes two issues: 
1. State acccessors were allowed to query for archival *accessory* state during tx execution
2. Queries for pruned archival data would often incorrectly return `None` when they should return an error instead

Fixing these two issues required minor adjustments to the ApiStateAccessor and the `Storage/NativeStorage` traits.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-db/src/historical_state.rs
```diff
@@ -1,70 +1,134 @@
...
     /// The [`SlotNumber`] that will be used for the next batch of writes to the DB
     /// This [`SlotNumber`] is also used for querying data,
     /// so if this instance of [`HistoricalStateReader`] is used as read-only, it won't see newer data.
     next_version: SlotNumber,
 }
 
+/// A collection of changes to the state db. Includes versioned changes to user/kernel state, and a plain schema batch of changes to any other columns.
+#[derive(Debug, Clone, Default)]
+pub struct StateChanges {
+    pub(crate) user: Arc<VersionedSchemaBatch<NomtStateValues<UserNamespace>>>,
```
citizen-stig, line 36: Why do we need 3 Arcs here, instead of 1 around whole `StateChanges` ?

crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
```diff
@@ -295,7 +296,7 @@ where
         let state_overlay = state.into_state_overlay();
 
         let rockbound_snapshot = SnapshotGroup {
-            historical_state: Arc::new(historical_state),
+            historical_state,
```
citizen-stig, line 299: Maybe I am missing something, but why not to have Arcs to be consistent with other change sets, related to previous comment https://github.com/Sovereign-Labs/sovereign-sdk/pull/1430/files#r2253966131

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -5,23 +5,123 @@ use std::thread::JoinHandle;
...
+/// A database to store the flat state of the rollup (i.e. the raw key-value pairs)
+pub struct FlatStateDb {
+    user: VersionedDB<NomtStateValues<UserNamespace>>,
+    kernel: VersionedDB<NomtStateValues<KernelNamespace>>,
+    other: Arc<rockbound::DB>,
+}
+
+impl FlatStateDb {
+    const DB_NAME: &'static str = "state";
+    const DB_PATH_SUFFIX: &'static str = "state";
```
citizen-stig: Nit for consistency, suffix usually goes with `-db`

```
    const DB_PATH_SUFFIX: &'static str = "accessory";
    const DB_NAME: &'static str = "accessory-db";
```

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -5,23 +5,123 @@ use std::thread::JoinHandle;
...
+use crate::schema::namespace::{
+    NomtCommittedVersion, NomtHistoricalState, NomtPruningState, NomtStateValues,
+};
+use crate::schema::tables::{ModuleAccessoryState, StateRootHashes};
 use crate::state_db_nomt::{NomtSessionBuilder, NomtStateDb, StateOverlay};
 use crate::storage_manager::{update_ledger_finalized_height, InitializableNativeNomtStorage};
+use crate::DbOptions;
+
+/// A database to store the flat state of the rollup (i.e. the raw key-value pairs)
+pub struct FlatStateDb {
```
citizen-stig: Nit: Could we place this struct into top of `sov-db` crate, so it is similar to what is already there, plus `storage_manger` module contains only storage manager related code

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -5,23 +5,123 @@ use std::thread::JoinHandle;
...
+        let user = VersionedDB::<NomtStateValues<UserNamespace>>::from_db(other.clone())?;
+        let kernel = VersionedDB::<NomtStateValues<KernelNamespace>>::from_db(other.clone())?;
+        Ok(Self {
+            user,
+            kernel,
+            other,
+        })
+    }
+
+    #[allow(dead_code)]
```
citizen-stig: Why is it dead_code? If it is for test, I usually put explicit `cfg(test)`

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -5,23 +5,123 @@ use std::thread::JoinHandle;
...
+        self.kernel
+            .materialize(&kernel, &mut other_changes, version)?;
+
+        Ok(other_changes)
+    }
+
+    /// Coalesce all the changes into a single schema batch and write it atomically.
+    pub fn commit(&self, state: StateChanges) -> anyhow::Result<()> {
+        let commit = self.prepare_commit(state)?;
+        self.other.write_schemas(commit)?;
```
citizen-stig: just to clarify, all the data is actually stored in `.other` rocksdb, right? And we write everything atomically here?

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -61,25 +160,27 @@ where
...
-        self.state.commit(state)?;
-        self.accessory.write_schemas(&accessory)?;
+        self.merklized_state.commit(state)?;
+        self.accessory
+            .write_schemas(Arc::unwrap_or_clone(accessory))?;
         // Ledger goes last, as its data is used during the start.
         // So if ledger save failed, state and accessory will be synced from DA
-        self.ledger.write_schemas(&ledger)?;
+        self.ledger.write_schemas(Arc::unwrap_or_clone(ledger))?;
         // Historical data is committed the last, as in case of failure, it can be synced from the normal state,
```
citizen-stig: It looks like comments went ouf ot date.

crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
```diff
@@ -242,7 +259,7 @@ impl<S: Spec> EventContainer for ApiStateAccessor<S> {
 #[derive(Debug, thiserror::Error, Clone, PartialEq, Eq)]
 pub enum ApiStateAccessorError {
     /// The requested height is not accessible.
-    #[error("Impossible to get the rollup state at the specified height. Please ensure you have queried the correct height.")]
+    #[error("Impossible to get the rollup state at the specified height. The requested height may have been pruned, or it may be in the future. Please ensure you have queried the correct height.")]
```
citizen-stig, line 262: Nit: it is out of scope of this PR, so only if you have bandwith. But it would be great to have height as param of this error and return it. It might give us a little bit more signal in case of debugging.

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -82,22 +81,18 @@ where
...
             return None;
         }
         let next_version = self.historical_state.get_next_version();
         match version {
-            None => Some(
-                next_version
-                    .checked_sub(1)
-                    .expect("Next version for non empty storage should be above 0"),
-            ),
+            None => Some(None),
```
citizen-stig, line 93: yammie. Could we please add a comment in which cases we get Some(Some(n), and in which `Some(None)` and how it is different from plain `None`.

crates/full-node/sov-db/src/historical_state.rs
```diff
@@ -1,70 +1,134 @@
...
     const DB_NAME: &'static str = "historical-state-db";
 
-    /// Create a new instance of [`HistoricalStateReader`] from a given [`DeltaReader`].
-    pub fn with_delta_reader(reader: DeltaReader) -> anyhow::Result<Self> {
-        let next_version = Self::next_version_from(&reader)?;
-        tracing::trace!(?next_version, "Initialized historical state reader");
-        Ok(Self {
-            db: reader,
+    // Used for testing only.
+    #[allow(dead_code)]
```
citizen-stig: If it is only for testing, let's make it `cfg(test)` or `cfg(feature = "test-utils")

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -5,23 +5,123 @@ use std::thread::JoinHandle;
...
+    }
+
+    #[allow(dead_code)]
+    /// Get the underlying [`rockbound::DB`] for the historical state. Used for testing only.
+    pub fn get_db(&self) -> Arc<rockbound::DB> {
+        self.other.clone()
+    }
+
+    /// Get the underlying [`VersionedDB`] for the user state.
+    pub fn get_user_db(&self) -> &VersionedDB<NomtStateValues<UserNamespace>> {
```
citizen-stig: Do we want to keep these methods completely public instead of pub(crate)?

Asking because it is often gets abused and it is hard to backtrack on that.

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -30,19 +130,18 @@ pub(crate) struct DbGroup<H, K> {
 impl<H, K> DbGroup<H, K>
 where
     H: digest::Digest<OutputSize = digest::typenum::U32> + Send + Sync,
-    K: Eq + std::hash::Hash + Clone,
+    K: Eq + std::hash::Hash + Clone + std::fmt::Debug,
```
citizen-stig, line 38: Nit: Debug bound might be leftover from debugging, but I am fine to keep it.

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -384,21 +397,41 @@ where
...
+            Err(e) => {
+                // Historical errors are not allowed when fetching without a version
+                panic!("Database error while getting value for key {key}. error: {e:?}");
+            }
+        }
     }
 
-    fn get_accessory(&self, key: &SlotKey, version: Option<SlotNumber>) -> Option<SlotValue> {
-        self.read_value::<Accessory>(key, version)
+    // TODO: Split accessory get versioned into separate function
```
citizen-stig: Is this TODO still relevant?

citizen-stig (changes requested):
Great job, looking forward to see this merged.

My main question is around storage manager change sets handling. Happy to discuss.

crates/full-node/sov-db/src/historical_state.rs
preston-evans98: This is because we often want to access one item individually. For example, we need `other` to create a delta reader, but not user/kernel. Similarly, we need user or kernel without `other` in some instances.

preston-evans98: I didn't fully appreciate that at the time of the previous discussion, but having attempted the change it's much messier with a single Arc than with 3

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
preston-evans98: Fixed!

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
preston-evans98: Yep

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
preston-evans98: This wasn't changed in the PR, but now you've got me thinking - shouldn't the flat state be committed before the ledger (per this comment)?

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
preston-evans98: These are needed by `sov-test-utils` as well. Do you see a problem with them being public? Getting access to the individual db seemed like a reasonable operation to be `pub` to me, but I'm happy to hear pushabck if you have it

crates/module-system/sov-state/src/nomt/prover_storage.rs
preston-evans98: Nope, that's been done now! Removed

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
preston-evans98: Done

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
citizen-stig: I don't have strong pushback on this one. Wanted to make sure it is a deliberate decision

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
citizen-stig: I think so. Ledger is used to detech where to begin on startup.

I think historical DB being the last is an artifact of the time, when actual data has been written to NOMT.

preston-evans98: Fixed.

crates/module-system/sov-state/src/nomt/prover_storage.rs
preston-evans98: Done!

citizen-stig (approved):

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
preston-evans98: Leaving it since it's useful

crates/full-node/sov-db/src/historical_state.rs
preston-evans98: Fixed

crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
preston-evans98: Leaving this as is because it'll break consistency; we return this error in a few places where the height isn't easily accessible, and we also match on the error message in tests. Not too hard to fix, but I'd rather not make this PR any bigger :)

PR #1431, opened 2025-07-29
Theodore/sequencer failover squashed for new repo

theodorebugnet:
# Description

_Draft PR until it's properly rebased with commits fixed._

Adds failover using a postgres mechanism. Each instance generates a UUID on startup and tries to set itself as leader in the dedicated table for this; this succeeds only if the current leader's last heartbeat is old enough. The leader then periodically sends heartbeats, while all replicas periodically try to take over.

Notes on the implementation:
 * The database backend now returns a nullable result on all write operations; the sequencer's DB uses this to identify if the operation failed due to no longer being master, aborting the current operation (e.g. batch starting/ending, accept_tx, etc.). For simplicity though, we don't do anything else when this is detected - this is only to guard for race conditions until we receive the pg_notify event of the master node_id having changed (and transition to replica operation).
 * The timing parameters for failover are hardcoded for now; I've added a todo in #3100 to expose them as config values later.
 * Because a replica might succeed in taking over before its state is fully synced, the `ReplicationTask` has three state: replica, master, and transitioning to master. While transitioning, the sequencer will still report `is_master = false` in the API for now (there's a todo for this in the issue above).

## Linked Issues
- Related to #3100 

## Testing

## Docs


theodorebugnet:

PR #1440, opened 2025-07-30
Fix dead-lock in `NomtProverStorage::compute_state_update`

citizen-stig:
# Description

## Root cause

Nomt can commit data only if there are no live sessions. But in rare cases if NomtStorage manager tries to commit in very bad time, it will dead lock. Code explanation below.

```rust

pub struct NomtStorageManager<Da: DaSpec, H, S: InitializableNativeNomtStorage<H, Da::SlotHash>> {
    nomt_snapshots: Arc<RwLock<HashMap<Da::SlotHash, StateOverlay>>>,
    // other things
}

pub struct NomtSessionBuilder<H, K> {
    state_db: Arc<NomtStateDb<H>>,
    // In revered chronological order, as [`nomt::SessionParams::overlay`] expects
    relevant_snapshot_refs: Vec<K>,
    // Reference to snapshots for all blocks currently available in a storage manager.
    // Used to get reference to state overlays and build session.
    // Same thing from above
    all_snapshots: Arc<RwLock<HashMap<K, StateOverlay>>>,
}

impl NomtProverStorage {

    fn compute_state_update(&self, ...) {
        // Takes and releases the read lock
        let user_session = self.state_session_builder.begin_user_session()?;
        // Here storage manager takes write lock, and tries to commit several times,
        // but not succeeeds, because user_session is alive and compute_state_update never complietes, because line below waits for acquiring read lock
        let kernel_session = self.state_session_builder.begin_kernel_session()?;
        
    }
}
```

## Solution

Add new method, which starts both sessions from same ReadLock, so both sessions can finish.

## Testing
Existing tests are passing

## Docs
No updates

### DIFF
--- a/crates/module-system/sov-state/src/nomt/prover_storage.rs
+++ b/crates/module-system/sov-state/src/nomt/prover_storage.rs
@@ -601,14 +601,10 @@ where
         };
         let storage_root_historical = self.get_root_hash_unbound(version_to_use)?;
         if self.should_check_dbs_sync(version_to_use) {
-            let SessionsContainer {
-                user: user_session,
-                kernel: kernel_session,
-            } = self.state_session_builder.begin_both_sessions()?;
-            let user_root = user_session.prev_root();
-            let kernel_root = kernel_session.prev_root();
-            drop(user_session);
-            drop(kernel_session);
+            let session_container = self.state_session_builder.begin_both_sessions()?;
+            let user_root = session_container.user.prev_root();
+            let kernel_root = session_container.kernel.prev_root();
+            drop(session_container);
             let prev_root_nomt = StorageRoot::new(user_root.into_inner(), kernel_root.into_inner());
             assert_eq!(
                 storage_root_historical, prev_root_nomt,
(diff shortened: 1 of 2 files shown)


crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -599,11 +601,13 @@ where
...
+            let SessionsContainer {
+                user: user_session,
+                kernel: kernel_session,
+            } = self.state_session_builder.begin_both_sessions()?;
             let user_root = user_session.prev_root();
-            drop(user_session);
-            let kernel_session = self.state_session_builder.begin_kernel_session()?;
             let kernel_root = kernel_session.prev_root();
+            drop(user_session);
             drop(kernel_session);
```
bkolad: What if we don’t desugar user_session and simply call `drop(user_session)` instead of dropping the two sessions separately?

bkolad (approved):

crates/module-system/sov-state/src/nomt/prover_storage.rs
citizen-stig: Good point!

PR #1447, opened 2025-07-30
Fix resync edge case + fix replica initial sync

theodorebugnet:
# Description

This PR bundles three separate fixes:
 * Unrelated to replication: there was an edge case involving the transaction cache. During resync, `inner.process_wait_for_node_resync()` flushes the tx cache for every update_state. Once the sequencer is considered resynced, say at update state for slot `n`, the tx cache will have `next_tx_number` from slot `n`. When the next StateUpdateInfo for slot `n+1` comes in and goes into the replay_batches branch, the sequencer starts replaying pending batches starting with `n+2`. This means that transactions from slot `n+1` will not have been added to the tx cache - and it will error on when sanity checking for consecutive tx numbers.
     * This has been fixed by having `ReplaySoftConfirmationsOnTopOfNodeState` contain `is_startup_or_resync` rather than just `is_startup`, i.e. the first update_state after a successful resync will still have this as `true` (and then will set `inner.is_ready = Ok(())` so will be false henceforth). This is then special cased to flush the tx cache one again.
 * Replication: backfill used to run concurrently to the event pre-fetching futures. This was incorrect: it should run in order with `processed_events`. `create_backfill_future()` has been removed, and the backfill future has been replaced with one that resolves immediately; the actual backfill call has been moved to the ordered event processing loop. (This is backported from the failover PR, but was necessary to allow replicas to resync.)
 * Replication resync: replicas have access to the sequencer DB, but until their node is resynced, they cannot actually apply the pending batches because none of the events in the DB can be applied onto their executor's state. There was no proper functionality to wait for resync. This PR adds a notification channel from update_state which the replica_sync_task uses to a) avoid processing events until the executor is sufficiently caught up, and b) fast-forward event processing in case update_state actually fast-forwards the executor state beyond the current event cursor.
     * There is a known edge case remaining: if an event _starts_ being processed, but before it sends the message to `inner`, update_state finishes and replaces the executor with a future state - then there is no check for this, the past event will be re-applied to the state. This will cause the replica to crash with no other ill effects (`batch` events will cause a visible slot number mismatch, and `transaction` events will cause result in a duplicate transaction panic), and should be rare enough that restarting the replica should be sufficient mitigation for now.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/inner.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/inner.rs
@@ -7,10 +7,10 @@ use std::time::Duration;
 use anyhow::anyhow;
 use sov_blob_sender::BlobInternalId;
 use sov_blob_storage::SequenceNumber;
-use sov_modules_api::capabilities::{BlobSelector, RollupHeight};
+use sov_modules_api::capabilities::RollupHeight;
 use sov_modules_api::{
-    DaSyncState, FullyBakedTx, GasArray, GasSpec, KernelStateAccessor, Runtime, Spec,
-    StateCheckpoint, StateUpdateInfo, VersionReader, VisibleSlotNumber,
+    DaSyncState, FullyBakedTx, GasArray, GasSpec, Runtime, Spec, StateCheckpoint, StateUpdateInfo,
+    VersionReader, VisibleSlotNumber,
 };
 use sov_state::{NativeStorage, Storage};
 use tokio::sync::{mpsc, oneshot, watch};
@@ -57,12 +57,6 @@ const CHANNEL_SIZE: usize = 128;
 type AcceptTxRet<S, Rt> =
     Result<oneshot::Receiver<AcceptedTx<Confirmation<S, Rt>>>, AcceptTxError<S>>;
 
-/// The `Inner` sends the latest processed slot number when it finishes the starting sync.
-pub enum HasFinishedStartup {
-    No(watch::Sender<SequenceNumber>),
-    Yes,
-}
-
 /// A inner sequencer struct containing state that requires synchronized access.
 /// This struct accepts/rejects transactions, then hands them to the side effects task
 /// to be persisted.
@@ -86,6 +80,9 @@ where
     /// We need this rather than relying on `SequencerNotReadyDetails::Startup` because that state
     /// can be overwritten when the node is resyncing.
     has_finished_startup: bool,
+    /// Channel that sends a notification every time the executor is replaced after batch replay
+    /// finishes (at the end of update_state).
+    replay_notifier: watch::Sender<Option<SequenceNumber>>,
     metrics: Vec<PreferredSequencerChannelMetrics>,
     // Shared between sequencer and Inner.
     tx_queue_id: Arc<AtomicU64>,
@@ -780,7 +777,7 @@ pub(crate) fn create<S, Rt>(
     executor_events_sender: ExecutorEventsSender<S, Rt>,
     sequence_number_of_next_blob: SequenceNumber,
     in_flight_blobs: Arc<AtomicUsize>,
-    startup_notifier: oneshot::Sender<SequenceNumber>,
+    replay_notifier: watch::Sender<Option<SequenceNumber>>,
     stop_at_rollup_height: Option<RollupHeight>,
 ) -> (
     SynchronizedSequencerState<S, Rt>,
@@ -810,7 +807,8 @@ where
         executor_events_sender,
         sequence_number_of_next_blob,
         in_flight_blobs,
-        has_finished_startup: HasFinishedStartup::No(startup_notifier),
+        has_finished_startup: false,
+        replay_notifier,
         metrics: Vec::with_capacity(128),
         is_ready,
         stop_at_rollup_height,
@@ -1338,7 +1336,7 @@ where
         let ((batches_to_replay, fetch_batches_to_replay_metrics), is_startup) = {
             (
                 inner.completed_batches_to_replay(next_sequence_number_according_to_node, true),
-                matches!(inner.has_finished_startup, HasFinishedStartup::No(_)),
+                !inner.has_finished_startup,
             )
         };
         let is_resync = matches!(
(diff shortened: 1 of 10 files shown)


crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -1334,6 +1339,11 @@ where
                 !inner.has_finished_startup,
             )
         };
+        let is_resync = matches!(
```
theodorebugnet, line 1342: `is_resync` is used to fix the global resync edge case (issue 1)

crates/full-node/sov-sequencer/src/preferred/replica_sync_task.rs
```diff
@@ -21,8 +22,9 @@ use crate::preferred::{exit_rollup, DbEvent, PreferredSequencer};
 use crate::ProofBlobSender;
 
 /// Event type enum for type-safe parsing
-#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq)]
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq, sqlx::Type)]
```
theodorebugnet, line 25: This is a small bugfix, also separate maybe, backfill used to break because `sqlx` couldn't parse postgres `event_type` as string. So we derive sqlx::Type on EventType directly

crates/full-node/sov-sequencer/src/preferred/replica_sync_task.rs
```diff
@@ -100,48 +102,35 @@ type PendingEventFuture = Pin<Box<dyn Future<Output = anyhow::Result<CompletedEv
 #[derive(Debug)]
 enum CompletedEvent {
     Event(DbEvent),
-    Backfill,
+    Backfill {
```
theodorebugnet, line 105: This is part of the backfill bugfix (issue 2)

crates/full-node/sov-sequencer/src/preferred/replica_sync_task.rs
```diff
@@ -251,18 +240,47 @@ where
...
                         // Check for gaps and add backfill if needed
-                        if detect_gap(*latest_received_event_id, parsed_notification.event_id).is_some() {
-                            let backfill_future = create_backfill_future(
-                                sequencer.clone(),
-                                query_pool,
-                                *latest_received_event_id,
-                                parsed_notification.event_id,
-                            );
+                        if detect_gap(latest_received_event_id, parsed_notification.event_id).is_some() {
+                            let backfill_future: PendingEventFuture = Box::pin(async move {
```
theodorebugnet, line 271: The backfill_future change is for the backfill bugfix (issue 2) - everything else around is for the replica sync

crates/full-node/sov-sequencer/src/preferred/replica_sync_task.rs
```diff
@@ -280,10 +298,13 @@ where
             Some(completed_result) = pending_events.next() => {
                 match completed_result? {
                     CompletedEvent::Event(db_event) => {
-                        process_db_event(sequencer.clone(), db_event, query_pool).await?;
+                        process_db_event(sequencer.clone(), db_event, query_pool, &mut last_fully_processed_batch).await?;
                     }
-                    CompletedEvent::Backfill => {
-                        trace!("Backfill completed up to event_id");
+                    CompletedEvent::Backfill {
```
theodorebugnet, line 303: backfill bugfix (issue 2)

crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
```diff
@@ -146,6 +146,17 @@ impl<S: Spec, Rt: Runtime<S>> TransactionCache<S, Rt> {
         }
     }
 
+    pub async fn clean_and_overwrite_next_tx_number(&self, tx_number: u64) {
```
theodorebugnet, line 149: This is for issue 1, since we need to call it from the sequencer synchronously I think. Also not very clean since we already have this functionality above, maybe could be refactored better

crates/full-node/sov-sequencer/src/preferred/update_state.rs
```diff
@@ -75,6 +75,12 @@ where
         let node_state_root = tracing::trace_span!("root_hash")
             .in_scope(|| info.storage.get_root_hash(info.slot_number))?;
 
+        if is_startup_or_resync {
```
theodorebugnet, line 78: This is the main fix for issue 1

crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
```diff
@@ -1896,6 +1896,73 @@ async fn events_are_returned_in_tx_response() {
     assert_eq!(response.events.len(), 1);
 }
 
+#[tokio::test(flavor = "multi_thread")]
```
theodorebugnet, line 1899: Test for issue 1

crates/full-node/sov-sequencer/tests/integration/preferred_with_replicas.rs
```diff
@@ -88,7 +90,7 @@ async fn test_actions_against_replicas(
         run_actions_against_test_rollup(actions, master, &admin.clone(), state).await;
 
     // Ensure replicas have processed the database changes
-    tokio::time::sleep(Duration::from_millis(300)).await;
+    tokio::time::sleep(Duration::from_millis(3000)).await;
```
theodorebugnet, line 93: Wrong change, shouldn't be needed actually

theodorebugnet (commented):
Marked how to split the PR according to the issues - everything not explicitly mentioned is for the replica sync fix (bullet point 3)

crates/full-node/sov-sequencer/src/preferred/replica_sync_task.rs
```diff
@@ -251,18 +240,47 @@ where
...
+                                if last_fully_processed_batch.is_none_or(|seq| seq < last_replayed_sequence_number) {
+                                    // Query the batch_close event with this sequence number to find the event_id
+                                    // Set our latest_received_event_id to that event's id so we continue from there
+                                    let new_latest_event_id = query_event_id_for_batch_close(query_pool, last_replayed_sequence_number).await?;
+                                    if latest_received_event_id.is_some_and(|latest| latest >= new_latest_event_id) {
+                                        // Return an error rather than using assert! - this way the
+                                        // sequencer will call exit_rollup() cleanly
+                                        anyhow::bail!("Sanity check failed - update_state fast-forwarded our sequence number, but this would not have fast-forwarded our event ID");
+                                    }
+                                    latest_received_event_id = Some(new_latest_event_id);
```
bkolad, line 262: Is that ok that we override `latest_received_event_id` with `Some(parsed_notification.event_id)` in line 286

theodorebugnet: Yes that should be fine - we use this value to check for backfill for events between `new_latest_event_id` and `parsed_notification.event_id`. Once we've inserted the backfill, the latter is the correct value

PR #1449, opened 2025-07-31
fix(celestia): improve error handling in header conversion and mutex locking

DeVikingMark:
Replace unsafe unwrap() calls with unwrap_or_default() in TendermintHeader conversion to prevent panics when optional header fields are None. Add descriptive error message for mutex lock acquisition.

- Replace unwrap() with unwrap_or_default() for last_commit_hash, last_results_hash, evidence_hash
- Improve mutex lock error message in prev_hash() method


preston-evans98:
@DeVikingMark, I'm afraid many of these `unwrap_or_default` calls are unsound; these fields are not allowed to be `None`, so we would like to panic on error here. Closing this for now

PR #1457, opened 2025-08-01
Use `tokio::task::block_in_place` where LedgerDb lock is touched

citizen-stig:
# Description

To avoid tokio starvation

From Documentation: 

> Runs the provided blocking function on the current thread without blocking the executor.
In general, issuing a blocking call or performing a lot of compute in a future without yielding is problematic, as it may prevent the executor from driving other tasks forward. Calling this function informs the executor that the currently executing task is about to block the thread, so the executor is able to hand off any other tasks it has to a new worker thread before that happens. See the CPU-bound tasks and blocking code section for more information.
Be aware that although this function avoids starving other independently spawned tasks, any other code running concurrently in the same task will be suspended during the call to block_in_place. This can happen e.g. when using the join! macro. To avoid this issue, use spawn_blocking instead of block_in_place.
Note that this function cannot be used within a current_thread runtime because in this case there are no other worker threads to hand off tasks to. On the other hand, calling the function outside a runtime is allowed. In this case, block_in_place just calls the provided closure normally.
Code running behind block_in_place cannot be cancelled. When you shut down the executor, it will wait indefinitely for all blocking operations to finish. You can use shutdown_timeout to stop waiting for them after a certain timeout. Be aware that this will still not cancel the tasks — they are simply allowed to keep running after the method returns.

This is another iteration of https://github.com/Sovereign-Labs/sovereign-sdk-wip/pull/3295 

If we notice performance degradataion, we can switch to `spawn_blocking`

## Linked Issues
- Related to https://github.com/Sovereign-Labs/sovereign-sdk-wip/issues/1000

## Testing

## Docs
No updates

### DIFF
--- a/crates/full-node/sov-db/src/ledger_db/rpc.rs
+++ b/crates/full-node/sov-db/src/ledger_db/rpc.rs
@@ -556,16 +556,16 @@ impl LedgerStateProvider for LedgerDb {
     type Error = anyhow::Error;
 
     async fn get_head_slot_number(&self) -> Result<SlotNumber, Self::Error> {
-        self.get_rpc_reader()
+        let rpc_reader = tokio::task::block_in_place(|| self.get_rpc_reader());
+        rpc_reader
             .get_head_slot_number()
             .await
             .map(Option::unwrap_or_default)
     }
 
     async fn get_latest_finalized_slot_number(&self) -> Result<SlotNumber, Self::Error> {
-        self.get_rpc_reader()
-            .get_latest_finalized_slot_number()
-            .await
+        let rpc_reader = tokio::task::block_in_place(|| self.get_rpc_reader());
+        rpc_reader.get_latest_finalized_slot_number().await
     }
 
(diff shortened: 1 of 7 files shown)


bkolad (approved):

crates/full-node/sov-db/src/ledger_db/rpc.rs
```diff
@@ -556,16 +556,16 @@ impl LedgerStateProvider for LedgerDb {
     type Error = anyhow::Error;
 
     async fn get_head_slot_number(&self) -> Result<SlotNumber, Self::Error> {
-        self.get_rpc_reader()
+        let rpc_reader = tokio::task::block_in_place(|| self.get_rpc_reader());
```
bkolad, line 559: Would it be possible to use spawn_blocking everywhere? It's generally the preferred.

citizen-stig:
Closed in favour of https://github.com/Sovereign-Labs/sovereign-sdk/pull/1481

PR #1473, opened 2025-08-04
EIP 712 authenticator

LogvinovLeon:
# Description

This PR implements EIP712 authenticator. It's similar to the canonical one ans reuses some logic but verifies the EIP712 signatures.
It relies on the fact that Transaction can verify it's signature as EIP712.
This relies on Transaction implementing `SolValue`.
For now - I manually convert the TxDetails to a type that implements SolValue. So that we have something to sign. This is obviously unfinished and insecure as we only sign one field.
Future plans:
* Run cargo expand on UniversalWallet on some simple type and see what it produces
* Ammend the macro so that it also generates sol! definitions
* Implement a conversion from `impl UniversalWallet` to `impl SolValue`

## Linked Issues
- Related to # (issue) 

## Testing
There is a success test. Failure test coming next

## Docs
No docs yet

### DIFF
--- a/crates/module-system/sov-modules-api/src/transaction/mod.rs
+++ b/crates/module-system/sov-modules-api/src/transaction/mod.rs
@@ -231,14 +231,7 @@ impl<R: TransactionCallable, S: Spec> Transaction<R, S> {
         })?;
         serialized_tx.extend_from_slice(chain_hash);
 
-        match &self.versioned_tx {
-            VersionedTx::V0(inner) => {
-                MeteredSignature::new::<S>(inner.signature.clone())
-                    .verify(&inner.pub_key, &serialized_tx, meter)
-                    .map_err(TransactionVerificationError::from)?;
-            }
-        }
-        Ok(())
+        self.verify_inner(&serialized_tx, meter)
     }
 
     /// Check whether the transaction has been signed correctly using EIP712.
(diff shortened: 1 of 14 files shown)


crates/module-system/sov-modules-api/src/transaction/mod.rs
```diff
@@ -229,14 +231,33 @@ impl<R: TransactionCallable, S: Spec> Transaction<R, S> {
...
+    }
+
+    /// Check whether the transaction has been signed correctly using EIP712.
+    ///
+    /// # Errors
+    /// Returns an error if:
+    ///  * The signature is wrong
+    ///  * Serializing or hashing the transaction fails
+    ///  * Any operation runs out of gas
+    pub fn verify_eip712(
```
bkolad: Shouldn't this method be available only for EvmCryptoSpec, so it can be used exclusively by EIP-712 auth?

bkolad: Alternatively, we can define `eip712_signing_hash(tx_details, &DOMAIN)` and `verify_eip712(tx, ..)` as methods on the EIP-712 authenticator, passing the transaction as a parameter. This way, `verify_eip712` remains encapsulated within the authenticator.

bkolad (approved):

PR #1478, opened 2025-08-05
LedgerNotification service refactoring

citizen-stig:
# Description

Current implementation of LedgerNotificationService::send_notification has 2 issues:

1. It sends notification about a finalized slot with slot number of updated slot
2. It takes and drops 3 mutexes, meaning there might be concurrenct submission of different notifications. 

This PR changes that:

1. Explicitly pass 2 parameters about which slots to notify
2. Take a 3 mutexes at once, send all required notifications and drop them. I believe this would be easier to reason about.

This PR also adds trace level logging to notification submission, so we can have full view of what is happening there.

## Linked Issues
Related to debugging of `test_check_no_reorgs_longer` test flakyness

## Testing
Existing test should be passing

## Docs
No updates


citizen-stig:
Closed in favour of #1508

PR #1482, opened 2025-08-05
Sequencer: Cleanup code related to `RollupBlockExecutor

bkolad:
# Description

Before this PR, the responsibility for creating the `RollupBlockExecutor` was ambiguous. In some cases, the `RollupBlockExecutorConfig` was created within the `inner` process and then passed to the `sequencer`, which constructed the `RollupBlockExecutor`. In other cases, the roles were reversed.

This PR makes the following changes:
1. Simplifies the construction of `RollupBlockExecutor` by providing 2 consturctor instead of one constructo that accepts an Option.
2. Refactors `RollupBlockExecutorConfig` to include only values common to all executor types.
3.  Clearly assigns the responsibility for creating the `RollupBlockExecutor` solely to the `Inner` process.

No buisness logic has been chnged

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/block_executor.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/block_executor.rs
@@ -98,14 +98,13 @@ impl<S: Spec> RollupBlockExecutorError<S> {
 type StateRootReceiver<S> =
     oneshot::Receiver<(RollupHeight, <<S as Spec>::Storage as Storage>::Root)>;
 
-pub struct RollupBlockExecutorConfig<S: Spec, Rt: Runtime<S>> {
-    pub config: SequencerConfig<S::Address, PreferredSequencerConfig>,
+#[derive(Clone)]
+pub struct RollupBlockExecutorConfig<S: Spec> {
     pub da_address: <S::Da as DaSpec>::Address,
     pub shutdown_notifier: Sender<()>,
-    pub state_root_request_sender: tokio::sync::mpsc::Sender<StateRootComputeRequest<S>>,
     pub shutdown_receiver: watch::Receiver<()>,
     pub shutdown_sender: watch::Sender<()>,
-    pub startup_transaction_cache_writer: Option<TxResultWriter<S, Rt>>,
+    pub state_root_request_sender: Sender<StateRootComputeRequest<S>>,
 }
 
 pub struct RollupBlockExecutor<S, Rt>
@@ -114,11 +113,13 @@ where
     Rt: Runtime<S>,
 {
     pub checkpoint: StateCheckpoint<S>,
-    rollup_block_task_state: Option<BackgroundTaskState<S>>,
+    seq_config: SequencerConfig<S::Address, PreferredSequencerConfig>,
+    shutdown_receiver: watch::Receiver<()>,
+    shutdown_sender: watch::Sender<()>,
 
+    rollup_block_task_state: Option<BackgroundTaskState<S>>,
     next_event_number: u64,
     next_tx_number: u64,
-    config: SequencerConfig<S::Address, PreferredSequencerConfig>,
     da_address: <S::Da as DaSpec>::Address,
     // A sender notifying that this acceptor has successfully shut down. We give a handle to
     // each background task when it is spawned, ensuring that this channel remains open as long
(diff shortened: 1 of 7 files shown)


crates/full-node/sov-sequencer/src/preferred/replica_sync_task.rs
```diff
@@ -108,29 +108,14 @@ enum CompletedEvent {
...
-    latest_state_update: StateUpdateInfo<S::Storage>,
+    _latest_state_update: StateUpdateInfo<S::Storage>,
     latest_loaded_event_id: Option<u64>,
 ) -> JoinHandle<()>
 where
     S: Spec,
     Rt: Runtime<S>,
     Da: DaService<Spec = S::Da>,
 {
-    if let Err(e) = sequencer
```
bkolad, line 119: This code contains a bug, so instead of updating it, I removed it. A proper fix will be included as part of the replica workstream in upcoming PRs

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -98,14 +98,13 @@ impl<S: Spec> RollupBlockExecutorError<S> {
...
-    pub config: SequencerConfig<S::Address, PreferredSequencerConfig>,
+#[derive(Clone)]
+pub struct RollupBlockExecutorConfig<S: Spec> {
     pub da_address: <S::Da as DaSpec>::Address,
     pub shutdown_notifier: Sender<()>,
-    pub state_root_request_sender: tokio::sync::mpsc::Sender<StateRootComputeRequest<S>>,
     pub shutdown_receiver: watch::Receiver<()>,
     pub shutdown_sender: watch::Sender<()>,
-    pub startup_transaction_cache_writer: Option<TxResultWriter<S, Rt>>,
+    pub state_root_request_sender: Sender<StateRootComputeRequest<S>>,
```
theodorebugnet, line 107: This is down to preference but I like keeping the module name, i.e. `mpsc::Sender` since tokio has multiple different channels with identical looking Sender and Receiver structs

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -144,27 +141,49 @@ impl<S: Spec, Rt: Runtime<S>> RollupBlockExecutor<S, Rt> {
...
     ) -> RollupBlockExecutor<S, Rt> {
+        Self::new_with_tx_cache_writer_inner(
+            info,
+            Some(tx_cache_writer),
+            rollup_exec_config,
+            seq_config,
+        )
+    }
+
+    fn new_with_tx_cache_writer_inner(
```
theodorebugnet: Also very tiny nit: maybe just `new_inner` is clearer here, since it can be also called without a tx cache writer?

crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -429,16 +428,30 @@ where
...
         let next_sequence_number_according_to_node =
             get_next_sequence_number_according_to_node(info, &mut Rt::default());
+
         self.executor_events_sender
             .trigger_recovery(next_sequence_number_according_to_node, recovery_strategy)
             .await;
 
-        let executor_from_info = RollupBlockExecutor::<_, Rt>::new(info, rollup_exec_config);
+        // Creates a new executor  for recovery. This must *not* be called to create executors
+        // under other circumstances, since it causes side effects on the transaction cache.
```
theodorebugnet: Since this is no longer a doc comment on a function:
```suggestion
        // Creates a new executor for recovery, which will cause side effects on the transaction cache.
```

crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -1649,19 +1655,26 @@ where
...
-        transaction_cache_write_handle: TxResultWriter<S, Rt>,
         info: StateUpdateInfo<S::Storage>,
-        rollup_exec_config: RollupBlockExecutorConfig<S, Rt>,
         reason: &'static str,
     ) {
         let mut inner = self.get_inner_with_timing(reason).await;
-        let executor_from_info = RollupBlockExecutor::<_, Rt>::new(&info, rollup_exec_config);
+
+        // Creates a new executor  for recovery. This must *not* be called to create executors
+        // under other circumstances, since it causes side effects on the transaction cache.
```
theodorebugnet: Same as above
```suggestion
        // Creates a new executor for recovery, which will cause side effects on the transaction cache.
```

theodorebugnet (approved):
LGTM with non-blocking style suggestions

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
bkolad: I agree, but in other parts of the same file, we use just Sender. I wanted to keep it consistent. We'll need a separate PR to prefix all the channels across the codebase.

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
bkolad: fixed

PR #1487, opened 2025-08-06
Fix numbering issues in ledger API. Add regression test

preston-evans98:
# Description

This PR fixes several minor bugs in the ledger API which caused incorrect item numbers (slot number/tx number / batch number) to be returned when the `include_children` flag was passed. It also adds new snapshot tests on this behavior, and ensures that we skip serializing the "children" array if the user did not request children to be included.

### DIFF
--- a/crates/full-node/sov-ledger-apis/src/lib.rs
+++ b/crates/full-node/sov-ledger-apis/src/lib.rs
@@ -901,7 +901,8 @@ struct Transaction<TxReceipt: TxReceiptContents, E> {
     #[serde_as(as = "serde_with::base64::Base64")]
     pub body: Vec<u8>,
     pub receipt: TxEffect<TxReceipt>,
-    pub events: Vec<RuntimeEventResponse<E>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub events: Option<Vec<RuntimeEventResponse<E>>>,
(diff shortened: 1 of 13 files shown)


crates/full-node/sov-ledger-apis/src/lib.rs
```diff
@@ -862,20 +876,33 @@ struct Batch<B, TxReceipt: TxReceiptContents, E> {
...
+                let mut failed_conversion = false;
+                for (offset, tx) in txs.into_iter().enumerate() {
+                    if let ItemOrHash::Full(tx) = tx {
+                        output.push(Transaction::new(tx, batch.tx_range.start + offset as u64));
+                    } else {
+                        failed_conversion = true;
+                        break;
+                    }
+                }
+                if failed_conversion {
```
citizen-stig, line 898: In which case conversion is going to fail? Should we log it? Or is it logged inside ::new?

citizen-stig (approved):

crates/full-node/sov-ledger-apis/src/lib.rs
preston-evans98: This would fail if you did querymode::compact. But in that case, returning nothing matches the semantics of the API as closely as possible, so I think this is ok

PR #1488, opened 2025-08-06
Sequencer: Cleanup related to `api_ledger_db`

bkolad:
# Description

This PR moves api_ledger_db into Inner, simplifying the sequencer workflow by removing the need to pass it through methods.


crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -661,6 +667,25 @@ where
...
     }
+
+    pub(crate) async fn update_api_ledger(&self, info: &StateUpdateInfo<S::Storage>) {
+        let start = std::time::Instant::now();
+        tracing::trace!(
+            slot_number = %info.slot_number,
+            latest_finalized_slot_number = %info.latest_finalized_slot_number,
+            "Starting LedgerAPI storage update");
+        self.api_ledger_db
+            .replace_reader(info.ledger_reader.clone());
```
citizen-stig, line 678: I've been playing with this in `https://github.com/Sovereign-Labs/sovereign-sdk/pull/1478

Could you please add trace between replace reader and before send_notifications? That would be helpful.

citizen-stig: This is non-blocking, obviously.

citizen-stig (approved):

crates/full-node/sov-sequencer/src/preferred/inner.rs
bkolad: Will do it in the next PR.

PR #1489, opened 2025-08-06
[WIP] Add an example of using UniversalWallet schema to derive sol! definitions and use them to compute EIP712 hashes

LogvinovLeon:
# Description

Adds an example that has:
* A build script that takes some schema derived from Rust types and generates `alloy_schema.rs`
* `main.rs` that consumes alloy schema to generate EIP712 hash

This is just a scaffold. For now it only implements structs and uint256, but it already works for them.

## Linked Issues
- Related to # (issue) 

## Testing
Run `cargo run` in  `examples/universal-wallet`

## Docs


crates/universal-wallet/schema/src/ty/sol.rs
```diff
@@ -0,0 +1,50 @@
+use crate::{
```
theodorebugnet, line 3: Throughout most of the codebase I think we're using the `module_name/mod.rs` rather than `module_name.rs` style. I think it'd be good to be consistent

theodorebugnet (commented):
The implementation looks good to me! But is the `examples/universal-wallet` crate temporary until the rest of the schema types are implemented? If it is, then I feel like it'd be better to delay merging a bit until the implementation is finished, and then add an example and tests for it in e.g. the demo-rollup.

PR #1491, opened 2025-08-06
Return nonce in dedup endpoint

citizen-stig:
# Description

This PR enhances the dedup endpoint to correctly return nonces instead of generation numbers, and adds support for explicitly selecting between nonce and generation uniqueness identifiers through query parameters.

Corresponding web3-js PR: https://github.com/Sovereign-Labs/sovereign-sdk-web3-js/pull/204

## Key Changes

- **BREAKING**: Default behavior of `/rollup/addresses/{address}/dedup` has changed:
  - It returns `nonce` instead of `generation`
  - It accepts hex encoded credential_id, OpenAPI spec has changed.
- Added query parameter support: `?select=nonce` or `?select=generation` for explicit selection
- Renamed `NonceDeDupEndpoint` to `SovereignDeDupEndpoint` to reflect dual capability
- Added `next_nonce()` method to complement existing `next_generation()` functionality
- Enhanced response structure with conditional field serialization
- `HexString` now is more permissive and allows skipping leading `0x`

## Breaking Changes

The dedup endpoint default behavior has changed:

- **Before**: Returned generation number by default in the `nonce` field.
- **After**: Returns nonce by default, in the nonce field.
- **Migration**: Use `?select=generation` query parameter to maintain old behavior, but response structure changes, `generation` field now contains generation number

------

- [x] I have updated `CHANGELOG.md` with a new entry explaining the breaking change and migration path
- [x] I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. (Only added tracing dependency to sov-uniqueness for debugging)

## Linked Issues
- Related to nonce/generation uniqueness separation feature

## Testing

Added comprehensive integration test (`uniqueness.rs`) that validates:
- Independent tracking of nonces and generations per account
- Nonces require sequential ordering (no skipping)
- Generations allow skipping values
- Both counters can be used interchangeably in the same account
- Proper error handling for invalid nonce/generation values

Run tests with:
```bash
cargo test -p sov-sequencer --test integration uniqueness
```

## Docs

- OpenAPI spec updated with corrected description
- CHANGELOG includes breaking change notice with migration instructions


crates/full-node/sov-rollup-apis/src/endpoints/dedup.rs
```diff
@@ -60,45 +60,120 @@ pub trait DeDupEndpoint<S: Spec>: Clone + Send + Sync + 'static {
...
+/// The response of the Dedup implementation for both nonce and generation number.
+///
+/// Only one field will be populated based on the query parameter:
+/// - `nonce`: The next sequential nonce value (must be used in order)
+/// - `generation`: The next available generation number (can skip values)
 #[derive(serde::Serialize, Clone)]
-pub struct NonceResponse {
-    /// The current nonce assiociated with the requested address.
-    pub nonce: u64,
+pub struct DedupResponse {
```
bkolad, line 94: Shouldn’t this be an enum, since we never return both nonce and generation?

crates/module-system/module-implementations/sov-uniqueness/src/lib.rs
```diff
@@ -78,6 +78,23 @@ impl<S: Spec> Uniqueness<S> {
                     .unwrap_or_default())
             })?
     }
+
+    /// Retrieves the latest known nonce + 1 for a given credential id.
+    ///
+    /// # Errors
+    /// May return an error if the next nonce overflows or if state access fails.
+    pub fn next_nonce<Reader: StateReader<User>>(
```
bkolad, line 86: Does it return the next nonce to use?
Our nonce numbering starts from zero. So, if we send the first-ever transaction, we should use nonce 0, but this method returns 1. Is that okay?

citizen-stig: Good catch! I will add a test for that too

crates/full-node/sov-rollup-apis/src/endpoints/dedup.rs
citizen-stig: I also thought saw, but for simplicity in JS land @preston-evans98 suggested to use this approach.

crates/module-system/module-implementations/sov-uniqueness/src/lib.rs
citizen-stig: So that was what get caught by the web3-js test.

But I will need to add a test inside the SDK for that.

citizen-stig: Added a check for the default nonce in the uniqueness intergration test.

bkolad (approved):

crates/full-node/sov-rollup-apis/src/endpoints/dedup.rs
```diff
@@ -60,45 +64,125 @@ pub trait DeDupEndpoint<S: Spec>: Clone + Send + Sync + 'static {
...
+    pub generation: Option<u64>,
+}
+
+/// Query parameters for the dedup endpoint.
+#[derive(Debug, Clone, Default, serde::Deserialize)]
+pub struct DedupQuery {
+    /// Select which kind of uniqueness to return.
+    /// If omitted, returns Nonce
+    /// Example: `?select=nonce` or `?select=generation`
+    pub select: Option<SelectField>,
```
ross-weir, line 109: Is retrieving the nonce and generation an expensive operation? Could we always just return both values

citizen-stig: Nonce is cheap, direct state map access
Generation a little bit more expensive, state map + iteration over values. Probably not a big deal, but might be unpleasant for production.

ross-weir (approved):

crates/full-node/sov-rollup-apis/src/endpoints/dedup.rs
theodorebugnet, line 109: Very minor nit but thoughts on naming the query parameter something like `type`? E.g. `?type=nonce` and `?type=generation`.
It will make the rust variable name annoying though so maybe not worth it

theodorebugnet (approved):
LGTM

crates/full-node/sov-rollup-apis/src/endpoints/dedup.rs
citizen-stig: Yeah, it is going to be nont nice and have to do serde attributes.

PR #1492, opened 2025-08-06
Add test of bridged gas token using hyperlane

preston-evans98:
# Description
This PR makes the gas token config optional and runs a complete test using admin mode to set up the gas token bridged in via Hyperlane.


crates/module-system/module-implementations/sov-bank/src/genesis.rs
```diff
@@ -14,7 +14,8 @@ use crate::{config_gas_token_id, Amount, Bank, TokenId};
 #[schemars(bound = "S: Spec", rename = "BankConfig")]
 pub struct BankConfig<S: Spec> {
     /// Configuration for the gas token
-    pub gas_token_config: GasTokenConfig<S>,
+    #[serde(default)]
+    pub gas_token_config: Option<GasTokenConfig<S>>,
```
ross-weir, line 18: nit: Might be worth a comment explaining why/when this could be `None`

ross-weir (approved):
[examples/demo-rollup/tests/resync/data/mock_da.sqlite](https://github.com/Sovereign-Labs/sovereign-sdk/pull/1492/files#diff-94d99a9b178484d38799ef287fb02f832acf15d4bb1cc9f2979bea6f1c2ccc97) intentional?

crates/module-system/module-implementations/sov-bank/src/genesis.rs
preston-evans98: Good call. Will do in a quick follow-up

PR #1496, opened 2025-08-07
Remove `DaSyncState` from the sequencer.

bkolad:
# Description

This PR makes the sequencer more deterministic by removing `DaSyncStatus` from it. Previously, `DaSyncStatus` fetched DA information directly from the node and mutated its internal state, creating an implicit dependency between the node and the sequencer. In the new design, the sync information is provided through `StateUpdateInfo` and is immutable, eliminating that hidden coupling.

### DIFF
--- a/crates/full-node/sov-stf-runner/src/runner.rs
+++ b/crates/full-node/sov-stf-runner/src/runner.rs
@@ -9,6 +9,7 @@ use jsonrpsee::RpcModule;
 use sov_db::ledger_db::{LedgerDb, SlotCommit};
 use sov_db::schema::{DeltaReader, SchemaBatch};
 use sov_metrics::{MonitoringConfig, RunnerMetrics};
+
 use sov_rollup_interface::common::{RollupHeight, SlotNumber};
 use sov_rollup_interface::da::{BlobReaderTrait, BlockHeaderTrait, DaSpec};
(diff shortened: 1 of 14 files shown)


crates/full-node/sov-stf-runner/src/runner.rs
```diff
@@ -165,15 +166,12 @@ where
...
             SocketAddr::new(axum_config.bind_host.parse()?, axum_config.bind_port);
 
-        let next_item_numbers = ledger_db.get_next_items_numbers()?;
-        let last_slot_processed_before_shutdown = next_item_numbers.slot_number.saturating_sub(1);
-
-        let da_height_processed =
-            runner_config.genesis_height + last_slot_processed_before_shutdown.get();
+        let first_unprocessed_height_at_startup = sync_state
+            .synced_da_height
+            .load(std::sync::atomic::Ordering::Acquire)
```
citizen-stig, line 171: Should we have safe match here?

citizen-stig (approved):

crates/full-node/sov-stf-runner/src/runner.rs
bkolad: fixed

theodorebugnet (approved):

PR #1503, opened 2025-08-09
Allow pushing to accessory state vecs during tx execution

preston-evans98:
# Description

This PR allows users to push to accessory state vecs without requiring read permissions. This is sound, since users are not allowed to view the contents of the vec (including the length) by the signature of the method. This is required to allow real-time soft confirmations for EVM chains

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/sov-modules-api/src/containers/vec.rs
+++ b/crates/module-system/sov-modules-api/src/containers/vec.rs
@@ -3,7 +3,7 @@ use std::marker::PhantomData;
 
 use sov_state::codec::BorshCodec;
 use sov_state::namespaces::{Accessory, CompileTimeNamespace, Kernel, User};
-use sov_state::{EncodeLike, Prefix, StateCodec, StateItemCodec};
+use sov_state::{EncodeLike, Prefix, StateCodec, StateItemCodec, StateItemDecoder};
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


bkolad (approved):

crates/module-system/sov-modules-api/src/containers/vec.rs
```diff
@@ -320,6 +320,36 @@ where
...
+        let len_key = self.len_value.slot_key();
+        let index = if let Some(len_bytes) = state.get_value(Accessory::NAMESPACE, &len_key) {
+            self.len_value
+                .codec()
+                .value_codec()
+                .decode_unwrap(len_bytes.value())
+        } else {
+            0
+        };
+        self.len_value.set(&index.saturating_add(1), state)?;
```
bkolad: Shouldn’t we return an error (or `bool::false` rather than `()` in the `Ok` case) instead of using `index.saturating_add(1)` if the length overflows?

preston-evans98: Hmm, I think you're right that failing silently is a bad idea. This *should* be unreachable since we use a u64 for the index, but it seems much better to panic if that assumption is violated than to have silent data corruption. Fixed this and a few other instances!

PR #1507, opened 2025-08-11
`LedgerDB`: Save discarded blob hash with its `BlobDiscardReason`

bkolad:
# Description

Until now, the ledger DB only stored discarded blob hashes, but the BlobSender will now also need the discard reason.


crates/full-node/sov-db/src/ledger_db/mod.rs
```diff
@@ -341,7 +341,7 @@ impl LedgerDb {
         blob: StoredDiscardedBlob,
         schema_batch: &mut SchemaBatch,
     ) -> anyhow::Result<()> {
-        schema_batch.put::<DiscardedBlobByHash>(&blob.hash, &blob)
+        schema_batch.put::<DiscardedBlobByHash>(&blob.discarded_blob.hash.0, &blob)
```
theodorebugnet, line 344: It looks like we still only store the hash alone in the LedgerDB (unless I missed something). Is this intended?

theodorebugnet: Never mind - just realised it's now part of the StoredDiscardedBlob

theodorebugnet (approved):

PR #1508, opened 2025-08-11
Fix flakyness `test_check_no_reorgs_longer` by notifying only about processed finalized transitions

citizen-stig:
# Description

StateManager register notification for the latest seen finalized DA height, not the latest processed. This means that slot might be not available for WebSocket subcription.

Essence of this PR is in 1611c30093b427b843372104d1c635e342c72f8e

Other changes:

* More trace logging in various places for easier debug
* More tests similar to `test_check_no_reorgs_longer` but with longer/shorter finality
* Consistency renaming in `LedgerDb`: we had mix of `slot_num` vs `slot_number`, I've converted everything to `slot_number`

## Testing
New tests are added and existing are passing

## Docs
Updated

### DIFF
--- a/crates/full-node/sov-stf-runner/src/state_manager/mod.rs
+++ b/crates/full-node/sov-stf-runner/src/state_manager/mod.rs
@@ -316,8 +316,7 @@ where
         // ----
 
         let processing_finalized_transitions_start = std::time::Instant::now();
-        let (last_finalized_header, finalized_transitions) =
-            self.process_finalized_state_transitions(da_service).await?;
+        let finalized_transitions = self.process_finalized_state_transitions(da_service).await?;
         let processing_finalized_transitions_time =
             processing_finalized_transitions_start.elapsed();
         tracing::trace!(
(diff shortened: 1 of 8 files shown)


crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -332,25 +331,29 @@ where
...
+                    .saturating_sub(da_height_at_genesis),
+            );
+            tracing::trace!(
+                ?last_finalized_slot_number,
+                current_slot_number = ?slot_number,
+                "Going to materialize last finalized slot number"
+            );
+            let last_finalized_slot_update = self
+                .ledger_db
+                .materialize_latest_finalize_slot(slot_number, last_finalized_slot_number)?;
```
bkolad, line 348: Can we move it to line 383. We already have a loop there;
```
 for finalized_transition in &finalized_transitions {
 ...
 }

bkolad (approved):

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: This should happen before ledger changes set is saved, so I cannot move it into this loop, but I can move it to line 377

citizen-stig: I just thought it makes sense to keep it closer to place where ledger chage_set is created. 

Otherwise it is going be after aggregated proof data is added there.

PR #1511, opened 2025-08-11
Collect metrics on state accesses and storage reads

preston-evans98:
# Description

This PR instruments the stf blueprint with metrics which track...
- How many storage accesses occurred and how long was taken for each high level operation (Authenticate, resolve context, execute, etc.)
- How many items had to be retrieved from storage (i.e how many were not in the ProvableStorageCache)
- How long the worst storage accesses took

The bulk of this PR is adding scaffolding; the format of the metrics may need further revision to allow easy access from Grafana.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-metrics/src/lib.rs
```diff
@@ -35,3 +39,71 @@ macro_rules! save_elapsed {
...
+    /// The event is in progress.
+    InProgress(std::time::Instant),
+    /// The event has completed.
+    Completed(std::time::Duration),
+    #[default]
+    /// There is no event (because we're not in native mode).
+    None,
+}
+
+impl MaybeTimer {
```
bkolad, line 56: It might be easier to maintain if `MaybeTimer` had two separate impl blocks — one for the native case and one for the non-native case.

```
#[cfg(feature = "native")]
impl MaybeTimer {
 ...
}

```

```
#[cfg(not(feature = "native"))]
impl MaybeTimer {
 ...
}
```

crates/full-node/sov-metrics/src/influx_db_nonnative.rs
```diff
@@ -0,0 +1,213 @@
...
+        self.accesses.len()
+    }
+
+    /// Returns true if there are no state accesses since the last flush
+    pub fn is_empty(&self) -> bool {
+        self.accesses.is_empty()
+    }
+}
+
+// fn serialize_state_access_metric(buffer: &mut Vec<u8>, metric: &StateAccessMetric, prefix: &str) -> std::io::Result<()> {
```
bkolad: Let's remove commented code

bkolad (approved):

crates/full-node/sov-metrics/src/influx_db_nonnative.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+}
+
+#[cfg(feature = "native")]
+impl Metric for AuthAndProcessMetrics {
+    fn measurement_name(&self) -> &'static str {
+        "sov_rollup_auth_and_process_metrics"
+    }
+
+    // TODO: Finish serializing all the metrics
```
citizen-stig: Is this TODO still relevant?

crates/full-node/sov-metrics/src/lib.rs
```diff
@@ -35,3 +39,71 @@ macro_rules! save_elapsed {
...
+    /// Starts a timer if the `native` feature is enabled. Otherwise, does nothing.
+    #[cfg(feature = "native")]
+    pub fn start(&mut self) {
```
citizen-stig: Nit: but do we really need 2 starting methods? Won't `started` be enough?

crates/module-system/sov-modules-api/src/state/accessors/tests.rs
```diff
@@ -151,7 +153,7 @@ fn test_values<S: Spec>(state: &mut StateCheckpoint<S>) {
         let is_cached = state.is_value_cached(NAMESPACE, &state_value.slot_key());
         assert_eq!(is_cached, IsValueCached::Yes(AccessSize::Read(8)));
 
-        let maybe_size = state.get_size(NAMESPACE, &state_value.slot_key());
+        let maybe_size = state.get_size(NAMESPACE, &state_value.slot_key(), &mut metric);
```
citizen-stig, line 156: It is out of the scope of this test, but maybe we can make an exception and check that metrics are tracker properly?

citizen-stig (approved):

crates/full-node/sov-metrics/src/lib.rs
preston-evans98: Good idea. Done.

crates/full-node/sov-metrics/src/influx_db_nonnative.rs
preston-evans98: Nope, that's handled now

crates/module-system/sov-modules-api/src/state/accessors/tests.rs
preston-evans98: Unfortunately there's not really anything useful to assert here since the storage read has size 0 (which is the default). We could check that the time is non-zero, but that would be the case even if the metric wasn't passed properly.

crates/full-node/sov-metrics/src/lib.rs
preston-evans98: Unfortunately we do need this, yeah; for the `AuthAndProcessTimings` we want to construct the timers all at once (i.e. using default), but we don't want to start them until later

crates/full-node/sov-metrics/src/influx_db_nonnative.rs
preston-evans98: Done!

PR #1517, opened 2025-08-12
`BlobSelector`: Resubmit discarded blobs

bkolad:
# Description

This PR introduces the following changes to `BlobSende`r:
1. If a blob is discarded by` BlobSelector` for a reason other than `SequenceNumberTooLow`, `BlobSender` resubmits it.
2. If a blob is discarded with the reason `SequenceNumberTooLow`,` BlobSender` removes it from memory.
3. Fixed a bug in `BlobSelector` where blobs were rejected for the `InsufficientFunds` reason instead of `OutOfCapacity`.

### DIFF
--- a/crates/full-node/sov-blob-sender/src/lib.rs
+++ b/crates/full-node/sov-blob-sender/src/lib.rs
@@ -17,6 +17,7 @@ use sov_rollup_interface::common::HexHash;
 use sov_rollup_interface::node::da::{DaService, SubmitBlobReceipt};
 use sov_rollup_interface::node::ledger_api::{LedgerStateProvider, QueryMode};
 use sov_rollup_interface::node::{future_or_shutdown, FutureOrShutdownOutput};
+use sov_rollup_interface::stf::BlobDiscardReason;
 use tokio::sync::{broadcast, oneshot, watch, Mutex};
 use tokio::task::JoinHandle;
 use tokio::time::{interval, sleep};
@@ -55,15 +56,6 @@ pub trait BlobSenderHooks: Send + Sync + 'static {
 
     /// The blob was published and is now part of the DA's canonical chain.
     async fn on_published_blob(
-        &self,
-        _blob_id: BlobInternalId,
-        _blob_hash: [u8; 32],
-        _da_tx_id: <Self::Da as DaSpec>::TransactionId,
-    ) {
-    }
-
-    /// The blob was processed by the rollup node, but it may not be finalized yet.
-    async fn on_processed_blob(
         &self,
         _blob_id: BlobInternalId,
         _blob_hash: [u8; 32],
@@ -441,7 +433,10 @@ impl FinalizationManager for LedgerDb {
         {
             Some(batch) => (batch.slot_number, BlobSelectorStatus::Accepted),
             None => match self.get_discarded_blob_by_hash(blob_hash).await? {
-                Some(blob) => (blob.slot_number, BlobSelectorStatus::Discarded),
+                Some(blob) => (
+                    blob.slot_number,
+                    BlobSelectorStatus::Discarded(blob.discarded_blob.reason),
+                ),
                 None => return Ok(None),
             },
         };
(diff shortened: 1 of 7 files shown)


crates/full-node/sov-blob-sender/src/lib.rs
```diff
@@ -747,21 +734,41 @@ impl<Da: DaService, FM: FinalizationManager> TaskState<Da, FM> {
...
+                                    blob_id,
+                                    receipt.blob_hash.into(),
+                                    &receipt.da_transaction_id,
+                                )
+                                .await;
 
-                    // We won't shut down the rollup in case of this error, but we will log the error.
-                    let _ = self.remove_blob_or_err(blob_id).await;
-                    break;
+                            // We won't shut down the rollup in case of this error, but we will log the error.
```
citizen-stig, line 754: I suppose error is loggerd in the function, right?

crates/full-node/sov-blob-sender/tests/blob_sender.rs
```diff
@@ -292,6 +276,111 @@ async fn blob_sender_exit_if_blob_not_processed() -> anyhow::Result<()> {
...
+    let data_1 = {
+        let blob_id = 11u8;
+        let data = Arc::new([blob_id, 2, 3, 4, 5]);
+        blob_sender
+            .publish_batch_blob(data.clone(), blob_id as BlobInternalId)
+            .await?;
+        data
+    };
+
+    sleep(Duration::from_secs(1)).await;
```
citizen-stig, line 340: How do we know that this 1 second won't be flaky in CI

crates/full-node/sov-blob-sender/tests/blob_sender.rs
```diff
@@ -292,6 +276,111 @@ async fn blob_sender_exit_if_blob_not_processed() -> anyhow::Result<()> {
...
+    let submissions = blob_sender.nb_of_concurrent_blob_submissions();
+    assert_eq!(submissions, 1);
+
+    {
+        deps.da.produce_block_now().await?;
+        sleep(Duration::from_secs(1)).await;
```
citizen-stig, line 346: Same about flakyness

crates/module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -834,11 +834,26 @@ impl<S: Spec> BlobStorage<S> {
                 .get_sender_balance(preferred_sender, state)
                 .unwrap_or(Amount::ZERO);
 
+            if !blobs_to_select.can_accept_blob(SequencerType::Preferred, blob_with_id.blob_size())
+            {
+                tracing::error!(
+                    blob_id = hex::encode(blob_id),
+                    "The blob is discarded because it exceeds the size limit."
```
citizen-stig, line 841: Could we add blob size to the log message? Max allowed capacity also would be nice, but at least a size.

Ideally we would have both values in the error struct, but that's optional

crates/full-node/sov-blob-sender/src/lib.rs
bkolad: yes

crates/module-system/module-implementations/sov-blob-storage/src/capabilities.rs
bkolad: It is already logged in `can_process_blob`

crates/full-node/sov-blob-sender/tests/blob_sender.rs
bkolad: It is three times the slot duration, which should be enough to observe the changes. The other tests in this file use the same duration and did not encounter flakiness.

crates/full-node/sov-blob-sender/tests/blob_sender.rs
citizen-stig: Discuss

citizen-stig (approved):

Issue #1520, opened 2025-08-12
[TRACKER] Solana offchain message authenticator


theodorebugnet:
## Description

Add an authenticator that accepts transactions to that have been signed in solana wallets as offchain messages, supporting both the official format that Ledger uses (https://docs.anza.xyz/proposals/off-chain-message-signing#message-preamble) and raw signed messages (which Phantom will likely use).

## Checklist

 - [x] Implement authenticator #1522 
     - [x] Consider adding the chain_name to the JSON #1526 
 - [x] Implement web3 support https://github.com/Sovereign-Labs/sovereign-sdk-web3-js/pull/205
     - [x] Implement encoding for both solana message types
     - [x] Use a mock signer to test byte compatibility with the rust implementation
     - [x] Add Solana wallet support - at least Ledger, for the preamble message type
     - [ ] Investigate Phantom's behaviour in order to add detection for which serialization to use, and add support for it if needed

## Unprioritised
- [ ] Make UnsignedTransaction versioned, or at least SolanaOffchainUnsignedTransaction versioned (if this will allow unsigned transactions to become versioned under the hood without a consensus breaking change). This may not be high priority since adding a new transaction version is consensus-breaking anyway.

## Deprioritised
 - [ ] Review sequencer endpoint construction, to standardise extra authenticator endpoints, e.g. by defining them in the Authenticator trait (rather than every rollup defining their own non-standard `sequencer_additional_apis`). **OR** Switch the authenticator encoding to the client side, making the `sequencer/txs` authenticator-agnostic; this will likely require providing the authenticator enum in the schema, along with some way for clients to identify which enum variant to use.
 - [ ] Revisit bounding by the `Spec`'s crypto type, perhaps by adding an enum of supported crypto primitives to the Spec. Currently the authenticator can be used on non-Solana compatible rollups (e.g. using non-ed25519 signatures), it's up to the rollup developer to manually make sure they are compatible. This is semi-related to the workstream for encoding crypto types in the uiversal schema to allow the SDK to automatically select the correct signature type, encode/decode addresses and public keys etc., since both workstreams will probably require hardcoding an enum of "supported" crypto primitives.

PR #1522, opened 2025-08-12
Add solana offchain message authenticator

theodorebugnet:
# Description

Adds an authenticator module for solana offchain signed messages, and integration tests verifying transaction acceptance with a TestRollup. The transactions are JSON-serialized in order to be user-readable when displayed in the wallet, which is imperfect but functions well enough given that we can't have custom display logic and that it needs to be reversibly serializable.
Solana defines a special header (preamble) for offchain messages, which is what the Ledger app requires, but apparently not all wallets follow it so we support both spec-compliant messages (with the preamble) as well as "raw" signed messages, i.e. where the wallet just signs the provided JSON string and nothing more, nothing less.

At least 700-ish of the new lines are purely tests, and quite a bit more are boilerplate trait implementations, so actual functionality added is not as big as the line count would suggest.

Changes to highlight:
 * Implements a second `accept_solana_offchain_tx` endpoint to the sequencer APIs. Defining this in the FullNodeBlueprint for the integration tests required making several sequencer data structures public, whereas previously they were only privately used inside the `sov-sequencer::rest_api` module.
 * Adds serde `Serialize` and `Deserialize` bounds to the Spec trait. This causes some issues with structs generic in S that themselves derive Serialize and Deserialize, which requires sprinkling `#[serde(bound = S::Spec)]` because the Deserialize trait has a lifetime and rustc has a compiler limitation causing errors when the Deserialize bound is duplicated.
     * This is iffy in the `DispatchCall` and `Genesis` macros, where we override the bounds. In rare cases users might have to use `no_default_attrs` if they need to also override serde bounds on their own, but in the vast majority of cases it should be fine.

## Linked Issues
- Related to #1520 

## Testing

## Docs


crates/module-system/module-implementations/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -0,0 +1,484 @@
...
+use sov_modules_api::transaction::{
+    self, AuthenticatedTransactionAndRawHash, TransactionCallable, TxDetails, UnsignedTransaction,
+};
+use sov_modules_api::{
+    CryptoSpec, DispatchCall, GasMeter, GasSpec, MeteredJsonDeserialize,
+    MeteredJsonDeserializeError, MeteredSignature, ProvableStateReader, Spec, TxHash,
+};
+
+/// The application domain for Solana offchain messages (placeholder for now)
+pub const APPLICATION_DOMAIN: [u8; 32] = [0u8; 32];
```
theodorebugnet: Of note: what should we set this to?

theodorebugnet: As per discussion, #1526 will set this to the chain_hash.

crates/module-system/sov-modules-macros/src/dispatch/genesis.rs
```diff
@@ -20,9 +20,14 @@ impl GenesisMacro {
...
     ) -> syn::Result<proc_macro::TokenStream> {
-        let default_attrs = vec![quote::quote! {
-            #[derive(Clone, ::serde::Deserialize, ::serde::Serialize)]
-        }];
+        let default_attrs = vec![
+            quote::quote! {
+                #[derive(Clone, ::serde::Deserialize, ::serde::Serialize)]
+            },
+            quote::quote! {
+                #[serde(bound = "S: sov_modules_api::module::Spec")]
```
preston-evans98: This will break if the generic isn't named `S`. Also, we should probably add it to the list of defaults in the docs comment

preston-evans98: I'm not sure if we already have the constraint that the first generic must be named `S`. If not, maybe we should add a quick check here and throw an error if it has another name? Alternatively, we could get the ident of the first generic and put it into the bound instead of a hard-coded `S`

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -0,0 +1,483 @@
...
+    pub runtime_call: R::Call,
+    /// The uniqueness identifier
+    pub uniqueness: UniquenessData,
+    /// Data related to fees and gas handling.
+    pub details: TxDetails<S>,
+    /// The chain hash the transaction data was signed with. Because we're not actually using the
+    /// schema to display the transaction, we have to include and validate the hash explicitly.
+    #[serde_as(as = "serde_with::hex::Hex")]
+    pub chain_hash: [u8; 32],
+}
```
preston-evans98: Had we decided that this chain_hash was no longer required?

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -0,0 +1,483 @@
...
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct SolanaOffchainSimpleMessage<S: Spec> {
+    /// The message is a JSON-serialized SolanaOffchainUnsignedTransaction, unaltered.
+    pub signed_message: Vec<u8>,
+    pub pubkey: <S::CryptoSpec as CryptoSpec>::PublicKey,
+    pub signature: <S::CryptoSpec as CryptoSpec>::Signature,
+}
+
+/// The length of a preamble with a single 32-byte signer.
+pub const PREAMBLE_LEN: usize = std::mem::size_of::<RawSolanaOffchainMessagePreamble>();
```
preston-evans98: I don't think Rustc guarantees that `size_of` will be consistent across platforms or compiler versions, since it depends on the struct packing which is not guaranteed to be stable. This looks like it's working for now, but we may want to revisit if we run into issues.

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -0,0 +1,483 @@
...
+    }
+    // Format 0 is the ASCII, hw-wallet compatible format
+    if preamble.message_format != 0 {
+        return Err(FatalError::DeserializationFailed(
+            "Invalid message format in preamble".to_string(),
+        ));
+    }
+    if preamble.signer_count != 1 {
+        return Err(FatalError::DeserializationFailed(
+            "Invalid signer count in preamble".to_string(),
```
preston-evans98: nit: Let's include the invalid item

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -0,0 +1,483 @@
...
+    }
+    // 0 is the only supported header version
+    if preamble.header_version != 0 {
+        return Err(FatalError::DeserializationFailed(
+            "Invalid header version in preamble".to_string(),
+        ));
+    }
+    if preamble.application_domain != APPLICATION_DOMAIN {
+        return Err(FatalError::DeserializationFailed(
+            "Invalid application domain in preamble".to_string(),
```
preston-evans98: nit: Let's include the invalid domain and the expected domain

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -0,0 +1,483 @@
...
+    if expected_length != actual_message_length {
+        return Err(FatalError::DeserializationFailed(format!(
+            "Message length mismatch: expected {expected_length}, got {actual_message_length}"
+        )));
+    }
+
+    Ok(())
+}
+
+// TODO: json_bytes and signed_bytes will have a good amount of duplication, maybe we can reduce
```
preston-evans98: (Question): What's the status of this todo?

crates/module-system/sov-modules-api/src/gas/metered_utils.rs
```diff
@@ -247,6 +247,67 @@ pub trait MeteredBorshDeserialize<S: Spec>: Sized {
...
+    /// A gas error was raised when trying to deserialize the data.
+    #[error("A gas error was raised when trying to deserialize the data, {0}")]
+    GasError(GasMeteringError<GU>),
+    /// An io error occurred while deserializing the data.
+    #[error("IO error: {0}")]
+    SerdeError(serde_json::Error),
+}
+
+/// Charges gas for deserialization.
+pub trait MeteredJsonDeserialize<S: Spec>: Sized {
```
bkolad, line 262: I don't see any implementation for this trait, is it needed?

crates/module-system/sov-modules-api/src/transaction/mod.rs
```diff
@@ -310,6 +310,7 @@ impl<R: TransactionCallable, S: Spec> Transaction<R, S> {
 #[derive(
     derive_more::Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize, UniversalWallet,
 )]
+#[serde(bound = "R::Call: serde::Serialize + serde::de::DeserializeOwned")]
```
bkolad, line 313: Can't comment there (the file just above) but [solana_auth.rs](https://github.com/Sovereign-Labs/sovereign-sdk/pull/1522/files#diff-e966d305491dba02ad4ea4296f6963945f7d5f7015bb18510437da1046b699ab) is empty.

crates/module-system/sov-modules-api/src/transaction/mod.rs
```diff
@@ -379,6 +380,11 @@ impl<R: TransactionCallable, S: Spec> UnsignedTransaction<R, S> {
             self.details,
         )
     }
+
+    /// Returns a copy of the RuntimeCall from this unsigned transaction.
+    pub fn call(&self) -> R::Call {
```
bkolad, line 385: Looks like this method is unused

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -0,0 +1,483 @@
...
+            details: self.details,
+        }
+    }
+
+    fn unmetered_deserialize_inner(buf: &mut &[u8]) -> Result<Self, serde_json::Error> {
+        serde_json::from_slice::<SolanaOffchainUnsignedTransaction<R, S>>(buf)
+    }
+}
+
+impl<R, S> MeteredJsonDeserialize<S> for SolanaOffchainUnsignedTransaction<R, S>
```
bkolad: Why do we need the `MeteredJsonDeserialize` trait? The only usage I’ve seen is
`<SolanaOffchainUnsignedTransaction<D, S> as MeteredJsonDeserialize<S>>::deserialize`,
so it seems like we could just use the `SolanaOffchainUnsignedTransaction` type directly.

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -0,0 +1,483 @@
...
+    pub header_version: u8,
+    pub application_domain: [u8; 32],
+    pub message_format: u8,
+    pub signer_count: u8,
+    pub signer: [u8; 32],
+    pub message_length: [u8; 2],
+}
+
+/// Validates a Solana offchain message preamble
+fn validate_preamble(
```
bkolad: Nit, but this could be a method on `RawSolanaOffchainMessagePreamble` and the name could be shortened to `validate`

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -0,0 +1,483 @@
...
+            .map_err(MeteredJsonDeserializeError::SerdeError)
+    }
+}
+
+/// The envelope for a signed spec-compliant solana offchain message, where the signed message
+/// includes the preamble.
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct SolanaOffchainSpecCompliantMessage<S: Spec> {
+    /// The message is a JSON-serialized SolanaOffchainUnsignedTransaction with the standard preamble prepended.
+    pub signed_message: Vec<u8>,
```
bkolad: Let's rename to `signed_message_with_preamble`

crates/module-system/sov-solana-offchain-auth/src/capabilities.rs
```diff
@@ -0,0 +1,170 @@
...
+        <Self::Auth as TransactionAuthenticator<S>>::encode_authenticator_input(
+            &Self::add_solana_offchain_auth(tx),
+        )
+    }
+}
+
+/// See [`TransactionAuthenticator::Input`].
+#[derive(std::fmt::Debug, Clone, BorshDeserialize, BorshSerialize)]
+// #[borsh(use_discriminant=true)]
+// #[repr(u8)]
```
bkolad: Commented code

crates/module-system/sov-solana-offchain-auth/src/capabilities.rs
```diff
@@ -0,0 +1,170 @@
...
+use serde::Serialize;
+use sov_modules_api::capabilities::{
+    AuthenticationError, BatchFromUnregisteredSequencer, FatalError, TransactionAuthenticator,
+    UnregisteredAuthenticationError,
+};
+use sov_modules_api::{DispatchCall, FullyBakedTx, ProvableStateReader, RawTx, Runtime, Spec};
+
+/// Indicates that a runtime supports the `SolanaOffchain` transaction authenticator
+/// and provides suitable methods for encoding and decoding solana offchain message transactions.
+pub trait SolanaOffchainAuthenticatorTrait<S: Spec>: Runtime<S> {
```
bkolad, line 15: At some point, we discussed the idea of unifying all authenticators (though I may have misunderstood). Is there a way to achieve this functionality without implementing these additional traits?

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
theodorebugnet: Sorry for the confusion - those changes are still in #1526! I don't want to merge now since you've already reviewed this - on re-review, feel free to press merge there and just re-review the combined branch, or have it as a follow-up.

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
theodorebugnet: That's a great point, thanks for spotting!
The preamble size is constant, so I'll just hardcode it to 85 then, rather than trying to be too fancy.

crates/module-system/sov-modules-macros/src/dispatch/genesis.rs
theodorebugnet: Yeah good point.
This compiles just fine with `#[serde(bound = "")]` and tests seem to pass without issues. What do you think about setting it to this temporarily, and then improving the macro to generate correct bounds as a follow-up?

crates/module-system/sov-modules-api/src/transaction/mod.rs
theodorebugnet: Whoops haha

crates/module-system/sov-modules-api/src/transaction/mod.rs
theodorebugnet: I do use it, in `decode_solana_json_tx()`.

crates/module-system/sov-solana-offchain-auth/src/capabilities.rs
theodorebugnet: Not trivially, so I left it out of scope of this change and for now implemented this in the current way. Unifying them will definitely be nice in the future but it'll be separate work

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
theodorebugnet: Thanks for spotting - I forgot to come back to it! Refactored this to use a slice instead of copying the entire transaction twice over.

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
theodorebugnet: Simplified as per discussion on slack

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
theodorebugnet: Merged that PR in for cleanup - this should leave only one PR to review.

crates/module-system/sov-modules-macros/src/dispatch/genesis.rs
theodorebugnet: As discussed on slack, I added better bounds generation to the genesis and dispatchcall macros.
It's not 100% perfect in case users add a complicated generic that itself also needs overriding (and thus conflicts with our `#[serde(bounds)]` attribute), but at that point they can escape with `no_default_attrs` or we can look into improving the devex if and when someone starts having issues; for now this fix should cover the vast majority of usecases IMO.

crates/module-system/sov-solana-offchain-auth/src/capabilities.rs
```diff
@@ -0,0 +1,170 @@
...
+    SolanaOffchain(T),
+}
+
+/// Solana offchain message compatible transaction authenticator. See [`TransactionAuthenticator`].
+pub struct SolanaOffchainAuthenticator<S, Rt>(PhantomData<(S, Rt)>);
+
+impl<S, Rt> TransactionAuthenticator<S> for SolanaOffchainAuthenticator<S, Rt>
+where
+    S: Spec,
+    // S::Address: FromVmAddress<Base58Address>,
```
bkolad: commented code

bkolad (approved):

PR #1525, opened 2025-08-13
Speed up `seq_many_invalid_tx` by adjusting retry logic

citizen-stig:
# Description

Test `seq_many_invalid_tx` uses submit method with retry, which leads to very long test time, because default exponential backoff total waiting time is going to be a little bit more than 5 minutes. 

This PR changes that, and in order to keep closer to original intention:

* It will retry on most of the errors, except bad request
* It increases number of txs 10x to increase load on the sequencer.
* It removes creating block inside a task, so there's less between tasks, which on my machine lead 10x speed up in the test.

## Testing
Test is passing

## Docs
No updates

### DIFF
--- a/crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
+++ b/crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
@@ -1723,10 +1723,7 @@ async fn flaky_seq_back_pressure() {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn seq_many_invalid_txs() {
-    sov_test_utils::logging::initialize_or_change_logging_with_filter(
-        "warn,sov_metrics=error,sov=debug,integration=debug",
-    );
-    let txs = 100u64;
+    let txs = 1000u64;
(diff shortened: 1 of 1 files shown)


crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
```diff
@@ -1745,36 +1747,88 @@ async fn seq_many_invalid_txs() {
...
+                da_service.produce_block_now().await.unwrap();
+            }
+            assert!(res.is_err(), "Request has been accepted, when it shouldn't");
         }));
     }
 
     test_rollup
         .da_service
-        .produce_n_blocks_now(50)
+        .produce_n_blocks_now(txs as usize / 2)
```
preston-evans98, line 1819: Why do we change the number of blocks here?

citizen-stig: Before it was 50 when there was 100 tx, so I decided to keep the ratio. 

There was no explanation about the previous number. I can revert it back to have hard coded number

bkolad (approved):

PR #1538, opened 2025-08-18
Partial REVM/RETH upgrade

LogvinovLeon:
# Description

This PR is a first in a series of bringing EVM tooling up to the newest versions.
It upgrades RETH over two versions (~1 month)

Because of how interconnected all those deps are - I pinned some versions for now and I always move them into exact versions that work with each other. I'll unpin them when the full migration is done. This helps cargo resolution process

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


bkolad (approved):

crates/module-system/module-implementations/sov-evm/src/evm/conversions.rs
```diff
@@ -68,24 +71,24 @@ pub enum RlpConversionError {
...
     type Error = RlpConversionError;
 
     fn try_from(evm_tx: RlpEvmTransaction) -> Result<Self, Self::Error> {
-        let tx = convert_to_transaction_signed(evm_tx)?;
+        let tx: TransactionSigned = convert_to_transaction_signed(evm_tx)?;
         let tx = tx
-            .into_ecrecovered()
-            .ok_or(RlpConversionError::InvalidSignature)?;
+            .try_into_recovered()
+            .map_err(|_| RlpConversionError::InvalidSignature)?;
```
bkolad, line 91: Maybe it would be good to put the error `|_|` in `InvalidSignature` so we know why it failed

crates/module-system/module-implementations/sov-evm/src/evm/primitive_types.rs
```diff
@@ -78,19 +90,60 @@ impl SealedBlock {
     }
 }
 
+impl serde::Serialize for SealedBlock {
```
bkolad, line 95: Why can't we derive it?

crates/module-system/module-implementations/sov-evm/src/evm/primitive_types.rs
```diff
@@ -78,19 +90,60 @@ impl SealedBlock {
...
+            header: Sealed::new_unchecked(header, seal),
+            transactions,
+        })
+    }
+}
+
+#[serde_as]
 #[derive(Debug, PartialEq, Clone, serde::Serialize, serde::Deserialize)]
 pub struct Receipt {
+    #[serde_as(as = "ReceiptBincodeCompat")]
```
bkolad, line 142: Let's add a comment why do we need this

crates/module-system/module-implementations/sov-evm/src/evm/conversions.rs
LogvinovLeon: The return type is `Result<Recovered<Self>, Self>` so there is no additional information. They just return the tx itself

crates/module-system/module-implementations/sov-evm/src/evm/primitive_types.rs
LogvinovLeon: We can't because `serde_with` doesn't support telling - how to encode the nested type.
Here we have a field of the type Sealed<Header> and we want the Header to be encoded as BincodeCompatHeader, but Sealed itself doesn't have the adapter and it's just generic over header

PR #1557, opened 2025-08-24
EVM: Decouple `evm` and `sov` nonce

bkolad:
# Description

1. Adds a new` EthereumAddress::as_credential_id` method.
2. Refactors `fn extract_evm_authorization_data and fn get_transaction_count` to use the new method.
3. Introduces  `Uniqueness` module to the EVM. With this change, the `SOV` nonce and the `EVM` nonce are decoupled. The EVM nonce is now maintained solely for supporting the `CRATE` opcode.

## Linked Issues
- Related to # (issue)

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/call.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/call.rs
@@ -53,8 +53,12 @@ where
         let cfg = self.cfg(state)?.expect("Evm config must be set");
         let cfg_env = get_cfg_env(&block_env, cfg, None);
 
+        let nonce_to_use = self.get_nonce_to_use(signer, state)?;
+
         let evm_db: EvmDb<_, S> = self.get_db(state);
-        let result = executor::execute_tx(evm_db, &block_env, &evm_tx, signer, cfg_env);
+
+        let result =
+            executor::execute_tx(nonce_to_use, evm_db, &block_env, &evm_tx, signer, cfg_env);
 
         let previous_transaction = self.pending_transactions.last(state)?;
         let previous_transaction_cumulative_gas_used = previous_transaction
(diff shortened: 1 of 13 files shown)


crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -53,8 +53,12 @@ where
         let cfg = self.cfg(state)?.expect("Evm config must be set");
         let cfg_env = get_cfg_env(&block_env, cfg, None);
 
+        let nonce_to_use = self.get_nonce_to_use(signer, state)?;
```
LogvinovLeon: Why nonce_to_use and not just nonce?
If we want to distinguish between sov nonce and evm nonce maybe let's call it sov_nonce?

crates/module-system/module-implementations/sov-evm/src/evm/executor.rs
```diff
@@ -31,13 +31,14 @@ pub(crate) fn get_cfg_env(
 
 /// Execute an Ethereum transaction and commit it to the database.
 pub fn execute_tx<DB: Database<Error = Infallible> + DatabaseCommit>(
+    nonce_to_use: u64,
```
LogvinovLeon: Let's keep this interface clean and assign this new nonce into `tx`. That way this function doesn't need to know about this magic

LogvinovLeon (approved):

crates/module-system/module-implementations/sov-evm/src/evm/executor.rs
bkolad: `TransactionSigned` is an enum and doesn’t provide a way to override the nonce. Adding this would require additional plumbing code, which I believe introduces more complexity than simply adding the argument here.
For example `TransactionSigned::nonce` is implemented by a macro.

crates/module-system/module-implementations/sov-evm/src/call.rs
bkolad: Fixed

LogvinovLeon (approved):

PR #1558, opened 2025-08-25
`Bank`: Add support for querying DerivedHolder balance

bkolad:
# Description

1. Removes unused methods from `Bank`
2. Extends `route_balance` to allow querying `User/Module/DerivedHolder` balances
3. Extends a unit test to check if we can pass `DerivedHolder` to the balance endpoint.

Note: The `Bank::Transfer` call message currently only accepts user addresses, which limits testing and the bank functionality. We’ll revisit this on the daily call to decide whether to extend `Bank::Transfer` to support any token holder.

### DIFF
--- a/crates/module-system/module-implementations/sov-bank/src/query.rs
+++ b/crates/module-system/module-implementations/sov-bank/src/query.rs
@@ -105,11 +105,11 @@ impl<S: Spec> HasCustomRestApi for Bank<S> {
         axum::Router::new()
             .route("/tokens/gas_token", get(Self::route_gas_token))
             .route(
-                "/tokens/gas_token/balances/:address",
+                "/tokens/gas_token/balances/:holderStr",
                 get(Self::route_gas_token_balance),
             )
             .route(
-                "/tokens/:tokenId/balances/:address",
+                "/tokens/:tokenId/balances/:holderStr",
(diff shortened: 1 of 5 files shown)


crates/module-system/module-implementations/sov-bank/src/query.rs
```diff
@@ -74,22 +36,31 @@ impl<S: Spec> Bank<S> {
...
+                let amount = state
+                    .balances
+                    .get(&(holder.clone(), &token_id), &mut accessor)
+                    .unwrap_infallible()
+                    .ok_or_else(|| errors::not_found_404("Balance", holder))?;
+
+                Ok(Coins { amount, token_id }.into())
+            }
+            Err(err) => Err(errors::bad_request_400(
+                &format!("The given holder {holder_str} is not convertible to a TokenHolder"),
```
theodorebugnet: Given that this will be the error provided as the API response, I would assume a majority of people querying it will be querying for a user balance and might not know about the TokenHolder type, maybe we can make it slightly more user-friendly:
```suggestion
                &format!("The given parameter {holder_str} is not a valid address or token holder"),
```
Thoughts?

theodorebugnet (approved):

PR #1563, opened 2025-08-25
fix: MockCodeCommitment error message should say 8 bytes, not 32

MozirDmitriy:
MockCodeCommitment uses [u8; 8] but error message incorrectly stated "32 bytes".
Fixed documentation to match actual implementation.


MozirDmitriy:

PR #1570, opened 2025-08-26
Improve state access metrics

preston-evans98:
# Description

This PR simplifies and fixes several minor bugs with the new state access metrics. At Tristan's request, it also adds logging on slow transactions by default.


crates/full-node/sov-metrics/src/lib.rs
```diff
@@ -71,6 +71,11 @@ impl MaybeTimer {
     pub fn elapsed(&self) -> std::time::Duration {
         std::time::Duration::from_secs(0)
     }
+
+    /// Returns the elapsed time since the timer if the `native` feature is enabled. Otherwise does nothing.
+    pub fn stop_and_get_elapsed(&mut self) -> std::time::Duration {
```
bkolad, line 76: Looks like this method is only invoked from native code, so the stub is unnecessary.

crates/full-node/sov-metrics/src/influx_db_nonnative.rs
```diff
@@ -6,58 +8,160 @@ use crate::MaybeTimer;
...
+                display_fn: None,
+            },
             storage_read_size: None,
             duration: MaybeTimer::Completed(Duration::from_secs(0)),
+            access_type: StateAccessType::GetSize,
         }
     }
 }
 #[cfg(feature = "native")]
+#[allow(dead_code)]
```
bkolad: Do we need the dead_code here?

bkolad (approved):

crates/full-node/sov-metrics/src/lib.rs
preston-evans98: It fails to compile without it - possibly due to test dependencies

crates/full-node/sov-metrics/src/influx_db_nonnative.rs
preston-evans98: Nope, good catch! Leftover from some experimentation

Issue #1510, opened 2025-08-11
[TRACKER] EVM support


preston-evans98:
## Description

Tracks ongoing work on the EVM integration

## Checklist

 - [ ] Add support for EVM soft confirmations
   - [x] Allow pushing to AccessoryStateVec without read access #1503.
   - [x] Handle RPC queries for txs from the in-progress block without panicking
     - [x] Decide how to handle receipts (which need info about the whole block).
     - [x] Investigate possibility of returning receipts without creating a block. Not possible - `eth_getLogs` and `eth_getTransactionReceipt` both [require](https://www.quicknode.com/docs/ethereum/eth_getLogs) block information. 
       - [x] Investigate mini-blocks
    - [x] Implement quering form  in-progress txs
    - [x] Evem reverts
        - https://github.com/Sovereign-Labs/sovereign-sdk/pull/1608
        - https://github.com/Sovereign-Labs/sovereign-sdk/pull/1616
        - https://github.com/Sovereign-Labs/sovereign-sdk/pull/1620    
        - https://github.com/Sovereign-Labs/sovereign-sdk/pull/1632 

If not rejected, check whether the receipt exists in the EVM module.
   - [x] Test deployment with Uniswap or similar contract and Metamask
 - [ ] Indexing support (Mirror)
    - [ ] Make sure EVM block number = rollup height & Support for archival node
    - [x] Test the following  `eth_getBlockByNumber`, `eth_getBlockReceipts`, `eth_getTransactionReceipt` 
    - [x] debug_traceTransaction (We can reuse the EVM components, but not Reth, so we’ll need to implement some additional plumbing.)
 - [x] Integrate `sov-uniqueness` and the evm module. 
         We need to explore how to disable nonce checks in the EVM—possibly through the `Handler` trait or `EvmConfig`.
         - https://github.com/Sovereign-Labs/sovereign-sdk/pull/1557
         - https://github.com/Sovereign-Labs/sovereign-sdk/pull/1587
         - https://github.com/Sovereign-Labs/sovereign-sdk/pull/1588
         - https://github.com/Sovereign-Labs/sovereign-sdk/pull/1590
 - [x] Verify that `sov-bank` and `EVM` balances integrate seamlessly.
         - https://github.com/Sovereign-Labs/sovereign-sdk/pull/1562
         - https://github.com/Sovereign-Labs/sovereign-sdk/pull/1580
[...]

PR #1580, opened 2025-08-27
EVM: Remove balances fro evm genesis.

bkolad:
# Description

`EVM` account balances are now initialized in the `Bank` genesis instead of the `EVM` genesis. 

In the Bank genesis, we enforce that the total of all initial balances does not exceed `Token::supply_cap`. This check was not present in the EVM genesis, which allowed the possibility of overflowing bank balances. After this PR this is no longer possible.


crates/module-system/module-implementations/sov-evm/src/config.rs
```diff
@@ -46,7 +46,7 @@ impl Default for EvmChainSpec {
             coinbase: Address::ZERO,
             block_gas_limit: ETHEREUM_BLOCK_GAS_LIMIT_30M,
             block_timestamp_delta: SLOT_DURATION.as_secs(),
-            base_fee_params: alloy_eips::eip1559::BaseFeeParams::ethereum(),
+            base_fee_params: alloy_eips::eip1559::BaseFeeParams::new(1, 1),
```
bkolad, line 49: I will update #1510 to ensure we initialize all the EVM parameters correctly (these values are not set yet properly). We will set them up once we connect sov-gas

LogvinovLeon (approved):

PR #1587, opened 2025-08-27
EVM: Remove nonce form the evm genesis.

bkolad:
# Description

The nonces are managed by `sov-uniqueness` module, so we should not include them in the EVM genesis.

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/genesis.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/genesis.rs
@@ -21,8 +21,6 @@ pub struct AccountData {
     pub code_hash: B256,
     /// Smart contract code.
     pub code: Bytes,
-    /// Account nonce.
-    pub nonce: u64,
 }
 
 impl AccountData {
@@ -69,7 +67,7 @@ where
             AccountInfo {
                 balance: U256::ZERO,
                 code_hash: acc.code_hash,
-                nonce: acc.nonce,
+                nonce: 0,
                 code: None,
             },
         );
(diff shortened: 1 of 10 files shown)


crates/module-system/module-implementations/sov-evm/src/genesis.rs
```diff
@@ -69,7 +67,7 @@ where
             AccountInfo {
```
LogvinovLeon, line 67: AccountInfo::new(U256::ZERO, 0, acc.code_hash, None),

crates/module-system/module-implementations/sov-evm/tests/integration/genesis.rs
```diff
@@ -26,7 +26,7 @@ fn test_genesis_data() {
             &AccountInfo {
```
LogvinovLeon, line 26: Same

LogvinovLeon (approved):

crates/module-system/module-implementations/sov-evm/src/genesis.rs
bkolad: For some reason, `AccountInfo::new(balance: U256, nonce: u64, code_hash: B256, code: Bytecode)`, the constructor takes Bytecode instead of Option<Bytecode>, so we can't use it here.

crates/module-system/module-implementations/sov-evm/tests/integration/genesis.rs
bkolad: Same

PR #1588, opened 2025-08-28
EVM: cleanup + nonce test

bkolad:
# Description

This PR includes the following changes:
1. General cleanup and renaming
2. Updates `validate_against_state_and_deduct_caller` to handle nonce updates (based on the original revm implementation)
3.  Adds a nonce test

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/tests/integration/helpers.rs
+++ b/crates/module-system/module-implementations/sov-evm/tests/integration/helpers.rs
@@ -93,7 +93,7 @@ pub(crate) fn create_transfer_tx(
         nonce,
         ..Default::default()
     };
-    create_evm_tx(from, tx)
+    create_tx(from, tx)
 }
 
 pub(crate) fn create_deploy_tx(
@@ -106,7 +106,7 @@ pub(crate) fn create_deploy_tx(
         nonce,
         ..Default::default()
     };
-    create_evm_tx(account, tx)
+    create_tx(account, tx)
 }
 
 pub(crate) fn create_set_arg_tx(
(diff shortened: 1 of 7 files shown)


crates/module-system/module-implementations/sov-evm/tests/integration/helpers.rs
```diff
@@ -93,7 +93,7 @@ pub(crate) fn create_transfer_tx(
         nonce,
         ..Default::default()
     };
-    create_tx(from, tx)
+    create_evm_tx(from, tx)
```
LogvinovLeon: We're inside the EVM module so all tx's here are EVM tx's and therefore we can use the shorted name

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
```diff
@@ -100,3 +100,33 @@ fn test_failed_tx_doesnt_update_evm_module_state() {
...
+    let (mut runner, from, to) = setup();
+
+    let from_addr = from.address();
+    let value = 1;
+    let transfer_tx = create_transfer_tx(0, &from, &to, value);
+
+    runner.execute_transaction(TransactionTestCase {
+        input: transfer_tx,
+        assert: Box::new(move |_ctx, state| {
+            let evm = Evm::<S>::default();
```
LogvinovLeon: evm can be created once outside

citizen-stig (approved):

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
```diff
@@ -100,3 +100,33 @@ fn test_failed_tx_doesnt_update_evm_module_state() {
...
+            let mut db = evm.get_db(state);
+            let from_acc = db.basic(from_addr).unwrap().unwrap();
```
LogvinovLeon, line 117: BTW, is there a good reason beyond historical why TransactionTestCase assertion closure can't return Result so that we can ? instead of unwrap()? Seems like we unwrap quite a bit inside those tests. I understand - it's a bigger refactor as it's not only inside the EVM

LogvinovLeon (approved):

crates/module-system/module-implementations/sov-evm/tests/integration/helpers.rs
bkolad: fixed

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
bkolad: fixed

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
bkolad: I think we wanted to handle all the issues in one place inside the clojure. In principle, Result could work.

PR #1589, opened 2025-08-28
Metered state access in EVM DB with errors propagation

LogvinovLeon:
# Description

### Changes
  - **Database refactoring**: Moved EVM DB code into dedicated `db` submodule for better
  organization
  - **Metered state access**: Integrated metered state accessor throughout EVM components to
  enable resource monitoring
  - **Error propagation**: Enhanced error handling by propagating state access errors through
  the Database interface
  - **Code cleanup**: Simplified database commit logic and derived trivial constructors where
  applicable

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/db/commit.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/db/commit.rs
@@ -10,6 +10,7 @@ use sov_modules_api::prelude::UnwrapInfallible;
 use sov_modules_api::{InfallibleStateAccessor, Spec};
 
 use super::EvmDb;
+use crate::db::DbAccount;
 use crate::{to_rollup_address, to_rollup_balance};
 
 impl<Ws: InfallibleStateAccessor, S: Spec> DatabaseCommit for EvmDb<Ws, S>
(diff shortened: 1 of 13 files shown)


crates/module-system/module-implementations/sov-evm/src/db/commit.rs
```diff
@@ -0,0 +1,81 @@
...
+
+        let mut account = account.info;
+
+        self.bank_module
+            .override_gas_balance(
+                to_rollup_balance(account.balance),
+                &to_rollup_address::<S>(address),
+                self.state,
+            )
+            .expect("Failed to override gas balance");
```
bkolad: The issue is that we can’t panic here, but `EvmDb::commit` doesn’t allow returning early with `?`.
I think we’ll need to do the following:

1. Add a field to `EvmDb: early_finish: Option<Error>`.
2. If we encounter an error in this method, return from `commit` and set `early_finish`.
3. In `call.rs`, handle this error accordingly.

crates/module-system/module-implementations/sov-evm/src/db/commit.rs
```diff
@@ -0,0 +1,96 @@
...
+use sov_address::{EthereumAddress, FromVmAddress};
+use sov_modules_api::{Spec, StateAccessor};
+
+use super::EvmDb;
+use crate::db::{DbAccount, Error};
+use crate::{to_rollup_address, to_rollup_balance};
+
+/// EVM database commit interface.
+#[auto_impl(&mut, Box)]
+pub trait FallibleDatabaseCommit {
```
bkolad, line 14: Is the trait still necessary, given that `EvmDb` is the only implementation?

crates/module-system/module-implementations/sov-evm/src/db/mod.rs
```diff
@@ -1,72 +1,66 @@
...
-
-impl<Ws: InfallibleStateAccessor, S: Spec> Database for EvmDb<Ws, S>
+impl<'a, Ws: StateAccessor, S: Spec> Database for EvmDb<'a, Ws, S>
 where
     S::Address: FromVmAddress<EthereumAddress>,
 {
-    type Error = Infallible;
+    type Error = Error;
 
     fn basic(&mut self, address: Address) -> Result<Option<AccountInfo>, Self::Error> {
```
bkolad, line 51: What if we remove the custom `Error` and instead use `<Accessor as StateReader<User>>::Error`?
It seems that  all of the failure cases come from state access.
I also don’t expect us to ever need to pattern match on specific error variants.

crates/module-system/module-implementations/sov-evm/src/db/commit.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn commit(&mut self, changes: HashMap<Address, Account>) -> Result<(), Self::Error>;
+}
+
+impl<'a, Ws: StateAccessor, S: Spec> FallibleDatabaseCommit for EvmDb<'a, Ws, S>
+where
+    S::Address: FromVmAddress<EthereumAddress>,
+{
+    type Error = Error<Ws>;
+
+    fn commit(&mut self, changes: HashMap<Address, Account>) -> Result<(), Error<Ws>> {
```
bkolad: Let’s use `Self::Error` here and elsewhere. That way, if we ever want to change the error type, we only need to update it in `type Error = `

bkolad (approved):

PR #1590, opened 2025-08-28
`EVM`: Add test to verify that contracts deployed by the same sender are assigned different addresses

bkolad:
# Description

Like in the title

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
+++ b/crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
@@ -131,3 +131,69 @@ fn test_account_nonce() {
         }),
     });
 }
+
+// Check that if the same account deploys two contracts, each deployment results in a unique contract address
+#[test]
+fn test_deploy_many_contracts() {
+    let (mut runner, account, _) = setup();
+    let contract = SimpleStorageContract::default();
+    let contract_addr_1 = account.address().create(0);
+
+    // Deploy a contract and execute two transactions that update its storage.
+    {
+        let create_contract_tx_1 = create_deploy_tx(0, &contract, &account);
+        let set_value_tx_1 = create_set_arg_tx(5, 1, &contract, contract_addr_1, &account);
+        let set_value_tx_11 = create_set_arg_tx(7, 2, &contract, contract_addr_1, &account);
+
+        let evm = Evm::<S>::default();
+        runner.execute_batch(BatchTestCase {
+            input: vec![create_contract_tx_1, set_value_tx_1, set_value_tx_11].into(),
+            assert: Box::new(move |_ctx, state| {
+                let storage_value_1 = evm
+                    .get_storage(&contract_addr_1, &U256::ZERO, state)
+                    .unwrap()
+                    .unwrap();
+
+                assert_eq!(U256::from(7), storage_value_1);
+            }),
+        });
+    }
+
+    // Deploy another contract and execut single transaction that updates its storage.
+    {
+        let contract_addr_2 = account.address().create(3);
+        let create_contract_tx_2 = create_deploy_tx(3, &contract, &account);
+        let set_value_tx_2 = create_set_arg_tx(99, 4, &contract, contract_addr_2, &account);
+
+        let evm = Evm::<S>::default();
+        runner.execute_batch(BatchTestCase {
+            input: vec![create_contract_tx_2, set_value_tx_2].into(),
+            assert: Box::new(move |_ctx, state| {
+                // Contract have different addresses.
+                assert_ne!(contract_addr_1, contract_addr_2);
+
+                let mut db = evm.get_db(state);
+                let contract_1_account = db.basic(contract_addr_1).unwrap().unwrap();
+                let contract_2_account = db.basic(contract_addr_2).unwrap().unwrap();
+
+                // The contracts have the same code.
+                assert_eq!(contract_1_account.code_hash, contract_2_account.code_hash);
+
+                let storage_value_1 = evm
+                    .get_storage(&contract_addr_2, &U256::ZERO, state)
+                    .unwrap()
+                    .unwrap();
+
+                assert_eq!(U256::from(99), storage_value_1);
+
+                // The storage of the first contract didn't change.
+                let storage_value_1 = evm
+                    .get_storage(&contract_addr_1, &U256::ZERO, state)
+                    .unwrap()
+                    .unwrap();
+
+                assert_eq!(U256::from(7), storage_value_1);
+            }),
+        });
+    }
+}
--- a/crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
+++ b/crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
@@ -139,61 +139,51 @@ fn test_deploy_many_contracts() {
     let contract = SimpleStorageContract::default();
     let contract_addr_1 = account.address().create(0);
 
-    // Deploy a contract and execute two transactions that update its storage.
-    {
-        let create_contract_tx_1 = create_deploy_tx(0, &contract, &account);
-        let set_value_tx_1 = create_set_arg_tx(5, 1, &contract, contract_addr_1, &account);
-        let set_value_tx_12 = create_set_arg_tx(7, 2, &contract, contract_addr_1, &account);
+    let create_contract_tx_1 = create_deploy_tx(0, &contract, &account);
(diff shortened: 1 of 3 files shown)


crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
```diff
@@ -131,3 +131,69 @@ fn test_account_nonce() {
...
+fn test_deploy_many_contracts() {
+    let (mut runner, account, _) = setup();
+    let contract = SimpleStorageContract::default();
+    let contract_addr_1 = account.address().create(0);
+
+    // Deploy a contract and execute two transactions that update its storage.
+    {
+        let create_contract_tx_1 = create_deploy_tx(0, &contract, &account);
+        let set_value_tx_1 = create_set_arg_tx(5, 1, &contract, contract_addr_1, &account);
+        let set_value_tx_12 = create_set_arg_tx(7, 2, &contract, contract_addr_1, &account);
```
LogvinovLeon: Why do we need this second transaction?

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
```diff
@@ -131,3 +131,69 @@ fn test_account_nonce() {
...
+
+        let evm = Evm::<S>::default();
```
LogvinovLeon: Define once outside

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
```diff
@@ -131,3 +131,69 @@ fn test_account_nonce() {
...
+        let contract_addr_2 = account.address().create(3);
+        let create_contract_tx_2 = create_deploy_tx(3, &contract, &account);
+        let set_value_tx_2 = create_set_arg_tx(99, 4, &contract, contract_addr_2, &account);
+
+        let evm = Evm::<S>::default();
+        runner.execute_batch(BatchTestCase {
+            input: vec![create_contract_tx_2, set_value_tx_2].into(),
+            assert: Box::new(move |_ctx, state| {
+                // The two contracts have different addresses.
+                assert_ne!(contract_addr_1, contract_addr_2);
```
LogvinovLeon: This assertion doesn't test our code but tests the library.
It would make sense if we assert actual addresses of created contracts, but not statically precomputed addresses

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
```diff
@@ -131,3 +131,69 @@ fn test_account_nonce() {
...
+    }
+
+    // Deploy another contract and execut single transaction that updates its storage.
+    {
+        let contract_addr_2 = account.address().create(3);
+        let create_contract_tx_2 = create_deploy_tx(3, &contract, &account);
+        let set_value_tx_2 = create_set_arg_tx(99, 4, &contract, contract_addr_2, &account);
+
+        let evm = Evm::<S>::default();
+        runner.execute_batch(BatchTestCase {
```
LogvinovLeon: There doesn't seem to be any value in doing this in two batches. Let's combine into one batch.

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
```diff
@@ -131,3 +131,69 @@ fn test_account_nonce() {
...
+                assert_ne!(contract_addr_1, contract_addr_2);
+
+                let mut db = evm.get_db(state);
+                let contract_1_account = db.basic(contract_addr_1).unwrap().unwrap();
+                let contract_2_account = db.basic(contract_addr_2).unwrap().unwrap();
+
+                // The two contracts have the same code.
+                assert_eq!(contract_1_account.code_hash, contract_2_account.code_hash);
+
+                let storage_value_1 = evm
```
LogvinovLeon: I think this test would be much simpler and still as useful if we just check that the addresses are different and don't submit any set_value tx's. If addresses are different - then obviously those are two different contracts and we don't need to differentiate by observing behavior.

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
```diff
@@ -131,3 +131,69 @@ fn test_account_nonce() {
...
+                assert_eq!(U256::from(7), storage_value_1);
+            }),
+        });
+    }
+
+    // Deploy another contract and execut single transaction that updates its storage.
+    {
+        let contract_addr_2 = account.address().create(3);
+        let create_contract_tx_2 = create_deploy_tx(3, &contract, &account);
+        let set_value_tx_2 = create_set_arg_tx(99, 4, &contract, contract_addr_2, &account);
```
LogvinovLeon: Let's try to use simplest values. If we need two values - let's use 1 and 2. Then later the reader sees that those are obviously dummy values that don't have any additional semantics.
With 99 the reader might think - why did they choose this particular number?

LogvinovLeon (changes requested):

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
bkolad: EVM does not implement `Copy`, so we either need to clone it or instantiate it again. I chose to instantiate it again since the module itself is stateless.

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
bkolad: I wanted to test the case when bach has more than 1 tx.

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
bkolad: Later, we call get_storage with contract_addr_1 and contract_addr_2. I want to be absolutely sure that these calls are made to different addresses, regardless of how the addresses are computed. 

The addresses change on each test run.

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
bkolad: We updated the `DB::commit` method and the `EVM` handlers. Since changes to EVM behavior aren’t always obvious, keeping this assertion seems useful. Verifying that calls to different instances of the same contract don’t overwrite the same storage slot is also well within the scope of a contract deployment test.

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
bkolad: fixed

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
bkolad: fixed

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
LogvinovLeon: If we were testing opaque contracts that can use proxy patterns - sure. But here we test our dummy contract and we know it doesn't do anything like that.
In this test we want to check that we do infact increment nonce on contract deployment and we can check that by just checking the resulting addresses. That way if we ever break that behavior - we have a small test that tests just that and it's easier to locate and debug and reason about

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
LogvinovLeon: Why? How does that relate to the behavior that we test in this test case?
Here we test that we increment nonce on create. If we do but only at the end of the batch - then it will still fail even without those transactions

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
bkolad: If `contract_addr_1 == contract_addr_2`, then the assertion
assert_eq!(contract_1_account.code_hash, contract_2_account.code_hash); holds trivially.
The precondition for the test is that the addresses are diffrent. 

Or is there a way to access the contract account after deployment other than querying `db.basic(contract_addr_1)`?

LogvinovLeon (approved):

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
bkolad: This

PR #1592, opened 2025-08-29
EVM: Prune AccessoryState

bkolad:
# Description

This PR introduces a pruning mechanism to the EVM module. Once a defined threshold is reached, the following values are pruned:
```
pub(crate) blocks: AccessoryStateVec<SealedBlock, BcsCodec>,
pub(crate) block_hashes: AccessoryStateMap<B256, u64, BcsCodec>,
pub(crate) transactions: AccessoryStateVec<TransactionSignedAndRecovered, BcsCodec>,
pub(crate) transaction_hashes: AccessoryStateMap<B256, u64, BcsCodec>,
```

### DIFF
--- a/constants.testing.toml
+++ b/constants.testing.toml
@@ -10,9 +10,9 @@ freeze = [1, 1]
 CHAIN_ID = 4321
 CHAIN_NAME = "TestChain"
 # We will keep only this many blocks in the EVM module.
-EVM_BLOCK_PRUNING_THRESHOLD = 5
+EVM_BLOCK_PRUNING_THRESHOLD = 1000
 # We will keep only this many transactions in the EVM module.
-EVM_TRANSACTION_PRUNING_THRESHOLD = 20
+EVM_TRANSACTION_PRUNING_THRESHOLD = 100_000
 # How many rollup blocks to wait before terminating setup mode. While setup mode is enabled,
 # the rollup does not charge gas. This is useful when initializing the rollup with a bridged gas token.
 SETUP_MODE_TERMINATION_HEIGHT = 0 # Disable setup mode entirely by default.
--- a/constants.testing.toml
+++ b/constants.testing.toml
@@ -9,6 +9,10 @@ freeze = [1, 1]
 # We use the ID 4321 for demo purposes. Change this value before deploying!
 CHAIN_ID = 4321
 CHAIN_NAME = "TestChain"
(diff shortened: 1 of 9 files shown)


constants.toml
```diff
@@ -10,6 +10,10 @@ freeze = [1, 1]
 # We use the ID 4321 for demo purposes. Change this value before deploying!
 CHAIN_ID = 4321
 CHAIN_NAME = "TestChain"
+# We will keep only this many blocks in the EVM module.
+EVM_BLOCK_PRUNING_THRESHOLD = 1000
+# We will keep only this many transactions in the EVM module.
+EVM_TRANSACTION_PRUNING_THRESHOLD = 100_000
```
bkolad, line 16: The threshold values are read at compile time.

crates/module-system/module-implementations/sov-evm/src/hooks.rs
```diff
@@ -191,21 +192,14 @@ impl<S: Spec> FinalizeHook for Evm<S> {
...
         let mut block = self
             .pending_head
             .get(state)
             .unwrap_infallible()
             .unwrap_or_else(|| {
-                panic!("Pending head must be set to block {expected_block_number}, but was empty")
+                panic!("The impossible happened: the pending block should always be set.")
             });
 
-        assert_eq!(
```
bkolad, line 204: This assertion no longer holds as blocks can be pruned

constants.testing.toml
```diff
@@ -9,6 +9,10 @@ freeze = [1, 1]
 # We use the ID 4321 for demo purposes. Change this value before deploying!
 CHAIN_ID = 4321
 CHAIN_NAME = "TestChain"
+# We will keep only this many blocks in the EVM module.
+EVM_BLOCK_PRUNING_THRESHOLD = 1000
```
LogvinovLeon, line 13: This seems like a good first step, but if he have two separate values - we can and will end up in a situation where we keep some block where it's transactions are partially prunned.
Would be better to only define the number of blocks we keep and cascade to deleting all the transactions in prunned blocks. Could be done in the next PR.

LogvinovLeon (approved):

constants.testing.toml
bkolad: This is a good point. I will fix it in follow up

PR #1594, opened 2025-08-31
EVM: Prune AccessoryState part 2.

bkolad:
# Description
This is a follow-up to #1592.

The following changes have been implemented:
1. Removing the need for `EVM_TRANSACTION_PRUNING_THRESHOLD`
2. Previously, accessory data was stored in `AccessoryStateVec`. However, with pruning enabled, vector indexing became invalid. For example, if we removed genesis from the vector (at index 0), all subsequent elements would shift, and vec[0] would then incorrectly correspond to block1.

To address this, we replaced `AccessoryStateVec`s with `AccessoryStateMap`s. Unlike vectors, maps avoid this shifting issue. The tradeoff is that we now need to store an additional state value `block_numbers`. 

3. Instead of refactoring `test_transactions_receipts` I removed it and extended `test_executing_eth_transactions` (where the checks belong)


crates/module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -187,52 +192,33 @@ impl<S: Spec> Evm<S> {
...
+    pub fn block_numbers<Accessor: AccessoryStateReaderAndWriter>(
         &self,
         state: &mut Accessor,
-    ) -> Vec<SealedBlock> {
-        self.blocks.collect_infallible(state)
+    ) -> RangeInclusive<u64> {
+        self.block_numbers
+            .get(state)
+            .unwrap_infallible()
+            .expect("Block numbers must be set")
```
bkolad, line 219: This is set at genesis so is safe, but all the expect/unwraps will be tacked in another issue.

crates/module-system/module-implementations/sov-evm/tests/integration/state.rs
```diff
@@ -54,61 +56,3 @@ fn test_block_updates() {
...
-        .into(),
-        assert: Box::new(move |_result, state| {
-            let evm = Evm::<S>::default();
-            let txns = evm.transactions(state);
-            let signed_txns = txns
-                .iter()
-                .map(|tx| tx.signed_transaction().clone())
-                .collect::<Vec<_>>();
-
-            assert_eq!(signed_txns, vec![signed_tx1.clone(), signed_tx2.clone()]);
```
bkolad, line 107: Check from this test ware moved to `test_executing_eth_transactions`

crates/module-system/module-implementations/sov-evm/src/genesis.rs
```diff
@@ -55,7 +55,10 @@ where
         self.cfg.set(&chain_cfg, state)?;
         self.head.set(&block, state)?;
         #[cfg(feature = "native")]
-        self.pending_head.set(&block, state)?;
+        {
+            self.block_numbers.set(&RangeInclusive::new(0, 0), state)?;
```
LogvinovLeon, line 60: Why not: 0..=0?

crates/module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -344,3 +330,17 @@ where
 {
     S::Address::from_vm_address(EthereumAddress::from(address))
 }
+
+pub(crate) fn crate_range_from(
```
LogvinovLeon: What happens when the new_start actually makes the range narrower? I think we should add a test that showcases this

LogvinovLeon: Also - let's keep the main module tidy and move this and to_rollup_address into some utils module.

LogvinovLeon (approved):

crates/module-system/module-implementations/sov-evm/src/genesis.rs
bkolad: I think I prefer explicit syntax more.

crates/module-system/module-implementations/sov-evm/src/lib.rs
bkolad: I moved it to hoks.rs and made it private.
We test the case when start advences here https://github.com/Sovereign-Labs/sovereign-sdk/pull/1594/files#diff-45935130697b7741c0323a2f444513d26fb36562c5f7e1f21cf23fca145db4d0R67

PR #1595, opened 2025-09-01
fix: remove unused import

Fibonacci747:
use std::hash::Hash;
Import isn't needed for #[derive(Hash)] and it's not used anywhere


Fibonacci747:

PR #1596, opened 2025-09-01
Add support for transaction receipts from pending blocks

LogvinovLeon:
# Description

  This PR enables the EVM module to return transaction receipts for transactions that are
  still in pending blocks (not yet sealed). Previously, receipts were only available after
  block finalization.

### Changes
  - Add `LiveTxNumbers` to track current and first transaction numbers in the active block
  - Store transactions and receipts immediately when processed (native mode only)
  - Introduce `MaybeSealedBlock` enum to handle both sealed and pending blocks in RPC
  responses
  - Update receipt retrieval logic to check pending state when sealed block is not
  available

### Benefits
  - Users can get transaction receipts immediately after submission
  - Improves compatibility with Ethereum tooling that expects receipts for pending
  transactions
  - No impact on guest execution (pending block logic disabled in zkVM)

- [ ] I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.
- [ ] I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. (Are all new dependencies necessary? Is any module dependency leaked into the full-node (hint: it shouldn't)?)

## Linked Issues
- Fixes # (issue, if applicable)
- Related to # (issue) 

## Testing
Describe how these changes were tested. If you've added new features, have you added unit tests?

## Docs
Describe where this code is documented. If it changes a documented interface, have the docs been updated?

Rendered docs are available at `https://sovlabs-ci-rustdoc-artifacts.us-east-1.amazonaws.com/<BRANCH_NAME>/index.html`

PR #1608, opened 2025-09-02
Add support for transaction receipts from pending blocks

bkolad:
# Description

This is a port of #1596.

The main differences: 
1. Transactions & Receipts are kept in Maps rather than Vecs
2. No need for `LiveTxNumbers`

Tests are coming in the follow-up 

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/call.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/call.rs
@@ -123,6 +123,35 @@ where
         self.pending_transactions
             .push(&pending_transaction, state)?;
 
+        #[cfg(feature = "native")]
+        {
+            let head = self.head.get(state)?.unwrap();
+            let first_tx_index = head.transactions.end;
+
+            let pending_len = self.pending_transactions.len(state)?;
+
+            let tx_index = first_tx_index + pending_len - 1;
+
+            self.transactions
+                .set(&tx_index, &pending_transaction.transaction, state)?;
+
+            self.receipts
+                .set(&tx_index, &pending_transaction.receipt, state)?;
+
+            let hash = pending_transaction.transaction.signed_transaction.hash();
+            self.transaction_hashes.set(hash, &tx_index, state)?;
+
+            /*
+            self.receipts
+                .push_accessory(&pending_transaction.receipt, state)?;
+            self.transactions
+                .push_accessory(&pending_transaction.transaction, state)?;
+            let hash = pending_transaction.transaction.signed_transaction.hash();
+            self.transaction_hashes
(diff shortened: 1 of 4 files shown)


crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -301,18 +302,17 @@ where
...
-        let tx_number = self.get_tx_index_by_hash(&hash, state);
-
-        let receipt = tx_number.map(|number| {
-            let tx = self.transaction(number, state).unwrap();
-            let block = self.block(tx.block_number, state);
-            let receipt = self.receipt(tx_number.unwrap(), state).unwrap();
-
-            build_rpc_receipt(block, tx, tx_number.unwrap(), receipt)
-        });
+        let tx = self.transaction(number, state).unwrap();
```
bkolad, line 309: We will deal with the unwraps in a separate PR

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -123,6 +124,36 @@ where
...
+        // Fetch `head` and `pending_len` before the `native` code block.
+        // This ensures consistent gas charges between native and non-native execution.
+        #[allow(unused_variables)]
+        let head = self
+            .head
+            .get(state)?
+            .expect("Impossible happened: Head must be set.");
+
+        #[allow(unused_variables)]
+        let pending_len = self.pending_transactions.len(state)?;
```
LogvinovLeon, line 136: If it's only used in native - why not move it down into the block?

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -123,6 +124,36 @@ where
...
+
+        #[cfg(feature = "native")]
+        {
+            let first_tx_index = head.transactions.end;
+            let tx_index = first_tx_index + pending_len - 1;
+
+            self.transactions
+                .set(&tx_index, &pending_transaction.transaction, state)
+                .unwrap_infallible();
```
LogvinovLeon, line 145: This function returns anyhow::Result<()> so we can use ? here which is more concise

LogvinovLeon (approved):

crates/module-system/module-implementations/sov-evm/src/call.rs
bkolad: Will address in folloup

crates/module-system/module-implementations/sov-evm/src/call.rs
bkolad: Will address in folloup

PR #1616, opened 2025-09-03
EVM Add `test_executing_eth_transactions_several_blocks`

bkolad:
# Description

This is a follow up to #1608:
1. Adds `test_executing_eth_transactions_several_blocks` test
2. Cleanup review comments form #1608

## Linked Issues
- Related to #1608

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/call.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/call.rs
@@ -7,7 +7,6 @@ use crate::evm::executor::{self};
 use crate::evm::primitive_types::{Receipt, TransactionSignedAndRecovered};
 use crate::evm::RlpEvmTransaction;
 use crate::executor::get_cfg_env;
-use crate::Block;
 use crate::{Evm, PendingTransaction, SpecId};
 use alloy_primitives::Address;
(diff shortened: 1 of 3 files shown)


crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -136,23 +140,8 @@ where
...
-                .set(&tx_index, &pending_transaction.receipt, state)
-                .unwrap_infallible();
-
-            let hash = pending_transaction.transaction.signed_transaction.hash();
-            self.transaction_hashes
-                .set(hash, &tx_index, state)
-                .unwrap_infallible();
-        }
+        self.set_sccessory_state(head, &pending_transaction, pending_len, state)
+            .unwrap_infallible();
```
bkolad, line 143: We use unwrap_infallible() to indicate the operation cannot fail and does not charge gas since accessory state updates are free.

Accessing head and pending_len (lines 133 and 140) must be done outside set_accessory_state because they are state variables that incur gas costs. Doing this inside would charge gas only in native execution, leading to inconsistencies. To ensure gas is charged equally in both zk and native execution, these values must be fetched beforehand.

LogvinovLeon (approved):

PR #1620, opened 2025-09-03
EVM: Revert failed Transactions 

bkolad:
# Description

This PR makes the following changes:
1. Unsuccessful EVM transactions now get reverted.
2. Added a `SimpleStorage::always_revert` method to test reverted transactions.
3. Added a new high-level test: `evm_test_soft_confirmations`.
4. Added a setup method to `evm_test_helper` to make test setup easier.


examples/demo-rollup/tests/evm/evm_soft_conf.rs
```diff
@@ -0,0 +1,42 @@
...
+
+        test_rollup.wait_for_next_blocks(1).await;
+        let rec = evm_client.receipt(tx_hash).await.unwrap();
+        assert!(rec.block_hash.is_some());
+
+        assert_eq!(rec.block_number.unwrap().as_u64(), block_nr + 1);
+    }
+
+    // Check that invalid trsnacations are rejected.
+    {
```
LogvinovLeon, line 36: Can we move it to a separate test as this doesn't require sequencer start-stop logic?
Also - can we assert specific error so that we know if we start rejecting for the wrong reason?

LogvinovLeon (approved):

examples/demo-rollup/tests/evm/evm_soft_conf.rs
bkolad: Added an assertion for a specific error.

Regarding separate tests: in the demo rollup, we have high-level integration tests.  
These tests cover more functionality than the framework-based in the evm module tests but take longer to set up.  
If we split them too granularly, it will increase CI time.  

It's a balance — we want to group tests by functionality without making them overly fine-grained.  
Given this, I'll keep it in the same test.

PR #1621, opened 2025-09-04
Charge rollup gas for EVM execution

LogvinovLeon:
## Description
  - Introduce rollup gas metering for EVM execution
  - Configure EVM gas limit based on available rollup gas
  - Charge rollup gas proportionally to EVM gas consumed

  ## Changes

  ### Gas Metering
  - Add `gas_limit()` method to `GasMeter` trait to query remaining gas
  - Add `DEFAULT_GAS_TO_CHARGE_PER_EVM_GAS` constant (1:1 ratio)
  - Charge rollup gas after EVM execution based on `gas_used`

  ### EVM Configuration
  - Pass rollup gas limit to EVM transactions via `create_tx_env()`
  - Enable `disable_balance_check` to allow transactions without balance checks
  - Enable `disable_block_gas_limit` to use rollup gas instead of block limits

  ### Code Improvements
  - Extract `get_receipt()` and `handle_execution_error()` methods for better readability
  - Refactor executor to accept `TxEnv` directly instead of building it internally
  - Add context error handling throughout

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -46,84 +48,36 @@ where
...
-            .expect("Pending block must be set");
+            .context("Pending block must be set")?;
+        let transaction = TransactionSignedAndRecovered {
+            signer,
+            signed_transaction: tx,
+            block_number: block_env.number.to::<u64>(),
+        };
 
-        let cfg = self.cfg(state)?.expect("Evm config must be set");
+        let cfg = self.cfg(state)?.context("Evm config must be set")?;
```
bkolad: I think we should panic here instead of returning early. If cfg is not set, it indicates a serious issue we can't recover from, so panicking is appropriate.

bkolad: I’ve updated the tracking issue. At some point, we’ll review all `.expect` calls in the EVM module and add comments explaining why they’re justified (or remove them)

crates/module-system/sov-modules-api/src/gas/traits.rs
```diff
@@ -570,6 +570,15 @@ pub trait GasMeter {
         Ok(())
     }
 
+    /// Returns the amount of gas available. Used to set the EVM gas limit
+    fn gas_limit(
+        &mut self,
+    ) -> Result<<Self::Spec as Spec>::Gas, GasMeteringError<<Self::Spec as Spec>::Gas>> {
+        unreachable!(
+            "Default implementation should not be called. Override in the respective gas meter"
```
bkolad, line 578: We can remove the default implementation of the method, which will require the trait implementer to provide their own.

constants.toml
```diff
@@ -83,6 +83,9 @@ DEFAULT_GAS_TO_CHARGE_PER_BYTE_SIGNATURE_VERIFICATION = [1, 1]
 DEFAULT_FIXED_GAS_TO_CHARGE_PER_SIGNATURE_VERIFICATION = [100, 100]
 # --- End Gas parameters to specify how to charge gas for hashing ---
 
+# The cost of EVM execution
+DEFAULT_GAS_TO_CHARGE_PER_EVM_GAS = [1, 1]
```
bkolad: This should be [1,0] as EVM is concerned only with the first gas dimensionn

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -46,84 +48,36 @@ where
         // Inside the EVM, we use nonces only for the CREATE operation.
         // The uniqueness check was performed before the call was dispatched.
         let account_nonce = self.get_account_nonce(signer, state)?;
-
-        let evm_tx: TransactionSigned = convert_to_transaction_signed(message.rlp)?;
-
+        let gas_limit = state.gas_limit()?;
```
bkolad: The sov gas meter is initialized with:

1. `remaining_gas: S::Gas`
2. `remaining_funds: Option<Amount>`

Initially, `remaining_funds` is set to `tx.max_fee`. Since Ethereum also uses a `max_fee` concept, we should be able to support this as well. Maybe it works out of the box and we just need to replace the `tx.max_fee` with the one from `gas_meter.`

Can be done as a separate PR.

bkolad:
Looks great!
Here’s how we can test this change:
1. Execute an EVM transaction using the testing framework.
2. Get the batch gas usage as G1.
3. Override `DEFAULT_GAS_TO_CHARGE_PER_EVM_GAS = [0, 0]`.
4. Execute the same transaction in a fresh testing environment.
5. Record the batch gas usage as G2.

Verify that: `G1- G2 = [evm_reciept.gas_used, 0]`

Can be done as followup

bkolad (approved):

PR #1624, opened 2025-09-04
Avoid Replaying Txs During State Update in Most Cases

preston-evans98:
# Description

This PR implements a "fast track" for state updates which skips replay almost all of the time. The slow track (replaying txs) is now only necessary on startup, or when the node is resyncing/recovering.

This is implemented by adding an additional `rollup_height` field to each entry in the `StateCheckpoint` tracking when the entry was last modified. Then in `update_state` we simply swap the new `Storage` into the state checkpoint and prune any entries that are now outdated.


crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
```diff
@@ -2195,14 +2195,32 @@ async fn flaky_txs_that_enter_before_downtime_are_dropped() {
...
         .unwrap();
 
+    // Produce a new block that includes this first tx.
+    test_rollup
+        .da_service
+        .produce_n_blocks_now(1)
+        .await
+        .unwrap();
+    // Wait until the new batch is almost processed
+    sleep(Duration::from_millis(1200)).await;
```
bkolad, line 2207: Now we have `test_rollup.wait_for_height, `test_rollup.wait_for_next_blocks` methods. Maybe we could use one of them here and in other places instead of sleeping.

crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
```diff
@@ -160,7 +162,7 @@ pub struct ApiStateAccessor<S: Spec> {
     gas_price: <S::Gas as Gas>::Price,
     kernel_cache: ProvableStorageCache<namespaces::Kernel>,
     user_cache: ProvableStorageCache<namespaces::User>,
-    accessory_writes: HashMap<SlotKey, Option<SlotValue>>,
+    accessory_writes: HashMap<SlotKey, (u64, Option<SlotValue>)>,
```
bkolad: Let's add a comment explaining the meaning of the u64 in the tuple

crates/module-system/sov-state/src/cache.rs
```diff
@@ -113,6 +125,33 @@ mod internal {
             self.log.iter()
         }
 
+        #[cfg(feature = "native")]
+        pub(crate) fn prune_up_to(&mut self, rollup_height: u64) {
+            // We skip the expensive `retain` operation if the log is empty.
+            // According to its docs, `retain` runs in O(capacity) time rather than O(len); If the map is already empty from `.clear()`, that could be expensive!
```
bkolad, line 131: Good point!

bkolad (approved):

crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
preston-evans98: For some reason, waiting instead of sleeping makes the test *more* flaky. Since the flakiness predates this PR, I'm going to merge this as-is rather than spending time debugging.

crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
preston-evans98: Replaced this with a little helper struct; should be much clearer now

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -577,7 +580,7 @@ impl<S: Spec, Rt: Runtime<S>> RollupBlockExecutor<S, Rt> {
                 }
             };
             // Sanity check: the height we received should match the height we need.
-            if received_height != next_visible_rollup_height {
+            if received_height > next_visible_rollup_height {
```
theodorebugnet, line 583: The changed sanity check doesn't seem to match the comments. *If* we expect the state roots received to be contiguous, then from what I can understand it seems like we could do
```rust
let fetching_height = largest_known_root_height.saturating_add(1);
// ...
if received_height != fetching_height {
}
```

If we don't expect them to be contiguous, then maybe the comments below can be adjusted, i.e. "did not equal expected height" should be "exceeded maximum expected height" and maybe clarify the `// Sanity check` comment above to mention that we don't expect them to be contiguous but we do expect to not have anything above the maximum.

crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -1560,6 +1577,39 @@ where
...
+        // Atomically swap in the new storage and prune the old one.
+        let new_rollup_height = StateCheckpoint::new(info.storage.clone(), &Rt::default().kernel())
+            .rollup_height_to_access();
+        self.inner
+            .executor
+            .checkpoint
+            .replace_storage_and_prune(info.storage.clone(), &Rt::default().kernel());
+        tracing::info!("Storage has been replaced");
+        // Update the `inner`'s state to reflect the new storage.
+        // These steps should match `process_final_catchup` except for the need to drop the db_event_subscription.
```
theodorebugnet, line 1590: Would it make sense to extract the common parts into a shared function, e.g. `inner.set_info_and_mark_ready()`? Seems easy and would help keep the two in sync, assuming the `state_roots.retain()` call can be done after the `update_api_ledger()` call.

I.e. this entire section of `process_new_storage()` would be
```
self.inner.set_info_and_mark_ready().await;
self.inner.executor.state_roots.retain(...);
```
And the equivalent section in `process_final_catchup()` would become
```
inner.set_info_and_mark_ready().await;
drop(db_event_subscription);
drop(inner);
```

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
preston-evans98: Good catch. Updated!

crates/full-node/sov-sequencer/src/preferred/inner.rs
preston-evans98: Unfortunately, this is not possible due to some limitations of the borrow checker. In particular, it complains that `self` is borrowed mutably twice if we extract this code into a helper that takes `&mut self`; and we can't drop the first borrow because that would release the lock on self.inner - which would create a race condition

PR #1629, opened 2025-09-05
fix: correct "cargo version" to "rustc version" in error message

Snezhkko:
# Description

- Fix typo in does_rustc_match function error message
- Function works with rustc, not cargo, so error message should be accurate
- Improves clarity when debugging toolchain version mismatches


Snezhkko:

PR #1626, opened 2025-09-05
EVM: Remove panics and unwraps

bkolad:
# Description

Removes panic and unwraps from the evm code.

- [ ] I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.
- [ ] I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. (Are all new dependencies necessary? Is any module dependency leaked into the full-node (hint: it shouldn't)?)

PR #1631, opened 2025-09-05
EVM: Remove panics and unwraps part 2

bkolad:
# Description

This is a follow up to #1626

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/call.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/call.rs
@@ -16,6 +16,7 @@ use crate::evm::primitive_types::{Receipt, TransactionSignedAndRecovered};
 use crate::evm::RlpEvmTransaction;
 use crate::executor::get_cfg_env;
 use crate::{Evm, PendingTransaction};
+use anyhow::Context as _;
 
 /// EVM call message.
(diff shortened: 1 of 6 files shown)


crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -38,15 +39,13 @@ where
...
+            .expect("The impossible happened: block_env is not set.");
         let tx = convert_to_transaction_signed(message.rlp)?;
         // The signature was checked before the call was dispatched,
         // and the signer was recovered during the authentication process.
         let signer = *context
-            .get_sender_credential::<Address>()
-            .ok_or(anyhow::anyhow!(
-                "EVM transaction must be authenticated by the EVM authenticator"
-            ))?;
+            .get_sender_credential::<Address>().expect("The impossible happened: EVM transaction wasn't authenticated by the EVM authenticator");
```
preston-evans98: What happens if somebody tries to send an EVM transaction like a regular SDK (module system) transaction?

preston-evans98 (approved):
LGTM, other than the one question about sending EVM txs through the standard authentication path

crates/module-system/module-implementations/sov-evm/src/call.rs
bkolad: Oops fixing it.

PR #1636, opened 2025-09-08
BlobSender: Don't resend completed batches.

bkolad:
# Description
The `BlobSender` was attempting to send all batches that were completed and not pruned. However, due to a bug here https://github.com/Sovereign-Labs/sovereign-sdk/blob/nightly/crates/full-node/sov-sequencer/src/preferred/inner.rs#L382 It ended up sending too many batches. As we are not pruning everything we have to.

This PR refactors the `BlobSender` to prevent it from re-sending completed batches.  We will discuss on the daily what to do about the original bug.

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/side_effects.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/side_effects.rs
@@ -60,14 +60,14 @@ where
         batch: PreferredSequencerReadBatch,
         info_to_store: BatchToStore,
     ) -> anyhow::Result<()> {
-        self.db.terminate_batch(info_to_store).await?;
-        self.update_api_state(checkpoint);
-
         // Publish the batch.
         self.blob_sender
             .add_txs(batch.blob_id, batch.tx_hashes.clone())
             .await;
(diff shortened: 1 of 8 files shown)


crates/full-node/sov-sequencer/src/preferred/side_effects.rs
```diff
@@ -60,15 +60,15 @@ where
...
-        self.db.terminate_batch(info_to_store).await?;
-        self.update_api_state(checkpoint);
-
         // Publish the batch.
         self.blob_sender
             .add_txs(batch.blob_id, batch.tx_hashes.clone())
             .await;
         self.blob_sender.publish_batch(batch).await?;
 
+        self.db.terminate_batch(info_to_store).await?;
```
bkolad, line 69: We now call self.blob_sender.publish_batch(batch).await? before `terminate_batch`, ensuring that the blob_sender retains the batch in its internal db even if the sequencer crashes.

The only problem that remains is blob sender db corruption. If needed, we could move it to PostgreSQL.

citizen-stig (approved):

bkolad:
The underlying bug was fixed

PR #1653, opened 2025-09-10
EVM: Remove chain_id from genesis.

bkolad:
# Description

Removes chain_id from evm genesis.

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/authenticate.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/authenticate.rs
@@ -45,22 +45,24 @@ fn recover_evm_signer(
 }
 
 /// Creates the transaction details and tx hash for an EVM transaction.
-fn create_evm_tx_details_and_hash<S: Spec>(
+fn create_auth_tx_and_hash<S: Spec>(
     tx: &TransactionSigned,
-) -> Result<(TxDetails<S>, TxHash), AuthenticationError> {
+) -> Result<AuthenticatedTransactionAndRawHash<S>, AuthenticationError> {
     let tx_hash = TxHash::new(**tx.hash());
     let tx_chain_id = validate_chain_id(tx.chain_id(), tx_hash)?;
     let gas_limit = tx.gas_limit();
 
-    Ok((
-        TxDetails {
-            chain_id: tx_chain_id,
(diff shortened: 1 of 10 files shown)


crates/module-system/module-implementations/sov-evm/src/authenticate.rs
```diff
@@ -45,14 +45,35 @@ fn recover_evm_signer(
 }
 
 /// Creates the transaction details for an EVM transaction.
-fn create_evm_tx_details<S: Spec>(tx: &TransactionSigned) -> TxDetails<S> {
+fn create_evm_tx_details<S: Spec>(
+    tx: &TransactionSigned,
+) -> Result<(TxDetails<S>, TxHash), AuthenticationError> {
+    let tx_hash = TxHash::new(**tx.hash());
```
LogvinovLeon, line 51: Why did we move it here?
TxHash is not part of TxDetails and this move makes the function name not fully descriptive and complicates return value while saving 0 lines of code.

crates/module-system/module-implementations/sov-evm/src/authenticate.rs
```diff
@@ -45,14 +45,35 @@ fn recover_evm_signer(
...
+
+    if chain_id != rollup_chain_id {
+        return Err(AuthenticationError::FatalError(
+            FatalError::InvalidChainId {
+                expected: rollup_chain_id,
+                got: chain_id,
+            },
+            tx_hash,
+        ));
     }
```
LogvinovLeon: Extract to `validate_chain_id(&tx)`

LogvinovLeon (approved):

crates/module-system/module-implementations/sov-evm/src/authenticate.rs
bkolad: This is because the method now requires a `tx_has`h, which is derived from the `tx` already passed to it. Passing both `tx` and `tx_hash` separately would be more error-prone, as someone could provide a `tx_hash` that doesn’t actually match the given tx.

bkolad: Renamed to `create_evm_tx_details_and_hash`

LogvinovLeon (approved):

LogvinovLeon (approved):

PR #1659, opened 2025-09-10
Add tests for EVM gas estimation

LogvinovLeon:
# Description

*Summary of the changes...*

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -347,17 +347,17 @@ where
...
         let result = self.call(request, block_number, state)?;
         let gas_used = result.gas_used();
-        state
+        let gas_meter = state.try_as_basic_gas_meter().unwrap();
+        gas_meter
             .charge_linear_gas(
                 &<S as GasSpec>::gas_to_charge_per_evm_gas(),
                 gas_used as u32,
             )
             .unwrap();
```
bkolad: We now always justify panics with clear comments, such as "The impossible happened..".
Let's update this to `expect(..some message)` instead. 

A panic here is impossible because we initialized the` EVM` with gas from the `gas_meter` before executing the transaction. This guarantees that the EVM cannot consume more gas than what was provided by the `gas_meter`.

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -347,17 +347,17 @@ where
...
-        let gas_meter = state.try_as_basic_gas_meter().unwrap();
         let total_gas_used =
             gas_meter.initial_gas.as_ref()[0] - gas_meter.remaining_gas.as_ref()[0];
-        const RELATIVE_MARGIN: u64 = 100_000;
-        let gas_used_with_margins = (total_gas_used * 3) / 2 + RELATIVE_MARGIN; // gas * 1.5 + 100_000
+        const ABSOLUTE_MARGIN: u64 = 100_000;
+        let gas_used_with_margins = (total_gas_used * 3) / 2 + ABSOLUTE_MARGIN; // gas * 1.5 + 100_000
```
bkolad: Let's use `checked_mul and checked_div`. The EVM module should never panic, even with extremely large values.
If the checked mul/div returns None then we should return an error. 

Also, calculating it as `(total_gas_used / 2) * 3 + ABSOLUTE_MARGIN`  reduces the likelihood of overflo. As the division is done before multiplication

examples/demo-rollup/tests/evm/evm_gas_estimation.rs
```diff
@@ -0,0 +1,22 @@
...
+    let gas_estimation = test_client.eth_estimate_gas(simple_transfer).await;
+    assert_eq!(gas_estimation, 100_000); // Simple transfer consumes 0 EVM gas so we only see the ABSOLUTE_MARGIN
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn contract_deploy() -> anyhow::Result<()> {
+    let (_, test_client, _, _) = setup(0).await;
+
+    let deploy_tx = test_client.make_eip1559_tx(0, None, Some(test_client.contract.byte_code()));
```
bkolad, line 18: let_ = evm_test_helper::deploy_contract_check(&client)
        .await
        .unwrap();
```

File: examples/demo-rollup/tests/evm/evm_gas_estimation.rs [hunk 4]
```diff
@@ -0,0 +1,22 @@
+use crate::evm::evm_test_helper::setup;
+use ethers_core::abi::Address;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn simple_transfer() -> anyhow::Result<()> {
+    let (_, test_client, _, _) = setup(0).await;
+
+    let simple_transfer = test_client.make_eip1559_tx(0, Some(Address::zero()), None);
```
bkolad, line 8: We have `client.send_eth(reciever_address, eth_to_send).await;` helper

examples/demo-rollup/tests/evm/evm_gas_estimation.rs
```diff
@@ -0,0 +1,22 @@
...
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn contract_deploy() -> anyhow::Result<()> {
+    let (_, test_client, _, _) = setup(0).await;
+
+    let deploy_tx = test_client.make_eip1559_tx(0, None, Some(test_client.contract.byte_code()));
+    let gas_estimation = test_client.eth_estimate_gas(deploy_tx).await;
+    assert_eq!(gas_estimation, (208_049 * 3) / 2 + 100_000);
```
bkolad: This will become outdated very soon, so we try not to hardcode specific values. Let's just assert that it takes more than `100_000`

LogvinovLeon: I agree in general, but still think it's worth it in this case.
1. It's acting like a snapshot and will fail if I accidentally change the gas estimation logic.
2. It will be very visible on a PR - how did the changed logic affect the results.
3. It's only one test. If we had multiple - then sure

Having the abovementioned in mind - seems like cost is less than a benefit

bkolad (approved):

PR #1671, opened 2025-09-11
EVM: Add stub for `eth_subscribe`

bkolad:
# Description

1. Adds `pub_sub: alloy::RootProvider` to the EVM test client, allowing us to subscribe to logs using `eth_subscribe`.
2. Adds a stub implementation for` eth_subscribe`. The full implementation will be included in the next PR.

The `eth_subscribe` implementation in ethers is buggy, so I switched to `alloy`, which works correctly.
Unfortunately, this means we now have two separate clients:

`alloy::RootProvider` for subscriptions
`ethers::SignerMiddleware` for the rest of the functionality

The goal is to fully migrate from `ethers` to `alloy`, but attempting to do this in one step proved too time-consuming.
I propose a gradual migration approach:

First, make the EVM test client’s public interface independent of `ethers`. This means all public methods should only accept and return standard Rust types.

Once that refactor is complete, switching completely to` alloy` will be straightforward.

I've already started this refactoring, but there are still many methods left to migrate.

### DIFF
--- a/crates/full-node/sov-ethereum/src/handlers/mod.rs
+++ b/crates/full-node/sov-ethereum/src/handlers/mod.rs
@@ -1,11 +1,12 @@
-use std::sync::Arc;
-
+use alloy_primitives::Address;
 use alloy_primitives::{Bytes, B256};
-
+use alloy_rpc_types::Log;
 use jsonrpsee::types::{ErrorObjectOwned, Params};
 use jsonrpsee::Extensions;
+use jsonrpsee::PendingSubscriptionSink;
+use jsonrpsee::SubscriptionMessage;
+use reth_primitives::LogData;
 use sov_address::{EthereumAddress, FromVmAddress};
-
 pub use sov_evm::EthereumAuthenticator;
 #[cfg(feature = "local")]
 use sov_evm::Evm;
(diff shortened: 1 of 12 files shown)


crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -156,3 +159,50 @@ where
 
     Ok(tx_hash)
 }
+
+pub async fn eth_subscribe<S, Seq>(
```
bkolad, line 163: This will be implemented in the following PR

crates/full-node/sov-ethereum/src/lib.rs
```diff
@@ -62,6 +62,12 @@ where
         ready(Ok::<_, Infallible>(U256::ZERO))
     })?;
     rpc.register_async_method("eth_sendRawTransaction", handlers::eth_send_raw_transaction)?;
+    rpc.register_subscription(
+        "_eth_subscribe",
```
bkolad, line 66: prefixing with _ so it is not yet accessible by standard evm tools.

crates/utils/sov-eth-client/src/lib.rs
```diff
@@ -27,38 +34,52 @@ pub struct TestClient {
...
+        let ws_conn_str = &format!("ws://127.0.0.1:{}/rpc", http_addr.port());
+        let http_conn_str = &format!("http://127.0.0.1:{}/rpc", http_addr.port());
 
-        let rpc = WsClientBuilder::default()
-            .build(&format!("ws://127.0.0.1:{}/rpc", http_addr.port()))
-            .await
-            .unwrap();
+        let pub_sub = pubsub(ws_conn_str).await;
+
+        let client = {
```
bkolad, line 59: We will remove it once we remove ethers

LogvinovLeon (approved):

PR #1672, opened 2025-09-11
Fix sequence number calculation used to prune sequencer DB

preston-evans98:
# Description

This PR fixes the bug in sequencer DB pruning. Previously, we tried to fetch the latest *finalized*  sequence number from the node just by manually overriding the `true_slot_number` stored in a fresh kernel state accessor to the latest *finalized* slot number. This was incorrect. Since `next_sequencer_number` is a regular kernel state value (not a versioned value) the `true_slot_number` of the accessor has no effect on the returned value. 

I've fixed the logic by adding a new override to `KernelStateAccessor` which bypasses the state checkpoint and reads directly from the `HistoricalValues` table to fetch the value as of the requested slot number.

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/inner.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/inner.rs
@@ -379,13 +379,7 @@ where
 
         match latest_finalized_sequence_number(latest_state_info, &mut runtime) {
             Some(num) => {
-                // TODO(@neysofu): somehow, if we prune too close to the latest
-                // finalized sequence number, we get panics due to missing blobs
-                // and inconsistent state. There is clearly something wrong with
-                // the pruning height calculation height.
-                if let Some(num) = num.checked_sub(100) {
-                    self.executor_events_sender.prune(num).await;
-                }
+                self.executor_events_sender.prune(num).await;
             }
             None => {
                 // Nothing to prune because there's no sequence number history.
@@ -1362,6 +1356,16 @@ where
         let condition_nodes_sequence_number_is_fresher =
             next_sequence_number_according_to_node > next_sequence_number;
 
+        // There's an edge case on restart where the node hasn't synced to the chain tip yet but doesn't know it. We can check for it by seeing
+        // if the first batch to replay has a sequencer number that's more than 1 greater than the node's sequence number (because sequence numbers are contiguous, and
+        // we only prune a blob from the sequencer DB after it's been finalized on DA - so that blob must already exist on DA and the node just doesn't know that it isn't synced).
+        let oldest_unfinalized_sequence_number = batches_to_replay
+            .first()
+            .map(|b: &PreferredBatchToReplay| b.batch.inner.sequence_number);
+        let condition_node_is_unsynced_and_doesnt_know_it = (next_sequence_number_according_to_node
+            .saturating_add(1))
+            < oldest_unfinalized_sequence_number.unwrap_or(0);
+
         // Once we're this close to `deferred_slots_count`, we risk crossing the
         // `deferred_slots_count` threshold before the next call to
         // `update_state`. That's no good.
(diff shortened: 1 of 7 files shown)


crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -1362,6 +1356,16 @@ where
         let condition_nodes_sequence_number_is_fresher =
             next_sequence_number_according_to_node > next_sequence_number;
 
+        // There's an edge case on restart where the node hasn't synced to the chain tip yet but doesn't know it. We can check for it by seeing
```
theodorebugnet, line 1359: If this is a problem then it sounds like the node logic should really not assume it's synced without actually getting the tip height from the DaService first. In a perfect world I'd rather fix this in the node than here, though I don't know the complexity involved. But for example this edge case will still exist if the sequencer DB was also wiped: there could still be batches on DA that the node hasn't seen, but if the sequencer doesn't have any `batches_to_replay` saved it won't detect this condition.

Fixing it in the node might also need some redesign in DaSyncState and related machinery (because there's currently no way to represent the state of "I'm not synced because I don't know the tip height"), so I don't want to block the PR, just wanted to bring it up for consideration.

crates/module-system/sov-modules-api/src/state/accessors/kernel.rs
```diff
@@ -66,6 +66,9 @@ pub struct KernelStateAccessor<'a, S: Spec> {
     /// The inner working set
     pub checkpoint: &'a mut StateCheckpoint<S>,
     pub(crate) true_slot_num: SlotNumber,
+    /// Whether to read directly with `read_direct_from_storage` override enabled. This is a bug in the sequencer. Please report it. instead of using the checkpoint
```
theodorebugnet: accidental paste sneaked in

crates/module-system/sov-modules-api/src/state/accessors/kernel.rs
```diff
@@ -142,6 +153,56 @@ impl<S: Spec> KernelStateAccessor<'_, S> {
...
+    fn get_value(
+        &mut self,
+        namespace: Namespace,
+        key: &SlotKey,
+        metric: &mut StateAccessMetric,
+    ) -> Option<SlotValue> {
+        if self.read_direct_from_storage {
+            use sov_state::namespaces::{Kernel as KernelNamespace, User};
+            use sov_state::NativeStorage;
+            match namespace {
```
theodorebugnet, line 186: Thoughts on something like
```rust
match namespace {
    Namespace::User => self.checkpoint.delta.inner.get_historical::<User>(key, Some(self.true_slot_num), &Default::default()),
    Namespace::Kernel => self.checkpoint.delta.inner.get_historical::<KernelNamespace>(key, Some(self.true_slot_num), &Default::default()),
    Namespace::Accessory => self.checkpoint.delta.inner.get_accessory_historical(key, Some(self.true_slot_num)),
}.expect("Failed to read {namespace:?} value with `read_direct_from_storage` override enabled. This is a bug in the sequencer. Please report it.")
```
?
(Non-blocking, completely to preference)

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/inner.rs
bkolad: Let's discuss it for 3 min after the daily. The node takes the tip height from the DA see (in the normal case (not upgrading))`make_da_sync_state` 
```
 let target_da_height=    if stop_at_rollup_height.is_some() {
        da_service.get_last_finalized_block_header().await
    } else {
        da_service.get_head_block_header().await
    };

 let sync_state = Arc::new(DaSyncState {
        synced_da_height: da_height_processed.into(),
        target_da_height: AtomicU64::new(target_da_height),
       ...
    });
```

theodorebugnet: Blazej's right - as discussed on daily this is an issue with DA nodes so nothing we can do, in which case this best effort heuristic sounds like the best solution :+1:

crates/module-system/sov-modules-api/src/state/accessors/kernel.rs
preston-evans98: Fixed. Thanks!

crates/module-system/sov-modules-api/src/state/accessors/kernel.rs
preston-evans98: That is nice. Made the change here, but unfortunately we can't easily make the same change for `get_size` because of the need to `map` over heterogeneous types.

PR #1690, opened 2025-09-15
Add `/sequencer/ready` to OpenAPI spec and unflaky some more sequencer tests

theodorebugnet:
# Description
Adds more utilities to `TestRollup`. This should help unflaky both recovery tests (I was able to `cargo stress` both for 200+ runs without failure) and should allow the utilities to be used to going forward to reduce reliance on sleeps in tests that rely on the node syncing or the sequencer being ready/unready.

I also tried adding `wait_for_sequencer_ready()` to `TestRollup::start()` but this fails - I think we need to generate some DA blocks first before the sequencer can become ready. This can be explored in a follow-up PR if more related flakiness is observed.

- [ ] ~I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. (Are all new dependencies necessary? Is any module dependency leaked into the full-node (hint: it shouldn't)?)~

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/sov-test-utils/src/test_rollup.rs
+++ b/crates/module-system/sov-test-utils/src/test_rollup.rs
@@ -809,16 +809,17 @@ where
     }
 
     /// Generic helper for waiting on a condition with timeout and polling.
+    ///  * condition_string: inserted into "Timeout waiting for {condition_string}", format accordingly
     async fn wait_for_condition<F, Fut>(
         &self,
         mut condition_check: F,
-        condition_name: &str,
-        timeout_secs: u64,
+        condition_string: &str,
     ) -> anyhow::Result<()>
     where
         F: FnMut() -> Fut,
         Fut: std::future::Future<Output = anyhow::Result<bool>>,
     {
+        const TIMEOUT_SECS: u64 = 20;
         let wait_loop = async {
             loop {
                 match condition_check().await {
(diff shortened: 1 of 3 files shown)


crates/module-system/sov-test-utils/src/test_rollup.rs
```diff
@@ -791,24 +791,83 @@ where
...
 
+    /// Checks if the sequencer is ready without waiting.
+    pub async fn is_sequencer_ready(&self) -> bool {
+        self.client.client.is_ready().await.is_ok()
+    }
+
+    /// Polls the sequencer until is_ready() returns Err(). Useful when you expect the sequencer to
+    /// go into resync/recovery/startup mode, to avoid wait_for_sequencer_ready() from resolving
+    /// _before_ the sequencer becomes unready.
+    pub async fn wait_for_sequencer_not_ready(&self) -> anyhow::Result<()> {
```
citizen-stig, line 807: Would it timeout after some time or be running indifentely?

theodorebugnet: It times out after 20 seconds (hardcoded), as per the `wait_for_condition` helper - we could do custom timeouts but I don't think any test should expect to wait anywhere near 20 seconds

citizen-stig (approved):

crates/module-system/sov-test-utils/src/test_rollup.rs
citizen-stig: Can we make it more discoverable, like move timeout duration to public const and reference it in the documentation?

theodorebugnet: Sure, good ideas

PR #1694, opened 2025-09-15
Limit Memory Consumption of StateDB Pruning Task

preston-evans98:
# Description

This PR caps the memory usage of the `StateDB` Pruning task at a few hundred MB. Previously, pruning would try to handle the entire backlog in a single massive batch, causing slow/unpredictable progress and OOM issues. This PR prevents this by capping the size of the pruning batch and adding a flag to inform the caller whether the pruning task returned because it was actually finished, or whether its simply hit the size limit and should be re-run immediately. 

### DIFF
--- a/crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
+++ b/crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
@@ -26,6 +26,10 @@ use crate::storage_manager::{update_ledger_finalized_height, InitializableNative
 
 const GIGABYTE: usize = 1024 * 1024 * 1024;
 
+// 3 million keys * 32 bytes is about 100 MB. This should be a large enough batch size to keep up with state growth,
+// without consuming excessive memory.
+pub(crate) const MAX_INDIVIDUAL_PRUNING_BATCH_SIZE: usize = 3_000_000;
+
 pub(crate) struct DbGroup<H, K> {
     merklized_state: Arc<NomtStateDb<H>>,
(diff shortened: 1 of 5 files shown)


citizen-stig (approved):

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -163,101 +168,114 @@ where
...
+                if let Some(kernel_version) =
+                    current_kernel_version.and_then(|v| v.checked_sub(versions_to_keep as u64))
+                {
+                    let prunable_keys = kernel
+                        .iter_pruning_keys_up_to_version(kernel_version)?
+                        .take(MAX_INDIVIDUAL_PRUNING_BATCH_SIZE);
+                    for key in prunable_keys {
+                        // Prune the pruning table.
+                        let key = key?;
+                        batch.delete::<NomtPruningState<KernelNamespace>>(&key)?;
```
bkolad, line 234: This for loop looks very similar to the previous one. Let's extract the core logic to a helper method.

bkolad (approved):

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
preston-evans98: Going to leave this out of scope since it's not a new issue. Happy to handle in a follow-up, but let's get this fix merged cuz it could break zeta.

PR #1696, opened 2025-09-16
EVM: Fix for `block_timestamp` in pending block.

bkolad:
# Description

This PR fixes a bug where `block_timestamp` was not being populated for pending blocks.
Although the hash of the pending block is unknown, we do have access to both the `block_number` and `block_timestamp`, which are now check in tests.

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
@@ -395,29 +395,9 @@ where
         if let Some(block) = block {
             return MaybeSealedBlock::Sealed(block.into());
         }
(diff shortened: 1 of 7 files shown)


crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -505,6 +492,10 @@ where
         let header = alloy_consensus::Header {
             parent_hash: head_block.header.seal(),
             number: pending_block_number,
+            timestamp: current_block_env
```
bkolad, line 495: This is the fix. We didn't set timestamp.

LogvinovLeon (approved):

PR #1704, opened 2025-09-17
Extracting re-usable code into `sov-soak-lib`

citizen-stig:
# Description

So rollup starter does not depend on `demo-stf`

## Testing
All existing tests should pass

## Docs
No updates


crates/utils/sov-soak-lib/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+    name: TestRuntime,
+    modules: [paymaster: Paymaster<S>, synthetic_load: SyntheticLoad<S>],
+    operating_mode: sov_modules_api::runtime::OperatingMode::Zk,
+    minimal_genesis_config_type: MinimalZkGenesisConfig<S>,
+    gas_enforcer: paymaster: Paymaster<S>,
+    runtime_trait_impl_bounds: [],
+    kernel_type: sov_kernels::soft_confirmations::SoftConfirmationsKernel<'a, S>,
+    auth_type: sov_modules_api::capabilities::RollupAuthenticator<S, Self>,
+    auth_call_wrapper: |call| call,
+}
```
theodorebugnet: What if we make sov-soak-lib completely runtime-free and specific soak tests supply their own runtimes, for separation of concerns? I.e. move the TestRuntime definition back to the demo rollup soak tests.
That way if you import the lib somewhere else to test your own rollup, it doesn't generate a TestRuntime that you don't need.

Though if it doesn't affect dependencies much then this might not mattter.

theodorebugnet (approved):

crates/utils/sov-soak-lib/src/lib.rs
citizen-stig: Good point!

Issue #1656, opened 2025-09-10
Enrich `Tx receipt` log message for better debuggability


citizen-stig:
It would be great if this log message:

https://github.com/Sovereign-Labs/sovereign-sdk/blob/28b336e2dd489b992c2f5a04f07c8bb5cccb8ba3/crates/module-system/sov-modules-stf-blueprint/src/sequencer_mode/common.rs#L195-L203

* Sender
* authorizing credential ID
* tx type info (module/callmessage)

Same would be great for ignored tx receipt

https://github.com/Sovereign-Labs/sovereign-sdk/blob/28b336e2dd489b992c2f5a04f07c8bb5cccb8ba3/crates/module-system/sov-modules-stf-blueprint/src/sequencer_mode/common.rs#L205-L210

PR #1715, opened 2025-09-17
feat: enrich tx receipt logs with sender, credential and call type

radik878:
Add best-effort decoding of transaction bodies at log time to include sender (rollup address derived from credential), authorizing credential ID, and tx type (module/call message discriminant). Extend apply_batch_logs to be generic over runtime and accept optional ignored raw tx bodies for enriching ignored receipts. Plumb ignored raw tx bodies in registered path; update unregistered path call site. Decode is native-only and fails gracefully, preserving existing behavior if decoding isn't available. Fix borrow-after-move by cloning raw tx; silence unused param in non-native builds. No changes to stored receipt schema.

#1656


cemozerr:
@radik878 Thank you for your contribution! Could you sign the CLA please?

radik878:

radik878:
> @radik878 Thank you for your contribution! Could you sign the CLA please

Yes, didn't notice immediately

PR #1724, opened 2025-09-19
EVM: The pending block returns a zero hash.

bkolad:
# Description
The `get_sealed_block_by_number` method incorrectly sealed a pending block. Since a pending block has an unknown hash, whenever a hash is requested for it, it should return zeros as the block hash. This PR fixes the issue.


crates/module-system/module-implementations/sov-evm/src/evm/primitive_types.rs
```diff
@@ -205,6 +205,14 @@ impl MaybeSealedBlock {
             Self::Pending(pending) => pending.header.timestamp,
         }
     }
+
+    /// The blcok header.
```
LogvinovLeon: Typo

LogvinovLeon (approved):

crates/module-system/module-implementations/sov-evm/src/evm/primitive_types.rs
bkolad: fixed

PR #1726, opened 2025-09-19
`RollupBlockExecutor`: extract helper methods for reusability

bkolad:
# Description

1. Groups `start_rollup_bloc`k parameters into a `StartBlockData` structure.
2. Adds `update_kernel_with_user_state_root` helper method.

These changes lay the groundwork for the upcoming PR, which will implement cache warm-up functionality. The worker executors will share some of the above methods.


crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -596,9 +616,12 @@ impl<S: Spec, Rt: Runtime<S>> RollupBlockExecutor<S, Rt> {
...
-            self.state_roots
-                .insert(next_visible_rollup_height, next_visible_root);
+            state_roots.insert(next_visible_rollup_height, next_visible_root);
         }
+        state_roots
+    }
+
+    fn update_kernel_with_user_state_root(&mut self) {
         // take all roots greater than self.started_from
         for (height, root) in self.state_roots.iter() {
```
bkolad, line 626: @preston-evans98 I kept the behavior the same as in the original code, but should we be looping over all self.state_roots or only the new entries returned by `get_state_roots`? Will update the comment in 625.

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -532,6 +546,18 @@ impl<S: Spec, Rt: Runtime<S>> RollupBlockExecutor<S, Rt> {
         }
     }
 
+    fn update_kernel_with_user_state_root(&mut self) {
+        // take all roots greater than self.started_from
+        for (height, root) in self.state_roots.iter() {
```
bkolad, line 551: @preston-evans98  I kept the behavior the same as in the original code, but should we be looping over all self.state_roots or only the new one from populate_state_roots? Looks like the comment in 550 is outdated.

preston-evans98: Hmm, yeah I think we used to prune `self.state_roots` implicitly (because it would start out empty each time we ran `update_state`). Now that we hold on to the executor and just swap the storage, we may be accumulating a bunch of unneeded roots here.

et --hard
preston-evans98, line 1: This looks like maybe an accidental checkin?

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
bkolad: I printed state_roots.len(), and it's usually very small — typically between 2 and 5 in our tests.

et --hard
bkolad: fixing

preston-evans98 (approved):

PR #1727, opened 2025-09-19
Revamp handling of uncommitted state and sequencer state root calculations

preston-evans98:
# Description

## Core Change - Add Changeset Between StateCheckpoint and Storage
This PR introduces the concept of `uncomitted_changes` that exist outside of storage in the sequencer. After this change, the `StateCheckpoint` (and related accessors) will first check their set of `uncomitted_changes` for a write to each requested key before falling back to storage. This new abstraction allows us to directly convert a `StateCheckpoint` into a shared readable object without doing any type conversion - which means we don't have to iterate over the changeset during `end_rollup_block`. This should drastically reduce tail latencies and should give a modest bump in overall throughput.

### State Root Computation Changes
Unfortunately, this change in abstractions surfaced a pre-existing bug in our sequencer state root computation. 

In certain edge cases during resync, the full-node can get *ahead* of the sequencer while the sequencer is replaying batches. In these cases, the backing `NOMT` instance will no longer have the data needed to compute the state root for the sequencer's current batch (because it has been overwritten with "future" data from the full node). In theory this is easy to handle - we just need to detect the edge case and read the state root from disk rather than recomputing it. 

Unfortunately, our previous logic for this was incorrect. It relied only on slot numbers - but slot numbers have no predictable relationship to user-space state roots; there's one (user-space) state root per rollup block, but a rollup block may occur as often as once per slot or much less frequently. Before the aforementioned changes to the sequencer state, we were getting lucky most of the time and the sequencer would compute the wrong state root, but then have it overwritten by the node before it was read. These changes seem to have messed with the timing such that we no longer have such luck - so we had no choice but to fix the issue. 

The fix is relatively simple. I've added a couple of new methods to the `NativeStorage` and `KernelWithSlotMapping` traits to let us...
 - Get the most recent rollup height *on disk*. If this height is newer than the sequencer's height, we know we're in the edge case
 - Find the slot number associated with that height if possible


crates/module-system/sov-state/src/cache.rs
```diff
@@ -522,22 +560,25 @@ impl<N: ProvableCompileTimeNamespace> ProvableStorageCache<N> {
...
     }
 
     // This method can be called only once per given key.
-    fn add_read(&mut self, key: SlotKey, node: Option<NodeLeafAndMaybeValue>) {
-        self.revertable_ordered_reads
-            .push((key.clone(), node.as_ref().map(|n| n.leaf)));
-
-        self.cache.add_read(key, node);
+    fn add_read(
+        key: SlotKey,
```
bkolad, line 580: What is the reason this method (and others) no longer takes &mut self but instead accepts revertable_ordered_reads as a parameter?

crates/module-system/sov-state/src/sequencer_state.rs
```diff
@@ -0,0 +1,211 @@
+use std::{
```
bkolad, line 1: Shouldn't the whole file be native only?

bkolad (approved):

crates/module-system/sov-state/src/cache.rs
```diff
@@ -399,13 +355,52 @@ impl<N: ProvableCompileTimeNamespace> ProvableStorageCache<N> {
...
                 Ok(size)
             }
         }
     }
 
+    #[cfg(feature = "native")]
+    /// Get the size of the value.
+    pub fn get_size_or_fetch<S: Storage>(
+        &mut self,
+        uncomitted_changes: &Option<Box<dyn StateGetter>>,
```
bkolad, line 379: As we discussed on our call, it would be great to simplify the dependencies. Ideally, cache.rs wouldn’t depend on sequencer_state. If that’s not possible, let’s at least try to simplify the types.

crates/module-system/sov-state/src/cache.rs
preston-evans98: Ah, this is because we need to call `add_read` while borrowing `uncomitted_changes`. If we use a method that takes `&mut self`, rustc thinks we have overlapping borrows - so we have to borrow each field separately at the call site (with the appropriate mutability) instead.

crates/module-system/sov-state/src/sequencer_state.rs
preston-evans98: Yeah, I think it can be. We used to need the `MaybePresentValue` enum in zk as well, but I think that's been cleaned up. Trying!

crates/module-system/sov-state/src/cache.rs
preston-evans98: Added a new `CachedWrites` type which is just a simple subset of `ProvableStorageCache`. Hopefully that makes it clearer!

PR #1729, opened 2025-09-19
Add `CacheWarmupExecutor`  task to execute incoming txs in order to warm up the cache.

bkolad:
# Description

1. Added` CacheWarmupExecutor` task to pre-execute incoming transactions for cache warmup.
2. Glue code that allows sending open block events and transactions to the `CacheWarmupExecutor`.

- [x] I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.~


crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -265,18 +267,35 @@ where
...
 
         self.executor
-            .start_rollup_block(start_block_data, min_profit_per_tx)
+            .start_rollup_block(start_block_data.clone())
             .await;
+
+        let state_roots = self.executor.state_roots.clone();
+
+        let notification = StartBtachNotification {
+            state_roots,
```
bkolad, line 295: We send the `state_roots` map to the workers, with each worker cloning it in its own thread. The map is small (size ≤5), so the overhead is minimal.

I attempted to send the checkpoint obtained after calling  `start_rollup_block` but ran into panics due to missing `state_roots`. After timeboxing the issue, I chose to send older checkpoint together with the `state_roots`

Mid-term options:
1. Investigate the root cause of the panic, which will likely require changes to `rollup_block_task_body`.
2Use an `Arc` for `state_roots` since workers don’t modify it, though this would require adding custom logic to the execut

crates/full-node/sov-sequencer/src/preferred/mod.rs
```diff
@@ -834,6 +851,7 @@ where
             tokio::time::sleep(std::time::Duration::from_millis(delay_ms)).await;
             tracing::debug!(%tx_hash, "Transaction delay completed, proceeding with processing");
         }
+        self.cache_warm_up_executor.send_tx(baked_tx.clone());
```
bkolad, line 854: We send the transaction to the warmup cache executor after determining whether it is delayed.

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -0,0 +1,156 @@
...
+                                continue;
+                            }
+                            Err(tokio::sync::broadcast::error::RecvError::Closed) => {
+                                // Quit if channel closed.
+                                return
+                            }
+                        }
+                    }
+
+                    tx = tx_receiver.recv_async() => {
```
bkolad, line 116: Workers steal transactions from the same queue. Each worker’s transactions will have increasing nonces, but they won’t be strictly sequential. That's why we had to change the nonce checking logic.

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -0,0 +1,156 @@
...
+use sov_modules_api::{FullyBakedTx, Runtime};
+use std::collections::BTreeMap;
+use tokio::task::JoinHandle;
+
+// We have several work-stealing executor workers for a single main worker, so we don't expect the channels to become full.
+// Even if they do, the sender uses a non-blocking method, meaning a few updates may simply be skipped.
+const TX_CHANNEL_SIZE: usize = 16;
+const OPEN_BLOCK_CHANNEL_SIZE: usize = 16;
+
+pub(crate) struct StartBtachNotification<S: Spec> {
```
LogvinovLeon: Batch

LogvinovLeon (approved):
One big question I have is if it actually makes things faster. I guess we'll know that after I implement some real transaction flow generation for profiling

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2025-09-21
+- #1729 **BREAKING CHANGE** Adds `num_cache_warmup_workers` to `xxx_rollup_config.toml`. This field specifies the number of workers responsible for warming up the executor cache in the preferred sequencer.
```
preston-evans98: Can we avoid breakage by adding a default here?

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -427,31 +426,56 @@ impl<S: Spec, Rt: Runtime<S>> RollupBlockExecutor<S, Rt> {
...
+        self.checkpoint = checkpoint;
 
         assert!(
             self.rollup_block_task_state.is_none(),
             "Starting a rollup block, but there's already one in progress {:?}. This is a bug, please report it.",
             self.rollup_block_task_state
         );
 
+        if state_roots.len() > 5 {
+            panic!("LOOOL {:?}", state_roots.len());
```
preston-evans98: This looks like an accidental checkin :)

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    pub(crate) async fn spawn_execution_task<Rt: Runtime<S>>(
+        info: StateUpdateInfo<S::Storage>,
+        exec_config: RollupBlockExecutorConfig<S>,
+        seq_config: SequencerConfig<S::Address, PreferredSequencerConfig>,
+    ) -> (Self, Vec<JoinHandle<()>>) {
+        let (tx_sender, tx_receiver) = flume::bounded(TX_CHANNEL_SIZE);
+        let (start_block_notification_sender, _) =
+            tokio::sync::broadcast::channel(OPEN_BLOCK_CHANNEL_SIZE);
```
preston-evans98: Is there a reason we use broadcast instead of `watch` here? I would have thought that we didn't care about processing these in order and would rather just take the newest one.

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -0,0 +1,156 @@
...
+                        let baked_tx = match tx {
+                             Ok(tx) => tx,
+                             Err(flume::RecvError::Disconnected) => {
+                                // Quit if channel closed.
+                                return
+                            },
+                        };
+                        if is_started{
+                            // Ignore the result.
+                            _ = executor.apply_tx_to_in_progress_batch(&baked_tx).await;
```
preston-evans98, line 126: Should we add a `span` here so that we can easily silence the logs from the optimistic execution? And maybe we should silence them by default in the pre-configured RUST_LOG setting?

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -0,0 +1,156 @@
...
+
+impl<S: Spec> CacheWarmUpExecutor<S> {
+    pub(crate) fn send_batch_start_notification(&self, data: StartBlockNotification<S>) {
+        // This `send` does not block.
+        let _ = self.start_block_notification_sender.send(data);
+    }
+
+    pub(crate) fn send_tx(&self, tx: FullyBakedTx) {
+        // Skip update if consumer is too slow.
+        let _ = self.tx_sender.try_send(tx);
```
preston-evans98, line 51: (Non-blocking): We had discussed also sending a `oneshot` with the tx so that the optimistic executor could send back the read-set and we could put it in the `ProvableStorageCache` ahead of execution. Is that still planned for a follow-up?

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
preston-evans98: On further thought, maybe we should add a warning if state_roots gets over (say) 50 items large or something though

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -0,0 +1,156 @@
...
+    tx_sender: flume::Sender<FullyBakedTx>,
+}
+
+impl<S: Spec> CacheWarmUpExecutor<S> {
+    pub(crate) fn send_batch_start_notification(&self, data: StartBlockNotification<S>) {
+        // This `send` does not block.
+        let _ = self.start_block_notification_sender.send(data);
+    }
+
+    pub(crate) fn send_tx(&self, tx: FullyBakedTx) {
```
preston-evans98, line 49: (Non-blocking): we should probably add some metrics about queue depth or warn if the `try_send` fails so that we know to increase the number of workers

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -427,31 +426,56 @@ impl<S: Spec, Rt: Runtime<S>> RollupBlockExecutor<S, Rt> {
         }
     }
 
-    #[tracing::instrument(skip_all, level = "trace")]
-    pub async fn start_rollup_block(
+    pub async fn start_rollup_block_with_provided_state_roots(
```
preston-evans98, line 429: (Non-blocking): We may want to add a new `ExecutionMode` so that we can distinguish the optimistic execution from the canonical sequencer attempt in logs/metrics. If we do, we should also add an `ExecutionContext` field to our tx level timing [metric](https://github.com/Sovereign-Labs/sovereign-sdk/blob/bdbc4a89171b32c02dddfd83f194e6d4923e597a/crates/full-node/sov-metrics/src/influx_db_nonnative.rs#L316) so we can distinguish between node/sequencer/cache warmer

examples/demo-rollup/tests/evm/evm_logs.rs
```diff
@@ -64,7 +64,7 @@ async fn evm_test_get_logs_range() {
     let contract_address = evm_client.alloy_deploy_contract().await;
     test_rollup.wait_for_next_blocks(1).await;
 
-    let nb_of_txs = 10;
+    let nb_of_txs = 30;
```
preston-evans98: (Non-blocking) This change looks possibly unrelated. Was it meant to be included in this PR?

CHANGELOG.md
bkolad: done

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
bkolad: fixed

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
bkolad: I wanted all the workers to see the same checkpoint, but now I think this isn’t necessary, so I’m fixing it.

bkolad: fixed

examples/demo-rollup/tests/evm/evm_logs.rs
bkolad: fixed

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
bkolad: Will be fixed in #1743

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
bkolad: Will be fixed in #1743

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
bkolad: Will be fixed in #1743

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
bkolad: I introduced `ExecutionContext::SequencerWarmUp` and passed `ExecutionContext` to the uniqueness capabilities (and the metric part). This makes it possible to distinguish between cases where execution is happening in the main sequencer executor versus a worker thread.

If needed, we can extend this approach by adding more detailed configurations, for example:
`ExecutionContext::SequencerWarmUp(SomeConfig)` & `ExecutionContext::Sequencer(SomeOtherConfig)`

I chose this approach because `ExecutionContext` already modifies the `apply_batch` behavior (the sequencer processes ignored transactions differently than the node). Extending it to handle capabilities was a natural continuation of this pattern.

bkolad: I didn't extend ExecutionMode since it’s defined at the Rollup blueprint level, where all dependencies are assembled to create a new rollup. And this part is identical for `ExecutionContext::SequencerWarmUp` and `ExecutionMode::Sequencer` are identical.

preston-evans98: Oops, I had meant `ExecutionContext`! Good choice :)

crates/full-node/full-node-configs/src/sequencer.rs
```diff
@@ -173,11 +174,15 @@ impl Default for PreferredSequencerConfig {
             is_replica: false,
             db_event_channel_size: default_db_event_channel_size(),
             batch_execution_time_limit_millis: 6_000, // 6 seconds
-            num_cache_warmup_workers: 0,
+            num_cache_warmup_workers: default_num_cache_warmup_workers(),
```
preston-evans98, line 177: Thanks! Leo raised a good point that this will be hard for users to discover. Let's add a commented out example with an explanation in the rollup.toml for demo-rollup to make it easier.

preston-evans98 (approved):

crates/full-node/full-node-configs/src/sequencer.rs
bkolad: Added a comment in the .toml

PR #1746, opened 2025-09-23
Evm metrics

LogvinovLeon:
# Description

```
DbMetrics {
    account: DbAccess {
        duration: 620.316µs,
        count: 2,
    },
    code: DbAccess {
        duration: 0ns,
        count: 0,
    },
    storage: DbAccess {
        duration: 0ns,
        count: 0,
    },
    block_hash: DbAccess {
        duration: 0ns,
        count: 0,
    },
}
EvmTxMetrics {
    total_time: 2.534661ms,
    fetch_state_time: 587.035µs,
    get_db_time: 3.68µs,
    execution_time: 984.143µs,
    state_commit_time: 234.578µs,
    receipt_time: 63.58µs,
    set_accessory_state_time: 46.83µs,
    set_state_time: 98.499µs,
}
```

Adds some basic EVM metrics tracking different execution stages and DB access.
Some refactors.
DB time is within the execution time so from the data above we can deduct that just EVM execution took 360ms.
get_db_time is small now but does a lot of cloning inside so wanted to still track it

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-metrics/Cargo.toml
+++ b/crates/full-node/sov-metrics/Cargo.toml
@@ -23,6 +23,7 @@ http = { version = "1.0", optional = true }
 derivative = { workspace = true, optional = true }
 sov-rollup-interface = { workspace = true }
 derive_more = { workspace = true, features = ["deref_mut", "deref"], optional = true }
+derive-new = { workspace = true }
 
 anyhow = { workspace = true }
 risc0-zkvm = { workspace = true, default-features = false, features = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11906,6 +11906,7 @@ dependencies = [
  "sov-eth-dev-signer",
  "sov-evm",
  "sov-kernels",
+ "sov-metrics",
  "sov-mock-da",
  "sov-modules-api",
  "sov-rollup-interface",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12030,6 +12030,7 @@ dependencies = [
  "bincode",
  "chrono",
  "derivative",
+ "derive-new",
  "derive_more 1.0.0",
  "http 1.3.1",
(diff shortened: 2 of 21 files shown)


crates/full-node/sov-metrics/Cargo.toml
```diff
@@ -23,6 +23,7 @@ http = { version = "1.0", optional = true }
 derivative = { workspace = true, optional = true }
 sov-rollup-interface = { workspace = true }
 derive_more = { workspace = true, features = ["deref_mut", "deref"], optional = true }
+derive-new = { workspace = true }
```
preston-evans98, line 26: Because of the supply chain risks, we have a pretty strong norm against adding macro crates unless they're either going to be very widely used (both in our codebase and by other Rust projects) or enable us to do something that we couldn't easily do otherwise. 

Do we know about this crate? Is it likely already in our dependency tree?

LogvinovLeon: It's downloaded 19M times and has 563 stars so it's used in the community.

It's currently only used in `sov-evm` but it's useful and I would like (if possible) to keep using it.
We have multiple objects in the codebase that have 2-3 fields and so with this crate we can write:
```rust
let x = Struct::new(a, b);
```
instead of:
```rust
let x = Struct {
    a,
    b
};
```

It seems like a small difference but it adds up. It's also nice to have one logic step taking up one line and not 4 when possible.

What are the options to keep using it safely? I can pin it's version in Cargo.toml. I can copy it's code. It's 600 lines.
I could implement the constructors manually but it's repetitive code.

We already use derive_more.

preston-evans98 (approved):
Looks nice! Thanks @LogvinovLeon!

Issue #1743, opened 2025-09-23
Address code review feedback provided in PR #1729.


bkolad:
https://github.com/Sovereign-Labs/sovereign-sdk/pull/1729#discussion_r2369233030
https://github.com/Sovereign-Labs/sovereign-sdk/pull/1729#discussion_r2369258964
https://github.com/Sovereign-Labs/sovereign-sdk/pull/1729#discussion_r2369274233

bkolad:
https://github.com/Sovereign-Labs/sovereign-sdk/pull/1785#discussion_r2399050921

PR #1751, opened 2025-09-24
Cache: retrieve all reads that the transaction is accessing for the first time

bkolad:
# Description

This follows up on issue #1729. 
It introduces a way to retrieve all reads that the transaction is accessing for the first time.

see #1743

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
@@ -122,9 +122,16 @@ impl<S: Spec> CacheWarmUpExecutor<S> {
                             },
                         };
                         if is_started {
-                            // Ignore the result.
-                            _ = executor.apply_tx_to_in_progress_batch(&baked_tx).await;
+                            let res = executor.apply_tx_to_in_progress_batch(&baked_tx).await;
 
+                            match res{
+                                Ok((_, _tx_change_set)) => {
+                                    // TODO
+                                },
+                                Err(_) => {
+                                    continue;
+                                }
+                            }
                         }
 
                     }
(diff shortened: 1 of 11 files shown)


crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -122,9 +122,16 @@ impl<S: Spec> CacheWarmUpExecutor<S> {
...
-                            _ = executor.apply_tx_to_in_progress_batch(&baked_tx).await;
-
+                            let res = executor.apply_tx_to_in_progress_batch(&baked_tx).await;
+
+                            match res{
+                                Ok((_, _tx_change_set)) => {
+                                    // TODO
+                                },
+                                Err(_) => {
+                                    continue;
```
bkolad, line 132: Will take care of logging/reporting it in the next PR

crates/module-system/sov-modules-api/src/state/accessors/internals.rs
```diff
@@ -431,6 +438,20 @@ impl<T> RevertableWriter<T> {
     }
 }
 
+impl<S: Spec> RevertableWriter<StateCheckpoint<S>> {
+    /// Change set resulting from transaction execution.
+    pub fn changes(&self) -> TxChangeSet {
+        TxChangeSet {
```
preston-evans98, line 444: Do we need to handle the uncommitted writes from the `StateCheckpoint`? Or maybe it should be a param that we pass explicitly so that we can make the decision based on the call site (i.e. if we use this method *before* committing the revertable writer for the first time, this is ok; but if we've already committed any other revertable writers it may not be)

preston-evans98: On closer examination I see that this didn't change in this PR though, so maybe we should leave that out of scope

preston-evans98 (approved):

PR #1753, opened 2025-09-24
Reduce coverage job memory usage

citizen-stig:
# Description

Goal is to prevent OOM, that has been showing up recently.

*Details*:

* Do not run `switcheroo disable` and run coverage only on default memebers. That will reduce coverage, but coverage for non main crates are not that critical
* set [`CARGO_BUILD_JOBS`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#configuration-environment-variables) to limit parallelism for compilation. This will makes it longer, but more stable.
* Add Cargo profile `coverage`, which limits memory consumption further.

See comments with resources usage change comparison

Unrelated change, but was in TODO list for a while:

* pass `RUSTFLAGS: "-D warnings"` only to check and lint jobs. all test related jobs should not fail if there's warning. This will give more info about tests status in case of non idea linting.

## Testing
Coverage job should pass

## Docs
No updates


citizen-stig:
Here are resources usages before those changes:
Screenshot 2025-09-24 at 19 47 54
Screenshot 2025-09-24 at 19 47 42
Screenshot 2025-09-24 at 19 47 33
Screenshot 2025-09-24 at 19 47 22

citizen-stig:
~10x reduction with `CARGO_BUILD_JOBS=2`
Screenshot 2025-09-24 at 19 58 38

citizen-stig:
And about 5x reduction with `CARGO_BUILD_JOBS=2` with similar runtime:

Screenshot 2025-09-24 at 20 21 21

bkolad (approved):

PR #1754, opened 2025-09-24
`EVM`: Add test for a contract that creates another contract using the `create` opcode.

bkolad:
# Description

*Summary of the changes...*


crates/module-system/sov-test-utils/src/evm/test-data/Store.sol
```diff
@@ -35,4 +38,12 @@ contract SimpleStorage {
             emit SimpleLog(msg.sender, topic1, i, num);
         }
     }
+
```
LogvinovLeon, line 41: This contract is called `SimpleStorage`. This name was OK when it only had get/set/multi-set but now it's a testing framework so we should either split it into multiple contracts or rename it to `Test`

crates/utils/sov-eth-client/src/lib.rs
```diff
@@ -177,6 +178,25 @@ impl TestClient {
...
+    pub async fn deploy_inner_contract(
+        &self,
+        contract_address: H160,
+    ) -> Result<PendingTransaction<'_, Http>, Box<dyn std::error::Error>> {
+        let tx = self.make_tx(
+            Some(contract_address),
+            Some(
+                self.contract.deploy(
+                    ethers::types::Bytes::from_str(
+                        "0x600a600c600039600a6000f3602a60005260206000f3",
```
LogvinovLeon, line 190: What's this magic number?

crates/module-system/sov-test-utils/src/evm/test-data/Store.sol
```diff
@@ -35,4 +38,12 @@ contract SimpleStorage {
...
+    function deploy(bytes memory initCode, uint256 value) public returns (address addr) {
+        assembly {
+            addr := create(value, add(initCode, 0x20), mload(initCode))
```
LogvinovLeon, line 44: Let's try to avoid inline assembly unless unavoidable.
```sol
contract Empty {}

contract Factory {
    event ContractDeployed(address addr);

    function deploy(uint256 value) public returns (address addr) {
        Empty c = (new Empty){value: value}();
        addr = address(c);
        emit ContractDeployed(addr);
    }
}
```

examples/demo-rollup/tests/evm/create.rs
```diff
@@ -0,0 +1,33 @@
...
+        .unwrap();
+
+    // Verify that calling the CREATE opcode twice results in contracts being deployed at different addresses.
+    let inner_addr_1 = get_inner_contract_addr(contract_address, &evm_client).await;
+    let inner_addr_2 = get_inner_contract_addr(contract_address, &evm_client).await;
+
+    assert_ne!(inner_addr_1, inner_addr_2);
+}
+
+async fn get_inner_contract_addr(contract_address: H160, evm_client: &TestClient) -> H160 {
```
LogvinovLeon, line 23: If we return anyhow::Result here we can use ? and have this statement take one line

LogvinovLeon (changes requested):

PR #1759, opened 2025-09-25
`Executor`: Add `oneshot` channel to share transaction changesets between worker and main executor.

bkolad:
# Description

In the optimistic execution workflow, transactions are executed on worker threads before being processed by the main executor.
Once a worker executes a transaction, it needs a way to propagate the computed changeset to the main executor so that the main executor can reuse the fetched values.

This PR introduces a one-shot channel and the necessary plumbing to enable sharing the transaction changeset.
Note that the changeset is not yet applied by the main executor — this functionality will be added in a follow-up PR.


crates/full-node/sov-sequencer/src/preferred/async_batch.rs
```diff
@@ -66,11 +67,28 @@ impl<S: Spec> MaybeAsyncBatch<S> {
...
 }
 
 impl<S: Spec> InjectedControlFlow<S> for MaybeAsyncBatchControlFlow<S> {
+    fn try_warm_up_cache(&mut self, _scratchpad: &mut TxScratchpad<S, StateCheckpoint<S>>) {
+        match self {
+            MaybeAsyncBatchControlFlow::Async {
+                responder: _,
+                maybe_tx_change_set,
+            } => {
+                let _maybe_tx_change_set = maybe_tx_change_set.take();
```
bkolad, line 86: We will use `maybe_tx_change_set` in the next PR to update `_scratchpad`

crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -1573,16 +1574,20 @@ where
...
-        if !batch_size_tracker.can_fit_tx_bytes(baked_tx.data.len()) {
+        let tx_len = baked_tx.data.len();
+        if !batch_size_tracker.can_fit_tx_bytes(tx_len) {
             return Err(AcceptTxError::TxTooBig {
                 current_batch_size: batch_size_tracker.current_batch_size,
                 max_batch_size: batch_size_tracker.max_batch_size,
             });
         }
 
+        cache_warm_up_executor.send_tx(baked_tx.clone());
```
bkolad: I moved the `cache_warm_up_executor.send` here to prevent executing transactions that would be rejected by the prechecks in `process_accept_tx`.
The drawback is that this reduces the time gap between the main executor and the worker.
If necessary, we can move the worker executor back to its previous location.

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -46,9 +72,19 @@ impl<S: Spec> CacheWarmUpExecutor<S> {
         let _ = self.start_block_notification_sender.send(Some(data));
     }
 
-    pub(crate) fn send_tx(&self, tx: FullyBakedTx) {
+    pub(crate) fn send_tx(&self, tx: FullyBakedTx) -> FullyBakedTxWithMaybeChangeSet {
+        let (sender, receiver) = oneshot::channel();
+
         // Skip update if consumer is too slow.
-        let _ = self.tx_sender.try_send(tx);
+        let _ = self.tx_sender.try_send(FullyBakedTxWithTxChangeSetSender {
```
bkolad: Logging/warning will be added as a followup

crates/full-node/sov-sequencer/src/preferred/async_batch.rs
```diff
@@ -66,11 +67,28 @@ impl<S: Spec> MaybeAsyncBatch<S> {
 /// The control flow injector for an async batch.
 #[derive(Debug)]
 pub enum MaybeAsyncBatchControlFlow<S: Spec> {
-    Async(AsyncBatchResponder<S>),
+    Async {
+        responder: AsyncBatchResponder<S>,
+        // If this is the main sequencer executor, it may benefit from using a precomputed
+        // transaction change set from the worker warm-up executor.
+        maybe_tx_change_set: Option<oneshot::Receiver<TxChangeSet>>,
```
preston-evans98, line 74: non-blocking, nit: Maybe we could rename this to `tx_read_set_hint`, since it's supposed to be the set of items that were read, not the items that were changed.

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -8,14 +9,34 @@ use sov_modules_api::Spec;
...
 use std::collections::BTreeMap;
+use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
 
 // We have several work-stealing executor workers for a single main worker, so we don't expect the channel to become full.
 // Even if it does, the sender uses a non-blocking method, meaning a few updates may simply be skipped.
 const TX_CHANNEL_SIZE: usize = 16;
 
+/// A transaction with a way to retrieve the corresponding state changes it may produce.
+pub struct FullyBakedTxWithMaybeChangeSet {
```
preston-evans98, line 23: nit: This also looks like a misnomer to me since (IIUC) we're receiving the *read* set not the write set

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -46,9 +72,19 @@ impl<S: Spec> CacheWarmUpExecutor<S> {
...
         // Skip update if consumer is too slow.
-        let _ = self.tx_sender.try_send(tx);
+        let _ = self.tx_sender.try_send(FullyBakedTxWithTxChangeSetSender {
+            tx: tx.clone(),
+            sender,
+        });
+
+        FullyBakedTxWithMaybeChangeSet {
+            tx,
+            receiver: Some(receiver),
```
preston-evans98: Should this receiver be `None` if `try_send` failed?

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
```diff
@@ -114,21 +150,25 @@ impl<S: Spec> CacheWarmUpExecutor<S> {
...
+                            let baked_tx = FullyBakedTxWithMaybeChangeSet::new(tx_with_sender.tx);
+                            let res = executor.apply_tx_to_in_progress_batch(baked_tx).await;
 
                             match res{
-                                Ok((_, _tx_change_set)) => {
-                                    // TODO
+                                Ok((_, tx_change_set)) => {
+                                    // It ok safe to ignore the error if the receiver was dropped.
+                                    // This can happen if the transaction on the main executor has already finished.
+                                    let _ = tx_with_sender.sender.send(tx_change_set);
```
preston-evans98, line 172: Here it looks like we're sending the changes. For full optimistic execution we'll need the writes as well - but for the current PR we want to send the *reads* right?

bkolad: In the first version I want to handle only reads according to our previous message:
```
For each read in the set from the channel…
Check if the key has already been read/written. If so, ignore it
Put the read value into the cache
```
Once we start working on the full version, we will add the writes, but I would like to close the whole logical flow.

crates/full-node/sov-sequencer/src/preferred/cache_warm_up_executor.rs
bkolad: We will be `try_recv` on the other side of the channel, so this wasn’t an issue, as it won't block. However, I think your suggestion makes the code’s intention clearer. I’ve updated it to return None in case of an error.

preston-evans98 (approved):

PR #1762, opened 2025-09-25
EVM fix `get_code` rpc.

bkolad:
# Description

This PR fixes the get_code endpoint.

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
@@ -226,7 +226,6 @@ where
     ) -> RpcResult<Bytes> {
         debug!("EVM module JSON-RPC request to `eth_getCode`");
         let mut state = self.resolve_state(block_number, state)?;
-
         let code = self
             .accounts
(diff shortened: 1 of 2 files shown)


crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -236,9 +235,10 @@ where
                     .get(&account.code_hash, state.deref_mut())
                     .unwrap_infallible()
             })
+            .map(|code| code.bytecode().clone())
             .unwrap_or_default();
 
-        Ok(code.bytecode().clone())
+        Ok(code.clone())
```
preston-evans98, line 241: nit: I think this clone is unnecessary since we already own the code

preston-evans98 (approved):

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
preston-evans98: Discussed with @bkolad offline. Merging as-is and we'll fix this in a follow up.

PR #1768, opened 2025-09-26
fix: use JRpcParams and fallible parsing in realtime_sendRawTransaction

sashass1315:
- Switch realtime_send_raw_transaction to JRpcParams<'static> and replace unwrap() with parameters.one()? to avoid panics on invalid/missing params.
- Align error handling with other ETH handlers which already use JRpcParams and fallible parsing (returning JSON-RPC Invalid params).
- This unsafe parsing was only present in realtime_send_raw_transaction; all other handlers already used the correct pattern.


sashass1315:

preston-evans98 (approved):
Good catch, thanks @sashass1315!

PR #1769, opened 2025-09-26
Improve handling of archival queries for rollup block heights

preston-evans98:
# Description

This PR makes some fixes to archival state queries by rollup height. These fixes should ensure that it's possible to query for a rollup height even before it's processed by the node.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
```diff
@@ -148,6 +148,57 @@ impl DaLayerWithSubscription {
     }
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_archival_state_is_immediately_available() {
+    let (test_rollup, admin) = create_test_rollup(
```
bkolad, line 153: This is with Manula block production, right?

bkolad (approved):

crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
```diff
@@ -613,31 +623,46 @@ impl<S: Spec + 'static> ApiStateAccessor<S> {
...
                     true_slot_number
-                } else {
+                } else if let Some(max_height) = state
+                    .uncomitted_changes
+                    .as_ref()
+                    .and_then(|c| c.latest_rollup_height())
+                {
+                    // If the requested height is greater than our uncommitted changes, we can't access it.
+                    if max_height < height {
+                        return Err(ApiStateAccessorError::HeightNotAccessible);
```
citizen-stig, line 648: Optional non blocking: would be great to have those values to be in the error

crates/module-system/sov-state/src/storage.rs
```diff
@@ -451,56 +449,14 @@ pub trait StateGetter: core::fmt::Debug + Send + Sync {
     /// Get the value.
     fn get(&self, namespace: Namespace, key: &SlotKey) -> MaybePresentValue<SlotValue>;
 
-    /// Clones the state getter, returning a new type-erased object.
-    fn box_clone(&self) -> Box<dyn StateGetter>;
-}
+    /// Ignore changes after the given rollup height.
```
citizen-stig: Could we please elaborate a little bit more here. Does it mean that those reads won't be accessible to read value or to also read leaf? Does it includes given height? Is there any default value expected?

citizen-stig: Also, does it mean that values will be removed or just hidden temporarily?

For example, if there are values for height 10, 11, 12 and I first pass parameter to be 11 and then 15, would I be able to see values at 12 after second call?

citizen-stig (approved):

crates/module-system/sov-state/src/storage.rs
preston-evans98: I've expanded the comment. Hopefully its clearer now

crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
preston-evans98: Unfortunately, this isn't a simple change because we sometimes need to create this error where the heights are not both known. Leaving out of scope.

crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
preston-evans98: Yep!

PR #1770, opened 2025-09-27
Buffer evm txs with future nonces

preston-evans98:
# Description

This PR adds a simple retry strategy for EVM txs which have future nonces; we check the signature once to recover the account, and then (if the nonce was a near future nonce) poll to see whether the nonce has become current for up to 100ms. 

### DIFF
--- a/crates/full-node/sov-ethereum/src/handlers/mod.rs
+++ b/crates/full-node/sov-ethereum/src/handlers/mod.rs
@@ -114,10 +114,10 @@ where
             }
         }
         // tokio::time::sleep can have unreliable timing under load, so if the total time we've been retrying is too large we'll break the loop early.
-        if start.elapsed().as_millis() > MAX_RETRY_DURATION_MS {
+        if start.elapsed().as_millis() > MAX_BUFFER_DURATION_MS {
             break;
         }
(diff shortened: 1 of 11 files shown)


crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -147,17 +151,47 @@ where
...
+                let expected_nonce = sov_uniqueness::Uniqueness::<S>::default().nonce(&credential_id, &mut state)?.unwrap_or_default();
+                if nonce == expected_nonce {
+                    ethereum.sequencer.accept_tx(tx).await.map_err(|e| {
+                        to_jsonrpsee_error_object(
+                            format!("{} - '{}' ({:?})", e.status, e.message, e.details),
+                            ETH_RPC_ERROR,
+                        )
+                    })?;
+                    return Ok(tx_hash);
+                } else if nonce < expected_nonce {
```
bkolad: We could also add another condition:` if nonce > expected_nonce + 100, return an error`. This indicates that the nonce is too far in the future, so we won't wait for it.

preston-evans98: Good idea. Done

bkolad (approved):

PR #1771, opened 2025-09-29
Prevent metrics loss on shutdown & blob sender logging

citizen-stig:
# Description

Metrics publishing task is now handled in the regular shutdown process, this prevents loss of the metrics that were in the buffer at the time of shutdown:
 * Publishing task receives shutdown receiver and can stop when rollup stops
 * It fetches all metrics from the channel and submits all remaining buffered metrics, so nothing is dropped during shutdown

Various changes in logging:

* Blob sender uses proper wording when timeout for processing is exceeded
* Use structured logging instead of string interpolation. Should help for customers who uses dedicated logging facility
* Runner uses custom span and adds start/stop heights only when they are initialized

- [x] Internal change~~I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.~~

## Testing
Manually tested. It is tricky to test metrics shutdown, as it relies on OnceLock per process.

## Docs
No updates

### DIFF
--- a/crates/full-node/sov-blob-sender/src/lib.rs
+++ b/crates/full-node/sov-blob-sender/src/lib.rs
@@ -481,7 +481,10 @@ impl<Da: DaService, FM: FinalizationManager> TaskState<Da, FM> {
         let res = self.db.set_state(blob_id, state).await;
         if let Err(err) = &res {
             tracing::error!(
-                "BlobSender: unable to save blob state: {state:?}, error: {err}, blob_id: {blob_id}. Shutting down."
+                blob_id,
+                error = ?err,
+                ?state,
+                "BlobSender: unable to save blob state. Shutting down."
             );
         }
         res
@@ -500,7 +503,11 @@ impl<Da: DaService, FM: FinalizationManager> TaskState<Da, FM> {
 
         if let Err(err) = &is_finalized {
             tracing::error!(
-                "BlobSender: unable to check if blob is finalized: {state:?}, error: {err}, blob_id: {blob_id}. Shutting down."
+                blob_id,
+                %blob_hash,
+                error = ?err,
+                ?state,
+                "BlobSender: unable to check if blob is finalized. Shutting down."
             );
         }
 
@@ -533,7 +540,7 @@ impl<Da: DaService, FM: FinalizationManager> TaskState<Da, FM> {
                 ?da_tx_id,
                 blob_processing_timeout = ?self.blob_processing_timeout,
                 ?elapsed,
-                "BlobSender: elapsed time for blob submission exceeded the resubmit interval.",
+                "BlobSender: elapsed time for blob processing exceeded the timeout.",
             );
             return true;
         }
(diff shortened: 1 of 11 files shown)


crates/full-node/sov-blob-sender/src/lib.rs
```diff
@@ -533,7 +540,7 @@ impl<Da: DaService, FM: FinalizationManager> TaskState<Da, FM> {
                 ?da_tx_id,
                 blob_processing_timeout = ?self.blob_processing_timeout,
                 ?elapsed,
-                "BlobSender: elapsed time for blob submission exceeded the resubmit interval.",
+                "BlobSender: elapsed time for blob processing exceeded the timeout.",
```
bkolad, line 543: Shouldn't this be `Published`? We call `check_timeout` in` BlobSubmissionStatus::Processed` => branch

crates/full-node/sov-blob-sender/src/lib.rs
```diff
@@ -629,13 +636,18 @@ impl<Da: DaService, FM: FinalizationManager> TaskState<Da, FM> {
                                     MAX_NB_OF_BLOB_SUBMISSION_RETRIES,
                                     ?da_tx_id,
                                     %blob_hash,
-                                    "Shutting down the rollup. Blob submission failed."
+                                    "Shutting down the rollup. Blob processing wasn't completed on time."
```
bkolad, line 639: Same we are in Publish

crates/full-node/sov-metrics/src/influxdb/publisher.rs
```diff
@@ -113,33 +117,52 @@ pub(crate) async fn metrics_publisher_task(
...
+                    tracing::info!(
+                        drained_metrics = drained_count,
+                        "Drained remaining metrics from channel during shutdown"
+                    );
+                }
+                tracing::info!(
+                    remaining_buffer_size = buffer.len(),
+                    "Metrics task is going to shutdown, send remaining metrics"
+                );
+                if !buffer.is_empty() {
```
bkolad, line 154: Is this possible? `process_measurement` always clears the buffer at the end. And it looks like the buffer can be made an internal variable in `process_measurement`

bkolad (approved):

crates/full-node/sov-blob-sender/src/lib.rs
citizen-stig: It is in `BlobSubmissionStatus::Published`:

https://github.com/Sovereign-Labs/sovereign-sdk/blob/8acc2de3851140a6faf0207274ed12ba56496f2b/crates/full-node/sov-blob-sender/src/lib.rs#L632

So as I understand it, it was published, but not processed. It has tx-receipt. Unless published means something else, rather than being accepted by DA, then I am missing something. 

There's case with re-orgs, but still we cannot conclude that blob wasn't published, only that it wasn't processed, because we are waiting for processing.

We had this mistake several times: we see error "elapsed time for submission" and we look into DA, because we think that it wasn't submitted to DA at all.

crates/full-node/sov-metrics/src/influxdb/publisher.rs
citizen-stig: process_measurement clears buffer only after submission, and submission happens only if size threshold has been exceeded.

So it is clearly possible if there was not enough measurements already at the shutdown, so we force to submit everything we have, even if it is a single measurement.

crates/full-node/sov-blob-sender/src/lib.rs
citizen-stig: answered here https://github.com/Sovereign-Labs/sovereign-sdk/pull/1771#discussion_r2387713430

crates/full-node/sov-blob-sender/src/lib.rs
bkolad: That is correct

PR #1776, opened 2025-09-29
Distinguish between graceful and ungraceful shutdown in more cases

preston-evans98:
# Description

This PR improves (but does not fully fix) handling of errors in the preferred sequencer, making it possible to distinguish between graceful and ungraceful shutdown in more places. This should fix the flaky test `rollup_shuts_down_if_panic_is_triggered`. 

Future PRs should attempt to remove the `exit_rollup` entirely (or at least reduce its usage), and should replace `anyhow` with `thiserror` and pattern matching where possible.


crates/full-node/sov-sequencer/src/preferred/mod.rs
```diff
@@ -736,6 +742,7 @@ where
                 "check_readiness",
             )
             .await
+            .map_err(|_| SequencerNotReadyDetails::Shutdown)?
```
citizen-stig, line 743: Do you think it's worth logging error here?

citizen-stig (approved):
Looks good to me, but would be good to have extra pair of eyes to double check sequencer related changes.

crates/full-node/sov-sequencer/src/preferred/mod.rs
citizen-stig: Ok, this error cannot be printed. I really hope that if the value is `Unexpected` it is logged somewhere or we will be able to quickly find a root cause.

theodorebugnet (approved):
I think it looks good - as long as the new shutdowns reliably also shutdown the node and all other threads in the process, since that was the original reasoning behind `exit_rollup()`. I think `rollup_shuts_down_if_panic_is_triggered()` should cover this though so it looks like we're good.

PR #1781, opened 2025-09-30
Suppress any logs emitted while running in the `SequencerWarmUp` execution context.

bkolad:
# Description

Executors running in the SequencerWarmUp context are used solely for cache warm-up, and their logs are suppressed to avoid noise.

## Linked Issues
- Related to #1743

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/block_executor.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/block_executor.rs
@@ -806,10 +806,13 @@ where
         executor_context,
     } = ctx;
 
-    let _span = tracing::trace_span!(
-        "preferred_seq_bg_task",
-        checkpoint_height = %checkpoint.rollup_height_to_access(),
-    )
+    let _span = match executor_context {
+        ExecutionContext::Sequencer => tracing::trace_span!(ExecutionContext::SEQUENCER),
+        ExecutionContext::SequencerWarmUp => {
+            tracing::trace_span!(ExecutionContext::SEQUENCER_WARM_UP)
+        }
+        ExecutionContext::Node => tracing::trace_span!(ExecutionContext::NODE),
+    }
     .entered();
 
     let stf = StfBlueprint::<S, Rt>::new();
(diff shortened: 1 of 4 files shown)


crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -806,10 +806,13 @@ where
...
-    let _span = tracing::trace_span!(
-        "preferred_seq_bg_task",
-        checkpoint_height = %checkpoint.rollup_height_to_access(),
-    )
+    let _span = match executor_context {
+        ExecutionContext::Sequencer => tracing::trace_span!(ExecutionContext::SEQUENCER),
+        ExecutionContext::SequencerWarmUp => {
+            tracing::trace_span!(ExecutionContext::SEQUENCER_WARM_UP)
+        }
+        ExecutionContext::Node => tracing::trace_span!(ExecutionContext::NODE),
```
bkolad, line 814: I couldn’t move this into a method that takes `ExecutionContext` and returns a `const &'static str`, because Rust required the self parameter to be const as well. Instead, it works here with explicit pattern matching

crates/module-system/sov-modules-rollup-blueprint/src/native_only/logging.rs
```diff
@@ -5,11 +5,32 @@ use std::str::FromStr;
...
+#[derive(Clone, Copy)]
+struct IgnoreSpan(&'static str);
+
+impl<S> Filter<S> for IgnoreSpan
+where
+    S: tracing::Subscriber + for<'a> tracing_subscriber::registry::LookupSpan<'a>,
+{
+    fn enabled(&self, _meta: &tracing::Metadata<'_>, ctx: &Context<'_, S>) -> bool {
+        if let Some(current) = ctx.lookup_current() {
+            if let Some(current) = ctx.lookup_current() {
```
citizen-stig: I am missing it, but why there's 2 if let nested?

bkolad: Ohh mistake
fixed

citizen-stig (approved):

PR #1784, opened 2025-09-30
Add state assertion module

theodorebugnet:
# Description

Adds a test module for use in acceptance testing.
 * Provides a user state value whose old value is asserted on every transaction, so if there's any state mismatch (either between the sequencer and the node, or in during resync) the transactions will immediately start failing
 * Has a method that asserts the main user-visible kernel state value, so if there's a mismatch between the sequencer accepting the transaction and the node executing it the tests will fail
 * Has a method to update accessory state, intended to be checked through the API to again ensure it's identical between sequencer live state and node archival state, and remains correct on resync

Implements transaction generator for the module, for generating user state updates (the other two callmessages don't make sense to generate, as they're meant for manual checks once per block). To do this, `AccountState` in the generator gets a new field since `additional_info` is currently unusable with the way the harness macro is structured.
It will be integrated better into the soak testing lib in a follow-up PR.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/module-implementations/sov-test-state-consistency/src/call.rs
```diff
@@ -0,0 +1,162 @@
...
+        expected_state_root: Vec<u8>,
+        _context: &Context<S>,
+        state: &mut impl TxState<S>,
+    ) -> Result<()> {
+        let mut mismatches = Vec::<String>::new();
+
+        let rollup_height = state.rollup_height_to_access().get();
+        // Sanity check
+        let saved_rollup_height = self.latest_rollup_height.get(state)?.unwrap_or(0);
+        assert_eq!(rollup_height, saved_rollup_height, "Mismatch between rollup height saved from begin_rollup_block_hook: {saved_rollup_height} and the one read from TxState: {rollup_height}");
```
preston-evans98, line 111: Here we panic on failure. Is that really the behavior we want? Should this return an error instead?

preston-evans98: Ah nvm, this is a consistency check in the module not part of the test criteria.  That seems reasonable

theodorebugnet: Yeah, this should never happen: `saved_rollup_height` comes from the begin_rollup_block_hook. The `state` begin hook and the TxState in a transaction should both return the same `rollup_height_to_access()` (unless I'm wrong about this), and if the rollup height changes mid-block it seems like a pretty critical issue.

Since this module is intended for testing, it seems fine to fail ASAP and just crash the rollup if we detect this inconsistency. But even in production this seems like the kind of error that I'd crash the rollup over rather than continue with major state handling bugs.

theodorebugnet: Ah somehow github only showed me your first comment haha. :+1:

preston-evans98 (approved):

PR #1785, opened 2025-09-30
Warm up the main executor cache with `TxChangeset` from the worker executors.

bkolad:
# Description
1. Only the `ExecutionContext::SequencerWarmUp` executor populates reads in `TxChangeSet`. These reads are used to warm up the main executor’s cache.

2. Adds the necessary plumbing to populate the main executor’s cache. The primary entry point is `TxScratchpad::apply_change_set`.

### DIFF
--- a/crates/module-system/sov-modules-api/src/state/accessors/scratchpad.rs
+++ b/crates/module-system/sov-modules-api/src/state/accessors/scratchpad.rs
@@ -1010,7 +1010,7 @@ mod tests {
     }
 
     fn create_srcratchpad<S: Spec>(storage: S::Storage) -> TxScratchpad<S, StateCheckpoint<S>> {
-        let checkpoint = StateCheckpoint::new(stotage, &MockKernel::new(4, 1));
+        let checkpoint = StateCheckpoint::new(storage, &MockKernel::new(4, 1));
         checkpoint.to_tx_scratchpad()
     }
 
--- a/crates/module-system/sov-modules-api/src/state/accessors/scratchpad.rs
+++ b/crates/module-system/sov-modules-api/src/state/accessors/scratchpad.rs
@@ -929,7 +929,7 @@ mod tests {
         let storage_manager = SimpleStorageManager::new();
         let storage = storage_manager.create_storage();
 
-        let mut worker_scratchpad = create_srtatchpad::<TestSpec>(storage.clone());
+        let mut worker_scratchpad = create_srcratchpad::<TestSpec>(storage.clone());
 
         // Key and values.
         let init_reads = vec![(10, 1), (11, 2), (12, 3)];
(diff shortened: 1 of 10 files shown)


crates/module-system/sov-modules-api/src/state/accessors/scratchpad.rs
```diff
@@ -342,8 +345,27 @@ impl<S: Spec, I: StateProvider<S>> PerBlockCache for TxScratchpad<S, I> {
...
+
+    /// Applies changes to the TxScratchpad to warm up its cache.
+    pub fn apply_change_set(&mut self, changes: TxChangeSet) {
+        let mut reads = changes.reads.unwrap();
+
+        // If a write overrides a read, ignore the read.
+        self.filter_out_writes(&mut reads.user, Namespace::User);
+        self.filter_out_writes(&mut reads.kernel, Namespace::Kernel);
+
+        self.inner.inner.add_read_if_not_present(reads);
```
preston-evans98, line 360: nit: We're paying a bit more than we need to here because we take two passes over the array (one to remove the keys that were written, one to add the new reads) and each pass takes O(capacity) time since these are maps - so the filtering on the first pass doesn't help with the time complexity of the next one

crates/module-system/sov-modules-api/src/state/accessors/scratchpad.rs
```diff
@@ -890,4 +918,144 @@ mod tests {
...
+        main_scratchpad.apply_change_set(changeset_from_worker);
+
+        // The main_reads and main_writes were not overridden by the changes from the worker_scratchpad.
+        assert_scratchpad_contains_data(&main_reads, &mut main_scratchpad, namespace);
+        assert_scratchpad_contains_data(&main_writes, &mut main_scratchpad, namespace);
+        // The extra key–value pairs are inserted into the main_scratchpad.
+        assert_scratchpad_contains_data(&[(101, 111), (102, 112)], &mut main_scratchpad, namespace);
+    }
+
+    fn create_srtatchpad<S: Spec>(stotage: S::Storage) -> TxScratchpad<S, StateCheckpoint<S>> {
```
preston-evans98: ni: 
```suggestion
    fn create_srcratchpad<S: Spec>(storage: S::Storage) -> TxScratchpad<S, StateCheckpoint<S>> {
```

preston-evans98 (approved):
LGTM

crates/module-system/sov-modules-api/src/state/accessors/scratchpad.rs
bkolad: `reads` and `writes` here are vecs. Added to #1743. Will talk about this on our next catchup

preston-evans98: Ah, nvm then!

PR #1810, opened 2025-10-06
Refactor gas module structure and tests

LogvinovLeon:
# Description

Reorganized the gas module for better maintainability and improved test quality.

### Module Organization
- Split monolithic `traits.rs` (1300+ lines) into focused submodules
- Separated `GasUnit`, `GasPrice`, `Error`, and implementation macros into dedicated files
- Moved gas meters (`BasicGasMeter`, `SlotGasMeter`, `UnlimitedGasMeter`) to `meters/` subdirectory

### Type System Improvements
- Made `GasUnit`, `GasPrice`, and `BasicGasMeter` implement `Copy`
- Removed 20+ unnecessary `.clone()` calls throughout the codebase
- Added `Copy` bound to `GasArray` trait

### Test Improvements
- Refactored tests to use table-driven/array-based parameterization for conciseness
- Added test helpers and macros (`gas!()`, assertion helpers) to reduce boilerplate
- Extracted test constants to eliminate duplication
- Split large multi-scenario tests into focused individual tests
- Added comprehensive coverage for `charge_linear_gas` edge cases
- Deleted duplicate test file (`charge_gas.rs`)

-----

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -513,16 +513,11 @@ where
...
-            .unwrap_or_else(|| {
-                panic!(
-                    "Cannot overflow after dividing by {COMFORTABLE_GAS_LIMIT_DIVISOR} and multiplying by {COMFORTABLE_GAS_LIMIT_MULTIPLIER}",
-                )
-            });
-        let close_to_gas_limit = remaining_slot_gas.dim_is_less_or_eq(&comfortable_gas_limit);
+        // We want to close when remaining gas is at most 5% of the initial gas limit,
+        // which means we've used at least 95% of the gas.
+        let comfortable_gas_limit =
+            <S as GasSpec>::initial_gas_limit().scalar_division(COMFORTABLE_GAS_LIMIT_DIVISOR);
```
preston-evans98: It looks like we changed the calculation here unless we also adjusted `COMFORTABLE_GAS_LIMIT_DIVISOR` to be `COMFORTABLE_GAS_LIMIT_MULTIPLIER/ COMFORTABLE_GAS_LIMIT_DIVISOR`. Is that right?

crates/module-system/sov-modules-api/Cargo.toml
```diff
@@ -64,6 +64,7 @@ utoipa = { workspace = true, optional = true }
 utoipa-swagger-ui = { workspace = true, optional = true }
 unwrap-infallible = { workspace = true }
 nearly-linear = { workspace = true }
+arrayvec = { workspace = true, features = ["serde", "borsh"] }
```
preston-evans98: What do we use `arrayvec` for in the new PR? A quick grep didn't show any usage...

LogvinovLeon: It's a leftover. I started by replacing a String with a stack-String but later realized that we only use it in a macro that accepts 'static str. Will remove

crates/module-system/sov-modules-api/src/gas/tests/charge_gas.rs
```diff
@@ -1,50 +0,0 @@
...
-use sov_test_utils::MockZkvm;
-
-use crate::default_spec::DefaultSpec;
-use crate::execution_mode::Native;
-use crate::{Amount, BasicGasMeter, GasArray, GasMeter, GasPrice, GasUnit};
-
-type S = DefaultSpec<MockDaSpec, MockZkvm, MockZkvm, Native>;
-
-#[test]
-fn try_charge_gas() {
```
preston-evans98, line 11: It looks to me like these tests were completely removed. Is that correct?

crates/full-node/sov-sequencer/src/preferred/inner.rs
LogvinovLeon: This is in fact tricky and I spent a great amount of time figuring this our. 

The original signature of: `checked_scalar_product` is: `fn checked_scalar_product(&self, scalar: $u) -> Option<Self> {`
Therefore it does not mutate self. Because we don't (didn't) assign the result of this operation - we loose it. Compiler doesn't complain as it's not `#[must_use]`. If you enable it - tests start failing as they rely on this behavior. Current behavior is correct as we take initial limit and divide by 20 getting to 5%. We later check if we have less than 5% left and close if so. I'll delete unused const

crates/module-system/sov-modules-api/src/gas/tests/charge_gas.rs
LogvinovLeon: Yes. There were some tests that were duplicate of the tests that were in traits.rs. And some tests that were migrated. You can look commit by commit around the place where I touched this file to make sure I didn't loose any meaningful tests

preston-evans98 (commented):
This change seems really nice - thanks for the effort @LogvinovLeon!

Unfortunately, the PR is really difficult to review because of the large size and the mix of API changes with simple code moves. I caught one or two (minor) issues, but with such a sprawling diff I'm afraid some may have slipped through. It looks like this history contains a bunch of smaller, more atomic changes - would it be easy to split this into chunks where each chunk does one logical thing? 

As an example...
- One PR to make the types `copy` and replace the dead clones
- (If needed) One PR to change the API from `&mut self` to `self` for the `scalar` operations. (If small, this could easily be lumped in with the previous change)
- One PR to split the large file(s) into smaller units - ideally with *zero* other changes so that the code move checker works well :)
- One PR to make the test changes)

LogvinovLeon:
I'm happy to split it in parts but almost each commit here can be a separate PR.
It might be easier to just go commit by commit as I tried to not mix things up within the commit.
If you wish - I'll split it into separate PRs but managing separate stacked PRs can be painful as one relies on the changes from another and I'll need to constantly be rebasing if applying fixes to the prior ones.

crates/full-node/sov-sequencer/src/preferred/inner.rs
preston-evans98: Ah, I see. Maybe this would be clearer if we renamed `comfortable_gas_limit` to something `cushion_amount`? Then the comparison `remaining_slot_gas < cushion_amount` is easy to parse

LogvinovLeon: I've fixed it properly by using both constants in the last commit

preston-evans98:
If you want to split it into chunks that small you certainly can, but 31 PRs feels like overkill to me :)

If you can find a reviewer who's willing to sign off on this I also won't be a blocker, but I'm just not confident enough in my review to approve atm. Have to step away in a few min but if it's still open tomorrow I'll see if going thru the commits in smaller chunks makes it easier to follow!

bkolad (approved):
The refactoring looks good to me. I didn’t notice any functional changes, except that the arithmetic operations on GasArray have been made pure.

PR #1823, opened 2025-10-07
Set blockenv at genesis.

bkolad:
# Description

Before the first `block_env` was set by  `begin_rollup_block_hook`. 

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/genesis.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/genesis.rs
@@ -1,6 +1,8 @@
 use alloy_consensus::constants::KECCAK_EMPTY;
 use alloy_primitives::{Address, B256, U256};
 use alloy_primitives::{BlockNumber, Bytes};
+use revm::context::BlockEnv;
+use revm::context_interface::block::BlobExcessGasAndPrice;
 use revm::primitives::hardfork::SpecId;
 use revm::state::AccountInfo;
 use sov_address::{EthereumAddress, FromVmAddress};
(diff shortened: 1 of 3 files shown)


preston-evans98 (approved):

crates/module-system/module-implementations/sov-evm/src/genesis.rs
```diff
@@ -86,6 +92,24 @@ where
 
         Ok(())
     }
+
+    // `BlockEnv`` will be overridden in begin_rollup_block_hook.
+    fn env_from_block(&self, block: &Block) -> BlockEnv {
```
LogvinovLeon: There are now two places that create `block_env`. Here and in hooks. And if one wants to see - what's the difference between them - need to read both places line by line.
Let's create a helper function that will set:
* `basefee`, `gas_limit`, `blob_excess_gas_and_price`, `beneficiary`, other defaults (common things)
and accept as a parameter:
* `number`, `timestamp`, `prevrandao`.
Then in here we'll have:
`self.env_for_block(header.number, header.timestamp, None)`
and in hooks:
`self.env_for_block(parent + 1, chain_state.time(), Some(pre_state_user_root))`

Will be more clear - what's actual logic and what's just common setup.

LogvinovLeon (changes requested):

crates/module-system/module-implementations/sov-evm/src/genesis.rs
bkolad: fixed

LogvinovLeon (approved):

PR #1825, opened 2025-10-07
Switch to cache friendly SlotKey representation

preston-evans98:
# Description

This PR changes `SlotKey` to be an enum of `[u8;70]` and `Arc<Vec<u8>>`, with `[u8;70]` being by far the dominant case. This means that for all of our modules and *almost* all user modules, the keys will be stack allocated. This improves performance by about 15%. 

In order to make the new representation efficient, this PR changes the way that slotkeys are computed. Previously, we generated keys by concatenating the output of the `module_path!()` macro with the name of the struct and field. This was guaranteed by the compiler to generate a globally unique prefix, but at the cost of a long identifier - each one was on the order of 50 bytes without the user key. 

In the new version, users manually specify unique prefixes for each module in their `constants.toml`. (Note that a check for duplicate IDs has been added at genesis, since this could be a footgun). 

This allows us to prefix keys with a 1-byte module ID plus a 1-byte field ID, reducing the prefix length from ~50 to 2 bytes. Since keys themselves are typically less than 67 bytes, this means we can almost always fit the entire key in a 70-byte array - which we can afford to allocate on the stack. If the key is larger than 70 bytes (including metadata) we fall back to heap allocating.


constants.toml
```diff
@@ -142,3 +142,50 @@ ELASTICITY_MULTIPLIER = 2
...
+INTERCHAIN_GAS_PAYMASTER_DISCRIMINANT = 13
+MERKLE_TREE_HOOK_DISCRIMINANT = 14
+WARP_DISCRIMINANT = 15
+PAYMASTER_DISCRIMINANT = 16
+REVENUE_SHARE_DISCRIMINANT = 17
+
+
+# Test modules
+MODULE_WITH_VERSIONED_STATE_ACCESS_IN_SLOT_HOOK_DISCRIMINANT = 233
+GAS_TESTER_DISCRIMINANT = 233
```
bkolad, line 169: MODULE_WITH_VERSIONED_STATE_ACCESS_IN_SLOT_HOOK_DISCRIMINANT and GAS_TESTER_DISCRIMINANT have the same discriminant

crates/module-system/sov-state/src/storage.rs
```diff
@@ -25,46 +25,188 @@ use crate::{
...
-    #[derivative(
-        Debug = "ignore",
-        PartialEq = "ignore",
-        Hash = "ignore",
-        Ord = "ignore"
-    )]
-    #[sov_wallet(skip)]
-    display_fn: Option<ArcFormatFn>,
+    // // TODO: Handle extra long keys
+    // pub(crate) prefix: Prefix,
```
bkolad: Let's remove this comment.

constants.toml
preston-evans98: Good catch! Fixed!

bkolad (approved):

crates/module-system/sov-state/src/storage.rs
preston-evans98: Done

PR #1826, opened 2025-10-08
Move tx cache warm-up outside protected section and add parallel `EVM` soak test mode

bkolad:
# Description

1. The transaction  is now sent to the `cache_warm_up_executor` within `accept_tx`, rather than inside the `inner` protected section. This allows the `cache_warm_up_executor` to start executing the transaction sooner.
2. `ProvableStorageCache` now has a default capacity set to `DEFAULT_CACHE_CAPACITY`.
3. A new mode has been added to the EVM soak test. In this mode, simple EVM transfers are sent in parallel, each from a different address.


crates/utils/sov-evm-soak-testing/src/transfer.rs
```diff
@@ -0,0 +1,43 @@
...
+    foucet_private_key: &str,
+    rpc_addr: SocketAddr,
+) -> anyhow::Result<()> {
+    let priv_keys: Vec<_> = generate_priv_keys(count, foucet_private_key)?
+        .into_iter()
+        .map(|pk| (RpcClient::alloy_address(&pk), pk))
+        .collect();
+
+    // Send funds to all participants in the benchmark.
+    let foucet_client = RpcClient::new(foucet_private_key, rpc_addr).await;
```
preston-evans98: nit ```suggestion
    let faucet_client = RpcClient::new(faucet_private_key, rpc_addr).await;
```

File: crates/utils/sov-evm-soak-testing/src/transfer.rs [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+    loop {
+        let current_block_number = foucet_client.block_number().await;
+        if current_block_number > block_number {
+            break;
+        }
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+    }
+
+    // The participants simultaneously return a portion of the funds to the faucet.
+    let foucet_address = RpcClient::alloy_address(foucet_private_key);
```
preston-evans98: nit
```suggestion
    let faucet_address = RpcClient::alloy_address(faucet_private_key);
```

crates/full-node/sov-sequencer/src/preferred/mod.rs
```diff
@@ -832,14 +835,17 @@ where
             return Err(shut_down_error());
         }
 
+        let tx_len = baked_tx.data.len();
+        let baked_tx = self.cache_warm_up_executor.send_tx(baked_tx);
```
preston-evans98, line 839: I see that we send to the executor before checking the signature. What's the tradeoff for that vs checking the signature first?

preston-evans98 (approved):

crates/full-node/sov-sequencer/src/preferred/mod.rs
bkolad: The signature check is performed inside the protected section, so for non-EVM transactions, signatures are verified serially (inside the inner). We could also extract the sig check from inner.rs.

crates/utils/sov-evm-soak-testing/src/transfer.rs
bkolad: fixed

crates/utils/sov-evm-soak-testing/src/transfer.rs
bkolad: fixed

PR #1829, opened 2025-10-09
Sequencer (EVM): Add cursor support for log pagination.

bkolad:
# Description

1. `eth_getLogsWithCursor method`:
Adds the method with an empty implementation. The actual logic and corresponding tests will be implemented in the next PR.

2. Preliminary test for the new endpoint:
Adds a test that simply verifies the `eth_getLogsWithCursor` endpoint is callable. Comprehensive tests will follow in the next PR.

3. Cursor support in the Ethereum sequencer:
Introduces a `cursor` mechanism to enable log pagination. When a client requests more logs than the configured `max_log_limit` through the `eth_getLogsWithCursor` endpoint, a partial result will be returned along with a cursor. This cursor can be used in subsequent queries to retrieve the remaining logs.


crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -207,3 +266,164 @@ where
...
+    _: Extensions,
+) -> Result<Vec<Log>, ErrorObjectOwned>
+where
+    S: Spec,
+    Seq: Sequencer<Spec = S>,
+    S::Address: FromVmAddress<EthereumAddress>,
+    Seq::Rt: HasKernel<S> + EthereumAuthenticator<S> + Default + Send + Sync + 'static,
+{
+    let _ = parameters.one::<FilterWithCursor>()?;
+
```
bkolad: This will be implemented in the following PR.

examples/demo-rollup/tests/evm/evm_subscribe.rs
```diff
@@ -124,6 +124,7 @@ async fn evm_test_log_subscription_with_pending_blcok() {
 // This behavior is used to verify the filtering mechanism.
 #[tokio::test(flavor = "multi_thread")]
 async fn evm_test_logs_filter() {
+    sov_test_utils::initialize_logging();
```
LogvinovLeon, line 127: Why do we need logging in tests?

examples/demo-rollup/tests/evm/evm_logs.rs
```diff
@@ -151,3 +137,123 @@ fn check_logs(filter: &Filter, logs: Vec<alloy_rpc_types_eth::Log>, expected_nb_
...
+    client: SimpleStorageClient,
+    contract_address: alloy_primitives::Address,
+}
+
+impl RollupAndClient {
+    async fn new(max_log_limit: usize) -> RollupAndClient {
+        let ext = SeqConfigExtension { max_log_limit };
+
+        let (test_rollup, evm_client, _) = setup_with_simple_storage(0, ext).await;
+        let contract_address = evm_client.alloy_deploy_contract().await;
```
LogvinovLeon, line 212: Let's not use the legacy way to deploy contracts. Let's use alloy wrappers generated by `sol!` macro.
Similar to how we do with `ERC20/Uni` in soak tests.

examples/demo-rollup/tests/evm/evm_logs.rs
```diff
@@ -151,3 +137,123 @@ fn check_logs(filter: &Filter, logs: Vec<alloy_rpc_types_eth::Log>, expected_nb_
...
+        test_rollup.wait_for_next_blocks(1).await;
```
LogvinovLeon: Why? This will probably hide some issues

examples/demo-rollup/tests/evm/evm_logs.rs
```diff
@@ -151,3 +137,123 @@ fn check_logs(filter: &Filter, logs: Vec<alloy_rpc_types_eth::Log>, expected_nb_
...
+            contract_address,
+        }
+    }
+
+    async fn produce_logs(
+        &self,
+        nb_of_txs: u32,
+        nb_of_logs_per_tx: u32,
+    ) -> Vec<alloy_primitives::TxHash> {
+        let mut tx_hashes = Vec::new();
```
LogvinovLeon: We know the capacity here

examples/demo-rollup/tests/evm/evm_logs.rs
```diff
@@ -151,3 +137,123 @@ fn check_logs(filter: &Filter, logs: Vec<alloy_rpc_types_eth::Log>, expected_nb_
...
+        for i in 0..nb_of_txs {
+            let hash = self
+                .client
+                .alloy_emit_logs(self.contract_address, i, nb_of_logs_per_tx)
```
LogvinovLeon, line 231: Let's use calls generated by `sol!` macro

examples/demo-rollup/tests/evm/evm_logs.rs
```diff
@@ -151,3 +137,123 @@ fn check_logs(filter: &Filter, logs: Vec<alloy_rpc_types_eth::Log>, expected_nb_
...
+                .await;
+            tx_hashes.push(hash);
+            if i % 3 == 0 {
```
LogvinovLeon: This seems arbitrary. If we try to emit logs here - let's emit logs. It's it's important for our test to intersperse logs with blocks - let's express it in test logic. I don't think that the caller of this function would expect it to do that

examples/demo-rollup/tests/evm/evm_logs.rs
```diff
@@ -151,3 +137,123 @@ fn check_logs(filter: &Filter, logs: Vec<alloy_rpc_types_eth::Log>, expected_nb_
...
+                self.test_rollup.wait_for_next_blocks(1).await;
+            }
+        }
+
+        tx_hashes
+    }
+
+    async fn get_tx_counet(&self) -> u64 {
```
LogvinovLeon: Typo

examples/demo-rollup/tests/evm/evm_logs.rs
```diff
@@ -151,3 +137,123 @@ fn check_logs(filter: &Filter, logs: Vec<alloy_rpc_types_eth::Log>, expected_nb_
...
+    (cursor.tx_index_absolute as u32) * nb_of_logs_per_tx + cursor.log_index_in_tx
+}
+
+fn new_fileter_for_all_logs() -> Filter {
+    Filter::new()
+        .from_block(0)
+        .to_block(BlockNumberOrTag::Latest)
+}
+
+struct RollupAndClient {
```
LogvinovLeon, line 201: I don't think that this structure is a good abstraction and I think that the name suggests it. Let's keep it in two separate variables. One for alloy contract and one for test_rollup.

crates/utils/sov-rpc-eth-types/src/filter_with_cursor.rs
```diff
@@ -0,0 +1,17 @@
...
+use alloy_rpc_types::Log;
+
+/// Filter for logs with cursor.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+pub struct FilterWithCursor {
+    pub cursor: Option<u128>,
+    pub filter: Filter,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
```
LogvinovLeon: WHat's the reason for using fully-qualified idents here? Why not `import serde::{Ser, Deser}` and derive `Serialize`?

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -207,3 +277,164 @@ where
...
+    S::Address: FromVmAddress<EthereumAddress>,
+    Seq::Rt: HasKernel<S> + EthereumAuthenticator<S> + Default + Send + Sync + 'static,
+{
+    let FilterWithCursor { cursor, filter } = parameters.one::<FilterWithCursor>()?;
+    let cursor = cursor.map(Cursor::unpack);
+    logs_for_filter(filter, cursor, ethereum).await
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+/// Cursor indicating where to start processing.
```
LogvinovLeon, line 294:
```
///   [ block_height (32b) ][ tx_index_absolute (64b) ][ log_index_in_tx (32b) ]
///   127──────────────96   95───────────────────32    31──────────────────0
```

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -207,3 +277,164 @@ where
...
+        let block_height = (v >> 96) as u32;
+        let tx_index_absolute = ((v >> 32) & 0xFFFF_FFFF_FFFF_FFFFu128) as u64;
+        let log_index_in_tx = (v & 0xFFFF_FFFFu128) as u32;
+
+        Self {
+            block_height,
+            tx_index_absolute,
+            log_index_in_tx,
+        }
+    }
```
LogvinovLeon:
```rs
    /// Packs `Self` into a `u128`.
    pub fn pack(self) -> u128 {
        let mut bytes = [0u8; 16];
        bytes[0..4].copy_from_slice(&self.block_height.to_be_bytes());
        bytes[4..12].copy_from_slice(&self.tx_index_absolute.to_be_bytes());
        bytes[12..16].copy_from_slice(&self.log_index_in_tx.to_be_bytes());
        u128::from_be_bytes(bytes)
    }

    /// Unpacks a `u128` into `Self`.
    pub fn unpack(v: u128) -> Self {
        let bytes = v.to_be_bytes();
        let block_height = u32::from_be_bytes(bytes[0..4].try_into().unwrap());
        let tx_index_absolute = u64::from_be_bytes(bytes[4..12].try_into().unwrap());
        let log_index_in_tx = u32::from_be_bytes(bytes[12..16].try_into().unwrap());
        Self { block_height, tx_index_absolute, log_index_in_tx }
    }

```

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -207,3 +277,164 @@ where
...
+    let FilterWithCursor { cursor, filter } = parameters.one::<FilterWithCursor>()?;
+    let cursor = cursor.map(Cursor::unpack);
+    logs_for_filter(filter, cursor, ethereum).await
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+/// Cursor indicating where to start processing.
+pub struct Cursor {
+    /// Starting block height for this cursor (inclusive). Top 32 bits.
+    pub block_height: u32,
```
LogvinovLeon: BlockNumber is u64 in alloy primitives: https://github.com/alloy-rs/core/blob/2ce907970edae54f66127af63e2c88594d4642d0/crates/primitives/src/aliases.rs#L96
We also index blocks with u64.
And there is a good reason. MegaETH has a block time of sub 10ms. At this speed - u32 overflows in around a year.

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -207,3 +277,164 @@ where
...
+    Seq::Rt: HasKernel<S> + EthereumAuthenticator<S> + Default + Send + Sync + 'static,
+{
+    let FilterWithCursor { cursor, filter } = parameters.one::<FilterWithCursor>()?;
+    let cursor = cursor.map(Cursor::unpack);
+    logs_for_filter(filter, cursor, ethereum).await
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+/// Cursor indicating where to start processing.
+pub struct Cursor {
```
LogvinovLeon, line 295: Let's move to a separate module. This is not part of the handlers. It's more like `rpc-types.rs`

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -207,3 +277,164 @@ where
...
+    /// Starting block height for this cursor (inclusive). Top 32 bits.
+    pub block_height: u32,
+    /// Absolute index of the first transaction to process (inclusive). Middle 64 bits.
+    pub tx_index_absolute: u64,
+    /// Index of the first log within that transaction to process (zero-based, inclusive). Lowest 32 bits.
+    pub log_index_in_tx: u32,
+}
+
+impl Cursor {
```
LogvinovLeon, line 304: Let's impl `pub fn new(_, _, _)` and use it.

crates/utils/sov-rpc-eth-types/src/filter_with_cursor.rs
```diff
@@ -0,0 +1,17 @@
+use alloy_rpc_types::Filter;
+use alloy_rpc_types::Log;
+
+/// Filter for logs with cursor.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+pub struct FilterWithCursor {
```
LogvinovLeon, line 8: derive new

crates/utils/sov-rpc-eth-types/src/filter_with_cursor.rs
```diff
@@ -0,0 +1,17 @@
...
+    pub cursor: Option<u128>,
+    pub filter: Filter,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// Logs and the next cursor.
+pub struct LogsWithMaybeCursor {
```
LogvinovLeon, line 16: derive new

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -34,13 +37,16 @@ where
     S::Address: FromVmAddress<EthereumAddress>,
     Seq::Rt: HasKernel<S> + EthereumAuthenticator<S> + Default + Send + Sync + 'static,
 {
-    logs_for_filter(parameters.one::<Filter>()?, ethereum).await
+    logs_for_filter(parameters.one::<Filter>()?, None, ethereum)
+        .await
+        .map(|r| r.logs)
```
LogvinovLeon:
```suggestion
        .await?
        logs
```

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -34,13 +37,16 @@ where
...
 }
 
 async fn logs_for_filter<S, Seq>(
```
LogvinovLeon: Let's create a struct `EthLogsService`. Than we can express all the trait bouds in one place. And also move it to a separate submodule and not keep it in handlers

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -51,119 +57,182 @@ where
...
 
-    let start = get_block_nr(from_block, &evm, state)?;
+    let start = match maybe_cursor {
+        Some(cursor) => cursor.block_height as u64,
+        None => get_block_nr(from_block, &evm, state)?,
+    };
+
     let end = get_block_nr(to_block, &evm, state)?;
 
     // We just validated that `start` and `end` are not pending.
```
LogvinovLeon, line 153: Where did we validate that?

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -51,119 +57,182 @@ where
...
+        state,
+    )?;
+
+    Ok(LogsWithMaybeCursor {
+        logs: rpc_logs,
+        cursor: next_cursor.map(|c| c.pack()),
+    })
 }
 
 fn logs_for_blocks_range<S>(
```
LogvinovLeon, line 131: One is called logs_for_x and another logs_from_x. Let's standardise

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -207,3 +277,164 @@ where
...
+                    );
+
+                    return Err(to_jsonrpsee_error_object(msg, ETH_RPC_ERROR));
+                }
+
+                (
+                    Range {
+                        start: cursor.tx_index_absolute,
+                        end: block.transactions.end,
+                    },
```
LogvinovLeon:
```suggestion
                    cursor.tx_index_absolute..=block.transactions.end;
```

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
```diff
@@ -207,3 +277,164 @@ where
...
+        Self {
+            block_height,
+            tx_index_absolute,
+            log_index_in_tx,
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct CursorIndexses {
```
LogvinovLeon, line 343: I don't see a point in this struct. Can't we just implement this logic on the cursor itself?
Also - let's not mix RPC stuff with logic. This should return structured error that later gets converted to RPC error in the handler. That way we can unit-test this logic testing for  an error using assert_eq!

LogvinovLeon (changes requested):

examples/demo-rollup/tests/evm/evm_subscribe.rs
bkolad: I noticed the test is flaky. Added logging so we have more info it happens next time.

examples/demo-rollup/tests/evm/evm_logs.rs
bkolad: fixed

examples/demo-rollup/tests/evm/evm_logs.rs
bkolad: fixed

examples/demo-rollup/tests/evm/evm_logs.rs
bkolad: fixed

examples/demo-rollup/tests/evm/evm_logs.rs
bkolad: We already use `deploy_contract` and `alloy_deploy_contract` in `demo-rollup/examples/evm`. Let's not add a third option. This can be done as a separate change in another P.

examples/demo-rollup/tests/evm/evm_logs.rs
bkolad: Same. This should be done as another PR for all the tests if demo-rollup/tests/evm

examples/demo-rollup/tests/evm/evm_logs.rs
bkolad: It’s a private struct used only in this file. The tests were harder to read without it. I’m happy to rename it if you’ve got suggestions.

crates/utils/sov-rpc-eth-types/src/filter_with_cursor.rs
bkolad: We use both conventions across the code base for serde.  Changed here tough.

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: We would need to add `Ok` around as well. What is wrong with `map`?

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: `get_block_nr` checks it (which is called before)

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: fixed

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: Fixed.

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: Actually, this is an internal part of the handlers — API clients only see a u128. The struct is public because I convert u128 to a Cursor in tests for more detailed checks.

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: This approach would be more error-prone. Using `new` makes it easy to accidentally pass the arguments in the wrong order (since there are two `u32` values and one `u64`). With the explicit syntax, the field names make the code clearer and safer.

crates/utils/sov-rpc-eth-types/src/filter_with_cursor.rs
bkolad: The fields are public, and `new` wouldn’t add any additional logic. Why introduce a second way to initialize the struct when one is enough? I don’t see the benefit unless we make the fields private — which we don't want.

crates/utils/sov-rpc-eth-types/src/filter_with_cursor.rs
bkolad: Same

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: In that case, we won't be able to use u128. I will see what to do about it.

crates/utils/sov-rpc-eth-types/src/filter_with_cursor.rs
LogvinovLeon: 4 lines vs 1 line. We have a lot of functions that are long and:
```
Struct::new(a, b)
```
is more concise than
```
Struct {
    a,
    b
}
```

LogvinovLeon: Especially visible in test tables

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: This is the signature of:
```
fn logs_for_block<S>(
    ...
    block_height: u64,
    indexes_from_cursor: Option<CursorIndexes>
)
```
If I passed a `Cursor` instead of `CursorIndexes`, the `logs_for_block` function would have access to both `block_height` (which should always be used) and `cursor.block_height` (which should never be used). I wanted to avoid passing redundant data to a function—especially when one of the values should never be used.

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: Added EthLogsService.
Added `EthLogsService`.
What would be the point of moving it to a separate module?
1. It’s convenient that we don’t have to click through so many layers of methods.
2. If we decide to do that, it should be done in a separate PR, so we don’t mix logic changes with pure refactoring.

examples/demo-rollup/tests/evm/evm_logs.rs
bkolad: fixed

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
LogvinovLeon: Agree in general but not here. Here the order is clear.
`Point(1, 2, 3)` is obviously: `Point(X: 1, Y: 2, Z: 3)`
Same with `Color(255, 255, 255)`.
Here we have three levels of indexing each one within the previous one.
It's like `Address("Poland", "Mazowieckie", "Warszawa")`.
I think if someone writes: `Point(X, Z, Y)` it will still be visible on a review even without field names

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: fixed with new logic

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: Now the data is hex encoded. Prefer not to put the details of the implementation in the documentation.

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: fixed

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
LogvinovLeon: Agree here. Without OK and map it was more imperative and collapsed to a single line. But Ok spreads it out so let's leave as it.
I try to avoid maps on options/results whenever ? can be used as it's usually more straight-forward.
```
let x = op1()?;
let y = op2(x)?;
Ok(y)
```
vs:
```
op1().map(|x| op2(x))
```

In the first option I see that the operation can fail but I can immediately drop this context. My logic remains linear. In the second - my logic continues in a closure. If I get a second result there - it's harder to handle.

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
LogvinovLeon: LogsService should be separate from RPC. It should not have:
`eth_get_logs` and `eth_get_logs_with_cursor`. Those are RPC methods and they should live in handlers and be very simple. Conversion/validation.
On the other side: `logs_for_block_hash`, `logs_for_blocks_range`, `logs_for_block` can live on LogsService.
Later we can:
`LogsService::new(filter, limits, maybe_cursor, state).by_block_hash(block_hash);`
`LogsService::new(filter, limits, maybe_cursor, state).by_block_range(from_block, to_block);`

crates/full-node/sov-ethereum/src/handlers/get_logs.rs
bkolad: `Point` is a universal type across many domains, while `Cursor` is a local type that will be unknown to the person who sees it for the first time, so I don’t think the analogy applies. Even with `Point`, I don’t mind writing it explicitly. The main concern is that some people will use a constructor and others won’t, which will lead to inconsistent conventions across the codebase + I don't see that the ordering is obvious. I’m happy to discuss this on a call or with the rest of the team

crates/utils/sov-rpc-eth-types/src/filter_with_cursor.rs
bkolad: We only use `FilterWithCursor { .. }` in one place — inside `get_logs_with_cursor`.

I do have a broader question though: do you think every struct with public fields should have a constructor? I don’t think that should be the case. Personally, I find explicit field names make the code easier to read.

I don’t mind adding a constructor here, since it doesn’t make much difference for this type, but it does introduce a bit of redundancy.

LogvinovLeon (approved):

Issue #1518, opened 2025-08-12
[TRACKER] Nonce and Generation combo support


citizen-stig:
## Description

SDK customers needs to be able to have nonce based transaction at the same time as generation based, for instance to be able to quickly cancel all outstanding orders

## Checklist

 - [x] Modify Transaction V0 to have explicit parameter https://github.com/Sovereign-Labs/sovereign-sdk/pull/1477
 - [x] Modify Dedup endpoint to support both uniqueness types provided by the SDK https://github.com/Sovereign-Labs/sovereign-sdk/pull/1491
 - [x] Update simulate endpoint to support both https://github.com/Sovereign-Labs/sovereign-sdk/pull/1830
 - [ ] Notify customers about new feature and ensure that everything works as expected
 - [ ] Document the feature and **review the generated Rustdoc**.

PR #1830, opened 2025-10-09
Refactor simulation endpoint

### LINKED ISSUE #1537: Improve `/rollup/simulate` response
In the case of failures the response is like:

```
CustomError("Transaction simulation reverted: Receipt { receipt: ReceiptInner { outcome: \"reverted\", content: \"eyJnYXNfdXNlZCI6WzE4NjMzLDE4NjMzXSwicmVhc29uIjoiRmFpbGVkIHRvIHRyYW5zZmVyIHRva2VuX2lkPXRva2VuXzFueWwwZTB5d2VyYWdmc2F0eWd0MjR6bWQ4anJyMnZxdHZkZnB0empoeGtndXoyeHh4M3ZzMHkwN3U3In0=\" } }, reason: ")
```

We should return the `content` as plain text instead of needing to decode base64.

ross-weir:
# Description

Refactors simulation endpoints inputs & outputs to make it easier to use.

Inputs can simply be a `sender` in the form of a credential id and the json for a runtime call (no need to serialize). With more advanced simulation parameters optional.

Outputs include the gas used, events as JSON (previously base64 binary), and a string representation of the error that occurred (if any). The error loses important context about what actually caused the revert due to the way errors are currently implemented and returned in modules, I have a dream to fix this one day with https://github.com/Sovereign-Labs/sovereign-sdk/pull/1419 but for now it's not ideal.

The old endpoint will be removed once it's confirmed it's not used by zeta.

## Linked Issues
- closes #1537
- https://github.com/Sovereign-Labs/sovereign-sdk-wip/issues/1753

### DIFF
--- a/crates/full-node/sov-api-spec/openapi-v3.yaml
+++ b/crates/full-node/sov-api-spec/openapi-v3.yaml
@@ -465,6 +465,32 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/SimulateExecutionResponse"
+  /rollup/simulate-v2:
+    post:
+      summary: Simulate transaction execution (v2)
+      description: |-
+        Simulates transaction execution against the current blockchain state without committing changes.
+        This endpoint provides detailed gas estimation, event emission, and error handling.
+        Useful for estimating gas costs, testing transaction behavior, and debugging.
+      operationId: simulate_v2
+      tags:
+        - Rollup
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/SimulateParametersV2"
+      responses:
+        "200":
+          description: Simulation completed successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/SimulateOutcomeV2"
+        "400":
+          $ref: "#/components/responses/ApiErrorResponse"
+        "500":
+          $ref: "#/components/responses/ApiErrorResponse"
   /rollup/sync-status:
     get:
       summary: Get the current sync status of the rollup
@@ -856,10 +882,15 @@ components:
           type: integer
         hyperlane_domain:
           type: integer
+        address_prefix:
+          type: string
+          description: The HRP used for addresses in this rollup (e.g., "sov"). Doesn't apply to solana/eth addressed apps.
+          example: "sov"
       required:
         - chain_name
         - chain_id
         - hyperlane_domain
+        - address_prefix
     Hash:
       type: string
       title: Hash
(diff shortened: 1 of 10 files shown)


citizen-stig:
I think those also might be related:

* https://github.com/Sovereign-Labs/sovereign-sdk-wip/issues/1753
* Item in here: https://github.com/Sovereign-Labs/sovereign-sdk/issues/1518 about auth support. I don't know if it is still relevant.

crates/full-node/sov-api-spec/openapi-v3.yaml
```diff
@@ -856,10 +882,15 @@ components:
...
           type: integer
+        address_prefix:
+          type: string
+          description: The HRP used for addresses in this rollup (e.g., "sov"). Doesn't apply to solana/eth addressed apps.
+          example: "sov"
       required:
         - chain_name
         - chain_id
         - hyperlane_domain
+        - address_prefix
```
citizen-stig, line 893: If it does not applly for solana/eth, why it is marked as required?

crates/full-node/sov-api-spec/openapi-v3.yaml
```diff
@@ -1142,3 +1173,137 @@ components:
...
+      type: object
+      description: Optional sequencer configuration for simulation requests
+      properties:
+        rollup_address:
+          type: string
+          description: Custom rollup address for the sequencer
+          example: "sov1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq5k2jj4"
+        da_address:
+          type: string
+          description: Custom data availability address for the sequencer
```
citizen-stig, line 1237: Which address does it use if not provided? Currently running sequencer?

What happens if both da and rollup addresses are provided, but they don't match? Request fails or rollup address takes priority and da is ignored

citizen-stig (approved):

crates/full-node/sov-api-spec/openapi-v3.yaml
ross-weir: it's always defined in `constants` so we always just return it in the response

crates/full-node/sov-api-spec/openapi-v3.yaml
ross-weir: Correct, currently running sequencer

> What happens if both da and rollup addresses are provided, but they don't match? Request fails or rollup address takes priority and da is ignored

not sure tbh, these params existed for the old endpoint so I just kept them. I think they will almost never be used

ross-weir:
> I think those also might be related:
> 
> * [Improve rollup tx simulation endpoint interface sovereign-sdk-wip#1753](https://github.com/Sovereign-Labs/sovereign-sdk-wip/issues/1753)
> * Item in here: [[TRACKER] Nonce and Generation combo support #1518](https://github.com/Sovereign-Labs/sovereign-sdk/issues/1518) about auth support. I don't know if it is still relevant.

yep they are, thanks :+1:

PR #1837, opened 2025-10-09
Add tx priority level

preston-evans98:
# Description

This PR adds an optional get_transaction_priority method to the runtime trait. Under load, transactions with higher priority are always processed before lower priority txs, and priorities with equal priority are processed in FIFO order.

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/inner.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/inner.rs
@@ -1,4 +1,5 @@
 #![allow(dead_code)]
+use std::collections::BTreeMap;
 use std::num::NonZero;
 use std::ops::Deref;
 use std::sync::atomic::{AtomicU32, AtomicU64, AtomicUsize, Ordering};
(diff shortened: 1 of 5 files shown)


crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -1208,33 +1253,103 @@ where
...
+                    self.heap.insert(
+                        Priority {
+                            priority: msg.priority(&self.runtime),
+                            index,
+                        },
+                        msg,
+                    );
+                    index += 1;
+
+                    // If the heap is full after adding the new message, try to clear some space by dropping old messages
```
bkolad, line 1293: Why is it dropped instead of handling the highest-priority AcceptTx message to make room?

crates/module-system/sov-modules-api/src/runtime/mod.rs
```diff
@@ -95,6 +97,15 @@ pub trait Runtime<S: Spec>:
     fn get_transaction_delay_ms(&self, _call: &Self::Decodable) -> u64 {
         0
     }
+
+    /// Gets the priority level of a transaction. Higher priority transactions are processed first in the sequencer when possible.
+    /// Messages with equal priority are processed in the order they are received. If messages have a delay configured in [`Runtime::get_transaction_delay_ms`],
+    /// they are considered to be "received" after the delay period has elapsed.
+    // Returns a u32 so that the sequencer can represent priority as a u64 and have some reserved values that are greater than the maximum priority level of any transaction.
+    #[cfg(feature = "native")]
+    fn get_transaction_priority(&self, _call: &FullyBakedTx) -> u32 {
```
bkolad, line 106: Just a question:
Why not add a priority field to the transaction struct?
That way, we could support different priorities for the same call types, and the sender could choose a priority level when sending the transaction.
(Although, in that case, the caller would need to know the priority at the time of signing.)

crates/full-node/sov-sequencer/src/preferred/inner.rs
preston-evans98: When the heap is full, that means that the sequencer isn't able to process messages at the rate that it's receiving them so we need to shed some load

crates/module-system/sov-modules-api/src/runtime/mod.rs
preston-evans98: Discussed offline - we're holding off on this for now because we have no mechanism to keep users from just using max priority all the time. Since the method takes a `FullyBakedTx`, we can easily add support in a future tx version if needed

bkolad (approved):

PR #1839, opened 2025-10-10
fix: replace blocking sleep with cooperative sleep in soak testing

sashaodessa:
## Problem
The `prepare_and_send_txs` function in `sov-soak-testing-lib` was using `std::thread::sleep` within an async context, which blocks the Tokio runtime thread and reduces performance.

## Solution
Replaced the blocking `std::thread::sleep` with cooperative `tokio::time::sleep` to maintain async runtime efficiency.


preston-evans98 (commented):
Looks good to me. Thanks @sashaodessa

preston-evans98 (approved):

PR #1845, opened 2025-10-12
fix: cap sleep duration in set_value_and_sleep to prevent DoS

anim001k:
### Problem
The `set_value_and_sleep` method in `sov-value-setter` allows unrestricted user input for sleep duration, potentially causing:
- **DoS vulnerability**: attackers can block execution threads with excessive sleep times
- **Performance degradation**: blocking sleep in transaction execution path
- **Poor maintainability**: unclear dependency on user input for execution timing

### Solution
- Added 100ms upper bound for sleep duration using `std::cmp::min`
- Added explanatory comment about DoS risk mitigation
- Preserves original functionality while preventing abuse


anim001k:

preston-evans98:
Hi @anim001k, thanks for the PR! This module is used for testing only, and large sleep values are an intentional feature to let us easily simulate long-running transactions. Going to close this for now.

PR #1846, opened 2025-10-12
Logs with cursor refactor

LogvinovLeon:
# Description

Main logic is in: `crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs`
It has structured errors and well-structured logic.
Three levels. block, tx, log. Each level applies it's cursor if present and checks bloom filters.
Each level can return early and return the cursor if logs limit is reached.
Some other small refactors.

After refactoring this code it became easier for me to understand and find issues.
Therefore I realized that:
* Previous version didn't do proper cursor validation on block level. If cursor was present - range was overriden.
* Current EVM implementation has a lot of functions that can return `RpcResult<Option<Obj>>` that should error on invalid params but because `PendingOrBlock` has `Invalid` variant - that gets converted into None in `get_sealed_block_by_number` - we wrongfully return None where error should be returned.
* Previous version didn't do filtering on receipt/tx level bloom filters.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs
```diff
@@ -0,0 +1,330 @@
...
+            FilterBlockOption::Range {
+                from_block,
+                to_block,
+            } => self.by_range(from_block, to_block),
+        }
+    }
+
+    fn by_hash(mut self, block_hash: B256) -> Result<LogsWithMaybeCursor> {
+        if self.cursor.is_some() {
+            return Err(Error::CursorNotSupportedForBlockHash);
```
bkolad: Does it mean that we can't submit a cursor and a filter with a block hash? If so, then this would not be correct.

crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+}
+
+pub struct LogsService<S: Spec, Seq: Sequencer<Spec = S>> {
+    filter: Filter,
+    cursor: Option<Cursor>,
+    max_logs: usize,
+    state: ApiStateAccessor<S>,
+    evm: Evm<S>,
+    logs: Vec<Log>,
```
bkolad, line 76: What if we removed `state` and `logs` from the structure? That way, we could pass immutable `&self` everywhere, making many methods easier to unit test, since some of them don’t actually depend on the state.

crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs
```diff
@@ -0,0 +1,330 @@
...
+    fn scan_block_range(&mut self, range: RangeInclusive<u64>) -> Result<Option<Cursor>> {
+        let range = self.apply_block_level_cursor(range)?;
+        for block_number in range {
+            let block = self.get_block(block_number)?;
+            if !self.filter.matches_bloom(block.header.logs_bloom()) {
+                continue;
+            }
+            if let Some(cursor) = self.scan_block(block)? {
+                return Ok(Some(cursor));
+            }
```
bkolad, line 168: Just return `self.scan_block(block)`

crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs
```diff
@@ -0,0 +1,330 @@
...
+        receipt: Receipt,
+        header: &Sealed<Header>,
+    ) -> Result<Option<Cursor>> {
+        let logs = receipt.receipt.logs;
+        let log_range = (0 as u32)..(logs.len() as u32);
+        let logs_iter = logs.into_iter().enumerate();
+        let skipped_logs =
+            self.apply_log_level_cursor(log_range, tx_index_absolute, header.number())?;
+
+        // As logs iter is pre-enumerated - we keep correct indices
```
bkolad, line 251: Let’s add a test for this to prevent it from being missed during future refactoring.

bkolad:
Thx if this
> Previous version didn't do proper cursor validation on block level. If cursor was present - range was overriden.

was related to this comment
https://github.com/Sovereign-Labs/sovereign-sdk/pull/1846#discussion_r2425300424

Then it was not a bug, `eth_getLogsWithCursor` should support `by_hash` so we can get all the logs from a given block if we query by hash.

crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    fn scan_block(&mut self, block: SealedBlock) -> Result<Option<Cursor>> {
+        let mut tx_range_absolut = block.transactions.clone();
+        tx_range_absolut = self.apply_tx_level_cursor(tx_range_absolut, block.number())?;
+        for tx_idx_absolute in tx_range_absolut {
+            let receipt = self.get_receipt(tx_idx_absolute)?;
+            if !self.filter.matches_bloom(receipt.bloom()) {
+                continue;
+            }
+            if let Some(cursor) = self.scan_tx(tx_idx_absolute, receipt, &block.header)? {
```
bkolad, line 205: just return `self.scan_tx(..)`

crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs
LogvinovLeon: I initially forbid it for simplicity/defensiveness as there is no practical reason to using a cursor for querying logs from one block.
Our limit is 20_000. I highly doubt that any block will ever have that many logs.
But now that I think about it - it's actually simpler to support it so I removed it.

crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs
LogvinovLeon: This is within a loop. I only want to return early if the cursor was returned. If we're still within limits - I want to continue.

crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs
LogvinovLeon: Same as above

crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs
LogvinovLeon: I don't think it would be more clear or easier to unit-test.

Currently the methods that don't rely on state - use `&self`.
The methods for state fetching below - rely on mutable state.
The `scan_*` methods - rely on state and logs. If we change that - each method would gain two arguments and the return value will also be less clear.
It's also important to symbolize that you can't call those methods simultaneously as one relies on the result of the other. And `&mut` shows that.

LogvinovLeon: What I would like to do is to mark some fields `readonly`. `filter`, `cursor`, `max_logs`.
That would make my intent more clear. I was thinking `OnceCell` but it still allows obtaining mutable references. Maybe I'll just add comments. Open to proposals here

LogvinovLeon:
> Thx if this
> 
> > Previous version didn't do proper cursor validation on block level. If cursor was present - range was overriden.
> 
> was related to this comment [#1846 (comment)](https://github.com/Sovereign-Labs/sovereign-sdk/pull/1846#discussion_r2425300424)
> 
> Then it was not a bug, `eth_getLogsWithCursor` should support `by_hash` so we can get all the logs from a given block if we query by hash.

Replied in a comment. Changed

crates/full-node/sov-ethereum/src/handlers/get_logs/service.rs
LogvinovLeon: Actualy - I wrapped this data in Immutable wrapper. See last commit. Looks clean

bkolad: In that case, I prefer the original version since all the fields are private anyway. I thought the `apply_xx` methods could be unit tested, but if that’s not worth the tradeoff, let’s revert to the version without Immutable.

bkolad (approved):

PR #1851, opened 2025-10-13
Debug trace block by number

LogvinovLeon:
# Description

Implements `debug_traceBlockByNumber` with tests

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
@@ -1,6 +1,7 @@
 use alloy_consensus::Sealed;
 use alloy_consensus::{transaction::Recovered, Transaction as TransactionTrait, TxReceipt};
-use alloy_primitives::Address;
+use alloy_eips::BlockNumberOrTag;
+use alloy_primitives::{Address, BlockNumber};
 use alloy_primitives::{Bytes, TxKind, B256, U256};
 use alloy_rpc_types::{
     Block, BlockTransactions, Log, ReceiptEnvelope, ReceiptWithBloom, Transaction,
@@ -264,6 +265,24 @@ where
                 Err(_) => PendingOrBlock::Invalid(block_number_str),
             },
         }
+    }
+
+    /// Converts BlockNumberOrTag into number.
+    pub fn resolve_block_number(
+        &self,
+        block: BlockNumberOrTag,
+        state: &mut ApiStateAccessor<S>,
+    ) -> BlockNumber {
+        let block_numbers = self.block_numbers(state);
+        let block_number = match block {
+            BlockNumberOrTag::Earliest => *block_numbers.start(),
+            BlockNumberOrTag::Latest | BlockNumberOrTag::Finalized | BlockNumberOrTag::Safe => {
+                *block_numbers.end()
+            }
+            BlockNumberOrTag::Number(nr) => nr,
+            BlockNumberOrTag::Pending => *block_numbers.end() + 1,
+        };
+        block_number
     }
 
     /// Retrieves a sealed block by number.
--- a/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
@@ -274,13 +274,13 @@ where
         }
 
         // Trace the target transaction
-        Ok(self.trace_transaction_inner(
+        self.trace_transaction_inner(
             &block_env,
             replay_tx_env(&traced_tx),
             cfg_env,
             &mut evm_db,
             &opts,
-        )?)
+        )
     }
 
     fn trace_transaction_inner(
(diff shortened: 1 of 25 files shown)


crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -172,12 +175,120 @@ where
...
+        block: BlockNumberOrTag,
+        opts: GethDebugTracingOptions,
+        state: &mut ApiStateAccessor<S>,
+    ) -> Result<Vec<TraceResult>, EthApiError> {
+        let block_number = self.resolve_block_number(block, state);
+        // Get transaction and block data
+        let block = self
+            .blocks
+            .get(&block_number, state)?
+            .ok_or(EthApiError::PrunedHistoryUnavailable)?;
```
bkolad: It’s possible that someone provides a block reference far in the future using `BlockNumberOrTag::Number(nr)`. In this case, the block won’t exist yet because it hasn’t been created, so the message is wrong. Maybe, `resolve_block_number` should validate this scenario?

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -172,12 +175,120 @@ where
...
+        let mut archival_state = self.archival_state(block_number - 1, state)?;
```
bkolad: `resolve_block_number` can return 0. In that case, we will underflow here.

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
bkolad: Like it could check if `nr > *block_numbers.end() + 1` and return an error.

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -172,12 +175,120 @@ where
...
+        let traced_tx = self
+            .transaction(index, state)
+            .ok_or_else(|| EthApiError::PrunedHistoryUnavailable)?;
+
+        let block = self
+            .blocks
+            .get(&traced_tx.block_number, state)?
+            .expect("Transaction block not available");
+
+        let mut archival_state = self.archival_state(traced_tx.block_number - 1, state)?;
```
bkolad: We don’t include any transactions in the genesis block (block_number = 0), so the code is fine for now. Still, let’s be cautious and handle the case where traced_tx.block_number == 0.

examples/demo-rollup/tests/evm/evm_tracing.rs
```diff
@@ -0,0 +1,33 @@
...
+#[tokio::test(flavor = "multi_thread")]
+async fn debug_trace_block_by_number() -> anyhow::Result<()> {
+    let rollup = setup_test_rollup(0, EVM_EXTENSION).await;
+    rollup.wait_for_next_blocks(1).await;
+
+    let client = alloy_client(rollup.http_addr);
+    let usdc = Erc20::deploy(client.clone(), "Usdc".into(), "USDC".into()).await?;
+    let mint_tx = usdc.mint(Address::ZERO, parse_ether("1")?).send().await?;
+    rollup.wait_for_next_blocks(1).await; // Block nr 2 mined with 2 transactions
+    rollup.pause_preferred_batches().await;
```
bkolad: Why do we pause here? The rest of the test should still work even if we continue creating blocks, right?

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -172,12 +175,120 @@ where
...
+            .ok_or_else(|| EthApiError::PrunedHistoryUnavailable)?;
+
+        let traced_tx = self
+            .transaction(index, state)
+            .ok_or_else(|| EthApiError::PrunedHistoryUnavailable)?;
+
+        let block = self
+            .blocks
+            .get(&traced_tx.block_number, state)?
+            .expect("Transaction block not available");
```
bkolad: Let's do `ok_or_else(|| EthApiError::PrunedHistoryUnavailable)?;` instead of expect.

examples/demo-rollup/tests/evm/evm_tracing.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::evm::evm_test_helper::setup_test_rollup;
+use crate::evm::evm_test_helper::EVM_EXTENSION;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn debug_trace_block_by_number() -> anyhow::Result<()> {
+    let rollup = setup_test_rollup(0, EVM_EXTENSION).await;
+    rollup.wait_for_next_blocks(1).await;
+
+    let client = alloy_client(rollup.http_addr);
+    let usdc = Erc20::deploy(client.clone(), "Usdc".into(), "USDC".into()).await?;
```
bkolad, line 17: This is a nicer way of deploying contracts. However, in `tests/evm` we’re now using three different approaches: `deploy_contract`, `alloy_deploy_contract`, and this new one. Could you please remove one of the older methods before introducing a new way or reuse an existing one so the tests remain more consistent?
Can be done in another PR, but shouldn't wait too long.

LogvinovLeon: Removed alloy_deploy_contract

bkolad (approved):

PR #1866, opened 2025-10-13
Testing: Refactor `RollupBuilder` to work with any DA.

bkolad:
# Description

The `RollupBuilder` currently imposes a constraint on `da_service`, requiring it to be a `StorableMockDaService`. In most cases, this constraint is unnecessary.

This PR splits the `RollupBuilder` into two parts:

1. one that can work with any `DaService`
2. another that specifically uses `StorableMockDaService` functionality.

This change is necessary for the next PR, where `RollupBuilder` will be used with a different DA service implementation.

The second change removes the `StoragePath` generic from `RollupBuilder`. It has been replaced with an enum, simplifying the generic types throughout the code.

The RollupBuilder functionality is the same; it’s just a refactor.

### DIFF
--- a/crates/module-system/sov-test-utils/src/test_rollup.rs
+++ b/crates/module-system/sov-test-utils/src/test_rollup.rs
@@ -113,7 +113,6 @@ pub struct RollupBuilderConfig<S: Spec> {
 /// A one-stop shop for building entire rollups and starting them in the
 /// background to test node APIs.
 #[derive(Clone)]
-
 pub struct RollupBuilder<R: FullNodeBlueprint<Native>> {
     genesis: GenesisSource<R::Spec, R::Runtime>,
     da_config: <<R as FullNodeBlueprint<sov_modules_api::execution_mode::Native>>::DaService as DaService>::Config,
--- a/crates/module-system/sov-test-utils/src/test_rollup.rs
+++ b/crates/module-system/sov-test-utils/src/test_rollup.rs
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
 use crate::postgres::connection_string_from_postgres_container;
 use std::net::SocketAddr;
 use std::num::NonZero;
@@ -112,7 +113,7 @@ pub struct RollupBuilderConfig<S: Spec, StoragePath = Arc<tempfile::TempDir>> {
 #[derivative(Clone(bound = "StoragePath: Clone"))]
 pub struct RollupBuilder<R: FullNodeBlueprint<Native>, StoragePath = Arc<tempfile::TempDir>> {
     genesis: GenesisSource<R::Spec, R::Runtime>,
-    da_config: MockDaConfig,
+    da_config: <<R as FullNodeBlueprint<sov_modules_api::execution_mode::Native>>::DaService as DaService>::Config,
     config: RollupBuilderConfig<R::Spec, StoragePath>,
     postgres_container_opt: Option<Arc<ContainerAsync<PostgresImage>>>,
     with_secondary_sequencer: Option<MockAddress>,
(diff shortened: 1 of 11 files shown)


crates/module-system/sov-test-utils/src/test_rollup.rs
```diff
@@ -901,6 +870,34 @@ where
...
+where
+    R: FullNodeBlueprint<Native, DaService = StorableMockDaService> + Default + 'static,
+    R::Spec: Spec<Da = MockDaSpec>,
+{
+    /// Pauses batch production for the preferred sequencer.
+    ///
+    /// Transactions accepted by the preferred sequencer after this call (and
+    /// before [`TestRollup::resume_preferred_batches`]) will all be part of the
+    /// same batch.
+    pub async fn pause_preferred_batches(&self) {
```
citizen-stig, line 877: Why is this functionality is only for `StorableMockDaService`, and not totally generic?

crates/module-system/sov-test-utils/src/test_rollup.rs
```diff
@@ -68,10 +70,25 @@ pub enum GenesisSource<S: Spec, R: Runtime<S>> {
     CustomParams(GenesisParams<R::GenesisConfig>),
 }
 
+#[derive(Clone)]
+
```
citizen-stig: Extraspace

crates/module-system/sov-test-utils/src/test_rollup.rs
```diff
@@ -68,10 +70,25 @@ pub enum GenesisSource<S: Spec, R: Runtime<S>> {
...
+pub enum StoragePath {
+    Tmp(Arc<tempfile::TempDir>),
+    Buf(PathBuf),
+}
```
citizen-stig, line 77: Just a suggestion, but maybe adding those will improve ergonomics, so people won't have to import full type.

```rust
impl From<Arc<tempfile::Tempdir>> for StoragePath 


// Wrap into Arc inside.
impl From<tempfile::Tempdir> for StoragePath
```

Because it is also possible to misuse:

```rust

let dir = tempfile::tempdir().unwrapr():

// Which should I use?
// this:
let storage_path_a = StoragePath::Buf(dir.path());
// or this/
let storage_path_b = StoragePath::Tmp(Arc::new(dir));
```

crates/module-system/sov-test-utils/src/test_rollup.rs
```diff
@@ -108,37 +113,21 @@ pub struct RollupBuilderConfig<S: Spec, StoragePath = Arc<tempfile::TempDir>> {
 
 /// A one-stop shop for building entire rollups and starting them in the
 /// background to test node APIs.
-#[derive(Derivative)]
-#[derivative(Clone(bound = "StoragePath: Clone"))]
-pub struct RollupBuilder<R: FullNodeBlueprint<Native>, StoragePath = Arc<tempfile::TempDir>> {
+#[derive(Clone)]
+
```
citizen-stig: Extra space

citizen-stig (approved):

crates/module-system/sov-test-utils/src/test_rollup.rs
bkolad: fixed

crates/module-system/sov-test-utils/src/test_rollup.rs
bkolad: fixed

crates/module-system/sov-test-utils/src/test_rollup.rs
bkolad: fixed

crates/module-system/sov-test-utils/src/test_rollup.rs
bkolad: In the example above, both approaches will work the same way.
I would prefer to leave it as it is. With the additional `From`,  we’ll have three ways to accomplish the same thing.

citizen-stig: > In the example above, both approaches will work the same way.

But which one users should choose? That will just lead to more errors in writing tests, especially by external people, when they use option a and then rollup restart.

PR #1871, opened 2025-10-14
feat(sov-db): parameterize ForkDescription Arbitrary with ForkGenParams

VolodymyrBg:
- Introduce ForkGenParams to configure proptest ranges and recursion limits.
- Replace hardcoded ranges with parameterized values; provide Default matching previous behavior.
- Preserve invariants (min length >= 1); use saturating_sub for child start bounds and add debug asserts.

### DIFF
--- a/crates/full-node/sov-db/src/storage_manager/tests/arbitrary.rs
+++ b/crates/full-node/sov-db/src/storage_manager/tests/arbitrary.rs
@@ -21,33 +21,39 @@ pub struct ForkDescription {
 }
 
 impl Arbitrary for ForkDescription {
-    type Parameters = ();
+    type Parameters = ForkGenParams;
 
-    fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
-        let leaf = (1u64..=10u64, 5u8..=30).prop_map(|(start_height, length)| ForkDescription {
-            start_height,
-            length,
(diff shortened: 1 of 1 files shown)


crates/full-node/sov-db/src/storage_manager/tests/arbitrary.rs
```diff
@@ -21,33 +21,39 @@ pub struct ForkDescription {
...
-    fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
-        let leaf = (1u64..=10u64, 5u8..=30).prop_map(|(start_height, length)| ForkDescription {
-            start_height,
-            length,
-            child_forks: Vec::new(),
-        });
+    fn arbitrary_with(params: Self::Parameters) -> Self::Strategy {
+        // lengths must be >= 1 to satisfy invariants in ForkMap materialization
+        debug_assert!(*params.leaf_length.start() >= 1);
+        debug_assert!(params.inner_length.start >= 1);
```
preston-evans98: In case we run any tests in release mode (we do!), let's convert these to `assert!` instead of `debug_assert`

preston-evans98 (commented):
Looks good to me aside from the one nit! 

 @citizen-stig, would you take a look as well?

citizen-stig (approved):

PR #1872, opened 2025-10-14
Add multisig support

preston-evans98:
# Description

This PR adds multisig support by adding a new transaction type `V1` which contains...
- A vector of signatures
- A vector of unused public keys
- A signing threshold

We compute the credential ID for such a multisig as `hash(num_required_signers || sorted(pub_keys))`. This construction allows us to compute the credential ID and multisig membership for a given transaction completely statelessly. 

### Alternate Design: Store Multisig Info in the Accounts Module
An alternative approach was considered in which V1 transactions instead consist of...
- A vector of signatures
- A credential ID
... and then the signature was checked by first retrieving the multisig parameters (allowed signers, threshold) from state and then verifying the signatures against the retrieved parameters. This approach was discaded since it's both more expensive at runtime (we have to do a state access) and less convenient. 

The tradeoff in the current design is that transactions are slightly larger, since unused public keys have to be included. However, the additional cost is not very large, so I decided that the additional calldata cost was worthwhile. 

**Cost Analysis**
In a typical 2/3 multisig, the cost of adding the additional unused public key is exactly one byte greater than the cost of including the credential ID. (A credential ID would be 32 bytes, while we need a 32-byte public key and a 1-byte threshold).

 For a 3/5 multisig, we pay an extra 33 bytes. (A credential ID would take 32 bytes, while we pay 32 bytes each for two unused pub keys and 1 byte for a threshold). Note that the overhead of a 3/5 multisig is already 288 bytes (3 * 64-byte signatures and 3 * 32-byte pubkeys), so the extra 33 bytes is only a 12% increase in signature size.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/sov-modules-api/src/transaction/mod.rs
```diff
@@ -271,6 +419,24 @@ impl<R: TransactionCallable, S: Spec> Transaction<R, S> {
                     .verify(&inner.pub_key, msg, meter)
                     .map_err(TransactionVerificationError::from)?;
             }
+            VersionedTx::V1(inner) => {
+                for signature in inner.signatures.iter() {
+                    // Charge gas for the signature verification with out verifying
+                    MeteredSignature::new::<S>(signature.signature.clone())
```
bkolad, line 430: We charge for all signature verifications, even if the first one fails. This simplifies reasoning about the overall cost, but let's add a comment to make it clear that this behavior is intentional.

crates/module-system/sov-modules-api/src/runtime/capabilities/authentication.rs
```diff
@@ -315,6 +315,36 @@ pub fn extract_authorization_data<S: Spec, D: DispatchCall<Spec = S>>(
...
+        .collect::<Vec<_>>();
+    let multisg = Multisig::new(tx_v1.min_signers, pub_keys);
+    let credential_id = multisg.credential_id::<<S::CryptoSpec as CryptoSpec>::Hasher>();
+
+    Ok(AuthorizationData {
+        uniqueness: tx_v1.uniqueness,
+        tx_hash: raw_tx_hash,
+        credential_id,
+        credentials: Credentials::new(multisg),
+        default_address: credential_id.into(),
```
bkolad, line 344: What about adding `min_signers` to `AuthorizationData`, and later to the `Context`? This way, whichever module processes the transaction can check the minimal threshold of valid signatures.

crates/module-system/sov-modules-api/src/transaction/mod.rs
```diff
@@ -66,6 +71,127 @@ pub struct Version0<Call, S: Spec> {
...
+    pub signatures: SafeVec<PubKeyAndSignature<S>, MAX_SIGNERS>,
+    /// The credential IDs that are part of the multisig but not used to sign the transaction.
+    // This is used to compute the credential ID statelessly
+    pub unused_pub_keys: SafeVec<<S::CryptoSpec as CryptoSpec>::PublicKey, MAX_SIGNERS>,
+    /// The minimum number of signers required to sign the transaction. In a 3/5 multisig, this would be 3.
+    /// Note that...
+    ///  - The transaction will be valid if at least `min_signers` signers have signed the transaction, but more signers are allowed. (This is useful for record keeping in case of, say, a unanimous decision)
+    ///  - If an invalid signature is detected, the transaction will be invalid. This allows us to use batch verification.
+    // This is used to compute the credential ID statelessly.
+    pub min_signers: u8,
```
bkolad, line 129: Related to this comment: https://github.com/Sovereign-Labs/sovereign-sdk/pull/1872/files#r2431366049
The transaction sender provides this field, but I haven’t found a way for a module to impose constraints on `min_signers`.

crates/module-system/sov-modules-api/src/transaction/mod.rs
```diff
@@ -66,6 +71,127 @@ pub struct Version0<Call, S: Spec> {
...
+    /// The runtime call of the transaction.
+    #[sov_wallet(
+        bound = "Call: sov_rollup_interface::sov_universal_wallet::schema::UniversalWallet"
+    )]
+    pub runtime_call: Call,
+    /// Uniqueness identifier of this transaction. see [`UniquenessData`] for more details.
+    pub uniqueness: UniquenessData,
+    /// The transaction metadata. Contains gas parameters and the chain ID.
+    pub details: TxDetails<S>,
```
bkolad, line 138: If all supported transaction versions share the same `details, uniqueness, and runtime_call` fields, we can move these fields to the `Transaction` struct and reuse them across `V0` and `V1`.

ross-weir: what if V2 doesn't need these fields?

I'm not sure about having common fields on `Transaction` because with new versions what is considered a common field could change

bkolad: Yes, we should extract only the fields that are mandatory for processing the transaction. For example, I think `uniqueness` is one such field. We’ll always need to verify whether a transaction is unique, and if new ways are introduced, we will extend the `UniquenessData` enum accordingly. So this seems like a common field for all tx types.

bkolad (approved):

crates/module-system/sov-modules-api/src/transaction/mod.rs
preston-evans98: Done

crates/module-system/sov-modules-api/src/runtime/capabilities/authentication.rs
preston-evans98: Chatted offline; this is available in the `Credential`

crates/module-system/sov-modules-api/src/transaction/mod.rs
preston-evans98: Synced at standup. We're going to leave this as-is because we don't know that these fields will be shared across all future tx types

crates/module-system/sov-modules-api/src/transaction/mod.rs
```diff
@@ -70,6 +75,127 @@ pub struct Version0<Call, S: Spec> {
...
+    borsh::BorshDeserialize,
+    serde::Serialize,
+    serde::Deserialize,
+    borsh::BorshSerialize,
+    UniversalWallet,
+    PartialEq,
+    Eq,
+)]
+/// A signature and public key pair.
+pub struct PubKeyAndSignature<S: Spec> {
```
deuszx, line 91: You not always need the public key as it can be recovered from the sig and the message. Will save you some precious bytes

preston-evans98: Unfortunately the `Spec` trait doesn't currently allows us to express the guarantee that the inner type will use secp256 signatures. Since Ed25519 doesn't support the ecrecover operation, we have provide the public key here for now - although we may be able to find a nice way to express the guarantee later on.

 If you have any ideas, I'd welcome a PR! Currently all of mine rely on specialization....

PR #1873, opened 2025-10-14
BlobSender: await finished JoinHandles to surface errors and prevent silent failures

vtjl10:
### Description
- **Summary**: Replace passive cleanup with explicit handling of finished tasks. When a blob task finishes, we remove it from the map, await its JoinHandle, and log any error.
- **Problem**: Previously, JoinHandle results were discarded; task errors were silently lost and concurrency counters risked sticking on failure.
- **Change**: In `crates/full-node/sov-blob-sender/src/lib.rs`, replace `retain`-based cleanup with:
  - Collect finished IDs
  - Remove them under lock
  - Await their handles
  - Log errors if any

### DIFF
--- a/crates/full-node/sov-blob-sender/src/lib.rs
+++ b/crates/full-node/sov-blob-sender/src/lib.rs
@@ -333,8 +333,24 @@ where
             },
         );
 
-        // TODO: handle errors from the spawned tasks.
-        blobs.retain(|_, b| !b.handle.is_finished());
+        // Handle finished tasks: await their JoinHandle to surface any errors, then remove.
+        let finished_ids: Vec<_> = blobs
+            .iter()
+            .filter(|(_, b)| b.handle.is_finished())
+            .map(|(id, _)| *id)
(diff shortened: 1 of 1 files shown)


vtjl10:

crates/full-node/sov-blob-sender/src/lib.rs
```diff
@@ -333,8 +333,24 @@ where
             },
         );
 
-        // TODO: handle errors from the spawned tasks.
-        blobs.retain(|_, b| !b.handle.is_finished());
+        // Handle finished tasks: await their JoinHandle to surface any errors, then remove.
+        let finished_ids: Vec<_> = blobs
```
bkolad: Thanks for the PR! We're already logging all issues within the blob sender tasks, so we don’t need additional logs here. It’s fine to just remove the TODO.

preston-evans98 (approved):

vtjl10:
@preston-evans98 merge?

PR #1877, opened 2025-10-15
Add to/from bytes traits to crypto spec & use hex in transaction structure

ross-weir:
# Description

- Adds to/from bytes trait bounds to `CryptoSpec::Signature` & `CryptoSpec::PublicKey`
- This enables us to change the `serde` representation of the signature and public key fields on `Transaction` to always be a hex string as opposed to an arbitrary layout depending on how the `Signature` / `PublicKey` serialization was implementated
- This simplifies a lot of code in the typescript & new rust sdk as we no longer have to account for the fact these fields could be any shape, they are now known.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/adapters/mock-zkvm/src/crypto.rs
```diff
@@ -167,16 +196,24 @@ pub struct Ed25519Signature {
     #[schemars(with = "&[u8]", length(equal = "ed25519_dalek::Signature::BYTE_SIZE"))]
     #[sov_wallet(as_ty = "[u8; ed25519_dalek::Signature::BYTE_SIZE]")]
     pub msg_sig: DalekSignature,
+    pub(crate) bytes: Vec<u8>,
```
ross-weir, line 199: For some of these impls we need to cache the bytes as the underlying type doesn't expose the bytes by ref

crates/module-system/sov-modules-api/src/transaction/mod.rs
```diff
@@ -52,8 +52,12 @@ impl<D: DispatchCall> TransactionCallable for D {
 /// V0 transaction.
 pub struct Version0<Call, S: Spec> {
     /// The signature of the transaction.
+    #[serde(with = "hex_field_format")]
+    #[sov_wallet(as_ty = "[u8; 64]", display = "hex")]
```
ross-weir, line 56: These need to be fixed size to match native `borsh` serialization

It's tricky to get the fixed size of the underlying type because it's generic which doesn't work with `as_ty`.

Discussed with Preston and we're OK with using fixed sizes of 64/32 here for now and solving it later if we need different sizes

crates/module-system/hyperlane/tests/integration/with_agent/helpers/mod.rs
```diff
@@ -197,7 +197,7 @@ impl HyperlaneBuilder {
 
         // Current image is based on https://github.com/Sovereign-Labs/hyperlane-monorepo/tree/integration-2025-09-17-rebase branch
         let docker_image = docker_image
-            .unwrap_or_else(|_| "ghcr.io/sovereign-labs/hyperlane-agent:integration-5".into());
+            .unwrap_or_else(|_| "ghcr.io/ross-weir/hyperlane-agent:integration-6".into());
```
ross-weir, line 200: will change back in follow up, I currently don't have perms to push to sov labs ghcr

theodorebugnet (approved):

PR #1887, opened 2025-10-15
Add a CryptoSpec override generic to `Transaction`, enabling Eip712Authenticator to be used for all rollups. Improve eip712 ergonomics.

theodorebugnet:
# Description

EIP712 transaction authentication *requires* secp256k1 signatures, so a rollup that's not already using a secp256k1-based cryptospec cannot use the authenticator.
This PR adds an optional third generic to `Transaction` for the cryptospec, which defaults to `<S as Spec>::CryptoSpec`, i.e. the normal usage. The EIP712 authenticator then overrides it by requiring it to implement a marker trait identifying it as secp256k1-compatible. The `EvmCryptoSpec` implements the marker trait using `Self`, while other cryptospecs can use a helper wrapper type that exposes the EvmCryptoSpec as the secp256k1 alternative (or implement the marker trait themselves with their own types if desired).

The PR also adds a helper for building a stub EVM RPC which is necessary for metamask to be able to connect to non-EVM chains, validate the chain_id and therefore allow signing EIP712 messages. This is boilerplate and adding it here allows rollups to import it rather than re-write it every time.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/sov-modules-api/src/runtime/capabilities/authentication.rs
```diff
@@ -427,10 +429,19 @@ pub fn authenticate<
 /// Decode bytes as a Sovereign SDK transaction, returning the message and tx info.
 #[cfg(feature = "native")]
 pub fn decode_sov_tx<S: Spec, D: DispatchCall<Spec = S>>(
```
theodorebugnet, line 431: Rust doesn't support defaults for generics in functions so adding a wrapper is a workaround to avoid the 3rd generic in the normal default case

bkolad (approved):

PR #1891, opened 2025-10-16
refactor(ledger-db): remove unused tx/event counters from SlotCommit

MozirDmitriy:
Removed the num_txs and num_events counters from SlotCommit, along with their increments, because they were never read anywhere and only introduced maintenance overhead and risk of drifting from the actual data; materialization already derives counts and ranges directly from receipts, so eliminating these dead fields simplifies the code without changing behavior.


preston-evans98 (approved):
Looks good to me, assuming CI doesn't find any places that these were actually needed. Thanks @MozirDmitriy

PR #1899, opened 2025-10-16
fix(stf-runner): avoid duplicate CORS headers by applying CORS only in RPC router

kilavvy:
Description
- Summary: Remove top-level CORS layer in start_http_server to prevent duplicate Access-Control-* headers.
- Rationale: CORS was applied both at the top router and inside rpc_module_to_router for /rpc, causing header duplication and potential preflight inconsistencies.
- Change: CORS is now applied only within rpc_module_to_router.

### DIFF
--- a/crates/full-node/sov-stf-runner/src/http.rs
+++ b/crates/full-node/sov-stf-runner/src/http.rs
@@ -32,9 +32,8 @@ pub(crate) async fn start_http_server(
     let handle = tokio::spawn(async move {
         tracing::info!(%rest_address, "Starting HTTP server");
         let mut router = router.layer(axum::middleware::from_fn(measure_time));
-        if let CorsConfiguration::Permissive = cors_configuration {
-            router = router.layer(CorsLayer::permissive());
-        }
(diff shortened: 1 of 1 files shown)


kilavvy:

preston-evans98 (commented):
Looks good to me. CC @citizen-stig, I seem to recall you working on this - do you happen to know if there are any edge cases we need to be watching out for?

crates/full-node/sov-stf-runner/src/http.rs
```diff
@@ -32,9 +32,8 @@ pub(crate) async fn start_http_server(
     let handle = tokio::spawn(async move {
         tracing::info!(%rest_address, "Starting HTTP server");
         let mut router = router.layer(axum::middleware::from_fn(measure_time));
```
preston-evans98, line 34: Looks like this `mut` is now unused which is causing CI to fail

citizen-stig:
@preston-evans98 , I think it needs to be check manually. This setting is basically for devex. But I don't remember exact details, I need to manually double check this PR

PR #1915, opened 2025-10-20
`Replica:` Start processing soft confirmations only at `BatchStart` boundaries.

bkolad:
# Description

1. Refactored the `replica sync task` for better maintainability.
2. Added an additional condition to the `replica sync task`: after a replica restart, it will not apply any events from Postgres until it receives the first `BatchStart` event. This ensures that the replica does not start processing soft confirmations from the middle of a batch.

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
@@ -176,7 +176,7 @@ impl EventReceiver {
             // and use its event_id as the target.
             Some(start_id) => {
                 let notify = self.recv_notifications().await?;
-                assert!(notify.event_id > start_id);
+                assert!(notify.event_id >= start_id);
                 (start_id, notify.event_id)
             }
             // Otherwise, `None` indicates the replica has just started.
--- a/crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
@@ -100,9 +100,9 @@ impl EventReceiver {
                 }
 
                 match self.fetch_data(start_event_id, prev_event_type).await {
-                    Ok((next_event_id, event_type)) => {
+                    Ok((event_id, event_type)) => {
                         nb_of_consecutive_db_errors = 0;
-                        start_event_id = Some(next_event_id + 1);
+                        start_event_id = Some(event_id + 1);
                         prev_event_type = event_type;
                     }
                     Err(err) => {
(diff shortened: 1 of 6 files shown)


crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
```diff
@@ -234,40 +167,45 @@ impl EventReceiver {
...
+            Some(start_id) => {
+                let notify = self.recv_notifications().await?;
+                (start_id, notify.event_id)
+            }
+            // Otherwise, `None` indicates the replica has just started.
+            // Keep listening for events until a `BatchStart` notification is received,
+            // then use that event's ID as both the starting and target event ID.
+            None => loop {
+                let notify = self.recv_notifications().await?;
+                if matches!(notify.event_type, EventType::BatchStart) {
```
bkolad, line 187: After a restart, the replica sync task waits for EventType::BatchStart before continuing.

crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
```diff
@@ -1,178 +1,109 @@
...
-                    &mut listener,
-                    &mut db_data_sender,
-                )
-                .await
-                {
-                    Ok(next_event_id) => {
+                match self.fetch_data(start_event_id, prev_event_type).await {
+                    Ok((next_event_id, event_type)) => {
                         nb_of_consecutive_db_errors = 0;
                         start_event_id = Some(next_event_id + 1);
```
theodorebugnet: If we're doing `start_event_id = next_event_id + 1`, should next_event_id be called `current_event_id` (or just event_id) instead? `next + 1` makes it sound like we'll be skipping one

crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
```diff
@@ -1,178 +1,109 @@
...
-                unreachable!("EventReceiver: impossible happened rollup didn't exit");
-            }
-        };
-
-        let mut start_event_id = match latest_event_id(&query_pool).await {
-            Ok(latest_event_id) => latest_event_id,
-            Err(e) => {
-                error!("Failed to get latest event id: {e:?}. Replica shutting down.");
-                exit_rollup(&self.shutdown_sender).await;
+                exit_rollup(&shutdown_sender).await;
```
theodorebugnet, line 54: As per discussion on standup, can we return an error here from the task and have the process exit correctly? Non-blocking

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
bkolad: fixed

crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
bkolad: I would do it in separate PR for all the relevant places.

PR #1916, opened 2025-10-20
Demo rollup: use nomt storage by default

citizen-stig:
# Description

Use nomt storage by default as it is a main storage for a long time. Also ensures that ZK roundtrip works in CI with NomtVerifier storage

## Testing
CI Passes

## Docs
No update


examples/demo-rollup/provers/sp1/src/lib.rs
```diff
@@ -27,4 +27,12 @@ lazy_static! {
         "{}/guest-celestia/elf/riscv32im-succinct-zkvm-elf",
         env!("CARGO_MANIFEST_DIR")
     ));
+    pub static ref SP1_GUEST_MOCK_NOMT_ELF: &'static [u8] = load_elf(&format!(
+        "{}/guest-mock-nomt/elf/riscv32im-succinct-zkvm-elf",
+        env!("CARGO_MANIFEST_DIR")
+    ));
+    pub static ref SP1_GUEST_CELESTIA_NOMT_ELF: &'static [u8] = load_elf(&format!(
```
bkolad, line 29: Devin has a valid point here:
```All SP1 ELF lazy_statics share a single OnceLock, so only the first-loaded ELF is ever returned
The load_elf function at lib.rs:7-17 contains a static ELF: OnceLock<Vec<u8>> inside the function body. Because this is a single static item (not one per call-site), OnceLock::get_or_init will initialize it on the very first call and then return that same value for every subsequent call, regardless of the path argument.
```

With any given rollup we will load only one elf but just to be super safe let's maybe use differenr `OnceLock` for every elf.

bkolad (approved):

examples/demo-rollup/provers/sp1/src/lib.rs
citizen-stig: Good point!

examples/demo-rollup/configs/mock_rollup_config.toml
```diff
@@ -29,8 +29,8 @@ block_time_ms = 1_000
 # The path to the rollup's data directory. Paths that do not begin with `/` are interpreted as relative paths.
 path = "demo_data"
 state_cache_size = 4294967296 # 4 GiB. Default is 1 GiB.
-user_commit_concurrency = 4 # Number of concurrent commit workers for user state
-user_hashtable_buckets = 200_000 # Expected state size, cannot be changed after creation
+# user_commit_concurrency = 4 # Number of concurrent commit workers for user state
```
citizen-stig: Where this is coming from...

PR #1924, opened 2025-10-20
Source namespace from constants.toml

citizen-stig:
# Description

Simplifies devex for rollup operators.

Now batch and proof namespace can be specified like that in constants.toml

It can be defined as:

* normal string and its byte representation is going to be used as namespaces. String should be 10 bytes long, meaning it can be 10 ASCII characters, or less UTF-8 characters.
* hex encoded string, and target type is byte array of length 10, that is expected by celestia. This can be used to define any byte sequence

https://mocha.celenium.io/namespace/000000000000000000000000000000000000736f762d6d696e692d6d?tab=Blobs

```toml
# The namespace used by the rollup to store its data.
# Using byte_string to convert "sov-test-b" to ASCII bytes
BATCH_NAMESPACE = { byte_string = "sov-test-b", type = "[u8; 10]" }
# The namespace used by the rollup to store aggregated ZK proofs.
# This is "sov-test-p" encoded as ASCII bytes.
# "hex" format is used as an example, that it is possible to define namespace as any sequence of bytes.
# `python3 -c "import sys; print('0x' + sys.argv[1].encode().hex())" "sov-test-p"`
PROOF_NAMESPACE = { hex = "0x736f762d746573742d70", type = "[u8; 10]" }
```

----

## Testing
Existing tests are passing


cemozerr:
Don't want to block this unnecessarily, but can we easily make this a readable text, rather than a hexstring?

preston-evans98 (approved):

PR #1931, opened 2025-10-21
Specify target account in Transaction

preston-evans98:
# Description

This transaction changes the behavior of our account-abstraction. Previously, we kept a map from `CredentialId` (i.e. pubkey) to address. For each transaction, we looked for a registered address to use given the pubkey and, if none was found, generated a "default address" from the pubkey and registered it. This meant that each key could only associated with a single account - wich opened up a griefing vector where an adversary took a known pubkey from another chain and registered it with a dummy account - effectively bricking the "default" address associated with that pubkey. 

In this PR, we take a different approach. Now, we store a Vec<AllowedCredential> in each account, and transactions specify which account they want to invoke their transaction from. If the specified address is not the default, we check the state map to ensure that the action is allowed before proceeding. This means that...
- The most common path (a pubkey controlling its default address) no longer needs any state access to resolve the account
- credentials can be re-used across accounts if desired
- There is no longer a griefing vector - registering a pubkey on a random account makes no difference to that pubkeys behavior on any other accounts.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


citizen-stig:
Corresponding web3-js PR: https://github.com/Sovereign-Labs/sovereign-sdk-web3-js/pull/218

PR #1939, opened 2025-10-22
Add web3 crate

ross-weir:
# Description

Introduces a `RuntimeDiscriminate` trait that call messages can derive to allow them to be used with the `UnmanagedRuntimeCall` type. The purpose of the `UnmanagedRuntimeCall` structure is to decouple serialization of call/runtime messages from the runtime itself. This is particularly useful for zeta who want to keep their module business logic closed source and thus the actual Runtime call type won't be available. An example implementation of this trait can be seen for the bank module, basically it should just use the value defined in `constants.toml` - if a different value is set then serialization errors could occur, unfortunately there doesn't seem to be a good way to auto-impl the `RuntimeDiscriminate` trait at the moment so this could lead to errors. Currently this will mostly be used by zeta to keep their exchange module closed source so we can ensure its implemented correctly

A `web3` rust crate has also been added under `crates/web3`. I ended up using native rust types from `sov-modules-api` so at the moment the crate is quite simple and exposes a helper transaction builder/serialization.
An example of it's usage can be seen: https://github.com/zetamarkets/sov-rollup/pull/534/files#diff-2c70cd6bee61e063d7c17577bacbdcd7bb3559dd95abe79c16984413b7fd3783
Devs would liking create a wrapper crate such as `bullet_trading_sdk` in my linked example that would export public interface such as call messages, rollup spec, chain hash, etc & bind the generics in `TransactionBuilder` for their types for improved devex for their users.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


theodorebugnet (approved):
LGTM. Would be nice to generate the discriminant using a macro and it seems almost trivial, except that the name of the `CallMessage` type is not hardcoded, so makes sense to leave it to future work.

What do you think about putting a note in the changelog? It's not a breaking change but it might help zeta keep track of when it gets in

ross-weir:
> LGTM. Would be nice to generate the discriminant using a macro and it seems almost trivial, except that the name of the `CallMessage` type is not hardcoded, so makes sense to leave it to future work.

What did you have in mind?

ross-weir:
Will merge for now so we can keep this work moving but happy to discuss the macro more

PR #1942, opened 2025-10-22
fix: allow undefined optional fields in JSON to Borsh serialization

### LINKED ISSUE #1738: wallet: optional fields cannot be undefined and require explicit `null`
For example, call message like this one, where `relayer` field is optional:

https://github.com/Sovereign-Labs/sovereign-sdk/blob/59530e3d797211a4b40473819b7beadf032f3ea2/crates/module-system/hyperlane/src/warp/mod.rs#L141-L151

Following schema will be generated.

```typescript
export interface TransferRemote {
    /**
     * The amount to transfer.
     */
    amount: number;
    /**
     * The domain of the destination chain.
     */
    destination_domain: number;
    /**
     * A limit for the payment to relayer to cover gas needed for message delivery.
     */
    gas_payment_limit: number;
    /**
     * The recipient on the destination chain.
     */
    recipient: string;
    /**
     * Selected relayer
     */
    relayer?: null | string;
    /**
     * The route to use for the transfer.
     */
    warp_route: string;
    [property: string]: any;
}
```

But if field `relayer` won't be defined in call message, following error will be thrown by the typescript SDK:

```
failed to call rollup: SerializationError: Expected type or field __SovVirtualWallet_CallMessage_TransferRemote.relayer, but it was not present
```

It would be better for DevX to allow undefined fields too.

erayack:
## Description

This PR fixes issue #1738 by allowing optional fields to be undefined in JSON input when serializing to Borsh format.

## Problem

Previously, when using the TypeScript SDK, optional fields in call messages (like `relayer` in `TransferRemote`) required explicit `null` values even when they were optional. Omitting these fields would result in the error:

```
failed to call rollup: SerializationError: Expected type or field __SovVirtualWallet_CallMessage_TransferRemote.relayer, but it was not present
```

## Solution

Modified the `visit_struct` method in `json_to_borsh.rs` to:

1. Check if a missing field is an `Option` type
2. If it is optional, create a `Value::Null` instead of throwing an error
3. This allows the `visit_option` method to properly encode the `None` variant

## Changes

- Updated `visit_struct` to handle missing optional fields gracefully
- Added explanatory comment about optional field handling
- Improved error handling for truly missing required fields

## Testing

The fix ensures that TypeScript interfaces with optional fields (marked with `?`) can omit those fields in JSON input, improving the developer experience.

## Related Issue

Fixes #1738

### DIFF
--- a/crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
+++ b/crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
@@ -2,8 +2,9 @@
 //! thus test sequencer + node interactions.
 
 use crate::utils::{
-    generate_paymaster_tx, generate_txs, new_test_rollup, pause_update_state,
-    tempdir_inside_codebase_dir, tx_set_value_with_gas, ModuleWithVersionedStateAccessInSlotHook,
+    generate_paymaster_tx, generate_txs, new_test_rollup, new_test_rollup_with_backend,
+    pause_update_state, tempdir_inside_codebase_dir, tx_set_value_with_gas,
+    ModuleWithVersionedStateAccessInSlotHook, TestSequencerDbBackend,
     MAX_BATCH_EXECUTION_TIME_MILLIS,
 };
 use backon::Retryable;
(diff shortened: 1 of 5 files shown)


erayack:

preston-evans98 (approved):
Great change, thanks @erayack! I think this PR is racing with #1931 which makes a nearly identical change - but it looks like yours should merge first. 

Do you have bandwidth to add a test to confirm the new behavior? (Can be done as a follow-up). If not, @theodorebugnet can add one later in the week

erayack:
Hi @preston-evans98. I will take a look in an hour to add a test.

erayack:
Can you trigger the workflow? @preston-evans98 @theodorebugnet

preston-evans98:
The new test looks like it might be for the wrong code. I think all we need to do is call `json_to_borsh` on a hard-coded transaction json string that's missing a `null` field. Shouldn't need any test_rollup changes.

erayack:
Thanks, @preston-evans98. I'm a bit confused while addressing the workflow failure.

PR #1950, opened 2025-10-22
Allow separating archival state into its own Rocksdb instance

preston-evans98:
# Description

This PR adds a new config option `storage.separate_archival_state`. If set to true, archival state will go into its own db at `{storage_path}/archival-state-db`. This option is false by default for backwards compatibility. When not set, the old database layout is used.


crates/full-node/sov-db/src/config.rs
```diff
@@ -54,6 +54,10 @@ pub struct RollupDbConfig {
     pub pruner_versions_to_keep: Option<usize>,
     /// Maximum number of keys to prune in a single batch.
     pub pruner_max_batch_size: Option<usize>,
+
+    /// Whether to use a separate db for archival state.
+    #[serde(default)]
+    pub separate_archival_state: bool,
```
citizen-stig, line 60: Why cannot we do a breaking change and always use separate archival state?

crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -20,26 +22,71 @@ pub struct FlatStateDb {
...
+    #[allow(dead_code)]
+    // We don't technically need to store the archival db here - it's only accessed through the user/kernel versioned DB wrappers.
+    // We keep it here so that the internal structure is more legible by glancing at this struct. Note that unless the user has set the config to separate out the archival db,
+    // this will point to the same rockbound::DB as the `other` field.
+    pub(crate) archival: Arc<rockbound::DB>,
+}
+
+pub struct FlatDbCommitData {
+    pub user_archival: SchemaBatch,
+    pub kernel_archival: SchemaBatch,
```
citizen-stig: Stupid question, but why live data is in a single schema batch, both kernel and user, but archival is split into 2?

crates/full-node/sov-db/src/config.rs
preston-evans98: We could, but breaking DB changes are really painful cuz they'll brick any existing rollups.

crates/full-node/sov-db/src/flat_db.rs
preston-evans98: Not a dumb question at all. This was an artifact of the old API setup. Tweaked rockbound to let us unify them.

crates/full-node/sov-db/src/config.rs
citizen-stig: Totally understand problems of breaking compatibility!

But keeping it backwards compatible not free and adds some complexity to already complicated enough part of the SDK.

And we have PRs in progress that will break compatibility too:

* https://github.com/Sovereign-Labs/sovereign-sdk/pull/1951
* https://github.com/Sovereign-Labs/sovereign-sdk/pull/1931

crates/full-node/sov-db/src/flat_db.rs
citizen-stig: So would it be possible to use unified schema batch in this PR?

preston-evans98: Yeah, just haven't pushed the commit yet

examples/demo-rollup/tests/restart/mod.rs
```diff
@@ -133,6 +164,203 @@ async fn start_stop_empty(
...
+    // Check that the balance is what we expect - i.e. no writes have been lost due to the crash on commit.
+    assert!(
+        response.to_string().starts_with("5"),
+        "Balance after set-value: {response}",
+    );
+    Ok(())
+}
+
+fn tx_send_transfer(value_to_set: u128, key: &Ed25519PrivateKey, nonce: u64) -> RawTx {
+    // let msg = <TestRuntime<TestSpec> as DispatchCall>::Decodable::ValueSetter(
```
citizen-stig: Leftover probably

crates/full-node/sov-db/src/config.rs
```diff
@@ -88,6 +92,7 @@ impl RollupDbConfig {
             pruner_block_interval: Some(100),
             pruner_versions_to_keep: Some(20),
             pruner_max_batch_size: None,
+            separate_archival_state: false,
```
citizen-stig, line 95: Let's enable that on demo-rollup, so it it covered by README tests too

citizen-stig (approved):

examples/demo-rollup/tests/restart/mod.rs
```diff
@@ -133,6 +164,203 @@ async fn start_stop_empty(
...
+    <<TestSpec as Spec>::CryptoSpec as CryptoSpec>::Hasher,
+    NomtProverStorage<
+        DefaultStorageSpec<<<TestSpec as Spec>::CryptoSpec as CryptoSpec>::Hasher>,
+        MockHash,
+    >,
+>;
+
+/// This test intentionally crashes the rollup during a commit to ensure that the correct state is computed afterward.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_start_stop_with_crash() -> anyhow::Result<()> {
```
citizen-stig, line 178: I am afraid that setting environment variable can trigger crash of parallel test, if commit happens at the same time.

Can we move this test to a diffenet profile in nextest, or use serial, so we make sure this test runs in isolation.
I think it is not big price to pay in terms of runtime, but good trade off in terms of overall flakyness.

preston-evans98: Nextest runs each one in its own process, so we the env vars shouldn't be able to cross-contaminate each other: https://nexte.st/docs/design/why-process-per-test/

crates/full-node/sov-db/src/config.rs
preston-evans98: Good idea. Done.

examples/demo-rollup/tests/restart/mod.rs
preston-evans98: Yep, removed!

PR #1953, opened 2025-10-23
`DaService` wrapper with cache for recently finalized headers

citizen-stig:
# Reason
Screenshot 2025-10-23 at 10 45 25

## Follow up improvements

- Make number of cached headers dynamic: When sync starts, pre-fetchers also puts headers into cache. After sync is completed, map is resized to smaller size
- Add head fetching: this will remove storage preparation phase
- Bring back DA headers subscriptions and use it: faster update on finalized header, less network calls

### Sequencer end to end tests:

Rewamped test infrastructure in crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs:
  - added helper methods for better synchronization:
    - produce_enough_finalized_slots(): Produces blocks needed for finalized slot availability
    - tenderly_produce_blocks: Produces N blocks with interval so pull task will see each block separately
 - use existing helper methods for better stability and make tests more explanatory
    - wait_for_sequencer_ready(): Waits for sequencer to be ready instead of arbitrary sleeps
    - wait_for_node_synced(): Ensures node has caught up before proceeding
    - send_raw_tx_to_sequencer(): Simplified transaction submission
  - fixed flaky tests:
    - test_transaction_priority: Now uses synchronization helpers instead of sleep-based timing
    - seq_behind_deferred_slots_count_simple_lagging: Improved sequencer recovery test reliability
  - test improvements across the board:
    - Replaced sleep() calls with proper wait conditions
    - Better error messages with context
    - More deterministic test behavior
    - Removed race conditions in timing-sensitive tests
  - `DELAYED_TX_DELAY_MS` changed from `500` to `2500`, so it sleeps more than 1 batch and sequencer defentely has time to recover before tx wakes up
  - test `max_batch_execution_time` refactored: Now it does not wait for slot update between submissions, but submits all txs first, and then collects all slots to analyze the effect. It also uses `ProduceBlockOnSubmit` to avoid manually triggering new block
  - test `query_historical_values` expanded to distingiush between `None` and `Some(0)`
  - test `flaky_txs_that_enter_before_downtime_are_dropped` changed from imperative sending, to sending 2 txs in a loop until hitting desired effect. This is needed to avoid non-determinism regarding getting last finalized block header. This test has been verified by breaking actual code that it suppose to test.
  - **marked test `test_archival_state_is_immediately_available` as flaky**, because fetched value often 1 height older. It is going to be investigated in follow up PRs

------

## Testing
All tests pass or have been updated

## Docs
No updates


crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    tracing::info!("DA finalized header provider background task stopped");
+}
+
+#[cfg(test)]
+mod tests {
+    // use super::*;
+
+    // TODO
```
citizen-stig: This

crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
```diff
@@ -0,0 +1,411 @@
...
+/// - The DA service returns an error (which is logged and stops the task)
+/// - All receivers of the finalized header are dropped
+///
+/// # Thread Safety
+///
+/// This struct is `Clone` and can be shared across threads. All internal state
+/// is protected by appropriate synchronization primitives (`Arc`, `RwLock`, `watch`).
+#[derive(Debug, Clone)]
+pub struct DaServiceWithCachedFinalizedHeaders<Da: DaService> {
+    // TODO: Remove Arc, DaService already clone!
```
bkolad: What about this TODO?

crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
```diff
@@ -0,0 +1,411 @@
...
+    da_service: Arc<Da>,
+    /// Receiver for the last finalized header, updated by background task
+    last_finalized: tokio::sync::watch::Receiver<<Da::Spec as DaSpec>::BlockHeader>,
+    /// Cache of recently finalized headers, keyed by height
+    recent_headers: Arc<tokio::sync::RwLock<BTreeMap<u64, <Da::Spec as DaSpec>::BlockHeader>>>,
```
bkolad: Let's switch to `std::sync::RwLock`. The Tokio version is only necessary if we need to hold a lock across an .await point, but we should avoid doing that whenever possible because it can lead to various issues. It looks like we're not holding the lock across any awaits in this PR.

crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
```diff
@@ -0,0 +1,411 @@
...
+    ) -> Result<<Da::Spec as DaSpec>::BlockHeader, Da::Error> {
+        self.da_service.get_head_block_header().await
+    }
+}
+
+// TODO: Switch to subscription when it is brought back
+async fn background_header_fetch_task<Da: DaService>(
+    da_service: std::sync::Arc<Da>,
+    finalized_sender: tokio::sync::watch::Sender<<Da::Spec as DaSpec>::BlockHeader>,
+    recent_headers: std::sync::Arc<
```
bkolad: Nit not blocking: 
We could extract `Arc<tokio::sync::RwLock<BTreeMap<u64, <Da::Spec as DaSpec>:: BlockHeader>>` as a separate method and put the following in a method:
```
let recent_header_read = recent_headers.read().await;
                            if !recent_header_read.contains_key(&height) {
                                drop(recent_header_read);
                                let mut recent_header_write = recent_headers.write().await;
                                recent_header_write.insert(height, finalized_header);
                                if recent_header_write.len() > MAX_RECENT_HEADERS {
                                    let evicted = recent_header_write.pop_first();
                                    tracing::trace!(?evicted, "Evicting older header");
                                }
                                tracing::trace!(finalized_height = %height, "Updated cached recent headers");
                            }
                        }
```
This way, the locking behaviour will be encapsulated in single method

bkolad (approved):

crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
citizen-stig: I want to do in a follow up: DaService is arc'ed in runner, so change will blow up the size of PR

crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
citizen-stig: Done

PR #1958, opened 2025-10-23
Replace unwraps with descriptive expects in EVM module & check each one

LogvinovLeon:
# Description

Replaces bare unwrap() calls with expect() containing clear messages explaining why operations are safe, and extracts repeated pruning logic into dedicated methods.

Manually checked each expect. Unless we have more than u64::MAX txs/blocks/logs they should not happen.

Also:
  - Extract `prune_block()` and `prune_tx()` helper methods to reduce duplication
  - Add `head()` helper method for consistent head block access
-------

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/call.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/call.rs
@@ -118,12 +118,11 @@ where
         start_timer!(get_head_t);
         // Fetch `head` and `pending_len` before the `native` code block.
         // This ensures consistent gas charges between native and non-native execution.
-        #[allow(unused_variables)]
+        #[allow(unused_variables, clippy::expect_used)]
         let head = self
             .head
             .get(state)?
-            // Justified, we set it at `genesis` and leter only override it.
-            .expect("Impossible happened: Head must be set.");
+            .expect("Head is set in genesis and never deleted");
         save_elapsed!(get_head_time SINCE get_head_t);
 
         #[cfg(feature = "native")]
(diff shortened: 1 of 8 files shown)


crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -207,8 +207,7 @@ where
             tx.receipt
                 .log_index_start
                 .checked_add(tx.receipt.receipt.logs.len() as u64)
-                // Justified, we will never have that many logs.
-                .expect("Impossible happened: Log index overflow.")
+                .expect("we will never have that many logs")
```
bkolad: I think the previous message was clearer.

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -276,11 +275,8 @@ where
         let first_tx_index = head.transactions.end;
 
         let tx_index = first_tx_index
-            .checked_add(pending_tx_len)
-            .expect("The impossible happened: Tx index overflow.")
-            .checked_sub(1)
-            // Justified, can't underflow because `pending_tx_len` is greater than 0.
-            .expect("The impossible happened: Tx index underflow.");
+            .checked_add(pending_tx_len - 1)
+            .expect("We will never have that many transactions");
```
bkolad: Same not sure what is "that many" here

crates/module-system/module-implementations/sov-evm/src/hooks.rs
```diff
@@ -46,8 +41,7 @@ impl<S: Spec> BlockHooks for Evm<S> {
             .header
             .number
             .checked_add(1)
-            // This is justified. We will never have so many blocks.
-            .expect("The impossible happened: Block number overflow");
+            .expect("We will never have so many blocks");
```
bkolad: Same the previous message looks more informative

crates/module-system/module-implementations/sov-evm/src/hooks.rs
```diff
@@ -150,8 +132,7 @@ impl<S: Spec> BlockHooks for Evm<S> {
 
         let end_tx_index = start_tx_index
             .checked_add(pending_transactions.len() as u64)
-            // This is justified. We will never have that many txs.
-            .expect("The impossible happened: Tx count overflow");
+            .expect("We will never have that many transactions");
```
bkolad: Same

bkolad (approved):

crates/module-system/module-implementations/sov-evm/src/call.rs
LogvinovLeon: I think the expect message should tell the user - why this in fact will never happen and not - what could have happened. https://doc.rust-lang.org/std/error/index.html#common-message-styles
When I read: "Log index overflow" it doesn't have any additional context. Yeah - `checked_add` can overflow. That's why we use `checked_add`. 
I propose a compromise:
`We should never have more than u64::MAX logs`
Then the user can access: Yeah. Quintillion logs is probably enough.

PR #1961, opened 2025-10-23
Replica add `validate_db_data_from_replica` method.

bkolad:
# Description

1. Replaces `exit_rollup` with `panic` in the replica sync task.
2. Fixes a minor bug in the db pruning height calculation.
3. Adds a `validate_db_data_from_replica` method that checks whether data from the DB can be applied on the executor.

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/inner.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/inner.rs
@@ -823,6 +823,11 @@ enum Message<S: Spec, Rt: Runtime<S>> {
     SimpleStateUpdate {
         info: StateUpdateInfo<S::Storage>,
     },
+
+    CloseCurrentBatch {
+        reason: &'static str,
+    },
+
     DoBatchStartMsg {
         resp: oneshot::Sender<Result<(), ReplicaError<S>>>,
         batch_from_master: BatchToStore,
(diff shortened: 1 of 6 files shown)


crates/full-node/sov-sequencer/src/preferred/replica/event_handler.rs
```diff
@@ -48,13 +48,29 @@ where
 {
     #[allow(clippy::match_same_arms)]
     async fn on_db_event(&self, data: DbData) -> Result<(), DBDataRejected> {
-        match data {
-            DbData::BatchStart(_) => {}
-            DbData::Transaction(_, _, _) => {}
-            DbData::BatchEnd(_) => {}
-            DbData::NewProof => {}
+        let res: Result<(), ReplicaError<S>> = match data {
+            DbData::BatchStart(_) => Ok(()),
```
bkolad, line 52: In the next PR we will plug `process_do_batch_start_replica/process_do_new_tx_replica/process_close_current_batch_replica` here

crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -823,18 +823,27 @@ enum Message<S: Spec, Rt: Runtime<S>> {
...
     DoNewTx {
         resp: oneshot::Sender<Result<(), ReplicaError<S>>>,
+        seq_nr_from_master: u64,
         tx_hash: TxHash,
         baked_tx: FullyBakedTx,
         reason: &'static str,
     },
-    CloseCurrentBatch {
+
+    DoCloseCurrentBatch {
```
theodorebugnet, line 844: Only small comment on naming - from what I can see the `DoX` messages are used only by replicas, right? Maybe they should be changed to e.g. `ReplicaCloseCurrentBatch` or similar. Same for DoNewTx etc.
Non-blocking for this PR ofc

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/inner.rs
bkolad: yep will do in the next pr

PR #1965, opened 2025-10-24
Integrate replica sync task and `PreferredSequencer`

bkolad:
# Description

1. Renamed replica-related messages from `DoXX` to `ReplicaXX`.
2. Completed the implementation of `SequencerStateUpdator::on_db_event`. Replica-specific state update methods are now fully integrated.
3. Added handling for the `PreferredSeqOperation::Unreachable` case, which is now applicable to replicas. 

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/inner.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/inner.rs
@@ -311,13 +311,9 @@ where
 
         // Creates a new executor for recovery. This must *not* be called to create executors
         // under other circumstances, since it causes side effects on the transaction cache.
-        let recovery_executor = RollupBlockExecutor::<_, Rt>::new_with_tx_cache_writer(
-            info,
-            self.tx_cache_writer.clone(), // Recovery executor fills the cache
-            self.rollup_exec_config.clone(),
-            self.seq_config.clone(),
-            Default::default(), // Since we're entering recovery, we don't re-use any of the uncommitted changes
-        );
+
+        // Since we're entering recovery, we don't re-use any of the uncommitted changes
+        let recovery_executor = self.new_executor_with_empty_uncommitted_changes(info);
 
         self.force_overwrite_state(info.clone(), recovery_executor)
             .await;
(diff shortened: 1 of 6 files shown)


crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -1716,7 +1738,15 @@ where
             condition_are_there_batches_to_replay,
             condition_node_is_unsynced_and_doesnt_know_it,
         ) {
-            (true, _, _, true, _) => PreferredSeqOperation::Unreachable,
+            (true, _, _, true, _) => {
```
bkolad, line 1741: We’ll also need to handle other edge cases — for example, clarifying what it means for a replica to recover. These will be addressed in upcoming PRs.

examples/demo-rollup/tests/replica/mod.rs
```diff
@@ -154,21 +154,9 @@ async fn test_replica_receives_txs_from_postgres() {
...
-        tokio::time::sleep(std::time::Duration::from_millis(
-            sov_test_utils::TEST_DEFAULT_MOCK_DA_BLOCK_TIME_MS,
-        ))
-        .await;
-    }
 
-    test_rollup
-        .wait_for_height(height_before_tx.get() + gap)
-        .await;
+    tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
bkolad, line 159: In this test, block production is done manually, and we’ve replaced block creation with a timeout. As a result, the replica can only observe updates through PostgreSQL. This test verifies that it behaves as intended.

crates/full-node/sov-sequencer/src/preferred/inner.rs
```diff
@@ -1989,7 +2006,17 @@ where
...
 
+        let recovery_executor = inner.new_executor_with_empty_uncommitted_changes(&info);
+
+        inner
+            .force_overwrite_state(info.clone(), recovery_executor)
+            .await;
+
         inner.update_api_ledger(&info).await;
+
+        if inner.is_replica() && info.sync_status.distance() <= distance {
```
theodorebugnet, line 2017: What's the reason for special-casing this here? IIRC, the master sequencer will wait for an extra state update before setting ready = Ok(). Are there any issues with having replicas use the same codepath?

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/inner.rs
bkolad: Yes, without this, the replica was entering an infinite resync loop. The issue is that after the extra state update, the inner executor still lags behind the node executor (so replica continues resyncing). If we mark it as OK here, we can start applying the PG transactions on the inner executor, which then races ahead of the node. I’ll look into making this cleaner.

PR #1966, opened 2025-10-24
Fix UniversalWallet serialization for web3 SDK

theodorebugnet:
# Description

Add UniversalWallet overrides at the type level for every implementer of CryptoSpec, rather than overriding it at the field level inside Transaction. This ensures the correct byte array length is used, whereas in the Transaction definition, it was not possible to use a generic length.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/sov-modules-api/src/transaction/mod.rs
```diff
@@ -59,11 +59,9 @@ impl<D: DispatchCall> TransactionCallable for D {
 pub struct Version0<Call, S: Spec, C: CryptoSpecExt = <S as Spec>::CryptoSpec> {
     /// The signature of the transaction.
     #[serde(with = "hex_field_format")]
-    #[sov_wallet(as_ty = "[u8; 64]", display = "hex")]
```
ross-weir, line 62: I would have thought we still need `hex` display to pass as hex string, is that not the case?

ross-weir (approved):

crates/module-system/sov-modules-api/src/transaction/mod.rs
theodorebugnet: You're right actually! Thanks for catching, I'll re-add that

PR #1973, opened 2025-10-26
chore: remove unused import from Risc0 ZKVM module

gap-editor:
# Description

This PR performs a small cleanup in the Risc0 cryptography module:

* Removed an unused import:

  ```rust
  use std::hash::Hash;
  ```

  which was redundant since `Hash` is already derived via `#[derive(Hash)]` on `Risc0PublicKey`.


gap-editor:

PR #1975, opened 2025-10-27
Add module associated error type

ross-weir:
# Description

Adds `Module::Error` associated type to modules. Set to `anyhow::Error` to preserve current behaviour except for `sov_bank` was changed to use it's own `Error` type. This was done in this PR because with the changes in this PR we would lose the `anyhow` error chain and thus many bank tests would be broken.

Adds a `ErrorDetail` trait the `Module::Error` should implement. This trait decides how errors are returned from the REST api.

This PR has a large diff but I couldn't see a sensible way to split it into multiple PRs so will list some points of interest below:

- New error traits/utilities/etc: https://github.com/Sovereign-Labs/sovereign-sdk/pull/1975/files#diff-415c56381b3f95e7f7fe5f966b7fa526921510f155604b69f0ac102d9f10b597
- Bank module error: https://github.com/Sovereign-Labs/sovereign-sdk/pull/1975/files#diff-5f98eb7f09ddbce3523f743be0fa835d7e6f8713a6ea471c301447faaa951d87 (the bank module changes in general would also be good to look over)

---

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/module-implementations/sov-attester-incentives/tests/integration/unbonding.rs
+++ b/crates/module-system/module-implementations/sov-attester-incentives/tests/integration/unbonding.rs
@@ -2,7 +2,7 @@ use sov_attester_incentives::{CustomError, UnbondingInfo};
 use sov_mock_da::MockAddress;
 use sov_modules_api::prelude::UnwrapInfallible;
 use sov_modules_api::registration_lib::RegistrationError;
-use sov_modules_api::Error::ModuleError;
+use sov_modules_api::Error;
 use sov_modules_api::{Amount, Spec, StateAccessorError};
 use sov_rollup_interface::common::SlotNumber;
 use sov_test_utils::runtime::TestRunner;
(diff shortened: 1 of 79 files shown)


crates/module-system/module-implementations/sov-attester-incentives/tests/integration/unbonding.rs
```diff
@@ -228,18 +224,16 @@ fn try_skip_two_phase_unbonding() {
...
-                                MockAddress,
-                                MockAddress,
-                                StateAccessorError<<S as Spec>::Gas>,
-                                _,
-                            >::Custom(
-                                CustomError::AttesterIsNotUnbonding(addr)
-                            )
-                            .into(),
-                        ),
+                        reason.reason.to_string(),
```
ross-weir, line 226: Note to self

Should compare 2 `ModuleError` instances if possible instead of strings

Although under the hood it also just does a to_string cmp this allows us to change how we cmp in the future with minimal updates required

ross-weir: can do later if needed, already spent ednought time on this PR

bkolad (approved):

PR #1979, opened 2025-10-27
Fix critical invariant check in NomtStateDb commit flow

prestoalvarez:
The commit flow in `NomtStateDb` uses `debug_assert_eq!` to verify that the commit flag status is `Completed` before proceeding. This check is completely removed in release builds, allowing commits to proceed from unsafe states.


prestoalvarez:

Issue #1630, opened 2025-09-05
[TRACKER] Celestia production setup


citizen-stig:
## Description

Partially based on https://github.com/Sovereign-Labs/sovereign-sdk-wip/issues/2682

## Important

 - [x] Perform long, stable soak run performance matching MockDA
   - [x] 8MB blobs with vanilla celestia node
   - [x] Remove inner lock on Celestia Service, needs this https://github.com/celestiaorg/celestia-core/pull/2558 https://github.com/Sovereign-Labs/sovereign-sdk/pull/2111
   - [x] ~~Twinkle connection https://github.com/Sovereign-Labs/sovereign-sdk/pull/1805~~
 - [x] Add custom adapter metrics
   - [x] Submit metrics into telegraf https://github.com/Sovereign-Labs/sovereign-sdk/pull/1633
   - [x] Custom dashboard for that metrics
 - [x] ~~Test keystorage option for celestia node for production setup~~
   - [x] ~~Document how it can be configured and what options are supported (related https://github.com/celestiaorg/docs/issues/1968)~~
 - [x] Ensure fees are high enough for fast inclusion. Users can parametrize that: https://github.com/Sovereign-Labs/sovereign-sdk/pull/1783 . Exact values can be configured by providing custom gas price endpoint to celestia node.
 - [x] https://github.com/celestiaorg/celestia-node/issues/4583
 - [x] Upgrade to V6 https://github.com/Sovereign-Labs/sovereign-sdk/pull/1745
 - [x] Move processing finalized blocks out of main loop to increase throughput: https://github.com/Sovereign-Labs/sovereign-sdk/pull/1953
 - [x] Config: API key from env variable for secure kubernetes setup https://github.com/Sovereign-Labs/sovereign-sdk/pull/1981
 - [x] https://github.com/eigerco/lumina/issues/816
 - [ ] Document the feature and **review the generated Rustdoc** for celestia adapter
 - [x] Run soak test on celestia mainnet
 - [x] Update documentation in starter on testnet and mainnet setup with providers and best practices https://github.com/Sovereign-Labs/rollup-starter/pull/73

## Nice to have

 - [ ] Benchmark celestia risc0 verification (w/o rollup logic)
[...]

PR #1981, opened 2025-10-27
Use `celestia-client` in celestia adapter

citizen-stig:
# Description

Switching to `celestia-client` allows working directly with RPC node providers without need of running light node. Although it is still possible, but signing of blob and request is done by rollup node.

This PR adds only native code. `celestia_client` relies on `celestia_types`, to ensure compatibility with existing prover.

## Changes in the config

- Renamed fields with backward compatibility:
  - celestia_rpc_address → rpc_url (with alias)
  - celestia_rpc_auth_token → rpc_auth_token (now optional, also can be taken from env)
  - celestia_rpc_timeout_seconds → request_timeout_secs (with alias)
- Removed fields:
  - max_celestia_response_body_size (no longer needed)
  - signer_address (now auto-detected from client)
- New fields:
  - grpc_url (optional, required for blob submission)
  - grpc_auth_token (optional, from env SOV_CELESTIA_GRPC_AUTH_TOKEN)
  - signer_private_key (optional, from env SOV_CELESTIA_SIGNER_KEY)
- Custom Debug implementation that redacts sensitive fields
- Support for read-only mode when grpc_url/signer are not set

## Using testcontainers instead of wiremock

- Replaced wiremock dependency with testcontainers for integration tests
- Added CelestiaDevNode test helper that spins up real Celestia validator and bridge containers
- Uses published docker images: ghcr.io/sovereign-labs/celestia-validator-devnet:v5.0.2-mocha and ghcr.io/sovereign-labs/celestia-bridge-devnet:v0.27.4-mocha
- Disabled 3 error-case tests that require toxiproxy (planned for follow-up PR)
- Tests now use actual Celestia protocol instead of mocks for higher fidelity

## Other changes

* Mutex on submit client is now hold between retries on error. This way it avoids spiraling nonce issue if one request accidentally fails, while other senders were waiting. This makes submission truly single threaded.
* `test_helper/files` moved to separate file. Tests of the verifier use function to extract blobs and verify, so no need to initialize DaService. 
* `test_helper/serialization` moved to separate file
* `CelestiaConfig` debug implementation changed, so keys won't accidentally leak into logs
* Hack applied to fix error below. Opionately select ring backend. Root cause will be fixed later.
  ```
  Could not automatically determine the process-level CryptoProvider from Rustls crate   features.
      Call CryptoProvider::install_default() before this point to select a provider manually, or make sure exactly one of the 'aws-lc-rs' and 'ring' features is enabled.
      See the documentation of the CryptoProvider type for more information.
  ```

## Disabled tests

Following tests are checking error cases and to make them work with test container, we need to add `toxiproxi` into scenario, which is moved to follow up PR

* `test_submit_blob_application_level_error`
* `test_submit_blob_internal_server_error`
* `test_submit_blob_response_timeout`

## Items for the follow up

* Optimize `get_block_at` and query block header concurrently with namespace data
* Enable disabled tests
  * Add test for subscription
  * More tests with testcontainers:
    * blob get all for proofs
    * several senders, same namespace, different namespaces.
    * different blob sizes
* Better retry logic: currently all errors are Transient 
* Suggestion: Change DaService::get_signer to return Option, for DaService that cannot publish.
* More sophisticated private key management:
  * Seed phrase support
  * Remote signer
* https://github.com/eigerco/lumina/issues/791 Timeout on client level, currently we wrap whole future
* testcontainer CelestiaDevNode improvements:
  * Fix failed submission on first couple blocks (account does not exist)
  * Tune logs and disable logs for `2025-10-29T08:57:49.124Z	WARN	share/discovery	discovery/discovery.go:188	error advertising	{"rendezvous": "/full/v0.1.0", "err": "failed to find any peer in table"}`
  * Helper for printing logs
* Extend config examples:
  * Community RPC endpoint and light node: Free, but unreliable
  * Fully Sovereign: Your own bridge node and consensus node

## Linked Issues
- Related to #1630  

## Testing
Tests are adapted and fixed.

## Docs
Following documentation:
 - [x] Config examples for quicknode
 - [x] Instruction on how to upgrade to newer celestia versions


crates/adapters/celestia/src/da_service/mod.rs
```diff
@@ -96,10 +95,14 @@ impl CelestiaService {
         };
         debug!(bytes, namespace = ?ns, "Sending raw data to Celestia");
 
+        let Some(signer) = &self.signer_address else {
+            // TODO: Better error when switched to thiserror.
```
citizen-stig: Fix this: otherwise it will keep retrying

crates/adapters/celestia/src/da_service/mod.rs
```diff
@@ -520,8 +496,14 @@ impl DaService for CelestiaService {
         .await
     }
 
+    // TODO: Should this become and Result or Option?
```
citizen-stig: In the future

bkolad (approved):

PR #1988, opened 2025-10-28
Sequencer: add operation_for_master and operation_for_replica methods.

bkolad:
# Description

1. Use the `select!` macro in the replica sync task to wait for the shutdown signal.
2. Add `operation_for_master and operation_for_replica` methods. These methods determine the sequencer state. Unfortunately, replicas can enter different states than masters, so we need to distinguish between them. At this point, the implementations are nearly identical (only the first case differs), but in the next PR we’ll add proper handling for replica-specific states, and the methods will diverge further.

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
@@ -83,18 +83,21 @@ impl EventReceiver {
 
     pub(crate) async fn spawn_db_data_fetcher(mut self) -> JoinHandle<()> {
         let mut nb_of_consecutive_db_errors = 0;
-        let shutdown_receiver = self.shutdown_sender.subscribe();
+        let mut shutdown_receiver = self.shutdown_sender.subscribe();
 
         tokio::spawn(async move {
             let mut start_event_id = None;
             let mut prev_event_type = None;
 
             loop {
-                if shutdown_receiver.has_changed().unwrap_or(true) {
-                    break;
-                }
+                let res = tokio::select! {
+                     _ = shutdown_receiver.changed() => {
+                        break
+                    },
+                    res = self.fetch_data(start_event_id, prev_event_type) => res
+                };
 
-                match self.fetch_data(start_event_id, prev_event_type).await {
+                match res {
                     Ok((event_id, event_type)) => {
                         nb_of_consecutive_db_errors = 0;
                         start_event_id = Some(event_id + 1);
--- a/crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
@@ -4,6 +4,8 @@ use crate::preferred::replica::db_data::DbData;
 use crate::preferred::replica::db_data::EventType;
 use crate::preferred::replica::db_data::EventsNotificationPayload;
 use crate::preferred::replica::db_data::ParsingError;
+use sov_rollup_interface::node::future_or_shutdown;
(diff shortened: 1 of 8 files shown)


crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/mod.rs
```diff
@@ -228,3 +229,24 @@ pub(crate) struct ProcessFinalCatchupData {
     pub(crate) transactions_count: usize,
     pub(crate) batch_is_in_progress: bool,
 }
+
+struct TrueTable {
```
theodorebugnet: Some naming thoughts: `TruthTable` would be more correct, but maybe even something like `ConditionsTable`?
If we go with the latter the bools can have the `condition_` prefix removed. So something like:
```suggestion
struct ConditionsTable {
    nodes_sequence_number_is_fresher: bool,
    too_close_to_deferred_slots_count_for_comfort: bool,
    node_is_lagging: bool,
    are_there_batches_to_replay: bool,
    node_is_unsynced_and_doesnt_know_it: bool,
}
```

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/mod.rs
```diff
@@ -228,3 +229,24 @@ pub(crate) struct ProcessFinalCatchupData {
...
+    condition_nodes_sequence_number_is_fresher: bool,
+    condition_too_close_to_deferred_slots_count_for_comfort: bool,
+    condition_node_is_lagging: bool,
+    condition_are_there_batches_to_replay: bool,
+    condition_node_is_unsynced_and_doesnt_know_it: bool,
+}
+
+#[derive(Debug)]
+struct InitialConditions {
```
theodorebugnet: Most of these bools are just extracted from `inner.is_ready()`, so there's some conceptual duplication - but is_startup is different (comes from `inner.has_finished_startup`) so we can't just implement `should_flush_tx_cache()` on is_ready directly. Maybe we can implement it on `inner`? I'm not sure - I'm just thinking there might be a way to reduce the amount of different types and variables carrying the same data; but if it ends up adding too much complexity then it's not worth it.

theodorebugnet (approved):
LGTM with non-blocking comment + naming suggestion

crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
```diff
@@ -83,18 +83,21 @@ impl EventReceiver {
...
             let mut start_event_id = None;
             let mut prev_event_type = None;
 
             loop {
-                if shutdown_receiver.has_changed().unwrap_or(true) {
-                    break;
-                }
-
-                match self.fetch_data(start_event_id, prev_event_type).await {
+                let res = tokio::select! {
```
citizen-stig: Why not `FutureOrShutdownOutput` which has nice ergonimics?

bkolad: fixed

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/mod.rs
bkolad: fixed

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/mod.rs
bkolad: I'll revisit this in the next PR. Since it's a large refactor, I don't want to introduce new code right now. I'll take a look soon.

PR #1991, opened 2025-10-28
Fix foundry: coerce 'latest' -> 'pending' for getBlockByNumber,

preston-evans98:
Coerce 'latest' -> 'pending' for getBlockByNumber, and allow querying the 'pending' block by requesting height N+1 in EVM RPC handlers

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
@@ -346,7 +346,9 @@ where
                 match pending_or_block_nr {
                     PendingOrBlock::Pending => MaybeArchivalState::Current(state),
                     PendingOrBlock::Number(number) => {
-                        if number == state.rollup_height_to_access().get() || (number == state.rollup_height_to_access().get() + 1) {
+                        if number == state.rollup_height_to_access().get()
+                            || (number == state.rollup_height_to_access().get() + 1)
+                        {
                             return Ok(MaybeArchivalState::Current(state));
                         }
                         let archival_state = state
--- a/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
@@ -16,7 +16,7 @@ use revm::context::{BlockEnv, CfgEnv};
 use sov_address::{EthereumAddress, FromVmAddress};
 use sov_modules_api::macros::config_value;
 use sov_modules_api::prelude::UnwrapInfallible;
-use sov_modules_api::{ApiStateAccessor, Spec};
+use sov_modules_api::{ApiStateAccessor, Spec, VersionReader};
 use sov_rollup_interface::common::RollupHeight;
 use sov_rpc_eth_types::{EthApiError, RpcInvalidTransactionError};
 
@@ -253,6 +253,12 @@ where
         block_number: Option<String>,
         state: &mut ApiStateAccessor<S>,
     ) -> Option<MaybeSealedBlock> {
+        if let Some(block_number) = &block_number {
+            if block_number == "latest" {
+                tracing::warn!("Overwriting latest block with pending block");
+                return Some(MaybeSealedBlock::Pending(self.pending_block(state)));
+            }
+        }
         let pending_or_block_nr = self.str_to_block_nr(block_number, state);
 
         match pending_or_block_nr {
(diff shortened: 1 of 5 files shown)


crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -334,6 +346,9 @@ where
                 match pending_or_block_nr {
                     PendingOrBlock::Pending => MaybeArchivalState::Current(state),
                     PendingOrBlock::Number(number) => {
+                        if number == state.rollup_height_to_access().get() || (number == state.rollup_height_to_access().get() + 1) {
```
bkolad: This is just a minor nit, and the PR can be merged without addressing it. Inside `str_to_block_nr`, we detect the `latest` block number using `self.block_numbers(state).end()`, so we could reuse the same logic here for consistency. Or `str_to_block_nr` could even return an enum `PendingOrLatestBlock` and we just pattern match on it.

bkolad (approved):

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -253,6 +253,12 @@ where
         block_number: Option<String>,
         state: &mut ApiStateAccessor<S>,
     ) -> Option<MaybeSealedBlock> {
+        if let Some(block_number) = &block_number {
+            if block_number == "latest" {
+                tracing::warn!("Overwriting latest block with pending block");
+                return Some(MaybeSealedBlock::Pending(self.pending_block(state)));
+            }
+        }
```
LogvinovLeon, line 261: We have three places where we resolve block numbers:
* One is here where we do the override as an early exit before calling `str_to_block_nr` and emit a warning
* Another is in `resolve_state` where we first call `str_to_block_nr` and then special-case this situation but with no warning.
* Last one is in `resolve_block_number` where we don't deal with this case

Maybe let's decide on one way of doing it, extract it to a helper utility where we add a comment explaining the reasoning and a warning. Later use it in all three places consistently?

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -340,6 +346,11 @@ where
                 match pending_or_block_nr {
                     PendingOrBlock::Pending => MaybeArchivalState::Current(state),
                     PendingOrBlock::Number(number) => {
+                        if number == state.rollup_height_to_access().get()
```
LogvinovLeon, line 349: This is confusing. So both `height` and `height + 1` are represented by the current state?
The block height+1 never has state changes?
Let's add a comment explaining the logic here.

examples/demo-rollup/tests/evm/evm_soft_conf.rs
```diff
@@ -26,11 +26,8 @@ async fn evm_test_soft_confirmations() -> anyhow::Result<()> {
...
 
-            assert_eq!(pending_block.parent_hash, latest_block.hash.unwrap());
-            assert_eq!(
-                pending_block.number.unwrap(),
-                latest_block.number.unwrap() + 1
-            );
+            assert_eq!(pending_block.parent_hash, latest_block.parent_hash);
+            assert_eq!(pending_block.number.unwrap(), latest_block.number.unwrap());
             assert!(pending_block.transactions.is_empty());
             assert!(latest_block.transactions.is_empty());
```
LogvinovLeon, line 32: If latest and pending now resolve to the same block - we can replace those 4 assertions by those two.
```rs
assert_eq!(pending_block, latest_block);
assert!(pending_block.transactions.is_empty());
```

examples/demo-rollup/tests/evm/evm_tx.rs
```diff
@@ -55,7 +55,7 @@ async fn sanity_checks(test_client: &SimpleStorageClient) {
 
     assert!(latest_block.number.unwrap().as_u64() > 0);
     assert!(latest_block.number > earliest_block.number);
-    assert!(pending_block.number > latest_block.number);
+    assert!(pending_block.number == latest_block.number);
```
LogvinovLeon, line 58: Same here. It's not enough to make this test pass. We've changed the semantics and therefore those assertions now don't make sense any more.
We used to have two objects in hand and checked how they relate. Now we have one object that can be accessed in two different ways. So let's first check that both ways of accessing return the same object and then do all assertions on one instance of it.

LogvinovLeon (commented):
Don't want to block this PR but added some comments

PR #1997, opened 2025-10-29
Fix: eth_sendTransaction fails for contract deployments without gas limit

LogvinovLeon:
# Description

Fixes `eth_sendTransaction` failing with "Transaction conversion error" when deploying contracts without specifying gas limit.

## Problem

Contract deployments via `eth_sendTransaction` (no `to` field, no `gas` field) fail because alloy's `build_typed_tx()` requires `to: Some(TxKind::Create)` but the JSON-RPC spec uses `to: null`.

## Solution

Convert `to: None` to `to: Some(TxKind::Create)` before calling `build_typed_tx()` in the `eth_send_transaction` handler.

---

## Linked Issues
- Related to # (issue) 

## Testing
- Added test `examples/demo-rollup/tests/evm/evm_no_gas_limit.rs` to verify fix

## Docs

### DIFF
--- a/crates/full-node/sov-ethereum/src/handlers/mod.rs
+++ b/crates/full-node/sov-ethereum/src/handlers/mod.rs
@@ -49,7 +49,6 @@ where
     Seq::Rt: HasKernel<S> + EthereumAuthenticator<S> + Default + Send + Sync + 'static,
     F: Fn(B256, Arc<Ethereum<S, Seq>>) -> Result<T, ErrorObjectOwned>,
 {
-    dbg!();
     let raw_evm_tx = RlpEvmTransaction { rlp: data.to_vec() };
     let (tx_hash, raw_message) = ethereum
         .make_raw_tx(raw_evm_tx)
--- a/crates/full-node/sov-ethereum/src/handlers/mod.rs
+++ b/crates/full-node/sov-ethereum/src/handlers/mod.rs
@@ -1,5 +1,7 @@
 mod get_logs;
 mod subscribe;
+#[cfg(feature = "local")]
+use alloy_primitives::TxKind;
 use alloy_primitives::{Bytes, B256};
 use alloy_rpc_types::TransactionReceipt;
 pub use get_logs::{Cursor, LogHandlers};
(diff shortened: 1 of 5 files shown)


bkolad (approved):

crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -47,6 +49,7 @@ where
     Seq::Rt: HasKernel<S> + EthereumAuthenticator<S> + Default + Send + Sync + 'static,
     F: Fn(B256, Arc<Ethereum<S, Seq>>) -> Result<T, ErrorObjectOwned>,
 {
+    dbg!();
```
bkolad: Let's remove it

crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -208,6 +211,13 @@ pub(crate) mod signer {
             )?;
             transaction_request.gas = Some(estimated_gas.to::<u64>());
 
+            // For contract deployments, convert `to: None` to `to: Some(TxKind::Create)`
+            // The JSON-RPC spec uses `null` or omitted `to` field for contract deployments,
+            // but alloy's `build_typed_tx()` requires `Some(TxKind::Create)`
+            if transaction_request.to.is_none() {
```
bkolad, line 216: Don't we need the same for eth_sendRawTransaction?

examples/demo-rollup/tests/evm/evm_no_gas_limit.rs
```diff
@@ -0,0 +1,47 @@
...
+        "from": from,
+        "maxFeePerGas": "0x1",
+        "maxPriorityFeePerGas": "0x1",
+        "data": bytecode,
+    }]);
+
+    // Should auto-estimate gas and deploy the contract
+    let tx_hash: B256 = client
+        .client()
+        .request("eth_sendTransaction", &params)
```
bkolad, line 36: What about eth_sendRawTx?

bkolad (commented):
Approving, provided we’re certain the fix has to apply only to eth_sendRawTx.

crates/full-node/sov-ethereum/src/handlers/mod.rs
LogvinovLeon: sendRawTransaction does not construct the TX from TxRequest but gets an already constructed TX and therefore does not trigger this code path in alloy

PR #2002, opened 2025-10-30
demo-rollup: Add `ExternalMock`

bkolad:
# Description

This PR introduces the following updates to the demo-rollup:
1. Adds a new rollup type, `ExternalMock,` which integrates with an external mock DA.
2. Moves the rollup configuration into a dedicated directory.


examples/demo-rollup/src/main.rs
```diff
@@ -257,6 +289,35 @@ async fn new_rollup_with_mock_da_and_jmt(
         .await
 }
 
+async fn new_rollup_with_external_mock_da_and_jmt(
```
citizen-stig, line 296: Do we actually need this one? I think NOMT is a preferred way forward.

citizen-stig: Or is it now existing mock rollups rely on external mock da?

citizen-stig (approved):

examples/demo-rollup/src/main.rs
bkolad: We need this one for unit tests because the testing framework worworksks with jmt

PR #2003, opened 2025-10-30
Report errors to ws clients when they occur

preston-evans98:
# Description

This PR implements a small feature requests from Zeta; now, websockets will report errors to the client before closing the connection whenever possible.

### DIFF
--- a/crates/full-node/sov-ledger-apis/src/lib.rs
+++ b/crates/full-node/sov-ledger-apis/src/lib.rs
@@ -667,8 +667,13 @@ where
     ) -> impl IntoResponse {
         ws.on_upgrade(|socket| async move {
             let subscription = state.ledger.subscribe_proof_saved().map(|data| {
-                AggregatedProof::try_from(data)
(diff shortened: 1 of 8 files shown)


crates/full-node/sov-ledger-apis/src/lib.rs
```diff
@@ -666,7 +668,7 @@ where
         ws.on_upgrade(|socket| async move {
             let subscription = state.ledger.subscribe_proof_saved().map(|data| {
                 AggregatedProof::try_from(data)
-                    .context("Failed to convert proof to REST API representation")
+                    .map_err(|_| WsLedgerError::AggregatedProofConvertFailed)
```
citizen-stig: Can we log original error at least on info level?

citizen-stig (approved):

crates/full-node/sov-ledger-apis/src/lib.rs
preston-evans98: Good catch. Done

Issue #1992, opened 2025-10-28
Out-of-order nonce queue handling in the sequencer


theodorebugnet:
## Description

The goal is to implement a per-account queue in the sequencer which will allow reordering of nonce-based transactions up to some limited threshold, to ensure they get in even if the sequencer receives them out of order.

The core of the implementation will be a new TxNonceQueue structure that will look roughly like this:
```rust
struct QueuedTx {
    tx: FullyBakedTx,
    // ...additonal data TBD...
    accepted_sender: oneshot::Sender,
}

struct TxNonceQueue {
  queue: Map<Address, ConcurrentOrderedVec<QueuedTx>> // exact data structure TBD. The values should have locking for concurrency - per-account (not for the while data structure) since different addresses have independent nonce tracking
}

impl TxNonceQueue {
  fn add_tx(address: S::Address, tx: FullBakedTx) -> oneshot::Receiver {
    let (sender, receiver) = oneshot::new();
    let queued_tx = QueuedTx::new(tx, sender);
    receiver
  }

  fn evict_tx(address: S::Address, nonce: u64);
  fn has_prerequisites_to_nonce(address: S::Address, nonce: u64) -> bool;

  async fn trigger_next_tx_if_valid(address: S::Address, next_valid_nonce: u64) {
    let Some(queued_tx) = maybe_get_next_tx_if_valid(nonce) else { return; } // removes the tx from self.queue
    // then we queue the new tx message and forward the return directly to the original task that will be awaiting on this sender - which corresponds to the original `accept_tx()` top-level API call
    tokio::spawn(async move {
      let ret = synchronised_state_updator.accept_tx_msg(queued_tx.tx);
      queued_tx.accepted_sender.send(ret);
    })
  }
}
```

Other than the queue, the bulk of the logic will be for awaiting the transaction result and handling timeouts in `accept_tx()`. The current implementation is roughly
```rust
async fn accept_tx() {
  // pre-checks...
  let rx = self.synchronised_state_updater.accept_tx_msg(tx).await;
  rx.await
}
```

The new implementation will look as follows:
```rust
async fn accept_tx() {
  // pre_checks...
  let locked_account_queue = self.tx_nonce_queue.lock(address).await;
  let current_nonce = <Rt::default() as TransactionAuthorizer<S>>::get_current_nonce();
  match TxNonceQueue::categorize_nonce(current_nonce, tx.nonce) {
    Valid => {
      drop(locked_account_queue);
[...]

PR #2004, opened 2025-10-31
Sequencer transaction queue for reordered nonces

theodorebugnet:
# Description

Implements a queue in the sequencer's `accept_tx()` which allows accepting nonce-based transactions out of order, and executing them in their nonce order. The design of this feature is explained in #1992.

A very large chunk of the linecount is tests. The important implementation is the sequencer mod.rs changes (~250 lines) and the impl part of tx_nonce_queue.rs (~260 lines). Also of note is the removal of the nonce retry in the sov-ethereum handler.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,722 @@
...
+            queues: DashMap::new(),
+        }
+    }
+
+    /// Lock the queue for a specific address for atomic nonce check + enqueue
+    pub fn lock_for_address(
+        &self,
+        credential_id: CredentialId,
+    ) -> dashmap::mapref::entry::Entry<CredentialId, AddressQueue<S, Rt>> {
+        self.queues.entry(credential_id)
```
bkolad, line 235: There’s something I don’t quite understand about DashMap. The documentation for `entry` says:

```
Locking behaviour: May deadlock if called when holding any sort of reference into the map.
```

Does this mean that if one thread has already locked an entry, and another thread tries to lock it again, it could lead to a deadlock?
My understanding is that the first thread would still be holding a reference into the map’s internal structure, which might be causing this issue. But maybe I don't read the documentation properly WDT?

theodorebugnet: Yeah the documentation is confusing, I thought the same too at first. But it says at the top
```
Documentation mentioning locking behaviour acts in the reference frame of the calling thread. This means that it is safe to ignore it across multiple threads.
```
So it's safe for multithreading, you just have to be careful within a single thread

bkolad: But we are inside an async. So 2 tokio tasks can run in the same thread.

crates/full-node/sov-sequencer/src/preferred/mod.rs
```diff
@@ -529,6 +538,138 @@ where
...
+                                let current_nonce = self.get_current_nonce_for_user(&credential_id);
+                                err_invalid_nonce::<S, Rt>(
+                                    tx_hash,
+                                    tx_nonce,
+                                    current_nonce,
+                                    credential_id,
+                                )
+                            });
+                        },
+                        _ = tokio::time::sleep(std::time::Duration::from_millis(
```
bkolad: My understanding is that in this situation:
1. The queue is locked, and we’re holding a reference to some of its internal data.
2. Both the Tokio `sleep` and the `queue_receiver` are inside a select!, so the thread running this task can pause it and start processing a new request.
3. A new request handled by the same thread may then try to lock the queue again, which can lead to a deadlock. WDT?

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,722 @@
...
+
+        if tx_nonce < lower_bound {
+            // The transaction can never be valid.
+            false
+        } else if lower_bound <= tx_nonce && tx_nonce <= upper_bound {
+            // There is uncertainty about the user account's real nonce. In this range, leniently
+            // treat transactions as valid (the uncertainty will eventually resolve itself as API
+            // state updates).
+            true
+        } else if tx_nonce > upper_bound {
```
bkolad: Let's remove `if tx_nonce > upper_bound` and unreachelble branch on line 112

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,722 @@
...
+    }
+
+    /// Remove a transaction by nonce
+    pub fn evict(&self, credential_id: &CredentialId, nonce: u64) -> Option<QueuedTx<S, Rt>> {
+        let mut entry = self.queues.get_mut(credential_id)?;
+        let tx = entry.remove(nonce)?;
+
+        // Clean up empty queue
+        if entry.is_empty() {
+            drop(entry);
```
bkolad: Would it deadlock without this drop?

crates/full-node/sov-sequencer/src/preferred/mod.rs
theodorebugnet: No, the queue lock is consumed by `enqueue_from_lock()`, so the queue should no longer be locked at the point of the select!. So this shouldn't happen.

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
theodorebugnet: Good point actually. But I don't think we hold a queue reference anywhere across an `await` point.
 * Inside the `TxNonceQueues`, all functions are sync, except `drain_any_ready_transactions()`. But the only await there is on the transaction execution and we drop the lock before that.
 * Inside `handle_nonce_based_tx()`, we get a lock at the beginning of the function, and then we drop it at the start of every arm in the `match` statement, and there are no awaits between the lock and the match arms

So it should be correct.

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
theodorebugnet: Yeah - because we have awaits afterwards. It would also affect performance even if we used any other queue library since it would block any other accept_tx from a user on the same locked shard until the accept_tx_message() completes.

crates/full-node/sov-sequencer/src/preferred/mod.rs
```diff
@@ -529,6 +538,138 @@ where
...
+    /// transaction is pulled from the queue it also forwards the results to the original oneshot,
+    /// so the original `accept_tx()` task receives the data and can return it to the user.
+    ///
+    /// If the original task does not get a response across the oneshot before the timeout, the
+    /// transaction is evicted from the queue, *unless* at that point there's already a queued
+    /// transaction for every nonce (all the pre-reqs are satisifed). In that case the assumption
+    /// is that it will get executed soon, provided all the txs are valid, so we don't evict. If
+    /// any pre-req is rejected, the following txs will be evicted on the next timeout (assuming a
+    /// new replacement pre-req with that nonce isn't submitted in the meantime of course).
+    async fn handle_nonce_based_tx(
```
bkolad: It appears that the queue-related code depends only on the `accept_tx_msg` function from `synchronized_state_updator`.
To improve modularity, we could do the following:

1. Define a trait that declares a single method matching the signature of `accept_tx_msg`.
2. Implement this trait for `synchronized_state_updator.` and use it in the sequencer
3. In tests, we can provide custom implementations to simulate and verify various edge cases (handle_nonce_based_tx could be unit-tested as a separate component)

Additionally, the `handle_nonce_based_tx` function could be moved to `preferred_tx_nonce... rs`, and with these changes, the queue would become a self-contained component with its own internal locking mechanisms, fully decoupled from the sequencer. So nothing about the locking policy leaks to the sequencer.

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,1475 @@
...
+
+use super::sync_sequencer_state::{
+    AcceptTxError, SequencerStateUpdator, SequencerStateUpdatorError,
+};
+use super::{err_invalid_nonce, Confirmation};
+
+pub(crate) type TransactionReceiverResult<S, Rt> =
+    Result<oneshot::Receiver<AcceptedTx<Confirmation<S, Rt>>>, AcceptTxError<S>>;
+
+pub struct QueuedTx<S: Spec, Rt: Runtime<S>> {
```
bkolad: Looks like now this can be made private

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,1475 @@
...
+    pub tx: FullyBakedTx,
+    pub tx_hash: TxHash,
+    pub original_tx_queue_id: u64,
+    pub result_sender:
+        oneshot::Sender<Result<TransactionReceiverResult<S, Rt>, SequencerStateUpdatorError>>,
+}
+
+pub struct AddressQueue<S: Spec, Rt: Runtime<S>> {
```
bkolad: Same looks like we use it only in this file

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,1475 @@
...
+    fn is_empty(&self) -> bool {
+        self.txs.is_empty()
+    }
+}
+
+/// This trait encodes the functionality that the TxNonceQueues need for handling submission of
+/// queued transactions.
+pub trait TxExecutionBackend<S: Spec, Rt: Runtime<S>> {
+    fn get_current_nonce_for_user(&self, credential_id: &CredentialId) -> u64;
+    fn execute_tx(
```
bkolad: Can we make this method async so the tupe signature becomes simpler?

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,1475 @@
...
+    /// Drain all ready transactions starting from expected_nonce until a gap or error
+    pub async fn drain_any_ready_transactions(
+        &self,
+        credential_id: CredentialId,
+        mut expected_nonce: u64,
+    ) {
+        loop {
+            // Lock and check if next tx is ready
+            let queued_tx = {
+                let mut queue = match self.queues.get_mut(&credential_id) {
```
bkolad: Let's rename it to queue_lock

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
theodorebugnet: I don't think we can because the Send bound needs to be explicitly added.

theodorebugnet: Actually I just realised using `#[async_trait]` might fix this

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,1475 @@
...
+                ))));
+            }
+
+            // Track execution
+            self.executed_txs.lock().unwrap().push((tx_hash, tx_nonce));
+
+            // Simulate successful execution
+            let (tx, rx) = oneshot::channel();
+            let baked_tx_clone = baked_tx.clone();
+            tokio::spawn(async move {
```
bkolad, line 1110: Why do we need to spawn here?

theodorebugnet: good point, we actually don't

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
bkolad: yeah was thinking about async_trait

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,1475 @@
...
+        loop {
+            // Lock and check if next tx is ready
+            let queued_tx = {
+                let mut queue = match self.queues.get_mut(&credential_id) {
+                    Some(queue) => queue,
+                    None => return, // No queue for this address
+                };
+
+                // Check if head has the expected nonce
+                let tx = loop {
```
bkolad: It would be ideal if we could separate the queue and async code more. 
Let's refactor this code block `queued_tx = {}` to a separate method.

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,1475 @@
...
+
+            // Check if we should continue draining
+            let should_continue = result.as_ref().is_ok_and(|r| r.is_ok());
+
+            // Send the raw result back to the original API handler task, so errors can be
+            // correctly propagated to the user. If the receiver is no longer listening, ignore.
+            let _ = queued_tx.result_sender.send(result);
+
+            // After the waiting task has been notified, clean up the queue if it's empty
+            if let Some(queue) = self.queues.get_mut(&credential_id) {
```
bkolad: Let's make it a method on self or queue. So the reference to queue internals doe not lek too much

crates/full-node/sov-sequencer/src/preferred/tx_nonce_queue.rs
```diff
@@ -0,0 +1,1475 @@
...
+        // be a race condition if a new transaction were to execute after this tx checked its nonce
+        // but before it added itself to the queue.
+        let locked_user_queue = self.lock_for_address(credential_id);
+        let current_nonce = self.submitter.get_current_nonce_for_user(&credential_id);
+        let max_accepted_nonce = current_nonce + self.maximum_future_nonce_delta;
+
+        match tx_nonce {
+            nonce if nonce == current_nonce => {
+                // Transaction has a valid nonce, execute immediately.
+                drop(locked_user_queue); // Release the queue lock as we won't be using it
```
bkolad: We could separate the queue code and async code in the following way:
In one method we do the pattern matching and lock managment
```
enum WaitForNonce{ <- maybe there is a better name for this enum
  Done,
  WaitForResult(..)
}

fn foo() =>  .. {
match tx_nonce {
    nonce if nonce == current_nonce => {
         drop(locked_user_queue);
         Ok(WaitForNonce::Done)
   }
   nonce if nonce > current_nonce && nonce < max_accepted_nonce => {
           let (result_sender, mut results_receiver) = oneshot::channel();
                Self::enqueue_and_unlock(
                    locked_user_queue,
                    result_sender,
                    baked_tx,
                    tx_hash,
                    tx_nonce,
                    original_tx_queue_id,
                );  
               Ok(WaitForNonce::WaitForResult(result_receiver))

  }
  _ => {Err(..)}

}
}

and later
match foo(){
   Done => { self
                    .submitter
                    .execute_t() ...
    }
  WaitForNonce => {
       loop {
                    tokio::select! {...}
       }
 }
}
"
```

This way, the queue locks are only handled in non-async methods, so we are sure they won't cross the await point by accident

crates/full-node/sov-sequencer/src/preferred/mod.rs
```diff
@@ -870,22 +889,45 @@ where
...
+
+        let (outer_res, is_nonce_based) = match uniqueness {
+            UniquenessData::Generation(_) => (
+                self.synchronized_state_updator
+                    .accept_tx_msg(&baked_tx, tx_hash, original_tx_queue_id, "accept_tx")
+                    .await,
+                false,
+            ),
+            UniquenessData::Nonce(tx_nonce) => (
+                self.tx_nonce_queues
```
bkolad, line 917: @preston-evans98 has recently noticed that for the evm `accept_tx` isn't run in "tokio::spawn"
https://sovereign-labs.slack.com/archives/C09MBHGG0GG/p1762830703696579?thread_ts=1762812477.973469&cid=C09MBHGG0GG

This means that the Future in `handle_new_tx` can be canceled, potentially leaving the queue in an inconsistent state. The fix is straightforward: at the beginning of accept_tx, we can simply spawn a tokio::task. Let’s check with @preston-evans98 to see if he’s already provided a fix elsewhere, so we don’t duplicate.

bkolad (approved):

crates/full-node/sov-sequencer/src/preferred/mod.rs
theodorebugnet: For future record, as discussed on slack I added a tokio spawn to ethereum, and this can be refactored into the sequencer in a follow up PR

PR #2013, opened 2025-10-31
Fix `eth_getLogsWithCursor` deserialization and response size limit

preston-evans98:
# Description

This PR makes and tests two small changes to `eth_getLogsWithCursor`. 
1. Adds a `#[serde(flatten)]` directive on the Filter in `FilterWithCursor`. This prevents us diverging from other chains by requiring `{ "cursor:" "foo", "filter": {"address": "…"}}` instead of the more standard flat structure with inline filter fields
2. Estimates the size of the current response in `getLogsWithCursor` and automatically paginates before reaching 1MB.

### DIFF
--- a/examples/demo-rollup/tests/evm/evm_logs.rs
+++ b/examples/demo-rollup/tests/evm/evm_logs.rs
@@ -282,7 +282,6 @@ async fn evm_test_get_logs_at_max_response_size() {
     let nb_of_logs_per_tx: u32 = 5000;
 
     let rollup_and_client = RollupAndClient::new(20_000).await;
-    let start_tx = rollup_and_client.get_tx_count().await as u32;
 
     rollup_and_client
         .produce_logs(nb_of_txs, nb_of_logs_per_tx, Some(3))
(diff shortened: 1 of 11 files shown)


examples/demo-rollup/tests/evm/evm_logs.rs
```diff
@@ -165,6 +165,67 @@ fn check_logs(filter: &Filter, logs: Vec<alloy_rpc_types_eth::Log>, expected_nb_
     }
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn evm_test_get_logs_with_cursor_and_filter() {
+    let max_log_limit = 9;
+    let nb_of_txs = 20;
+    let nb_of_logs_per_tx: u32 = 7;
+
+    let rollup_and_client = RollupAndClient::new(max_log_limit).await;
```
bkolad: Let's make `RESPONSE_SIZE_LIMIT` a configurable parameter, just like `max_log_limit.` This will allow us to test these conditions with much lower limits.

bkolad (approved):

examples/demo-rollup/tests/evm/evm_logs.rs
preston-evans98: Good idea. Done

PR #2025, opened 2025-11-05
Making `DaSerive::get_signer` return Option of the Address

citizen-stig:
# Description

This way it is possible to initialize DaService without any signer when only node is suppose to run.

This is not currently supported, but "NoOpSequencer" can be easily added to have "only-node" mode

## Testing
All Existing tests are passing

## Docs
Docstrings have been updated

### DIFF
--- a/crates/module-system/sov-modules-rollup-blueprint/src/native_only/mod.rs
+++ b/crates/module-system/sov-modules-rollup-blueprint/src/native_only/mod.rs
@@ -230,7 +230,9 @@ pub trait FullNodeBlueprint<M: ExecutionMode>: RollupBlueprint<M> {
                     background_handles,
                     proof_sender: sequencer,
                     api_ledger_db: api_ledger_db.clone(),
-                    da_address: da_service.get_signer().await.context("Full node with standard sequencer require DaService with signer support")?,
+                    da_address: da_service.get_signer().await.context(
+                        "Full node with standard sequencer require DaService with signer support",
+                    )?,
                 })
             }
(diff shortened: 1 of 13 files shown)


crates/module-system/sov-modules-rollup-blueprint/src/native_only/mod.rs
```diff
@@ -230,7 +230,7 @@ pub trait FullNodeBlueprint<M: ExecutionMode>: RollupBlueprint<M> {
                     background_handles,
                     proof_sender: sequencer,
                     api_ledger_db: api_ledger_db.clone(),
-                    da_address: da_service.get_signer().await,
+                    da_address: da_service.get_signer().await.context("XX")?,
```
bkolad: "XX": in context

crates/module-system/sov-modules-rollup-blueprint/src/native_only/mod.rs
```diff
@@ -260,7 +260,7 @@ pub trait FullNodeBlueprint<M: ExecutionMode>: RollupBlueprint<M> {
...
```
bkolad: Same

bkolad (approved):

ross-weir (approved):

crates/module-system/sov-modules-rollup-blueprint/src/native_only/mod.rs
citizen-stig: Good catch, forgot this one

PR #2028, opened 2025-11-05
feat(avail): handle all MultiAddress variants in transaction parsing

crStiv:
# Description

Implement handling for all MultiAddress enum variants (Address32, Raw, Index, Address20) in AvailBlobTransaction::new().

- Address32: direct conversion to AvailAddress
- Raw: conversion if 32 bytes, error otherwise
- Index: returns error (requires account lookup)
- Address20: returns error (cannot convert 20 bytes to 32 bytes)

Resolves TODO at line 54.


crStiv:

crStiv:
recheck

PR #2042, opened 2025-11-10
Fix fetch block reorg aware

### LINKED ISSUE #2037: `fetch_block_reorg_aware` blocks rollup on too low `polling_interval`
## Summary

Function `fetch_block_reorg_aware` is used by sov-stf-runner to react quickly on re-orgs

https://github.com/Sovereign-Labs/sovereign-sdk/blob/52505a9fdfae199d444fd0304a12dea8644db054/crates/full-node/sov-stf-runner/src/da_utils.rs#L16-L22

For example:

runner requests height 100. Just before network request fired, re-org happens and new head is 90. 
Runner can only speculate about behaviour of DaService. It can block until DA reaches that height. And resulted height will probably be different, so time is wasted.
Or it can keep failing and needlessly. 

Current implementation uses tokio select and if `DaService::get_block_at` hasn't returned on time, it checks if tip of DA has changed below requested height. If this happens it changes target height to be DA head.

If whole function does not succeed in `da_total_timeout`, it fails.

## Problem

If polling interval is set to be too small, so absolutely correct network call cannot complete on time, tokio::select will keep cancelling `DaService::get_block_at` before it has chance to actually complete.

## Goal

This function should be reworked, with all new information in mind:

* total_timeout should be kept in place, there's no Contract on how long DaService::get_block_at might take. DaService will do retries and sleeps, etc. But still runner should have its own way to emergency stop
* Mechanism of detecting changed head should probably happen in parallel and `get_block_at` should be cancelled only if target_height trully wen't below expected.

## Other notes

* Idea of removing manually set `da_polling_interval_ms`. This value should be set based on configuration of `da` section of the rollup and it places mental burden on rollup operator. We can improve that. Instead of custom value, we can add `DaService::approximate_block_time(&self) -> std::time::Duration` and then runner will use third of this time for polling. No configuration from rollup operator needed.

* Subscriptions: when rollup reaches tip of DA, it would be better to use head websocket subscription to fetch head. And if new head has non incremental height, it will trigger re-org

citizen-stig:
# Description

Changes behavior of `fetch_block_reorg_aware`.

Previous implementation could cancel future of `DaService::get_block_at` if it fails to complete under `da_polling_interval`, which might be not the case under normal operation, but with some network delays.

`fetch_block_reorg_aware` is completely rewritten, **so it is better to review file `crates/full-node/sov-stf-runner/src/da_utils.rs` as whole**

### How it works now:

* `DaService::get_block_at` is wrapped in tokio::timeout with parameter from `runner.total_da_timeout` to have upper limit of how long `get_block_at` can actually take.
* Concurrent future is started, which check if head rolled back below requested. This futures terminates on this case. So if head has rolled back below requested height before `DaService::get_block_at` is completed, it will try to fetch new head instead.

### Minor changes:

* Runner: Instead of polling, it subscribes to changes from `DaSyncState`.
* Correctly initialize `DaSyncState` with values from DB. Previously first instance of SyncStatus in channel was "Start", which is target and synced == 0. Which lead to unnecessary reorg like behaviour in runner. It was previously masked, but new approach uncovered it.
* MockDa: Fix in StorableMockDa**Service**, so it actually waits for new block even if reorg happend between waiting and fetching.

## Linked Issues
- Fixes #2037

## Testing

## Docs
No updates

### DIFF
--- a/examples/demo-rollup/tests/restart/mod.rs
+++ b/examples/demo-rollup/tests/restart/mod.rs
@@ -154,6 +154,7 @@ async fn start_stop_empty(
         (Level::WARN, "Skipping pruning of sequence number because it's already been pruned".to_string()),
         (Level::WARN, "The node is unsynced and doesn't know it. This probably means that you wiped the node DB and are resyncing.".to_string()),
         (Level::WARN, "Metics have been initialized outside of the rollup blueprint, some measurements can be lost on shutdown".to_string()),
+        (Level::WARN, "DA head rolled back below requested height, retrying with new head".to_string()),
     ];
 
     let mut recorded_errors_warnings =
@@ -549,6 +550,7 @@ async fn test_start_prover_manual() -> anyhow::Result<()> {
             Level::WARN,
             "Metics have been initialized outside of the rollup blueprint, some measurements can be lost on shutdown".to_string(),
         ),
+        (Level::WARN, "DA head rolled back below requested height, retrying with new head".to_string())
     ];
     recorded_errors_warnings.retain(|e| !known.contains(e));
     // We could've checked `.is_empty`, but in case of failure, we will see errors immediately.
--- a/examples/demo-rollup/tests/restart/mod.rs
+++ b/examples/demo-rollup/tests/restart/mod.rs
@@ -15,8 +15,8 @@ use sov_bank::{config_gas_token_id, Coins};
 use sov_db::storage_manager::NomtStorageManager;
 use sov_demo_rollup::{mock_da_risc0_host_args, MockDemoRollup};
 use sov_mock_da::storable::layer::StorableMockDaLayer;
-use sov_mock_da::BlockProducingConfig;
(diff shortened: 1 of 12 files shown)


examples/demo-rollup/tests/restart/mod.rs
```diff
@@ -154,6 +154,7 @@ async fn start_stop_empty(
         (Level::WARN, "Skipping pruning of sequence number because it's already been pruned".to_string()),
         (Level::WARN, "The node is unsynced and doesn't know it. This probably means that you wiped the node DB and are resyncing.".to_string()),
         (Level::WARN, "Metics have been initialized outside of the rollup blueprint, some measurements can be lost on shutdown".to_string()),
+        (Level::WARN, "DA head rolled back below requested height, retrying with new head".to_string()),
```
citizen-stig: Where  these are coming from??? Investigate

bkolad (approved):

PR #2052, opened 2025-11-11
Schema based rust web3

ross-weir:
# Description

Adds a `schema` based implementation of our rust `web3` crate. This version will be used for language bindings and if customers prefer to do runtime type serialization instead of comptime/generics. An advantage of runtime schema is that customers won't need to make their `Spec` / `CryptoSpec` / `RuntimeCall` generic types available.

I can see both approaches potentially being useful depending on the situation so have chosen to keep the existing `rust` based implementation for now.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/web3/README.md
```diff
@@ -0,0 +1,116 @@
...
+let unsigned_tx = TransactionBuilder::new(call)
+    .chain_id(1234)
+    .max_fee(100000u128)
+    .build()?;
+
+// Serialize for signing
+let bytes_to_sign = unsigned_tx.bytes_for_signing(&serializer)?;
+
+// Create signed transaction
+let signed_tx = unsigned_tx.to_signed(pub_key_bytes, signature_bytes);
```
theodorebugnet, line 102: Should `signature_bytes` be `bytes_to_sign` here?

crates/web3/src/schema.rs
```diff
@@ -0,0 +1,565 @@
...
+    JsonSerialization(#[from] serde_json::Error),
+    /// The provided schema is invalid or malformed.
+    #[error("Invalid schema: {0}")]
+    InvalidSchema(#[source] serde_json::Error),
+    #[error("Failed to calculate chain hash: {0}")]
+    ChainHash(#[source] sov_universal_wallet::schema::SchemaError),
+    /// Error occurred during HTTP request to fetch schema from URL.
+    #[error("HTTP request error: {0}")]
+    HttpRequest(#[from] reqwest::Error),
+    #[error("schema object from http response was invalid")]
```
theodorebugnet:
```suggestion
    #[error("HTTP schema request returned an invalid response")]
```
"Schema object invalid" is a bit confusing when there's an `InvalidSchema` error right above, for actually invalid schema objects

crates/web3/src/schema.rs
```diff
@@ -0,0 +1,565 @@
...
+/// # Returns
+///
+/// Returns `UniquenessData::Generation` with the current Unix timestamp
+///
+/// # Panics
+///
+/// Panics if the system clock is set to a time before the Unix epoch.
+pub fn default_uniqueness() -> Result<UniquenessData, TransactionBuilderError> {
+    use std::time::{SystemTime, UNIX_EPOCH};
+    let now = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
```
theodorebugnet: As per recent discussions I think we decided this should instead be in millis by default. (If this is updated, doc comment above should be too.)

theodorebugnet (approved):
The default uniqueness is important, other comment are non-blocking minor nits.
Approving since everything else looks good and I think it's good to merge if the uniqueness is adjusted to millis (unless there's a reason not to)

crates/web3/src/schema.rs
ross-weir: Thanks! Updated to millis

crates/web3/README.md
ross-weir: updated

PR #2043, opened 2025-11-10
Refactors WebSocket RPC handling code for improved readability and organization.

LogvinovLeon:
# Description

## Changes

- **Module structure**: Extract WebSocket handling from `http.rs` into dedicated `http/ws.rs` module
- **Variable naming**: `sender/receiver` → `ws_writer/ws_reader`, `socket_requests/socket_responses` → `msg_tx/msg_rx`
- **Function naming**: `handle_socket_read/write` → `socket_reader_task/socket_writer_task`
- **Code cleanup**: Extract `to_ws_message` helper, add shutdown handling to subscription forwarder

No functional changes - purely a refactoring for better code clarity.

----

- [ ] I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.
- [ ] I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. (Are all new dependencies necessary? Is any module dependency leaked into the full-node (hint: it shouldn't)?)

## Linked Issues
- Fixes # (issue, if applicable)
- Related to # (issue) 

## Testing
Describe how these changes were tested. If you've added new features, have you added unit tests?

## Docs
Describe where this code is documented. If it changes a documented interface, have the docs been updated?

Rendered docs are available at `https://sovlabs-ci-rustdoc-artifacts.us-east-1.amazonaws.com/<BRANCH_NAME>/index.html`

PR #2065, opened 2025-11-12
WS Speedup. Replaces: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2051

LogvinovLeon:
# Description

After https://github.com/Sovereign-Labs/sovereign-sdk/pull/2043 got merged - https://github.com/Sovereign-Labs/sovereign-sdk/pull/2051 got conflicts. This PR fixes those.

Also - uses `try_join` to stream and produce logs in parallel resulting in no `channel lagged` messages (even after milling of logs).
In the version before - we were getting a huge `channel lagged: {a lot}` message at the beginning.
This allows us treating lags as errors and panicking on them as lags are unacceptable from the perspective of the user using this API to get all the logs.

Also - increased the capacity of the internal channels in jsonrpsee from 1 to 128 which brought another 2x speedup. But more importantly - subscription now finishes in a time very close that it takes to produce logs.

For example: Streamed 10M logs in 62.3s. Produced in 61.4s. (6 micros for a log). 0 lagged

-----

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-stf-runner/src/http/ws.rs
+++ b/crates/full-node/sov-stf-runner/src/http/ws.rs
@@ -1,12 +1,13 @@
 use axum::extract::ws::{Message, WebSocket};
 use axum::extract::WebSocketUpgrade;
 use axum::response::IntoResponse;
+use axum::Error;
 use futures::stream::{SplitSink, SplitStream};
 use futures_util::sink::SinkExt;
 use futures_util::stream::StreamExt;
 use jsonrpsee::core::JsonRawValue;
 use jsonrpsee::RpcModule;
-use sov_rollup_interface::consume_until_shutdown;
+use sov_rollup_interface::{consume_many_until_shutdown, consume_until_shutdown};
 use tokio::spawn;
 use tokio::sync::{mpsc, watch};
 use tracing::{debug, error, trace};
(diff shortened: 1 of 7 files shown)


crates/full-node/sov-stf-runner/src/http/ws.rs
```diff
@@ -82,13 +83,13 @@ async fn socket_reader_task(
...
-                    handle_rpc_message(&text, &msg_tx, &rpc_methods, false, &shutdown_receiver).await;
+                    spawn(handle_rpc_message(text, msg_tx.clone(), rpc_methods.clone(), false, shutdown_receiver.clone()));
                 }
                 Message::Binary(data) => {
                     // Parse binary frame as UTF-8 JSON-RPC request
-                    match std::str::from_utf8(&data) {
+                    match String::from_utf8(data) {
                         Ok(text) => {
-                            handle_rpc_message(text, &msg_tx, &rpc_methods, true, &shutdown_receiver).await;
+                            spawn(handle_rpc_message(text, msg_tx.clone(), rpc_methods.clone(), true, shutdown_receiver.clone()));
```
bkolad, line 95: It appears that previously, messages were handled in order, but after this change, they can be processed out of order. For example, if two messages are executed in quick succession, they may spawn simultaneously, and the second one could finish before the first. Is that ok?

crates/rollup-interface/src/node/mod.rs
```diff
@@ -91,3 +91,28 @@ macro_rules! consume_until_shutdown {
         }
     };
 }
+
+/// Same as consume_until_shutdown but consumes elements in batches
+#[macro_export]
+macro_rules! consume_many_until_shutdown {
```
bkolad, line 97: It is used only in one place, can we avoid another macro and just use the code directly?  Recently we had many problems with `select!` macro (we will discuss on the daily) so it would be good if ussage is explicit.

crates/full-node/sov-stf-runner/src/http/ws.rs
LogvinovLeon: JSONRPC explicitly allows us to do that:
https://www.jsonrpc.org/specification
Section 6. There is an id field that is used to match responses to requests.

bkolad (approved):

crates/rollup-interface/src/node/mod.rs
LogvinovLeon: I think it makes the code more readable by articulating - what is common plumbing here and what is logic.
Handling shutdown signal and printing log is not that important here and most of the users reading this piece wouldn't care.
Screenshot 2025-11-12 at 14 54 43
Screenshot 2025-11-12 at 14 54 12

bkolad: Most people reading the code already know what `select!` does since it's a common macro. However, to understand what `consume_many_until_shutdown` does, the reader has to jump to its definition — and it turns out to be just a macro wrapping another macro. Yesterday, we had a particularly tricky debugging session involving `select!`, and in the past, it has also caused us quite a few problems. Making it explicit we make it easier to flag all the places where we use this macro.

LogvinovLeon: If I have a function: `average(elements)` then the same argument can be made: Most people already know what a `for loop` does and here they need to jump to the function definition to know what it does.

Still - when an abstraction is a good one and a well named one - the value of it is that you don't need to jump to it as in 90% of the cases you're not working on abstraction itself but on the logic you plug within it.

Therefore the question here is: When a user reads: `consume_many_until_shutdown` - is it kinda clear what it does or do they need to read the code?

For me - select is a control flow structure and I don't expect to ever be grepping for all the cases when I use `select!`. It's like grepping for all the `if` or `continue`. At the same time - I'm curious - what was your use-case.

bkolad: I see it the opposite way. The situation would be if there were already an `average` function, and now we’re introducing a wrapper called `slightly_better_averag` e—perhaps with one argument changed—to simplify some usage. That would only make sense if it’s used in many places, not just once.

Unfortunately, `select!` isn’t just another control flow construct—it can be quite a footgun. For example, if we await two futures and one completes, the other is automatically canceled. That means we have to carefully review the code to ensure all futures used within `select!` are cancel-safe. We’ve run into issues with this before. There are also other problems, such as incorrect assumptions about the order in which the select! macro notifies its branches (this is the bug from yesterday).

PR #2076, opened 2025-11-13
Port `6ce5a4c`  to dev branch

bkolad:
# Description

This PR is a port of 6ce5a4c with the following changes:

Add `TimingOracle` [05ed7ed](https://github.com/Sovereign-Labs/sovereign-sdk/pull/2076/commits/05ed7edd5b6c1c56baa88dc4a6b67669a7702e5e):
1. Only the part related to TimingOracle was ported, the EVM part will be ported in a separate PR
2. Reorganized the code, separating logic into dedicated files etc...
3. Added a test for `TimingOracle` functionality.
4. Made `TimingOracleConfig` optional—if it’s not specified in the rollup config, the timing oracle is disabled.

Add `is_responsible_for_gating_admins` 
1. Brings the original changes [9c6e29c](https://github.com/Sovereign-Labs/sovereign-sdk/pull/2076/commits/9c6e29cdd4fe3daa809ccf8800717a4a4db64626)
2. Removes `is_master` flag from `do_batch_start` [8e3b8a4](https://github.com/Sovereign-Labs/sovereign-sdk/pull/2076/commits/8e3b8a403eb90d637a2b4476981b4ebabc616bbd)

Question: Why was this added in the first place?  

## Linked Issues
- Related to 6ce5a4c


crates/full-node/sov-sequencer/src/preferred/timestamp.rs
```diff
@@ -0,0 +1,153 @@
...
+        let time_oracle_config = maybe_time_oracle_config?;
+        let priv_key = || -> anyhow::Result<_> {
+            match &time_oracle_config.private_key_hex {
+                Some(key_str) => {
+                    let bytes = HexString::from_str(key_str).map_err(|_| anyhow::anyhow!("Invalid oracle private key hex - could not parse as hex. Check your preferred sequencer config file."))?.0;
+                    let key = <S::CryptoSpec as CryptoSpec>::PrivateKey::try_from(bytes).map_err(|_| anyhow::anyhow!("Invalid oracle private key hex - invalid private key bytes. Check your preferred sequencer config file."))?;
+                    Ok(key)
+                }
+                None => {
+                    let key = <S::CryptoSpec as CryptoSpec>::PrivateKey::generate();
```
bkolad, line 42: This will only work if the rollup has a `paymaster` configured; otherwise, the new address won’t have any funds. I added a warning about this. If we remove the ephemeral key functionality the only downside is that we’d need to put the private key in the config right?

examples/demo-rollup/tests/evm/evm_oracle.rs
```diff
@@ -0,0 +1,48 @@
...
+        max_fee: 1_000_000,
+        interval_millis: 50,
+        private_key_hex: Some(private_key_hex),
+    };
+
+    start_node(config, 0, Some(EVM_EXTENSION), Some(time_stamp_config)).await
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn evm_test_oracle_timestamp() {
```
bkolad, line 32: In the next PR, we will extend this test to check EVM logs timestamps.

crates/full-node/sov-sequencer/src/preferred/timestamp.rs
preston-evans98: Yeah, correct

crates/utils/sov-rpc-eth-types/src/lib.rs
```diff
@@ -10,3 +10,12 @@ pub use filter_with_cursor::{FilterWithCursor, LogsWithMaybeCursor};
...
 pub use rpc_invalid_transaction_error::RpcInvalidTransactionError;
 pub use to_rpc_error::ToRpcError;
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq, Eq, Hash)]
+pub struct LogWithExecutionTimestamp<L = alloy_rpc_types::Log> {
+    #[serde(flatten)]
+    pub log: L,
+    #[serde(rename = "timeExecutedMs")]
+    #[serde(with = "alloy_serde::quantity")]
+    pub time_executed_ms: u64,
```
preston-evans98, line 20: Looks like one EVM change snuck in - is that intentional?

preston-evans98 (commented):
Thanks for the cleanups, this is much improved over my draft. Looks good to me!

preston-evans98 (approved):

preston-evans98:
> Question: Why was this added in the first place?

Without this new config `is_responsible_for_gating_admins`, we fail on the second startup with timing oracle because...
- The first run generates one ephemeral key and uses adds it to the admins list in memory
- The second run generates a *different* ephemeral key - so any soft-confirmed txs from the previous oracle fail to be reapplied.

crates/utils/sov-rpc-eth-types/src/lib.rs
bkolad: Yeah, I will need it anyway in my next PR, so I will keep it.

PR #2085, opened 2025-11-17
Add hyperlane solana register module

ross-weir:
# Description

This is a module written for Zeta that allows for the following:

When a user registers on bullet an embeded wallet is automatically created for them, the embedded wallet is used internally and the user only sees their Solana address. This module enables the embedded wallet to be linked to the users solana address on the rollup so the solana address is used normally with the embedded wallet remaining opaque.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/module-implementations/extern/hyperlane-solana-register/Cargo.toml
```diff
@@ -0,0 +1,50 @@
+[package]
+name = "hyperlane-register-module"
+version = "0.3.0"
+edition = "2021"
+license = "MIT OR Apache-2.0"
+authors = ["Sovereign Labs <info@sovereign.xyz>"]
+publish = false
+rust-version = "1.81"
```
citizen-stig: Could we use those from the workspace?

crates/module-system/module-implementations/extern/hyperlane-solana-register/Cargo.toml
```diff
@@ -0,0 +1,50 @@
+[package]
+name = "hyperlane-register-module"
+version = "0.3.0"
+edition = "2021"
+license = "MIT OR Apache-2.0"
```
citizen-stig: And I think this one is important too.

crates/module-system/module-implementations/extern/hyperlane-solana-register/Cargo.toml
```diff
@@ -0,0 +1,50 @@
+[package]
+name = "hyperlane-register-module"
```
citizen-stig: Is this module going to be published? Maybe we can prefix it with `sov-` as all others?

crates/module-system/module-implementations/extern/hyperlane-solana-register/configs.json
```diff
@@ -0,0 +1,10 @@
+{
+  "mainnet": {
+    "chainId": 1399811149,
+    "programId": "4KdqVph6eMnS2omUBLBH2u4G6wwqxG5hzesZpsFcSWod"
+  },
+  "testnet": {
+    "chainId": 1399811150,
+    "programId": "4KdqVph6eMnS2omUBLBH2u4G6wwqxG5hzesZpsFcSWod"
```
citizen-stig: Why those cannot be genesis params? Is it possible that those values change on hyperlane side? This mean it will require publishing a new version of the crate, rather than updating genesis config.

crates/module-system/module-implementations/extern/hyperlane-solana-register/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+                registered_address: resolved_address.to_string(),
+            })
+        } else {
+            Ok(())
+        }
+    }
+}
+
+pub mod config {
+    include!(concat!(env!("OUT_DIR"), "/config.rs"));
```
citizen-stig, line 183: I wonder, why do we embed config this way?

As far as I understand, we have 4 ways to provide parameters for module:

1. regular genesis.json
2. module execution config
3. constants.toml
4. This one by emebedding data from build.rs

Why do we choose 4th for this module? Why other cannot be used?

ross-weir: 1 - it's not something that needs to be part of the rollup state, just needs to be constant in the module
2 - it needs to be accessed on chain
3 - we could but I don't like the idea of polluting `constants.toml`, particularly because this value will only be relevant to zeta. If there was a way for consants.toml values to be omitted or defaults provided, etc I would be more for this
4. Because of the above

I'm open to suggestions on better ways to provide the config or if you disagree with the above

citizen-stig: > it's not something that needs to be part of the rollup state, just needs to be constant in the module.

Does it also means that change in this file is hard-fork?

citizen-stig (approved):

crates/module-system/module-implementations/extern/hyperlane-solana-register/src/lib.rs
citizen-stig: This is non-blocking for the PR, but we propbably need to have a guideline on how to expose configuration to modules.

Or maybe start working on optional values in constants.toml

PR #2100, opened 2025-11-21
Port fixses for the evm part 2

bkolad:
# Description

Some fixses needed by the evm.

### DIFF
--- a/crates/full-node/full-node-configs/src/sequencer.rs
+++ b/crates/full-node/full-node-configs/src/sequencer.rs
@@ -247,7 +247,7 @@ pub struct TimingOracleConfig {
     pub priority_fee_percentage: u8,
 
     /// The maximum fee that the sequencer will pay for the timestamp oracle update tx.
-    pub max_fee: u128,
+    pub max_fee: u64,
 
     /// The interval in milliseconds at which the timestamp oracle update tx is submitted.
     pub interval_millis: u64,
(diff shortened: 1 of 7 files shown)


crates/full-node/full-node-configs/src/sequencer.rs
```diff
@@ -247,7 +247,7 @@ pub struct TimingOracleConfig {
     pub priority_fee_percentage: u8,
 
     /// The maximum fee that the sequencer will pay for the timestamp oracle update tx.
-    pub max_fee: u128,
+    pub max_fee: u64,
```
bkolad, line 250: Had to change it to u64 because we store this config in a TOML file, and `serde_toml` doesn’t support `u128`.

LogvinovLeon (approved):

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -374,6 +374,11 @@ where
         match pending_or_block_nr {
             PendingOrBlock::Pending => Ok(MaybeArchivalState::Current(state)),
             PendingOrBlock::Number(number) => {
+                if number == state.rollup_height_to_access().get()
```
bkolad, line 377: Looks like this breaks test_state_at_invalid_depth. Should we remove test? @LogvinovLeon

PR #2103, opened 2025-11-23
Fix silent error handling in indexer event fetching

maradini77:
The indexer was silently swallowing all errors when fetching events from the rollup API. Any failure (HTTP 500, serialization errors, network issues) would return an empty array without logging or triggering the recovery mechanism. This created a false sense of normal operation while events were actually being skipped.


maradini77:

PR #2105, opened 2025-11-24
Add support for pinning state in RAM

preston-evans98:
# Description

This PR introduces a new `PinnedCache` type. This cache maps key `buckets` to maps from key->value. When a bucket is present, any key which matches the bucket is *guaranteed* to be in the cache if it exists. This means that we can return `None` if a bucket exists for the item but no value is present.

This PR works by adding a new method to the `Runtime` specifying which buckets of k/v pairs should be pinned. A recent [Rockbound PR](https://github.com/Sovereign-Labs/rockbound/pull/39) added support for iterating over keys to enable this. 

At node startup and each time the sequencer creates a fresh executor (startup, resync, recovery), we iterate over the DB to load all needed values into the pinned cache.


crates/module-system/sov-state/src/cache.rs
```diff
@@ -406,10 +442,28 @@ impl<N: ProvableCompileTimeNamespace> ProvableStorageCache<N> {
...
+                        if let MaybePresentValue::Present(value) = maybe_value {
+                            value
+                        } else {
+                            storage.get_leaf::<N>(key, witness)
+                        }
+                    }
+                    None => {
+                        // Note: we currently skip adding the read to cache if the value is pinned (why cache it twice).
+                        // TODO: Decide if this is the right behavior.
+                        if let MaybePresentValue::Present(value) = self.check_pinned_cache(key) {
```
bkolad: We execute the same logic for both `Some(uncomitted_changes) and None`. We can just move the logic just before `match uncomitted_changes (line 444)` and return in a single place.

crates/module-system/sov-state/src/cache.rs
```diff
@@ -467,14 +521,40 @@ impl<N: ProvableCompileTimeNamespace> ProvableStorageCache<N> {
...
                 Ok::<_, Infallible>(match uncomitted_changes {
                     // NATIVE only: we might have some intermediate state that isn't yet in storage (this could be state from an optimistic execution, or uncomitted state from the sequencer).
                     // If so, check that state first and fall back to storage.
-                    Some(uncomitted_changes) => uncomitted_changes
-                        .get(N::NAMESPACE, key)
-                        .or_else(|| storage.get::<N>(key, witness)),
-                    None => storage.get::<N>(key, witness),
+                    Some(uncomitted_changes) => {
+                        if let MaybePresentValue::Present(value) =
+                            uncomitted_changes.get(N::NAMESPACE, key)
```
bkolad: In `get_size_or_fetch`, we check `check_pinned_cache_static` first and then `uncommitted_changes`. Here, we do the opposite. Shouldn't the order be the same in both cases?

crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
```diff
@@ -51,6 +52,9 @@ pub struct NomtChangeSet {
     pub state: StateFinishedSession,
     pub historical_state: StateChanges,
     pub accessory: SchemaBatch,
+    /// Use type erasure because the `pinned_cache` type is defined in sov-state, which depends on this crate.
+    /// No type other than `PinnedCache` makes sense here.
```
bkolad, line 56: What about bringing `PinnedCache` to `storage_manager`?
```
pub struct PinnedCache {
    item_caches: BTreeMap<Prefix, StateItemCache>,
}
```
`Prefix` is now a fairly simple type defined in a single file, so we could move it here and re-export it.
This was we could avoid donwcasting

crates/module-system/sov-state/src/storage.rs
```diff
@@ -775,6 +998,22 @@ pub trait NativeStorage: Storage {
...
+
+    /// Takes the pinned cache if one is present in this storage. See [`PinnedCache`] for more details.
+    ///
+    /// In the full node only, the pinned cache is passed from block to block through the storage manager.
+    /// On saving the previous block, the storage manager takes its pinned block; then when it creates the storage for the *first* child block,
+    /// that storage is passed along with it.
+    /// If a block has multiple children (i.e. the chain has a fork at some height), the pinned cache is only passed to the first child block to be created; other children have to rebuild it from db.
+    ///
+    /// Note that the sequencer passes the pinned cache directly between executors without this hack, so this method is only used in the full node.
+    fn take_pinned_cache(&mut self) -> Option<PinnedCache>;
```
bkolad: There’s also a method with the same name in `ProvableStorageCache`, which the sequencer uses. Let’s rename one of them to avoid confusion and improve the grepping experience.

crates/module-system/sov-state/src/cache.rs
preston-evans98: Fixed

crates/module-system/sov-state/src/cache.rs
preston-evans98: It shouldn't matter since these two should always be in sync - but it'll be faster to check the pinned cache first and it's easier to reason about if they match. Fixed!

bkolad (approved):

crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
preston-evans98: Took a stab at this, but unfortunately pinned cache depends on pretty much the whole sov-state crate transitively via `NativeStorage` and `SlotKey/SlotValue`. 

Chatted offline - we'll leave this for now.

crates/module-system/sov-state/src/storage.rs
preston-evans98: Renamed!

Issue #2104, opened 2025-11-23
[Tracker] RAM Pinning


preston-evans98:
## Description
Allows state items to be pinned in RAM, ensuring that we never touch disk when searching for their items.

### Feature Implementation
- [x] Implement db changes to support RAM pinning
- [x] Add new PinnedCache type
- [x] Add support for reading from the pinned cache in `ProvableStorageCache` instead of falling to disk
- [x] Add support for instantiating the pinned cache in the Runtime
- [x] Instantiate and pass PinnedCache correctly in sequencer
- [x] Instantiate and pass PinnedCache correctly in Node

### Feature Testing
- [x] Test RAM Pinning in node/sequencer under normal operation
- [x] Test RAM pinning still works after recovery
- [x] Test RAM pinning still works after resync

### Merging
- [x] Clean up and PR rockbound changes https://github.com/Sovereign-Labs/rockbound/pull/39
- [x] Rebase on latest dev https://github.com/Sovereign-Labs/sovereign-sdk/pull/2105

### Configuration
- [x] Rebase on top of #2102 to allow configuration of pinned buckets at startup
- [x] Add `privileged` addresses to EVM whose contracts are automatically pinned
   - [x] Add privileged addresses to base runtime config
   - [x] Add ability to modify and save dynamic runtime config between runs so that we can track the addresses created by privileged users and pin them automatically.

### Configuration Testing
- [x] Integration test RAM pinning with EVM

### Optimization
- [ ] Avoid re-creating the pinned cache on `resync` events from the sequencer

### Unfinished Items from Initial PRs
- Cleanup  Storage API: Currently the pinned cache is passed as state through the `Storage` in the full node. SInce it's not referentially transparent, its hard to reason about
- Cleanup and more obviously document handling of the cache in the sequencer.

PR #2102, opened 2025-11-21
Add "execution time" config for modules

ross-weir:
# Description

Adds a mechanism to pass runtime configuration to modules to perform any setup for off-chain operations (such as metric gathering).

Open to bikeshedding on the naming of `ExecutionConfig` - `RuntimeConfig` seemed better to me but would be ambiguous when used in the same context as the rollup Runtime. 

- [ ] I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.
- [ ] I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. (Are all new dependencies necessary? Is any module dependency leaked into the full-node (hint: it shouldn't)?)

## Linked Issues
- Fixes # (issue, if applicable)
- Related to # (issue) 

## Testing
Describe how these changes were tested. If you've added new features, have you added unit tests?

## Docs
Describe where this code is documented. If it changes a documented interface, have the docs been updated?

Rendered docs are available at `https://sovlabs-ci-rustdoc-artifacts.us-east-1.amazonaws.com/<BRANCH_NAME>/index.html`

PR #2124, opened 2025-11-25
Add optional persistent contract list for RAM pinning to the EVM

preston-evans98:
# Description

This PR uses the mechanism introduced in #2102 to allow configuring a list of privileged deployers and contract addresses in the EVM module. This mechanism is opt-in at the crate Runtime; if not enabled, the EVM behaves as always. 

When this mechanism is enabled, the EVM checks each transaction's sender and, if it's a privileged sender, adds any created contracts to its in-memory list and to a config file whose path is provided by the user. Any contracts found in the config are pinned in RAM.

## Linked Issues
- Related to #2104 

## Testing

## Docs


crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -102,8 +104,22 @@ where
...
+            &state_changes,
+            &tx.signer,
+            &mut db,
+        ) {
+            Ok(new_pinned_contracts) => new_pinned_contracts,
+            Err(err) => {
+                tracing::debug!(error = ?err, "Ran out of gas while getting checking pinned contract list updates");
+                anyhow::bail!("EVM transaction error: {:?}", err);
+            }
+        };
```
LogvinovLeon:
```suggestion
        #[cfg(feature = "native")]
        let new_pinned_contracts = get_pinned_contract_list_updates(
            &state_changes,
            &tx.signer,
            &mut db,
        )
        .inspect_err(|err| tracing::debug!(error = ?err, "Ran out of gas while getting checking pinned contract list updates"))
        .map_err(|err| anyhow::anyhow!("EVM transaction error: {err:?}"))?;
```

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -333,6 +358,114 @@ pub(crate) fn verify_contract_creation_allowlist(
...
+}
+
+#[cfg(feature = "native")]
+impl<S: Spec> Evm<S> {
+    pub(crate) fn update_pinned_contract_list(
+        &self,
+        new_pinned_contracts: &Vec<Address>,
+        state: &mut impl TxState<S>,
+    ) {
+        use crate::execution_config::EVM_EXECUTION_CONFIG;
```
LogvinovLeon, line 410: Why do we need to import inside the function here? Why not just feature-gate the import like we do in other places?

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -333,6 +358,114 @@ pub(crate) fn verify_contract_creation_allowlist(
...
+        }
+    }
+    Ok(new_pinned_contracts)
+}
+
+#[cfg(feature = "native")]
+impl<S: Spec> Evm<S> {
+    pub(crate) fn update_pinned_contract_list(
+        &self,
+        new_pinned_contracts: &Vec<Address>,
```
LogvinovLeon:
```suggestion
        new_pinned_contracts: &[Address],
```

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -333,6 +358,114 @@ pub(crate) fn verify_contract_creation_allowlist(
...
+                // Only pin addresses which didn't exist previously or had their code hash go from empty to non-empty.
+                if db
+                    .basic(*address)?
+                    .map(|acc| acc.code_hash == KECCAK_EMPTY)
+                    .unwrap_or(true)
+                {
+                    new_pinned_contracts.push(*address);
+                }
+            }
+        }
```
LogvinovLeon:
```suggestion
        let is_contract = account.info.code.is_some_and(|code| !code.is_empty());
        if is_contract {
            let was_contract = db
                .basic(*address)?
                .map(|acc| acc.code_hash == KECCAK_EMPTY)
                .unwrap_or(true);
            if was_contract {
                new_pinned_contracts.push(*address);
            }
        }
```

crates/module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -217,6 +217,113 @@ where
...
+                    .map(|(address, limit)| {
+                        let bucket_id = self.get_bucket_id_for_address(address);
+                        (bucket_id, *limit)
+                    })
+                    .collect(),
+            )
+        }
+    }
+
+    #[cfg(feature = "native")]
```
LogvinovLeon: This whole module is native only so no need here

crates/module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -217,6 +217,113 @@ where
     }
 }
 
+#[cfg(feature = "native")]
+/// Provides configuration of native execution; currently, this modules only includes configuration for pinning contract storage to RAM.
+pub mod execution_config {
```
LogvinovLeon: Let's put it into a separate file

crates/module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -217,6 +217,113 @@ where
...
+    use sov_modules_api::ModuleExecutionConfig;
+    use sov_state::pinned_cache::BucketId;
+    use std::collections::BTreeMap;
+    use std::sync::OnceLock;
+    use std::sync::RwLock;
+
+    use super::*;
+
+    /// Global storage for the EVM execution configuration.
+    pub static EVM_EXECUTION_CONFIG: OnceLock<RwLock<EvmExecutionConfig>> = OnceLock::new();
```
LogvinovLeon: It's within the `execution_config` context so no need to repeat the subject. Let's import it like this:
`execution_config::CONFIG`. EVM is also unnecessary as we're within sov-evm

crates/module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -217,6 +217,113 @@ where
...
+    #[derive(Clone, Debug)]
+    pub struct EvmExecutionConfig {
+        /// The contents of the execution configuration.
+        pub contents: EvmExecutionConfigContents,
+        /// The location of the execution configuration file on disk. The file will get updated during execution.
+        pub location: std::path::PathBuf,
+    }
+
+    /// The default size limit for any new pinned bucket.
+    pub const fn default_bucket_size_limit() -> usize {
```
LogvinovLeon: why a function and not a const?

crates/module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -217,6 +217,113 @@ where
...
+        100 * 1024 * 1024 // 100MB
+    }
+
+    impl<S: Spec> ExecutionInit for Evm<S> {
+        type Config = EvmExecutionConfig;
+        // Do nothing; the configure function handles everything we need.
+        fn init(_config: &Self::Config) -> Result<(), Box<dyn std::error::Error>> {
```
LogvinovLeon: Should we have default impl for init in the trait definition so that we don't need to write in in all the modules that don't use it?

crates/module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -217,6 +217,113 @@ where
...
+    }
+
+    #[cfg(feature = "native")]
+    impl ModuleExecutionConfig for EvmExecutionConfig {
+        type Input = std::path::PathBuf;
+        fn configure(input: &Self::Input) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+            if !std::fs::exists(input)? {
+                std::fs::write(
+                    input,
+                    serde_json::to_string_pretty(&EvmExecutionConfigContents {
```
LogvinovLeon: Let's implement Default

crates/module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -217,6 +217,113 @@ where
...
+                        default_bucket_size_limit: default_bucket_size_limit(),
+                        privileged_deployer_addresses: vec![],
+                        known_contracts_and_limits: BTreeMap::new(),
+                    })?,
+                )?;
+            }
+            let file = std::fs::read(input)?;
+            let config: EvmExecutionConfigContents = serde_json::from_slice(&file)?;
```
LogvinovLeon: Inference is smart enough to guess this type

crates/adapters/celestia/src/config.rs
```diff
@@ -125,7 +125,7 @@ impl CelestiaConfig {
             signer_private_key: None,
             request_timeout_secs: default_request_timeout_seconds(),
             safe_lead_time_ms: default_safe_lead_time_ms(),
-            tx_priority: None,
+            tx_priority: TxPriority::High,
```
preston-evans98: Doctests fail to build without this change. Not sure why this was working before, but this is the correct default value.

crates/module-system/sov-state/src/pinned_cache.rs
```diff
@@ -86,6 +86,12 @@ impl BucketId {
     }
 }
 
+impl std::fmt::Display for BucketId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
```
LogvinovLeon:
```suggestion
        self.0.fmt(f)
```

examples/demo-rollup/stf/src/runtime.rs
```diff
@@ -203,6 +207,21 @@ where
...
+    fn populate_pinned_cache(storage: &S::Storage) -> Option<PinnedCache> {
+        let buckets_and_limits =
+            sov_evm::Evm::<S>::default().get_pinned_cache_buckets_and_limits()?;
+        let mut pinned_cache = PinnedCache::default();
+        for (bucket_id, limit) in buckets_and_limits {
+            if let Err(e) =
+                pinned_cache.try_load_bucket_if_absent(bucket_id.clone(), storage, limit)
+            {
+                tracing::warn!(bucket_id = ?bucket_id, limit = ?limit, error = ?e, "EVM Failed to load bucket into pinned cache");
+            }
```
LogvinovLeon:
```suggestion
            let _ = pinned_cache.try_load_bucket_if_absent(bucket_id.clone(), storage, limit).inspect_err(|e|tracing::warn!(bucket_id = ?bucket_id, limit = ?limit, error = ?e, "EVM Failed to load bucket into pinned cache"));
```

LogvinovLeon: Also - are we OK ignoring the result here?
Seems like we might want to react to: NotSupportedByStorage

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
```diff
@@ -0,0 +1,109 @@
...
+            config.num_cache_warmup_workers = 0;
+        };
+    })
+    .start()
+    .await
+    .unwrap()
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_ram_pinning_config_updates() -> anyhow::Result<()> {
```
LogvinovLeon, line 59: Every test is a test so no need for a prefix.
The module is called `ram_pinning` so that is also redundant.
#[test]
fn config_updates()

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
```diff
@@ -0,0 +1,109 @@
...
+    sov_test_utils::initialize_logging();
+    let temp_dir = tempfile::tempdir().unwrap();
```
LogvinovLeon:
```suggestion
    let temp_dir = tempdir()?;
```

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
```diff
@@ -0,0 +1,109 @@
...
+    let exec_config_path = temp_dir.path().join("evm_execution_config.json");
+    let signer: PrivateKeySigner = SENDER_PRIV_KEY.parse().unwrap();
```
LogvinovLeon:
```suggestion
    let signer: PrivateKeySigner = SENDER_PRIV_KEY.parse()?;
```

LogvinovLeon: Also all other cases of using unwrap() in this test can be replaced with ?

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
```diff
@@ -0,0 +1,109 @@
...
+    };
+    std::fs::write(
+        &exec_config_path,
+        serde_json::to_string_pretty(&exec_config_contents).unwrap(),
+    )
+    .unwrap();
+    let test_rollup =
+        start_node_with_ram_pinning(RollupProverConfig::Skip, temp_dir, exec_config_path.clone())
+            .await;
+    test_rollup.wait_for_next_blocks(3).await;
```
LogvinovLeon: Why 3? 1 should be enough and faster

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
```diff
@@ -0,0 +1,109 @@
...
+    let client = alloy_client_with_signer(test_rollup.http_addr, SENDER_PRIV_KEY);
+
+    tracing::info!("Deploying contract");
+    let contract = SimpleStorage::deploy(client).await?;
+    let contract_address = contract.address();
```
LogvinovLeon: Inline. Is used in 1 place and doesn't provide any savings nor better name.

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
```diff
@@ -0,0 +1,109 @@
...
+    tracing::info!("Setting value");
+
+    let tx = contract.set(U256::from(1)).send().await?;
+    let _receipt = tx.get_receipt().await?;
```
LogvinovLeon: Why do we need to fetch receipts is we discard them? It will also work without

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
```diff
@@ -0,0 +1,109 @@
...
+            .await;
+    test_rollup.wait_for_next_blocks(3).await;
+    let client = alloy_client_with_signer(test_rollup.http_addr, SENDER_PRIV_KEY);
+
+    tracing::info!("Deploying contract");
+    let contract = SimpleStorage::deploy(client).await?;
+    let contract_address = contract.address();
+    tracing::info!("Setting value");
+
+    let tx = contract.set(U256::from(1)).send().await?;
```
LogvinovLeon: Do we need to call get and set? Seems like only deploy will be sufficient for this test

LogvinovLeon (changes requested):

LogvinovLeon (approved):

examples/demo-rollup/stf/src/runtime.rs
preston-evans98: Good question. Since this config is optional, I think you're right - it probably makes sense to panic on `NotSupported` since the user specified two incompatible configs; that's almost definitely a mistake, so better to make it obvious

crates/module-system/module-implementations/sov-evm/src/call.rs
preston-evans98: No reason. Moved.

crates/module-system/module-implementations/sov-evm/src/call.rs
preston-evans98: Done.

crates/module-system/module-implementations/sov-evm/src/call.rs
preston-evans98: Looks like the suggestion should be `if !was_contract`. Fixing but switching to your style.

crates/module-system/module-implementations/sov-evm/src/lib.rs
preston-evans98: Serde default requires a function

crates/module-system/module-implementations/sov-evm/src/lib.rs
preston-evans98: Happy to discuss more, but I prefer `EVM_EXECUTION_CONFIG` since it's much more greppable. If we name this struct `CONFIG`, we're highly likely to have colliding names. `EVM_EXECUTION_CONFIG` is clear, and also works nicely if somebody imports the struct name on its own (which Rust Analzyer is likely to fill even if you don't intend to).

```rust
use sov_evm::execution_config::CONFIG;

// a few hundred lines later
fn foo() { 
  if CONFIG.get().is_some() ...  // What was CONFIG again? I have to scroll to the top of the file to find out...
}
```

crates/module-system/module-implementations/sov-evm/src/lib.rs
preston-evans98: Done.

crates/module-system/module-implementations/sov-evm/src/lib.rs
preston-evans98: Going to leave this one out of scope for this PR since it's a change to the trait design. cc @ross-weir, do you have thoughts?

crates/module-system/module-implementations/sov-evm/src/lib.rs
preston-evans98: Good idea. Done.

crates/module-system/module-implementations/sov-evm/src/lib.rs
preston-evans98: Removed.

crates/module-system/sov-state/src/pinned_cache.rs
preston-evans98: Done.

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
preston-evans98: As with the execution config, I'm happy to discuss further - but I generaly prefer the longer but more greppable name

preston-evans98: Let's discuss at standup

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
preston-evans98: Removing for now! Will follow up with more tests later

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
preston-evans98: No, I think we can safely drop these. Done.

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
preston-evans98: done

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
preston-evans98: Done

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
preston-evans98: Done

crates/module-system/module-implementations/sov-evm/src/lib.rs
preston-evans98: Sadly Serde default only supports functions, not consts

examples/demo-rollup/tests/evm/evm_ram_pinning.rs
preston-evans98: Done

crates/module-system/module-implementations/sov-evm/src/lib.rs
ross-weir: I wasn't expecting there to be a case where `ExecutionInit::init` would be empty since it's a optional trait (if it's empty just dont impl the trait)

On a side note `ModuleExecutionConfig` was intended to be a higher level trait not implemented at the module level and include the initialization of multiple modules - I guess that might be why this case has occurred

examples/demo-rollup/stf/src/runtime.rs
```diff
@@ -99,7 +103,7 @@ where
     type GenesisInput = GenesisPaths;
 
     #[cfg(feature = "native")]
-    type ModuleExecutionConfig = ();
+    type ModuleExecutionConfig = EvmExecutionConfig;
```
ross-weir, line 106: The intention was for this to be something like `DemoRollupModuleConfig` which would then handle calling all the required modules `ExecutionInit::init`

Ok as is - no need to delay important GTE work though of course

PR #2134, opened 2025-11-27
Failover: Add `sequencer_leader` table to postgres

bkolad:
# Description

1. Makes the `is_replica` config value optional. Missing values now default to “primary,” though this behavior will be updated in a subsequent PR where the database will determine the primary node.
2. Adds a new `sequencer_leader` table in PostgreSQL along with `try_update_leader` methods and corresponding tests.

See https://github.com/Sovereign-Labs/sovereign-sdk-wip/pull/3168 for more details.


crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
```diff
@@ -1,16 +1,25 @@
+#![allow(dead_code)]
```
bkolad, line 1: Will be removed in the next PR once we start using `try_update_leader`

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
```diff
@@ -201,6 +210,55 @@ impl PostgresBackend {
             in_progress_batch,
         })
     }
+
+    async fn try_update_leader(
+        &self,
+        node_id: Uuid,
+        time_delta: Duration,
```
theodorebugnet, line 217: I think value should always be constant between all nodes, to ensure elections are fair, right? I'm not sure I like passing it as a parameter, which gives an option to override it at any callsite (which shouldn't be allowed, I think).

It's useful to override in tests but maybe it can be a parameter on the PostgresBackend constructor? That way in real code the sequencer just has to pass the correct const on construction, rather than at every call. And the unit tests can re-connect `db` when changing the value.

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
bkolad: Yeap I will pass it in the constructor and read form a config. It juts made testing easier for now but will change in following PRs.

PR #2138, opened 2025-11-28
Failover: Change `node_id` type to String

bkolad:
# Description

1. Changes node_id type from UUID to String
2. Adds `time_delta` to `PostgresBackend` (code review from the previous PR)

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
@@ -9,7 +9,6 @@
 //! invariants and we'd rather have an application crash due to broken
 //! invariants than to have bugs that result in subtle state inconsistencies.
 
-use uuid::Uuid;
 pub mod postgres;
 pub mod rocksdb;
 use crate::preferred::PostgresBackend;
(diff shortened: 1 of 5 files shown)


crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -387,34 +388,39 @@ impl PreferredSequencerDb {
...
+
+            match is_replica {
+                Some(true) => Ok((
+                    Self {
+                        backend: None,
+                        shutdown_sender: shutdown_sender.clone(),
+                    },
+                    true,
+                )),
+                Some(false) | None => {
```
bkolad: We handle is_replica = Some(false) and is_replica = None the same way for backward compatibility. This behavior will change in upcoming PRs, where None will indicate that the database determines the leader.

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
```diff
@@ -13,17 +13,18 @@ use sqlx::postgres::{PgPool, PgPoolOptions};
...
-    node_id: Uuid,
+pub(crate) struct SequencerLeader {
+    node_id: String,
     last_updated: OffsetDateTime,
 }
 
 pub struct PostgresBackend {
     pool: PgPool,
     backoff_policy: ExponentialBuilder,
+    time_delta: Duration,
```
theodorebugnet, line 26: Could we give this a slightly more descriptive name? E.g. `heartbeat_timeout`, `leader_timeout`, `time_till_leader_update_allowed` or similar. (Maybe in a future PR)

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
```diff
@@ -55,6 +56,15 @@ macro_rules! run_with_retries {
 
 impl PostgresBackend {
     pub async fn connect(connection_string: &str) -> anyhow::Result<Self> {
+        Self::connect_with_time_delta(connection_string, Duration::ZERO, "NODE_ID".to_string())
```
theodorebugnet, line 59: Is `Duration::ZERO` a good default? If multiple nodes were to use this function to start, it would basically cause constant leader swaps. Maybe `Duration::MAX` instead, i.e. if you don't specify the delta, then the node never participates in leader elections?

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
bkolad: I am already working on the next PR will update it there.

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
bkolad: Will change it in the next PR (Now it is unused anyway)

PR #2167, opened 2025-12-02
Replace nonce queue with single-threaded implementation

theodorebugnet:
# Description

Having the nonce buffers be single-threaded should make it significantly easier to reason about.

Integration tests pass. Unit tests from the old implementation will be ported in follow-up work.

The main `run()` function is also quite huge and badly needs refactoring, this will be done in immediate follow-up work.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+        // We don't throw an error if has_in_flight is already true - for instance, if two txs with
+        // the current valid nonce arrive near-simultaneously, we let the executor sort them out for
+        // simplicity. Thus there can be more than one tx in flight - they just all need to have
+        // the same nonce.
+        self.has_in_flight = true;
+    }
+
+    fn mark_inflight_execution_succeeded(&mut self, tx_nonce_check: u64) {
+        if !self.has_in_flight {
+            tracing::error!("Sequencer nonce buffer: non-persisted tracking: attempted to mark executed tx successful when has_in_flight is false!");
```
preston-evans98: This seems like it should probably be at least a debug assertion - otherwise, we won't catch bugs in our tests.

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+            false => self.user_nonce(),
+        }
+    }
+
+    fn mark_inflight(&mut self, tx_nonce_check: u64) {
+        if self
+            .user_nonce()
+            .is_some_and(|user_nonce| tx_nonce_check != user_nonce)
+        {
+            tracing::error!("Sequencer nonce buffer: non-persisted tracking: attempted to execute nonce that does not match tracked next nonce!");
```
preston-evans98: This should probably be a debug assertion as well. (Or, even better, a real assertion)

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+                    tx_nonce,
+                    tx_hash,
+                })
+                .await;
+        });
+    }
+
+    async fn run(&mut self) {
+        while let Some(input) = self.buffer_input.recv().await {
+            println!("Nonce buffer task processing input {input:?}");
```
preston-evans98: nit: This look like leftover debug output

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+                    original_tx_queue_id,
+                    result_sender,
+                } => {
+                    let queue = self.buffers.entry(credential_id).or_default();
+
+                    let user_nonce = queue.non_persisted.user_nonce().unwrap_or_else(|| {
+                        self.execution_backend
+                            .get_current_nonce_for_user(&credential_id)
+                    });
+                    println!("User nonce during NewTx: {user_nonce}");
```
preston-evans98: nit: leftover println

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+                                    tx_hash,
+                                    tx_nonce,
+                                    user_nonce,
+                                    old_tx.nonce_when_queued,
+                                    old_tx.queued_at,
+                                    credential_id,
+                                    InvalidNonceReason::Replaced,
+                                ));
+                            }
+                            self.schedule_timeout(credential_id, tx_nonce, tx_hash);
```
preston-evans98: I think there's an edge case here where the user accidentally submits the exact same tx (with the same nonce) twice. Then, we'll have multiple timeouts scheduled for the same job. Two easy fixes: 
1. compare the hashes of new and old_tx. Ignore new_tx if it's got the same hash 
2. Add a oneshot to the timeout job which gets dropped when we execute or evict a txs. Then timeouts will get cleaned up immediately.

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+                        }
+                        Action::Execute => {
+                            println!("Determined action: Execute (for nonce {tx_nonce})");
```
preston-evans98: nit: Leftover println

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+                                        credential_id,
+                                        tx_nonce,
+                                        tx_result,
+                                        result_sender,
+                                    })
+                                    .await;
+                            });
+                        }
+                        Action::Reject => {
+                            println!("Determined action: Execute (for nonce {tx_nonce})");
```
preston-evans98: nit: Leftover println

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+                        queue.non_persisted.mark_inflight_execution_failed(tx_nonce);
+                    }
+
+                    let _ = result_sender.send(tx_result); // If the receiver was dropped, ignore
+
+                    let user_nonce = queue
+                        .non_persisted
+                        .user_nonce()
+                        .expect("We just marked the nonce on the queue, so it cannot be None");
+                    println!("\nAfter execution, next nonce being queued is {user_nonce}");
```
preston-evans98: nit: leftover println

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+                    loop {
+                        let Some(head_entry) = queue.txs.first_entry() else {
+                            break;
+                        };
+                        match head_entry.key().cmp(&user_nonce) {
+                            Ordering::Less => {
+                                // Stale transaction in queue - evict and ignore.
+                                // This should not normally happen either, but we handle it to avoid a deadlock
+                                // if it does happen for any reason.
+                                head_entry.remove();
```
preston-evans98, line 524: nit: Let's log an error here so that we can detect bugs

preston-evans98: Even better, let's add a debug assertion

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+                    }
+                }
+                NonceBufferInput::TxPersisted { credential_id } => {
+                    let hash_map::Entry::Occupied(entry) = self.buffers.entry(credential_id) else {
+                        continue;
+                    };
+                    if entry.get().txs.is_empty() {
+                        let real_nonce = self
+                            .execution_backend
+                            .get_current_nonce_for_user(&credential_id);
```
preston-evans98: nit: I think we can avoid the problems of asynchrony between the backend and the tx queue here by just putting the executed nonce into the `TxPersisted` message. We already have access to it when we create the message, and it'll also let us save the state read

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+                        }
+                    }
+                }
+                NonceBufferInput::TxTimedOut {
+                    credential_id,
+                    tx_nonce,
+                    tx_hash,
+                } => {
+                    let queue = self.buffers.entry(credential_id).or_default();
+                    let state_nonce = self
```
preston-evans98: nit: we can skip the state read here if `user_nonce_to_use_as_prerequisite_start`. Let's put it in an `unwrap_or_else`

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+                        .execution_backend
+                        .get_current_nonce_for_user(&credential_id);
+                    let user_nonce_for_prerequisites = queue
+                        .non_persisted
+                        .user_nonce_to_use_as_prerequisite_start()
+                        .unwrap_or(state_nonce);
+                    if queue.has_contiguity_between(user_nonce_for_prerequisites, tx_nonce) {
+                        self.schedule_timeout(credential_id, tx_nonce, tx_hash);
```
preston-evans98: If we have contiguity we'll never actually time out the transaction, right? We'll just keep hitting this condition and scheduling a new timeout 20 seconds in the future; which means that (under load) the response time for these transactions is unbounded. Maybe we should pop them anyway?

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+    tx_nonce: u64,
+    expected_nonce: u64,
+    nonce_when_queued: u64,
+    instant_queued: std::time::Instant,
+    credential_id: CredentialId,
+    queue_rejection_reason: InvalidNonceReason,
+) -> TransactionReceiverResult<S, Rt> {
+    let was_queued_msg = format!("The sequencer queued the transaction for reordering {} ms ago, when the user's nonce was {nonce_when_queued}.", instant_queued.elapsed().as_millis());
+    let queue_error_msg = match queue_rejection_reason {
+        InvalidNonceReason::Invalid => "The sequencer did not attempt to queue the transaction as it was not within valid queue limits (either in the past, or beyond the max limit).".to_string(),
```
preston-evans98, line 733: nit: Let's include the tx_nonce, the account_nonce, and the max limit

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -0,0 +1,800 @@
...
+        InvalidNonceReason::Timeout => format!("{was_queued_msg} In that time, the sequencer did not receive all the transactions leading up to this tx's nonce, so it has timed out and is being evicted from the queue."),
```
preston-evans98: nit: let's change "did not receive" to "did not accept" so that we have flexibility to reject txs even if their prereqs have been enqueued.

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: I'm shying away from making it a real assertion because if it fails, the entire sequencer will crash, while the only error will have been slightly less reliable nonce queueing behaviour. IMO it's better to have extra rejections from the sequencer than complete rollup crashes. Of course it'd be good to have strict correctness guarantees but not at the cost of bringing down customer rollups.
LMK if you disagree though

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: I'm not sure it would've caused issues, but checking the old_tx hash doesn't add extra complexity and eliminates the issue so I'll go ahead and add it :+1:

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: I had it that way at first but had decided it'd be easier to just check the real nonce once - if txs are inserted in batches into the DB, the TxPersisted messages could come all at once in arbitrary order, and just reading the state lets us clean up the queue once and the rest will be no-ops. Doing it based on the message does save the state read though you're right.

But can you clarify what asynchrony you're worried about here? I don't see any issues other than the state read optimisation, so I wanna make sure I'm not missing anything

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: The tx_nonce and account_nonce are part of the full error message below ("Tx bad nonce..."). I'll add the max limit :+1:

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: Correct. That's an intentional feature of the design - also kind of meant to be used with a much shorter timeout than 20s. The response time is bounded by how long it can possibly take to execute the rest the pre-requisite transactions.

The behaviour under load is a good point. We could add a max retry counter, so even if pre-reqs are satisfied, if the tx still isn't executed after N timeouts it's evicted anyway. Removing the contiguity check entirely would force all transactions to execute within the timeout window which seems strict.

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: Actually adding the max_limit is quite annoying because it's a property of the NonceBufferTask, which means either
a) adding yet another parameter to the function, which already takes a lot and also complicates its callsite in the NonceBufferInputSender (which would probably need to take it as a param from the sequencer itself), or
b) making it a member function of NonceBufferTask which again causes issues for NonceBufferInputSender.

It's doable but IMO the information is not very useful - we know the config and we know the `expected_nonce`, and the max nonce will always be `expected_nonce + max_future_nonce_delta` which can always be had from the instance's config.toml.

Let me know if you have pushback, otherwise I'll skip this for now to avoid the extra complexity from passing it around.

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
preston-evans98: The asynchrony is just the fact that an arbitrary number of other txs might have been executed and persisted to the api state before we process this message; on the other hand, the message in the queue are guaranteed to be processed in order.

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
preston-evans98: Fine by me!

preston-evans98 (approved):

PR #2169, opened 2025-12-02
[WIP] Sequencing data

LogvinovLeon:
# Description

TODO

-------

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-rollup-apis/src/endpoints/simulate.rs
+++ b/crates/full-node/sov-rollup-apis/src/endpoints/simulate.rs
@@ -318,6 +318,7 @@ impl<S: Spec, R: Runtime<S>> SovereignSimulate<S, R> {
             credential_id,
             default_address: credential_id.into(),
             credentials: Credentials::new(credential_id),
+            sequencing_data: None,
         }
     }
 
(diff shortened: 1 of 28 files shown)


crates/module-system/sov-modules-api/src/transaction/types/v2.rs
```diff
@@ -0,0 +1,72 @@
...
+    #[sov_wallet(
+        bound = "Call: sov_rollup_interface::sov_universal_wallet::schema::UniversalWallet"
+    )]
+    pub runtime_call: Call,
+    /// Uniqueness identifier of this transaction. see [`UniquenessData`] for more details.
+    pub uniqueness: UniquenessData,
+    /// The transaction metadata. Contains gas parameters and the chain ID.
+    pub details: TxDetails<S>,
+    /// Sequencer metadata. Can contain things like high-precision timestamps
+    pub sequencing_data: Vec<u8>,
```
bkolad: What’s the reasoning behind creating a new transaction type instead of adding this functionality to `V0` and `V1`? If we need to do something similar for transactions that support multi-sig, we would need to introduce a V4. Is the concern simply that if some V0 and V1 txs don’t use this field, they would waste space storing an empty Vec?

bkolad (commented):
This PR LGTM and thx for the code cleanup, which improved the code structure. 
I had only one question

crates/module-system/sov-modules-api/src/transaction/types/v2.rs
LogvinovLeon: My reasoning was that it would be a breaking change and people would not be able to deserialize txs they already have in the database. If we're ok with that - sure. But then by this logic - shouldn't we also remove V0 and V1 and just have V0 with multisig that can have only 1 signature?

PR #2175, opened 2025-12-03
Only gather witness for NomtProverStorage with strict_mode

citizen-stig:
# Description

Sequencer computes state update only to get user-space root hash. But NomtProverStorage will collect witness anyway, which will be thrown away.

This PR repurposes `is_strict_mode` which was set only for node storage, so only node will collect witness, so prover logic remains unchanged.
The flag is renamed to `with_witness`. All consistency checks remains.

Changes:

* nomt::Session is initialized with default params if flag is not passed, which is `WitnessMode::disabled`
* All read value are not added to our own `ArrayWitness`

## Testing

## Docs
Updated doc for state_db_nomt

### DIFF
--- a/crates/full-node/sov-db/src/state_db_nomt.rs
+++ b/crates/full-node/sov-db/src/state_db_nomt.rs
@@ -299,7 +299,7 @@ where
     /// Should be called only when really needed.
     /// Will hold the read lock to all snapshots.
     /// **Commiting storage will be blocked until all built sessions are deallocated.**
-    /// Produces session does not collect witness!
+    /// Returned session does not collect witness!
     /// Use `Self::begin_both_sessions` if witness is needed
     #[tracing::instrument(skip(self))]
     pub fn begin_user_session(&self) -> anyhow::Result<nomt::Session<BinaryHasher<H>>> {
@@ -341,7 +341,7 @@ where
     /// Should be called only when really needed.
     /// Will hold the read lock to all snapshots.
     /// **Commiting storage will be blocked until all built sessions are deallocated.**
-    /// Produces session does not collect witness!
+    /// Returned session does not collect witness!
     /// Use `Self::begin_both_sessions` if witness is needed
     #[tracing::instrument(skip(self))]
     pub fn begin_kernel_session(&self) -> anyhow::Result<nomt::Session<BinaryHasher<H>>> {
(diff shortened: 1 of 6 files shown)


crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -342,6 +341,8 @@ where
     /// Should be called only when really needed.
     /// Will hold the read lock to all snapshots.
     /// **Commiting storage will be blocked until all built sessions are deallocated.**
+    /// Returned session does not collect witness!
+    /// Use `Self::begin_both_sessions` if witness is needed
     #[tracing::instrument(skip(self))]
     pub fn begin_kernel_session(&self) -> anyhow::Result<nomt::Session<BinaryHasher<H>>> {
```
preston-evans98: nit: Let's rename to `begin_kernel_session_without_witness` to make this clear at the call site

crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -299,6 +299,8 @@ where
     /// Should be called only when really needed.
     /// Will hold the read lock to all snapshots.
     /// **Commiting storage will be blocked until all built sessions are deallocated.**
+    /// Returned session does not collect witness!
+    /// Use `Self::begin_both_sessions` if witness is needed
     #[tracing::instrument(skip(self))]
     pub fn begin_user_session(&self) -> anyhow::Result<nomt::Session<BinaryHasher<H>>> {
```
preston-evans98: nit: let's rename this to `begin_user_session_without_witness` to make this clear at the call site.

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -511,7 +519,7 @@ where
         let current_prev_root = StorageRoot::new(current_prev_user_root, current_prev_kernel_root);
 
         // Check staleness, pre-computation:
-        if self.is_strict_mode && current_prev_root != prev_state_root {
+        if self.with_witness && current_prev_root != prev_state_root {
```
preston-evans98: nit: It's not obvious from the name that `with_witness` will also enable consistency checks. Maybe it would be clearer if we said something like `strict_with_witness`? Open to suggestions if you have any!

preston-evans98 (approved):
Core change looks great! Left some nits related to naming, but feel free to merge when ready

PR #2177, opened 2025-12-03
Fix queue behaviour on sequencer resync.

theodorebugnet:
# Description

The queue assumes that all transactions go through it. On resync, this isn't true, and the API state will increase independently of the queue. We detect this now by a) checking the original_tx_queue_id and b) detecting SequencerNotReady errors; both of these cases invalidate the entire nonce queue and cause a wipe.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -134,22 +135,13 @@ impl NonPersistedTxs {
...
             let msg = "Sequencer nonce buffer: non-persisted tracking: attempted to mark executed tx successful when has_in_flight is false!";
             tracing::error!(msg);
             debug_assert!(false, "{msg}"); // See comment in mark_inflight
         }
         self.has_in_flight = false;
-
-        // If this condition is false, this was likely a concurrent transaction for the current
-        // nonce which failed. The successful transaction with the same nonce will have set
-        // last_successfully_executed correctly.
-        if self.last_successfully_executed.is_some_and(|l| {
```
theodorebugnet, line 148: This was a no-op, left in from a couple of rounds of simplifying the pre-req checks. If/when pre-req logic is added back, and simultaneous same-nonce execution is fixed as per the other comments nearby, then this can be added back as a debug assertion.

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -134,22 +135,13 @@ impl NonPersistedTxs {
...
     }
 
-    fn mark_inflight_execution_failed(&mut self, tx_nonce_check: u64) {
+    fn mark_inflight_execution_failed(&mut self) {
         if !self.has_in_flight {
             let msg = "Sequencer nonce buffer: non-persisted tracking: attempted to mark executed tx successful when has_in_flight is false!";
             tracing::error!(msg);
             debug_assert!(false, "{msg}"); // See comment in mark_inflight
         }
         self.has_in_flight = false;
```
preston-evans98, line 142: If we have two txs in flight concurrently (because the caller submitted the same nonce twice), will this be incorrect?

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -323,6 +321,26 @@ impl<E: TxExecutionBackend<S, Rt> + Clone + Send + Sync + 'static, S: Spec, Rt:
...
+    /// Wipe the entire queue, e.g. because the sequencer had to resync so the queued transactions
+    /// are no longer valid/relevant.
+    /// We drop all the buffers, which will drop the oneshot senders from the
+    /// NonceBufferInputSender tasks waiting for a response, where the drop can be handled
+    /// appropriately.
+    async fn wipe(&mut self) {
+        self.buffers = Default::default();
+        let mut keep_messages = Vec::new();
+        while let Ok(m) = self.buffer_input.try_recv() {
+            if matches!(m, NonceBufferInput::TxExecuted { .. }) {
```
preston-evans98: nit: Let's put an explicit match on all of the other branches here so that adding an enum variant doesn't create a bug later

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -431,41 +459,52 @@ impl<E: TxExecutionBackend<S, Rt> + Clone + Send + Sync + 'static, S: Spec, Rt:
...
                     result_sender,
                 } => {
+                    // If the tx was rejected because the sequencer is down (syncing,
+                    // recovering etc.), the nonce queue will be stale and needs to be wiped.
+                    // At minimum the non_persisted tracking for all users needs wiping for
+                    // correctness, but pre-downtime transactions are invalidated by the
+                    // sequencer anyway so we wipe everything.
+                    if is_notready_error(&tx_result) {
+                        self.wipe().await;
+                        let _ = result_sender.send(tx_result);
```
preston-evans98: Let's send first, then wipe?

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -601,25 +637,11 @@ impl<E: TxExecutionBackend<S, Rt> + Send + 'static, S: Spec, Rt: Runtime<S>>
...
-            err_invalid_nonce::<S, Rt>(
-                tx_hash,
-                tx_nonce,
-                current_nonce,
-                nonce_when_queued,
-                queued_at,
-                credential_id,
-                InvalidNonceReason::EvictedBeforeExecution,
-            )
+            // The oneshot sender was dropped. This might happen on shutdown, but will also happen
```
preston-evans98: Should we explicitly send a response when we drop the channel rather than relying on implicit behavior? I'm worried that we might have other reasons to drop the channel, and this error will mask the actual behavior

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -691,10 +708,15 @@ fn err_invalid_nonce<S: Spec, Rt: Runtime<S>>(
     ))))
 }
 
+fn is_notready_error<S: Spec, Rt: Runtime<S>>(result: &TransactionReceiverResult<S, Rt>) -> bool {
+    matches!(result, Ok(Err(AcceptTxError::NotFullySynced(_))))
```
preston-evans98: Same here, let's match explicitly on each arm so that adding a new variant doesn't silently break any assumptions

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: Actually great spot, thank you! For the same reason that this check isn't done in mark_succeeded. I had overlooked the second case here. Removing

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: Actually I don't think this affects correctness - the inflight TX is not going to be in the queue (we pop before executing), and the results_sender is owned by the message extracted from the channel and not `self.buffers`.
But I'll send first so we can reduce latency to the caller, there's no reason not to.

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: Replying on Slack
Edit: on further thought I agree

PR #2184, opened 2025-12-05
Rate limiting: Implement a rate-limiting mechanism using the token bucket algorithm.

bkolad:
# Description
This PR introduces a rate-limiting mechanism. The mechanism is not yet integrated into the sequencer (this will be done in the upcoming PR).
1. `rate_limiter/resource.rs` – Defines the resources whose usage will be rate-limited.
2. `rate_limiter/limiter.rs` – Contains the type definitions and the generic RateLimiter algorithm.
3. `rate_limiter/mod.rs` – Implements SovRateLimiter, a rate-limiter that operates on IP addresses and SOV-addresses.

Description form the doc:
```
/// This is an adaptation of the token bucket algorithm with the following differences:
///
/// 1. The resource being tracked is multidimensional, not just a simple request counter.
///    (This is abstracted behind the [`Resource`] struct.)
/// 2. We do not know how many resources a request will consume until after the request
///    has been executed.
///
/// Therefore, we provide two methods:
/// a) `allow` — checks at the beginning of a request whether the key has not yet exceeded
///    its allowed resource budget.
/// b) `update` — updates the resource usage after the request has completed.
///
/// Because of this separation, it is possible that after `update` a given key will exceed its
/// resource limit. When this happens, subsequent calls to `allow` will reject new requests
/// for that key until enough time has passed for resources to drain below the threshold.
```

Custom limits for special IP/addresses will be added in the next PR


crates/full-node/sov-sequencer/src/preferred/rate_limiter/resource.rs
```diff
@@ -0,0 +1,251 @@
...
+        if self.execution_time_micros > other.execution_time_micros {
+            return Err(LimitExceeded::ExecutionTime {
+                total_accumulated: self.execution_time_micros,
+                max_allowed: other.execution_time_micros,
+            });
+        }
+
+        // Gas is not included until all gas-related issues are resolved:
+        // - proper constants
+        // - proper gas estimation
```
bkolad, line 118: Gas will be added here when all the relevant gas issues are resolved.

crates/full-node/sov-sequencer/src/preferred/rate_limiter/limiter.rs
```diff
@@ -0,0 +1,465 @@
...
+use sov_modules_api::Gas;
+use sov_modules_api::Spec;
+use std::hash::Hash;
+use std::time::Duration;
+use std::time::Instant;
+
+/// Configuration for the rate limiter.
+#[derive(Debug, Clone)]
+pub(crate) struct RateLimiterConfig<S: Spec> {
+    pub(crate) ttl_in_milis: u64,
```
bkolad, line 14: This will be addressed later (once we start setting values), but I think the value should be on the order of several block times.

crates/full-node/sov-sequencer/src/preferred/rate_limiter/limiter.rs
```diff
@@ -0,0 +1,465 @@
...
+    pub(crate) max_allowed_resources: TotalResources<S::Gas>,
+    pub(crate) drain_rate: DrainRatePerMillis<S::Gas>,
```
bkolad: I would estimate it as:

```
let drain_rate =  MaxBatchResourceCapacity/MaxBatchExecutionTime*0.1
```
This means we start rate-limiting any actor that begins using around 10% of the available resources.
We will also have custom limits for preferred actors (Next PR)

This is just an initial proposal.

preston-evans98: Happy to discuss more, but I'd imagine we want rates to be even lower than that - probably 1% or less; otherwise, you could DOS the chain with just a small handful of addresses

preston-evans98: But that's out of scope for now :)

crates/full-node/sov-sequencer/src/preferred/rate_limiter/limiter.rs
```diff
@@ -0,0 +1,465 @@
...
+        self.drain_total_resource_used(now, drain_rate);
+        self.total_resource_used.allow(max_allowed)
+    }
+
+    /// Increase the resource used by a given trottler.
+    fn throttle(&mut self, resource_used: ResourceUsed<G>) {
+        self.total_resource_used.combine(&resource_used);
+    }
+
+    fn drain_total_resource_used(&mut self, now: Instant, drain_rate: &DrainRatePerMillis<G>) {
```
preston-evans98: nit: terminology for token bucket is usually "refilling" when you're crediting people for not using resources and `draining` when they consume resources

preston-evans98: Oh, maybe the terminology is correct but the usage is backwards? It looks like we're draining based on how long it's been since the last time someone consumed resources. (`mul_by_millis(since_last_refill)`). Shouldn't we be first crediting their account with the refil accumulated in that time, then draining a fixed amount for their new consumption?

preston-evans98:
High level, looks good at first review. AFAICT, the implementation is solid. That being said, I'm afraid I may have mixed things up due to the terminology being inverted from standard token bucket. (Usually, draining the bucket is a bad thing (it means you're consuming resources) and filling up the bucket is a good thing).

I hate to block on something non-substantive, but do you mind inverting the terminology and I'll re-review? I'm afraid the risk of stupid mistakes on my side is quite high given the current form :)

bkolad:
> High level, looks good at first review. AFAICT, the implementation is solid. That being said, I'm afraid I may have mixed things up due to the terminology being inverted from standard token bucket. (Usually, draining the bucket is a bad thing (it means you're consuming resources) and filling up the bucket is a good thing).
> 
> I hate to block on something non-substantive, but do you mind inverting the terminology and I'll re-review? I'm afraid the risk of stupid mistakes on my side is quite high given the current form :)

Yes, there is an important difference that I didn’t explain in the comments. We have to handle overflows, while the standard algorithm does not. The reason is that we only know our resource usage after executing a request, whereas the standard algorithm knows it before execution.

1. In the standard approach, the system starts with `MAX_ALLOWED` tokens, and each request decreases that amount while tokens are simultaneously refilling it at rate R. If an IP currently has X tokens, any request that requires more than X is rejected.

2. In our case, however, we only learn the actual resource usage after executing the request. That means an IP may have X tokens, but the executed request might consume more than X. Subtracting those tokens would result in negative value. While this in itself is ok, it becomes problematic for gas accounting because gas cannot be negative—so we would need an additional flag to detect when the calculation goes below zero etc...

Instead, I chose to track how many tokens were used rather than how many remain (Opposite to the standard algorithm). This approach avoids negative values entirely, but it also requires inverting some of the naming conventions.

I agree that departing from the standard nomenclature is confusing, so I updated the PR and applied the following change:
We still track `ResourceUsed` (not `ResourceAllowed`), but I renamed `drain` to `refill_tokens`.

```
impl<G: Gas> TotalResources<G> {
    #[must_use]
    fn refill_tokens(&self, how_much_to_refill: &Resource<G>) -> Self {
        // In [`Throttler`] we track the resources used by a given key.
        // Therefore, refilling means decreasing the amount of resources used.
        Self {
            inner: self.inner.saturating_sub(how_much_to_refill),
        }
    }
}
```
So now the meaning of a token is that it decreases the resources_used by a given IP (which is logically equivalent to increasing the resources_allowed - good thing, but also avoids underflows). This way, we only need to handle the conversion in a single method, and all higher-level methods are renamed from `drain_` to `refill_` and have the usual meaning.


Also extended documentation with an example:
https://github.com/Sovereign-Labs/sovereign-sdk/pull/2184/files#diff-5b162b44f5d28c7f4a66dfa46dbe353983e30cac219bf3d6b442d673ea5fa4aaR128

crates/full-node/sov-sequencer/src/preferred/rate_limiter/limiter.rs
```diff
@@ -0,0 +1,544 @@
...
+            inner: self.inner.checked_add(&used.inner)?,
+        })
+    }
+
+    #[must_use]
+    fn refill_tokens(&self, how_much_to_refill: &Resource<G>) -> Self {
+        // In [`Throttler`] we track the resources used by a given key.
+        // Therefore, refilling means decreasing the amount of resources used.
+        Self {
+            inner: self.inner.saturating_sub(how_much_to_refill),
```
bkolad, line 64: The standard token bucket algorithm that tracks `resource_allowed` rather than `resource_used` would use addition instead of subtraction here. But now the difference shows up only in a single method.

crates/full-node/sov-sequencer/src/preferred/rate_limiter/limiter.rs
```diff
@@ -0,0 +1,544 @@
...
+            last_refill: now,
+        })
+    }
+
+    /// Increase the resource used by a given throttler.
+    #[must_use]
+    fn throttle<K: Debug>(&self, resource_used: ResourceUsed<G>, key: &K) -> Self {
+        let used = match self.total_resource_used.combine(&resource_used) {
+            Some(used) => used,
+            None => {
```
bkolad, line 124: Overflow can only occur if something is misconfigured, since in practice it’s impossible to consume more than `u64::MAX` bytes or `u64::MAX` microseconds etc...  for a single request. Decided not to shut down, only issue an error.

crates/full-node/sov-sequencer/src/preferred/rate_limiter/limiter.rs
preston-evans98: Very helpful, thanks!!

crates/full-node/sov-sequencer/src/preferred/rate_limiter/limiter.rs
```diff
@@ -0,0 +1,543 @@
...
+pub(crate) struct RateLimiter<K, S: Spec> {
+    data: Cache<K, Throttler<S::Gas>>,
+    max_allowed_resources: TotalResources<S::Gas>,
+    refill_rate: RefillRatePerMillis<S::Gas>,
+}
+
+impl<K: Hash + Eq + Debug, S: Spec> RateLimiter<K, S> {
+    pub(crate) fn new(config: RateLimiterConfig<S>) -> Self {
+        let data = Cache::builder()
+            .time_to_live(Duration::from_millis(config.ttl_in_milis))
```
preston-evans98, line 218: One thing we should think about here is misconfig where the TTL is too short. Since accounts are credited with a full budget when they're not in cache, having a ttl that's shorter than the time it takes a bucket to refill will cause very unintuitive results; maybe we  an add a check in the constructor and warn or even error?

preston-evans98 (approved):

crates/full-node/sov-sequencer/src/preferred/rate_limiter/limiter.rs
bkolad: I think we should calculate TTL based on the total resources that can be consumed in a block. Will add it to the issue/

PR #2194, opened 2025-12-08
Adding a metric for number of reads/writes

citizen-stig:
# Description

TPS comparison does not work when comparing different call messages. Closer approximation between different rollups would be number of state reads/writes in a single batch.

This PR adds a metric for it. Dashboard panel will be added if we decide to proceed with such metric.

I've decided not to add number of bytes read/written, because it introduces a bigger performance penalty, so we don't need that now.

## Testing

## Docs
No updates

### DIFF
--- a/crates/module-system/sov-state/src/nomt/prover_storage.rs
+++ b/crates/module-system/sov-state/src/nomt/prover_storage.rs
@@ -532,15 +532,8 @@ where
             .context("kernel state")?
         };
 
-        // Metric
         let user_reads = state_accesses.user.ordered_reads.len();
         let user_writes = state_accesses.user.ordered_writes.len();
-        // let user_write_bytes = state_accesses
-        //     .user
-        //     .ordered_writes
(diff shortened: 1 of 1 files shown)


crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -531,6 +532,28 @@ where
...
 
+        // Metric
+        let user_reads = state_accesses.user.ordered_reads.len();
+        let user_writes = state_accesses.user.ordered_writes.len();
+        // let user_write_bytes = state_accesses
+        //     .user
+        //     .ordered_writes
+        //     .iter()
+        //     .map(|(_, slot_value)| slot_value.as_ref().map(|s| s.size()).unwrap_or_default())
+        //     .sum::<u32>();
```
preston-evans98: nit: commented code

preston-evans98 (approved):

PR #2203, opened 2025-12-09
Celestia background stat

citizen-stig:
# Description

Celestia adapter now spins up the task, that collects some information every 30 seconds:

* Signer balance
* Gas Price
* Sync distance of the connected node.

This PR adds extra 6 API calls per minute, but in the future it will allow to build more comprehensive alerting

Screenshot 2025-12-09 at 14 29 59

## Linked Issues
- Related to #1630 

## Testing
Dashboard updated and tested manually

## Docs
No updates

### DIFF
--- a/crates/adapters/celestia/src/da_service/mod.rs
+++ b/crates/adapters/celestia/src/da_service/mod.rs
@@ -32,7 +32,6 @@ use sov_rollup_interface::da::{DaProof, DaSpec, RelevantBlobs, RelevantProofs};
 use sov_rollup_interface::node::da::{
     run_maybe_retryable_async_fn_with_retries, DaService, MaybeRetryable, SubmitBlobReceipt,
 };
-use sov_test_utils::ledger_db::sov_api_spec::tokio_tungstenite::tungstenite::client;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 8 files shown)


ross-weir (approved):

crates/adapters/celestia/src/da_service/mod.rs
```diff
@@ -192,6 +199,20 @@ impl CelestiaService {
             );
         }
 
+        let tx_priority = config.tx_priority.clone().into();
+        if let Ok(signer) = client.address() {
+            let bg_client = config
+                .build_client()
+                .await
+                .expect("Failed to build celestia-client for background task");
+            tokio::spawn(stat_collection_task(
```
bkolad, line 208: Shouldn't we register the join handle of this long-running task?

bkolad (approved):

crates/adapters/celestia/src/da_service/mod.rs
citizen-stig: I thought about it, but we don't need to wait for this task on restart, because it does not interact with the rest of the system, only metrics. 

So the worst case scenario, there might be 2 metrics around restart, and only in test.

PR #2224, opened 2025-12-12
`sov-modules-api`: remove `alloy-sol-types`

citizen-stig:
# Description

Removing alloy-sol-types as it is not needed anymore

## Testing

## Docs
No updates


bkolad (approved):

theodorebugnet (commented):
IIRC this was part of the initial exploration into EIP712 signing, and it's not needed by now and can be removed. So I don't think we even need to feature gate it.

Issue #2173, opened 2025-12-03
[Tracker] Rate limiting


bkolad:
## Description
The goal of this workstream is to introduce a custom rate-limiting mechanism based on `IP` and `CredentialId` Access to the following resources should be limited:
- tx_exec_time
- tx_size
- tps
- gas_used

## Checklist

 - [x] Pass SocketAddr from axum handler to `process_accept_tx` https://github.com/Sovereign-Labs/sovereign-sdk/pull/2170
 - [x] Adjust the above logic to work with nginx
 - [x] We need to apply rate limiting to failed transactions too. For unsuccessful transactions, pass both `apply_tx_to_in_progress_batch_inner` and `gas_used` into `apply_tx_to_in_progress_batch_inner`. https://github.com/Sovereign-Labs/sovereign-sdk/pull/2172/files
 - [x] Implement `RateLimiter` service that rate limits based on `IP` & `CredentialId`.
 - [x] Add handling for `x-forwarded-for` header https://github.com/Sovereign-Labs/sovereign-sdk/pull/2199
 - [x] Add config that respects cutom limits: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2230
 - [x] Add custom limits for special CredentialId & IP #2230
 - [x] Add relevant metrics #2241
 - [x] Test everything on AWS and specifically verify IP-based rate limiting.
 - [x] Enable the service in the `dev` branch

PR #2230, opened 2025-12-14
Rate Limiting: Add `SovRateLimiterConfig` and `Runtime::resolve_address`

bkolad:
# Description
1. The rate limiter now operates on `Spec::Address` instead of `CredentialId`. This change was necessary because users want to define rate limits based on addresses `Spec::Address`, which are more user-facing than CredentialId.
2. We added a `Runtime::resolve_address` method that converts a `CredentialId` into an `S::Address`. This allows the sequencer to perform the conversion and pass an address directly to the rate limiter.
3. We introduced `SovRateLimiterConfig`, which allows users to configure custom rate limits.

See: https://github.com/Sovereign-Labs/sovereign-sdk/issues/2173

## Linked Issues
- Related to https://github.com/Sovereign-Labs/sovereign-sdk/issues/2173

## Testing

## Docs


crates/full-node/full-node-configs/src/sequencer.rs
```diff
@@ -178,8 +181,8 @@ pub struct PreferredSequencerConfig {
     #[serde(default)]
     pub timing_oracle: Option<TimingOracleConfig>,
     /// Configuration for rate-limiting the sequencer.
-    #[serde(default)]
-    pub rate_limiter: Option<SovRateLimiterConfig>,
+    #[serde(default = "default_rate_limiter::<Address>")]
```
bkolad, line 183: #[serde(default)] was causing some problems in higher level macros.

crates/full-node/full-node-configs/src/sequencer.rs
```diff
@@ -267,15 +274,31 @@ pub struct TimingOracleConfig {
     pub private_key_hex: Option<String>,
 }
 
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, JsonSchema)]
+pub struct SovRateLimiterConfig<Address: Copy> {
+    pub default_limits: Limits,
+    pub address_custom_limits: Vec<(Address, Limits)>,
```
bkolad, line 279: We will set the custom limits in the next PR. This PR adds only structural changes. The rate limiter is still diabled in prod.

crates/full-node/full-node-configs/src/sequencer.rs
```diff
@@ -1,20 +1,23 @@
-use std::num::NonZero;
+use std::{net::IpAddr, num::NonZero};
 
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 /// See [`SequencerConfig::sequencer_kind_config`].
 #[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case")]
+//#[serde(bound = "Address: Copy + serde::Serialize + serde::de::DeserializeOwned")]
```
citizen-stig: nit: comment

crates/full-node/sov-sequencer/src/preferred/rate_limiter/limiter.rs
```diff
@@ -561,4 +641,12 @@ mod tests {
             },
         }
     }
+
+    impl RefillRatePerMillis<Gas> {
+        fn zero() -> Self {
```
citizen-stig, line 646: Non relevant here, but might be useful in the future if all those were const:

* `RefillRatePerMillis::zero`
* `Resource::zero`
* `Gas::zero`

crates/full-node/sov-sequencer/src/preferred/rate_limiter/mod.rs
```diff
@@ -5,15 +5,18 @@ mod resource;
...
+use sov_modules_api::BasicAddress;
 use sov_modules_api::{CredentialId, Spec};
+use std::collections::HashMap;
+use std::hash::Hash;
 use std::{net::IpAddr, time::Instant};
 
 use crate::preferred::sync_sequencer_state::comfortable_gas_limit;
 
 #[derive(Debug)]
 pub(crate) struct Token<S: Spec> {
```
citizen-stig: Also not relevant here, but might be confusing to grep it with `Bank::Token`

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -804,9 +801,21 @@ where
...
+        let checkpoint = &mut inner.executor.checkpoint;
+        let mut rt = Rt::default();
+
+        let address = rt
+            .resolve_addrss(
+                &ip_and_credential.default_address,
+                &ip_and_credential.credential_id,
+                checkpoint,
+            )
+            // We don't meter StateCheckpoint.
```
citizen-stig: Why?

bkolad: The `StateCheckpoint` has gas meter that is `NoOp`. So it can be used in situation where we don't track gas (like here)

crates/module-system/sov-modules-api/src/lib.rs
```diff
@@ -123,6 +123,7 @@ pub use sov_rollup_interface::zk::{
 #[cfg(feature = "native")]
 pub use sov_rollup_interface::StateUpdateInfo;
 pub use sov_rollup_interface::{execution_mode, BasicAddress, TxHash};
+pub use sov_state::User;
```
citizen-stig, line 126: Why do we need exporting this?

crates/utils/sov-test-utils/src/runtime/macros.rs
```diff
@@ -165,6 +165,15 @@ macro_rules! generate_runtime_without_capabilities {
                 }
             }
 
+            fn resolve_addrss<ST: sov_modules_api::StateAccessor>(
```
citizen-stig: Typo:
```suggestion
            fn resolve_address<ST: sov_modules_api::StateAccessor>(
```

crates/utils/sov-test-utils/src/runtime/macros.rs
```diff
@@ -165,6 +165,15 @@ macro_rules! generate_runtime_without_capabilities {
...
+                &mut self,
+                default_address: &S::Address,
+                credential_id: &sov_modules_api::CredentialId,
+                state: &mut ST,
+            ) -> Result<S::Address, <ST as sov_modules_api::StateWriter<sov_modules_api::User>>::Error>{
```
citizen-stig: Do we actually need this StateWriter::error? Do we do anything with it? Maybe anyhow will be enough.

crates/module-system/sov-modules-api/src/runtime/mod.rs
```diff
@@ -164,10 +163,17 @@ pub trait Runtime<S: Spec>:
     }
 
     /// Populates the pinned state cache for the given storage if supported
-    #[cfg(feature = "native")]
```
citizen-stig, line 167: White is not native anymore?

crates/full-node/sov-sequencer/src/preferred/rate_limiter/limiter.rs
bkolad: Unfortunetly `Gas::zero` can't be made const as it it trait function (and rust doe not allow it). The other mrthods dpend on `Gas::zero`

crates/module-system/sov-modules-api/src/runtime/mod.rs
```diff
@@ -139,7 +139,6 @@ pub trait Runtime<S: Spec>:
     /// Messages with equal priority are processed in the order they are received. If messages have a delay configured in [`Runtime::get_transaction_delay_ms`],
     /// they are considered to be "received" after the delay period has elapsed.
     // Returns a u32 so that the sequencer can represent priority as a u64 and have some reserved values that are greater than the maximum priority level of any transaction.
-    #[cfg(feature = "native")]
```
citizen-stig, line 142: Why this is not native anymore?

crates/module-system/sov-modules-api/src/runtime/mod.rs
bkolad: It is, #[cfg(feature = "native")] is declared on the `impl` (not visible in the diff) so it is not needed on the method.

crates/module-system/sov-modules-api/src/runtime/mod.rs
bkolad: same

crates/utils/sov-test-utils/src/runtime/macros.rs
bkolad: I think in this case caller should decide. It is always possible to convert it on the caller side. In our case we just unwrap because StateCheckPoint can't run out of gas. But this method can work with anys state accessor

crates/module-system/sov-modules-api/src/lib.rs
bkolad: `Runtime::resolve_address` uses it, and we blank implement it in some testing macros. Inside the macro, it’s better to access` User` via `sov_modules_api::User`, since `sov_modules_api` is already in the macro’s scope.

crates/full-node/full-node-configs/src/sequencer.rs
bkolad: fixed

crates/utils/sov-test-utils/src/runtime/macros.rs
bkolad: fixed

crates/full-node/sov-sequencer/src/preferred/rate_limiter/mod.rs
bkolad: Fixed

crates/full-node/full-node-configs/src/sequencer.rs
citizen-stig: Probably its worth adding this comment into the code, so someone won't "optimize" it later

citizen-stig: non-blocking

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
citizen-stig: Can we please add this as a comment in the code, in case if another audit happens

crates/utils/sov-test-utils/src/runtime/macros.rs
citizen-stig: Ok, I see.

Another thing, we don't we use `::` here like in other methods

```suggestion
            ) -> Result<S::Address, <ST as ::sov_modules_api::StateWriter<::sov_modules_api::User>>::Error> {
```

crates/module-system/sov-modules-api/src/lib.rs
citizen-stig: So this also means, that resolve address can only work with Kernel/User state and it won't work if we add any other (like with privacy, etc)

citizen-stig (approved):

crates/module-system/sov-modules-api/src/lib.rs
bkolad: So this limitations comes from `Accounts` module. We already use `Accounts::resolve_sender_address` in stf blueprint and if we adde anything else we will need to change the `Accounts ` functionality

bkolad: So this limitations comes from `Accounts` module. We already use `Accounts::resolve_sender_address` in stf blueprint and if we adde anything else we will need to change the `Accounts ` functionality

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
bkolad: fixed

crates/utils/sov-test-utils/src/runtime/macros.rs
bkolad: fixed

PR #2231, opened 2025-12-14
Replace per-tx timeout tokio tasks with FIFO timeout queue. 

theodorebugnet:
# Description

Replaces the individual tokio tasks spawned for every transaction with a single FIFO task pulling timeouts out of an mpsc.

Additionally adds metrics on both the timeout queue mpsc and the main buffer task input mpsc, so we get more visibility into the amount of TXs hanging in the queue.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -33,12 +37,64 @@ use super::Confirmation;
...
+        }
+        Err(mpsc::error::TrySendError::Closed(message)) => Err(mpsc::error::SendError(message)),
+    };
+
+    if result.is_ok() {
+        metrics.queue_depth = sender.max_capacity() - sender.capacity();
+        sov_metrics::track_metrics(|t| {
+            t.submit(metrics);
+        });
+    }
```
preston-evans98: This is going to result in a *lot* of metrics - one for each transaction. Maybe we can aggregate the way we do with the sequencer inner? What we do there is pre-allocate batch of 128 metrics in a Vec, and then impl `Metric` for that vec so that we only have to submit once per 129 metrics

preston-evans98 (approved):

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: Good point - I'll go ahead and implement that before merging

PR #2235, opened 2025-12-16
Refactor: Simplify Sequencer Architecture and Improve Type Naming

LogvinovLeon:
## Summary

This PR refactors the preferred sequencer implementation to improve code organization, simplify type names, and create a cleaner separation between primary and replica sequencer modes. The changes are purely structural with no functional modifications.

## Motivation

This refactoring addresses several opportunities to improve the codebase:

1. **Type naming**: Simplify type names by removing redundant prefixes where module context already provides clarity
2. **Replica mode handling**: Move from runtime `Option` checks to compile-time type system guarantees
3. **Code organization**: Extract initialization logic to improve modularity
4. **Type safety**: Use the type system to enforce primary vs replica behavior differences

### 1. Database Layer Refactoring

Split the database implementation into explicit primary and replica variants:

**Before**:
```rust
pub struct PreferredSequencerDb {
    backend: Option<Box<dyn PreferredSequencerDbBackend>>,
    shutdown_sender: watch::Sender<()>,
}
// Methods had to check: if let Some(backend) = &self.backend { ... }
```

**After**:
```rust
pub trait Db: Send + Sync {
    async fn initial_data(&self) -> Result<(SequenceNumber, Cache)>;
    async fn bulk_insert_txs(...) -> Result<()>;
    // ... other operations
}

pub struct PrimarySequencerDb {
    backend: Box<dyn Backend>,  // Always present
    shutdown_sender: watch::Sender<()>,
}

pub struct ReplicaSequencerDb {
    shutdown_sender: watch::Sender<()>,
}
```

**Benefits**:
- Replica DB has explicit no-op implementations instead of `None` checks
- Type system enforces correct behavior
- Clearer intent - you can see at a glance what replicas do vs don't do
- Eliminates repetitive `if let Some(backend)` patterns

### 2. Type Name Simplification

Streamlined type names where the module context already provides sufficient namespacing:

| Old Name | New Name | Rationale |
|----------|----------|-----------|
| `PreferredSequencerDbBackend` | `Backend` | Within `preferred::db` module |
| `PreferredSequencerReadBlob` | `ReadBlob` | Context is clear from module path |
| `PreferredSequencerReadBatch` | `ReadBatch` | Same |
| `PreferredSequencerCache` | `Cache` | Same |
| `DbSnapshotData` | `SnapshotData` | "Db" prefix redundant in db module |
| `DbEvent` | `Event` | Same |

### 3. Cache Warmup Executor Simplification

**Before**:
```rust
pub struct CacheWarmUpExecutor<S: Spec> {
    inner: Option<CacheWarmUpExecutorInner<S>>,
}
// Every method: let Some(inner) = &self.inner else { return; };
```

**After**:
```rust
pub struct CacheWarmUpExecutor<S: Spec> {
    start_block_notification_sender: watch::Sender<...>,
    tx_sender: flume::Sender<...>,
    size: Arc<AtomicU64>,
}
// If the executor exists, it's active - no need for Option wrapper
```

The executor is now only instantiated for primary sequencers during initialization, so replicas simply don't have one rather than having a disabled one.

### 4. Configuration Type Cleanup

Simplified `is_replica` to a plain boolean:

```rust
// Before
pub struct PreferredSequencerConfig {
    pub is_replica: Option<bool>,
}

// After
pub struct PreferredSequencerConfig {
    pub is_replica: bool,  // Defaults to false
}
```

### 5. Code Organization

Extracted logic into focused modules:

- **`db/primary.rs`** (173 lines): Primary sequencer database with Postgres/RocksDB backend
- **`db/replica.rs`** (72 lines): Replica sequencer database with no-op implementations
- **`initialization.rs`** (312 lines): Builder pattern for sequencer initialization
- **`rpc_errors.rs`** (53 lines): RPC error construction helpers

----

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/full-node-configs/src/sequencer.rs
```diff
@@ -174,7 +174,7 @@ pub struct PreferredSequencerConfig<Address: Copy> {
     /// When enabled, the sequencer runs in replica mode and cannot accept transactions.
     /// It will sync from the master sequencer's database but remain read-only.
     #[serde(default)]
-    pub is_replica: Option<bool>,
+    pub is_replica: bool,
```
bkolad: Currently, the sequencer can only be in a Primary or Replica state, but we plan to introduce a third state, Transitioning, resulting in three possible variants. We may replace the Option with a custom enum for this, but we’ll leave it as is for now.

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -151,19 +151,19 @@ pub enum PreferredSequencerReadBlob<Inner = PreferredSequencerReadBatch> {
...
+            ReadBlob::Proof {
                 sequence_number, ..
             } => *sequence_number,
         }
     }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub(crate) enum DbEvent {
+pub(crate) enum Event {
```
bkolad: Actually, I prefer DbEvent since it’s easier to grep for in our codebase.

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -174,16 +174,16 @@ pub(crate) enum DbEvent {
     ProofBlobAccepted(SequenceNumber),
 }
 
-pub struct PreferredSequencerCache {
-    completed_blobs: VecDeque<PreferredSequencerReadBlob>,
+pub struct Cache {
```
bkolad: Let’s keep the old name, since we have many layers of caches, a more descriptive name is helpful.

crates/full-node/full-node-configs/src/sequencer.rs
LogvinovLeon: Ok. But we still should not have Option<State> but State?
As the sequencer is always in some state

crates/full-node/sov-sequencer/src/preferred/preferred_blob_sender.rs
```diff
@@ -14,65 +14,52 @@ use tokio::task::JoinHandle;
...
         storage_path: Box<Path>,
         tx_status_manager: TxStatusManager<Da::Spec>,
         shutdown_sender: watch::Sender<()>,
         blob_processing_timeout: Duration,
         blobs_sender_channel: broadcast::Sender<BlobExecutionStatus<Da::Spec>>,
-        is_replica: bool,
-    ) -> anyhow::Result<(Self, Option<JoinHandle<()>>)> {
+    ) -> anyhow::Result<(Self, JoinHandle<()>)> {
         let nb_of_concurrent_blob_submissions = Arc::new(AtomicUsize::new(0));
-        if is_replica {
```
bkolad, line 39: This would be a good refactoring, but not for the case where we need to manage replica-to-primary state transitions. The intended pattern is as follows: `PreferredBlobSender, CacheWarmUpExecutor, and PreferredSequencerDb` will handle the transition internally. To do this, they keep an inner: `inner: Option<_>`. When a replica becomes primary, `None` is updated to `Some(..)`. We can’t move this logic outside, as it would become too complicated. So let's move the flag and the Option back into all these tree components. (Soon the `Option` will become a 3 leg enum where we will also have `IsTransitioning` case)

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -376,196 +375,52 @@ impl From<BatchToStore> for StoredBlob {
...
-        }
+/// Factory function to create the appropriate database implementation.
+pub(crate) async fn create_preferred_sequencer_db(
+    shutdown_sender: watch::Sender<()>,
+    is_replica: bool,
+    storage_path: &Path,
+    postgres_config: &Option<PostgresConfig>,
+) -> Result<Box<dyn Db>> {
+    if is_replica {
+        Ok(Box::new(ReplicaSequencerDb::new(shutdown_sender)))
```
bkolad: I get the idea, but this makes the pattern for handling the replica different from `BlobSender and CacheWarmUpExecutor`. We need a uniform approach, since all three components will go through the same state transition when a replica becomes primary. To reduce complexity, the code should handle this consistently across components. `Let’s bring back backend: Option<Box<dyn PreferredSequencerDbBackend>>.`

bkolad (commented):
Most of the naming and file-splitting changes look good to me. However, we still have ongoing work on failover, which will require changes to how the `is_replica` flag is handled. For that reason, the current structure related to `is_replica` is preferable. I suggest we keep the naming and file splitting, but avoid changing how components handle the `is_replica` flag.

crates/full-node/sov-sequencer/src/preferred/preferred_blob_sender.rs
LogvinovLeon: Makes sense. I just don't like the idea of having objects that silently do nothing.
When reading code that does `self.blob_sender.store_blob()` I would expect it to actually store the blob.
In this case - it checks that it's a replica within and noops. In an ideal world - we would have `PrimarySequencer ` and `ReplicaSequencer` that would share some logic but also implement the differences and have smth like `replica.promote() -> Primary`, but agree that this is not the place to do this so reverted.

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
LogvinovLeon: Renamed to `BlobsCache`. Agree?

crates/full-node/full-node-configs/src/sequencer.rs
bkolad: Right now, we read is_replica from the config, but (soon) all nodes will start in a None state and attempt to insert themselves into the database as primary. Only one will succeed; the rest will become replicas.

I’m not yet sure whether, during this competition phase, we should treat those nodes as replicas, None, or introduce another enum, so I’d prefer not to change it until we have more visibility. I’ll start working on this soon.

bkolad (approved):

PR #2239, opened 2025-12-16
Switch API state to concurrent state checkpoint

preston-evans98:
# Description

This PR replaces the `StateCheckpoint` used to serve API state with a new `ConcurrentStateCheckpoint` based on the `concread` crate. This new checkpoint allows writes to proceed while reads are ongoing without locking (using MVCC). This new model significantly reduce the time spent cloning checkpoints in both the side effects task and when accessing API state. 


crates/module-system/sov-modules-api/src/state/accessors/concurrent_state_checkpoint.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+impl<S: Spec> ConcurrentStateCheckpoint<S> {
+    /// Create a `ConcurrentStateCheckpoint` containing the same changes as the given `StateCheckpoint`.
+    pub fn from_state_checkpoint(mut state_checkpoint: StateCheckpoint<S>) -> Self {
+        let map = concread::hashmap::HashMap::new();
+        state_checkpoint.delta.commit_revertable_storage_cache();
+        let mut writer = map.write();
+        for (key, value) in state_checkpoint.delta.user_cache.get_writes() {
+            writer.insert((key.clone(), Namespace::User), value.cloned());
```
bkolad: Nit: Because key/value structures are cheap to clone:  Since we consume `state_checkpoint`, we could expose an API that also consumes `user_cache, kernel_cache, and accessory_writes`, avoiding the need to clone here.

crates/module-system/sov-modules-api/src/state/accessors/concurrent_state_checkpoint.rs
```diff
@@ -0,0 +1,67 @@
...
+        }
+        for (key, value) in state_checkpoint.delta.kernel_cache.get_writes() {
+            writer.insert((key.clone(), Namespace::Kernel), value.cloned());
+        }
+        for (key, value) in state_checkpoint.delta.accessory_writes.iter() {
+            writer.insert((key.clone(), Namespace::Accessory), value.value.clone());
```
bkolad: If we do `state_checkpoint.delta.accessory_writes.into_iter()` then we don't need to call key.clone()/value.clone()

crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
```diff
@@ -264,17 +210,21 @@ pub(crate) enum StateToAccess {
...
+    local_user_writes: HashMap<SlotKey, Option<SlotValue>>,
+    local_accessory_writes: HashMap<SlotKey, AccessoryWrite>,
     uncomitted_changes: Option<Box<dyn StateGetter>>,
     temp_cache: TempCache,
+    // IMPORTANT: Do not re-order the checkpoint before the read_txn - otherwise the read_txn will potentially  be invalidated for a split second on drop.
+    #[debug(skip)]
+    read_txn: HashMapReadTxn<'static, (SlotKey, Namespace), Option<SlotValue>>,
+    // IMPORTANT: Do not re-order the checkpoint before the read_txn - otherwise the read_txn will potentially  be invalidated for a split second on drop.
+    // This is undefined behavior
+    state_checkpoint: Arc<ConcurrentStateCheckpoint<S>>,
```
bkolad: What about encapsulating `read_txn` and `state_checkpoint` in a separate struct, and making `unsafe fn lengthen_lifetime` a method on that struct, so all the tricky logic lives in one place?

bkolad (approved):

crates/module-system/sov-modules-api/src/state/accessors/concurrent_state_checkpoint.rs
```diff
@@ -0,0 +1,67 @@
...
+            writes: Arc::new(map),
+            visible_slot_num: state_checkpoint.visible_slot_num,
+            rollup_height: state_checkpoint.rollup_height,
+        }
+    }
+
+    /// Apply the given `TxChangeSet` to the `ConcurrentStateCheckpoint`.
+    pub fn apply_tx_changes(&self, changeset: TxChangeSet) {
+        let mut writer = self.writes.write();
+        for ((key, namespace), value) in changeset.writes {
```
bkolad: `concread::hashmap::HashMap` has an `extend` method:
```
 let mut writer = self.writes.write();
 writer.extend(changeset.writes.into_iter());
 writer.commit();
```

crates/module-system/sov-modules-api/src/state/accessors/concurrent_state_checkpoint.rs
preston-evans98: Done.

crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
preston-evans98: Done.

crates/module-system/sov-modules-api/src/state/accessors/concurrent_state_checkpoint.rs
preston-evans98: Done!

crates/module-system/sov-modules-api/src/state/accessors/concurrent_state_checkpoint.rs
preston-evans98: Done

PR #2244, opened 2025-12-17
Add some logging to warp metric collection

ross-weir:
# Description

*Summary of the changes...*

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/hyperlane/src/warp/hooks.rs
+++ b/crates/module-system/hyperlane/src/warp/hooks.rs
@@ -15,7 +15,11 @@ impl<S: Spec> Warp<S> {
         let route_ids = self.get_monitored_route_ids();
         let visible_slot = state.current_visible_slot_number();
 
-        tracing::trace!(route_count = route_ids.len(), "Emitting rate limiter metrics at slot {}", visible_slot);
+        tracing::trace!(
+            route_count = route_ids.len(),
+            "Emitting rate limiter metrics at slot {}",
+            visible_slot
+        );
 
         for route_id in route_ids {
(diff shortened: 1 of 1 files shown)


crates/module-system/hyperlane/src/warp/hooks.rs
```diff
@@ -15,11 +15,21 @@ impl<S: Spec> Warp<S> {
         let route_ids = self.get_monitored_route_ids();
         let visible_slot = state.current_visible_slot_number();
 
+        tracing::trace!(
+            route_count = route_ids.len(),
+            "Emitting rate limiter metrics at slot {}",
```
citizen-stig: String interpolation will blow up structured loggin, such as loki

ross-weir: how?

ross-weir: should I use `format!` instead?

citizen-stig: Because it will index body, so instead having single searchable body `            "Emitting rate limiter metrics"` it will have it at which slot, which is increasing.

I would suggest to use it as attribute, same way we have `route_count`

ross-weir: The body will be searchable with `Emitting rate limiter metrics.*`

It's a high cardinality field, is it OK to add it as an attribute?

citizen-stig: > It's a high cardinality field, is it OK to add it as an attribute?

Yep, that's the idea.

citizen-stig (approved):

PR #2248, opened 2025-12-17
Reduce num generations stored for timing oracle

preston-evans98:
# Description

Currently, the timing oracle can be a bottleneck due to very frequent tx submission using generation numbers. Since we store several seconds worth of tx history, the timing oracle accrues a very large account which takes hundreds of micros to deserialize. 

This PR reduces the number of generations that we'll store by artificially multiplying the generation number by 100. 

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/timestamp.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/timestamp.rs
@@ -109,6 +109,8 @@ where
                 .try_into()
                 .expect("Converting unix timestamp to i64 number of milliseconds failed");
 
+            // Artificially inflate the generation number to reduce the number of tx hashes we need to store
+            let generation = timestamp.saturating_mul(100);
             let message = Rt::maybe_set_oracle_timestamp(&runtime, timestamp).expect(
                 "Oracle support must be checked before before spawning update_timestamp_task",
             );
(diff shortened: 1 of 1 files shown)


crates/full-node/sov-sequencer/src/preferred/timestamp.rs
```diff
@@ -109,6 +109,8 @@ where
                 .try_into()
                 .expect("Converting unix timestamp to i64 number of milliseconds failed");
 
+            // Artificially inflate the generation number to reduce the number of tx hashes we need to store
+            let generation = timestamp.saturating_mul(100);
```
theodorebugnet, line 113: Very minor nit: maybe we should do 


```suggestion
            let generation = timestamp.checked_mul(100).unwrap_or_else(|| { tracing::error!("Overflow multiplying timestamp by 100"); timestamp });
```
? If overflow ever happens it'd be nice to get explicit logs of this rather than slowing down to a crawl with every single transaction getting the same max generation number.

Not a practical concern though so non-blocking

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/timestamp.rs
preston-evans98: Going to leave this as is since it shouldn't happen for the next 3 million years or so :). Plus, we'll get a hard error anyway when the bucket fills up.

PR #2253, opened 2025-12-18
Failover: DB rejecting queries from replica.

bkolad:
# Description

1. Updates the SQL statements to ensure the replica cannot modify the sequencer database.
2. If a rollup detects that it has transitioned from `Leader` to `Replica`, it shuts itself down.

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
@@ -1,5 +1,3 @@
-#![allow(dead_code)]
-
 //! Database for sequencer-related data.
 //!
 //! TODO(@neysofu): Remove *all* blocking code inside async functions.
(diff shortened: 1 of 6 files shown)


crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -420,30 +451,41 @@ impl PreferredSequencerDb {
...
+                    tracing::error!(
+                        "initial_data: The primary has become a replica. Shutting down."
+                    );
+                    exit_rollup(&self.shutdown_sender).await;
+                    unreachable!()
+                }
+            }
         } else {
             Ok((
                 0, // TODO this will be revisited when we enable the replica sync task.
```
theodorebugnet, line 485: By the way, not related to this PR, but I think we already have the replica sync task enabled right? If the comment is out of date let's just remove it (but if we need more changes then just a heads-up for a future PR)

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
bkolad: I will continue working on it

PR #2260, opened 2025-12-20
Failover: Improve database error details

bkolad:
# Description

This PR adds more context to PostgreSQL database errors by including the current leader in the error messages for the `Prune` and `CurrentData` operations. This additional detail is intentionally omitted for other operations, as they may be executed in tight loops and we want to avoid adding complexity or overhead to their SQL queries.

## Linked Issues
- Related to # (issue)

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
@@ -36,19 +36,19 @@ use crate::preferred::{exit_rollup, track_in_progress_batch_size};
 #[derive(Debug)]
 pub(crate) enum DbReadOutcome<T> {
     Success(T),
-    AbortedBecauseReplica { db_replica: Option<String> },
+    AbortedBecauseReplica { db_leader: Option<String> },
 }
 
 #[derive(Debug, PartialEq, strum::Display)]
-pub(crate) enum Operation {
+pub(crate) enum FailedOperation {
     BeginBlock,
     BatchAddTxs,
     AddTx,
     EndBlock,
-    Prune { db_replica: Option<String> },
+    Prune { db_leader: Option<String> },
     ReadBatch,
     AddProof,
-    CurrentData { db_replica: Option<String> },
+    CurrentData { db_leader: Option<String> },
 }
 
 #[derive(Debug)]
@@ -56,7 +56,7 @@ pub(crate) enum DbError {
     Database(anyhow::Error),
     ReplicaDisallowed {
         self_node_id: String,
-        operation: Operation,
+        operation: FailedOperation,
     },
 }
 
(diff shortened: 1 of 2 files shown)


crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
```diff
@@ -289,28 +295,34 @@ impl PostgresBackend {
...
-        .await
+        .await?;
+
+        Ok(maybe_leader.map(|l| l.node_id))
     }
 
-    fn is_leader(&self, maybe_leader: Option<SequencerLeader>) -> bool {
-        match maybe_leader {
-            Some(leader) => leader.node_id == self.node_id,
+    fn is_leader(&self, maybe_node_id: &Option<String>) -> bool {
```
theodorebugnet:
```suggestion
    fn is_leader(&self, maybe_leader_id: &Option<String>) -> bool {
```

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -36,13 +36,28 @@ use crate::preferred::{exit_rollup, track_in_progress_batch_size};
 #[derive(Debug)]
 pub(crate) enum DbReadOutcome<T> {
     Success(T),
-    AbortedBecauseReplica,
+    AbortedBecauseReplica { db_replica: Option<String> },
```
theodorebugnet: This is going to have the leader ID, right? So maybe
```suggestion
    AbortedBecauseReplica { db_leader: Option<String> },
```

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -36,13 +36,28 @@ use crate::preferred::{exit_rollup, track_in_progress_batch_size};
...
+}
+
+#[derive(Debug, PartialEq, strum::Display)]
+pub(crate) enum Operation {
```
theodorebugnet: This `Operation` looks like it will only be used in error messages. Otherwise having the `db_replica` param on some operations doesn't make sense in a general context - only for the error messages. WDYT? If you agree then maybe a clearer name would be `FailedOperation`, so it makes sense it's a structure for error messages and not a general selector between operations

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
bkolad: fixed

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
bkolad: fixed

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
bkolad: fixed

PR #2265, opened 2025-12-22
Allow posting reverted txs on chain

preston-evans98:
# Description

*Summary of the changes...*

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/execution_config.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/execution_config.rs
@@ -22,7 +22,7 @@ pub static EVM_EXECUTION_CONFIG: OnceLock<RwLock<EvmExecutionConfig>> = OnceLock
 pub struct EvmExecutionConfigContents {
     /// Whether to publish reverted transactions to the DA layer.
     #[serde(default)]
-    pub publish_reverted_txs: bool,
+    pub preferred_sequencer_publish_reverted_txs: bool,
     /// The default size limit for any pinned bucket.
     #[serde(default = "default_bucket_size_limit")]
     pub default_bucket_size_limit: usize,
@@ -37,7 +37,7 @@ pub struct EvmExecutionConfigContents {
 impl Default for EvmExecutionConfigContents {
     fn default() -> Self {
         Self {
-            publish_reverted_txs: false,
+            preferred_sequencer_publish_reverted_txs: false,
             default_bucket_size_limit: default_bucket_size_limit(),
             privileged_deployer_addresses: vec![],
             known_contracts_and_limits: BTreeMap::new(),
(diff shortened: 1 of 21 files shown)


crates/module-system/sov-modules-api/src/module/spec.rs
```diff
@@ -164,6 +165,10 @@ pub struct Context<S: Spec> {
     sequencing_data: Option<Bytes>,
     /// The rollup address that pays the gas fees for the transaction.
     gas_refund_recipient: S::Address,
+    /// The execution context of the transaction.
+    execution_context: ExecutionContext,
```
bkolad, line 169: Would it be possible to remove `is_preferred_sequencer` and add another case  to `ExecutionContext` like:
```

pub enum ExecutionContext {
    SequencerWarmUp,
    PreferredSequencer,
    StdSequencer
    Node,
}

```

bkolad (approved):

crates/module-system/sov-modules-api/src/module/spec.rs
preston-evans98: Synced with blazej offline. We decided it's preferable to keep these separate because the SequencerType (Preferred vs not) is the same for each transaction even across ExecutionContexts. Replacing the boolean with an enum instead.

crates/module-system/module-implementations/sov-evm/src/execution_config.rs
```diff
@@ -20,6 +20,9 @@ pub static EVM_EXECUTION_CONFIG: OnceLock<RwLock<EvmExecutionConfig>> = OnceLock
 /// Any addresses specified in `known_contracts_and_limits` will be pinned with the specified size limit.
 #[derive(Clone, Debug, serde::Serialize, serde::Deserialize, Eq, PartialEq)]
 pub struct EvmExecutionConfigContents {
+    /// Whether to publish reverted transactions to the DA layer.
+    #[serde(default)]
+    pub publish_reverted_txs: bool,
```
theodorebugnet: If I'm understanding right, this changes the default behaviour for standard sequencers - now they will _always_ include reverted TXs. Is that right?
I feel like that results in `publish_reverted_txs = false` having surprising behaviour for anyone not running a preferred sequencer, as the config value is ignored. I don't have the full context on what behaviour would be correct in the EVM, so assuming this change is desired, how about naming it
```rust
    pub preferred_sequencer_reject_reverted_txs: bool,
```
And then invert the check (`&& preferred_sequencer_reject_reverted_txs` rather than `!publish_reverted_txs`) - since the preferred sequencer having the option to reject them is the special case now, and including them seems to be the default case.

preston-evans98: Good call, I've renamed to `preferred_sequencer_publish_reverted_txs`.

PR #2266, opened 2025-12-22
Failover: Replace `is_replica` with `NodeRole` and add `ReplicaNoLeaderSync` mode

bkolad:
# Description

This PR replaces the `is_replica` flag with a `NodeRole` enum.

We already have three distinct node role cases, and we may want to introduce additional roles in the future, so enum is more appropriate.

```
    /// The node is the leader.
    NodeRole::Leader,
    /// The node is a replica and syncs from the leader.
    NodeRole::Replica,
    /// The node is a replica, but syncing from the leader via Postgres is disabled.
    NodeRole::ReplicaNoLeaderSync,
```

We also moved this setting `PostgreSQL` configurations, since we don't have multi-sequencer setups for RocksDB sequencers.

Here are related PRs in other repos:

1. https://github.com/Sovereign-Labs/sov-rollup-cdk-starter/pull/52
2. https://github.com/Sovereign-Labs/rollup-starter-ansible/pull/51
3. https://github.com/Sovereign-Labs/rollup-starter/pull/92

## Linked Issues
- Related to # (issue)

### DIFF
--- a/crates/full-node/full-node-configs/src/sequencer.rs
+++ b/crates/full-node/full-node-configs/src/sequencer.rs
@@ -129,10 +129,14 @@ pub enum RecoveryStrategy {
     TryToSave,
 }
 
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, JsonSchema)]
+#[derive(Debug, Copy, Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, JsonSchema)]
 pub enum NodeRole {
-    Replica,
+    /// The node is the leader.
     Leader,
+    /// The node is a replica and syncs from the leader.
+    Replica,
+    /// The node is a replica, but syncing from the leader via Postgres is disabled.
+    ReplicaNoLeaderSync,
 }
 
 /// Postgres DB config.
@@ -142,7 +146,7 @@ pub struct PostgresConfig {
     pub postgres_connection_string: String,
     /// Id of the node.
     pub node_id: String,
-    ///
+    /// The role of the node.
     pub node_role: NodeRole,
 }
 
--- a/crates/full-node/full-node-configs/src/sequencer.rs
+++ b/crates/full-node/full-node-configs/src/sequencer.rs
@@ -129,6 +129,12 @@ pub enum RecoveryStrategy {
     TryToSave,
 }
 
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, JsonSchema)]
+pub enum NodeRole {
+    Replica,
+    Leader,
+}
+
 /// Postgres DB config.
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, JsonSchema)]
 pub struct PostgresConfig {
@@ -136,6 +142,8 @@ pub struct PostgresConfig {
     pub postgres_connection_string: String,
     /// Id of the node.
     pub node_id: String,
+    ///
+    pub node_role: NodeRole,
 }
 
 /// Configuration for [`PreferredSequencer`].
@@ -171,10 +179,6 @@ pub struct PreferredSequencerConfig<Address: Copy> {
     pub recovery_strategy: RecoveryStrategy,
     /// Target time in milliseconds to spend executing all the txs in a single batch. Batches will be closed when they exceed this value.
     pub batch_execution_time_limit_millis: u64,
-    /// When enabled, the sequencer runs in replica mode and cannot accept transactions.
-    /// It will sync from the master sequencer's database but remain read-only.
-    #[serde(default)]
-    pub is_replica: Option<bool>,
     #[serde(default = "default_num_cache_warmup_workers")]
     /// The number of workers that warm up the main executor cache.
     pub num_cache_warmup_workers: usize,
@@ -207,7 +211,6 @@ impl<Address: Copy> Default for PreferredSequencerConfig<Address> {
             disable_state_root_consistency_checks: false,
             ideal_lag_behind_finalized_slot: default_ideal_lag_behind_finalized_slot(),
             recovery_strategy: RecoveryStrategy::None,
-            is_replica: Some(false),
             db_event_channel_size: default_db_event_channel_size(),
             batch_execution_time_limit_millis: 6_000, // 6 seconds
             num_cache_warmup_workers: default_num_cache_warmup_workers(),
--- a/crates/full-node/full-node-configs/src/sequencer.rs
+++ b/crates/full-node/full-node-configs/src/sequencer.rs
@@ -133,11 +133,13 @@ pub enum RecoveryStrategy {
 pub enum NodeRole {
     /// This node runs as the leader. The leader is responsible for producing new batches.
     Leader,
-    /// This node runs as a replica, syncing state from the leader.
-    /// Replicas do not publish blobs to the DA layer.
+    /// This node runs as a replica, syncing its state from the leader.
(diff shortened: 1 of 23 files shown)


crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -430,35 +432,32 @@ pub struct PreferredSequencerDb {
...
-                SequencerRole::Replica,
-            ));
-        }
-
-        let backend: Option<Box<dyn DbBackend>> = {
+        let (backend, role): (Option<Box<dyn DbBackend>>, _) = {
             if let Some(postgres_config) = &postgres_config {
-                Some(Box::new(PostgresBackend::connect(postgres_config).await?))
+                match postgres_config.node_role {
+                    NodeRole::ReplicaNoLeaderSync => (None, SequencerRole::ReplicaNoLeaderSync),
```
bkolad, line 441: NodeRole may look similar to SequencerRole, but they are intentionally split so they can evolve independently in the future. It’s likely that, with failover support, SequencerRole will need additional cases.

crates/full-node/full-node-configs/src/sequencer.rs
```diff
@@ -129,13 +129,25 @@ pub enum RecoveryStrategy {
     TryToSave,
 }
 
+#[derive(Debug, Copy, Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, JsonSchema)]
+pub enum NodeRole {
+    /// The node is the leader.
+    Leader,
+    /// The node is a replica and syncs from the leader.
+    Replica,
+    /// The node is a replica, but syncing from the leader via Postgres is disabled.
```
citizen-stig: From slack:

> It doesn't do any sequencing. The database and blob sender are None so it is guaranteed to not write anything to any db.
>  So replica can't mess anything

Can we add the purpose of this mode as you've described

crates/full-node/full-node-configs/src/sequencer.rs
```diff
@@ -129,13 +129,25 @@ pub enum RecoveryStrategy {
...
+}
+
 /// Postgres DB config.
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, JsonSchema)]
 pub struct PostgresConfig {
     /// Connection string.
     pub postgres_connection_string: String,
     /// Id of the node.
     pub node_id: String,
+    /// The role of the node.
```
citizen-stig: Nit: I think this is the case where we can just put `allow(missing_docs)` at the top of the struct =)

examples/demo-rollup/configs/mock_rollup_config.toml
```diff
@@ -110,10 +110,13 @@ ideal_lag_behind_finalized_slot = 3 # Ideal buffer of finalized slots to maintai
 # db_event_channel_size = 10000 # Event buffer size while update_state is running
 
 
-is_replica = false
+
 
 #[sequencer.preferred.postgres_config]
 #postgres_connection_string = "postgresql://user:pass@host/db" # HIGHLY RECOMMENDED: Use Postgres instead of RocksDB to soft confirmed transactions
+#node_id = "node_1"
+#node_role = "Leader"
```
citizen-stig: This is often used as a reference, can we add other options here with brief explanation.

examples/demo-rollup/tests/replica/replica_gets_txs_from_master.rs
```diff
@@ -2,15 +2,33 @@ use super::*;
...
+        Err(CreatePostgresError::DockerError(e)) => {
+            panic!("Failed to create Postgres container: {e}");
+        }
+    };
 
-    let token_id = config_gas_token_id();
+    let key_and_address = read_private_key::<S>("tx_signer_private_key.json");
+    let (da_service, da_shutdown, addr) = create_da_service_periodic().await;
+    // Ideal lag and stuff
+    da_service.wait_for_height(10).await.unwrap();
```
citizen-stig: Where this number is coming from?

citizen-stig: Why do we need it here, before starting a test rollup? Why ` test_rollup.wait_for_sequencer_ready().await.unwrap();` is not enough?

crates/full-node/full-node-configs/src/sequencer.rs
bkolad: Fixed.

crates/full-node/full-node-configs/src/sequencer.rs
bkolad: Fixed

examples/demo-rollup/configs/mock_rollup_config.toml
bkolad: I added the other options, but didn’t expand the comment. The docs here and in the rust docs code tend to fall out of sync quickly, so I prefer not to duplicate them.

examples/demo-rollup/tests/replica/replica_gets_txs_from_master.rs
bkolad: `wait_for_sequencer_read`y does not take the ideal lag into account. The sequencer must wait for this ideal lag before producing new batches. Most of the time, this doesn’t matter, since transactions are accepted and executed in the sequencer’s memory, but for this test, we need batches to be published on-chain, so I make sure the ideal-lag condition is satisfied.

crates/full-node/full-node-configs/src/sequencer.rs
citizen-stig: Maybe github hasn't refreshed it yet, but I see following:

```rust
    /// This node runs as a replica, syncing state from the leader.
    /// Replicas do not publish blobs to the DA layer.
    Replica,
    /// This node runs as a replica, but Postgres-based syncing from the leader is disabled.
    /// It only receives transactions via the DA layer.
    ReplicaNoLeaderSync,
}
```

Which explains what, but not why. And from the slack conversation, "Why not just replica or standard sequecner" is that this mode won't touch any DB except rollup state, so it is the purest for sync/backup or any similar operations

examples/demo-rollup/tests/replica/replica_gets_txs_from_master.rs
citizen-stig: Yes, that's correct. And common patter in other tests is:

```
    test_rollup.produce_enough_finalized_slots().await;
    test_rollup.wait_for_sequencer_ready().await.unwrap();
```

Where function `produce_enough_finalized_slots` handles all needed cases based on config and not hard-coded value in the test.

Because after a while we have a lot of tests with hard-coded value and any small changes makes it super hard to debug and fix.

crates/full-node/full-node-configs/src/sequencer.rs
citizen-stig: And another imporant point worth stating in the comment, that this `ReplicaNoLeaderSync` will have different state root comparing to regular `Replica`, because it won't see not yet submitted txs. It might be clear for devs, but not for rollup operator

citizen-stig (approved):

examples/demo-rollup/tests/replica/replica_gets_txs_from_master.rs
bkolad: fixed

crates/full-node/full-node-configs/src/sequencer.rs
bkolad: fixed

PR #2269, opened 2025-12-26
FlatStateDb: commit kernel and user changesets atomically

### LINKED ISSUE #2267: Prevent database corruption on ungraceful shutdown
## Background

## Design

```rust
enum CommitFlag { 
   Successful { height: u64 },
}

```

On commit: 
3. Commit kernel NOMT.
5. Commit user NOMT
8. Update flag to Successful

bkolad:
# Description

rockbound update: https://github.com/Sovereign-Labs/rockbound/pull/41

This PR refactors the `FlatStateDb::commit` logic so that `kernel and user` data are now applied atomically to `live_db` and `archival_db`, instead of being updated separately.

The most significant change is in `FlatStateDb::commit`.
Now, the commit flow works as follows:

1. Create `live_db_batch: WriteBatch and archival_db_batch: WriteBatch`
2. Lock the common `kernel/user cache`
3. Apply `kernel` updates to `live_db_batch`
4. Apply `user` updates to `live_db_batch`
(steps 3 and 4 are repeated for `archival_db_batch`)
5. Atomically commit both `archival_db_batch` and `live_db_batch`

This required the following changes in `[rockbound](https://github.com/Sovereign-Labs/rockbound)`:

- Added `VersionedDB::<_, DbCache>::update_versioned_db_batch` method, which accepts a mutable reference to a `WriteBatch`

- The cache for both `kernel` and `user` is now protected by a single lock. As a result, the cache must be created outside of `rockbound::VersionedDB` This was achieved by abstracting the injected component via the `CacheForVersionedDB` trait.

From the above, it becomes clear that we should merge

```
pub(crate) user: Arc<VersionedDB<NomtStateValues<UserNamespace>, DbCache>>,
pub(crate) kernel: Arc<VersionedDB<NomtStateValues<KernelNamespace>, DbCache>>,
```
into a single object (something also briefly discussed with @preston-evans98). In fact, it looks like `FlatStateDb` has  turned into that unified object.

We could remove `live_db and archival_db from VersionedDB` and instead inject them as method parameters—similar to what we now do for cache injection. After additional cleanup and renaming, `FlatStateDb` could be moved to `rockbound`.

I initially attempted this refactor in this PR, but it required extensive changes across `rockbound` and the `sov-db` crate. Given the scope, I decided not to proceed further at this time. If this consolidation is still a priority, we should revisit it once the crash-consistency issue is resolved and all relevant unit tests are passing.

see also https://github.com/Sovereign-Labs/rockbound/pull/41

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-db/src/flat_db.rs
+++ b/crates/full-node/sov-db/src/flat_db.rs
@@ -168,7 +168,7 @@ impl FlatStateDb {
         let user_cache = &inner.user_cache;
 
         // 2. Update batches with kernel values.
-        let metrics_kernel = VersionedDB::<_, DbCache>::update_verioned_db_batch(
+        let metrics_kernel = VersionedDB::<_, DbCache>::update_versioned_db_batch(
             &mut live_db_batch,
             &mut archival_db_batch,
             &state.kernel,
@@ -182,7 +182,7 @@ impl FlatStateDb {
         let archival_serialized_size_kernel = archival_db_batch.size_in_bytes();
 
         // 3. Update batches with user values.
-        let metrics_user = VersionedDB::<_, DbCache>::update_verioned_db_batch(
+        let metrics_user = VersionedDB::<_, DbCache>::update_versioned_db_batch(
             &mut live_db_batch,
             &mut archival_db_batch,
             &state.user,
--- a/crates/full-node/sov-db/src/flat_db.rs
+++ b/crates/full-node/sov-db/src/flat_db.rs
@@ -1,12 +1,6 @@
 //! A database to store the flat state of the rollup (i.e. the raw key-value pairs)
 //! used by NOMT.
 
-use std::sync::Arc;
-
-use rockbound::{
-    default_cf_descriptor, rocksdb::ColumnFamilyDescriptor, versioned_db::VersionedDB,
-};
-
 use crate::metrics::nomt::FlatStateCommitMetric;
 use crate::{
     historical_state::StateChanges,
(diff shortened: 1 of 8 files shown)


crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -139,10 +157,133 @@ impl FlatStateDb {
...
+            &mut live_db_batch,
+            &mut archival_db_batch,
+            &state.kernel,
+            version,
+            kernel_cache,
+            &self.live_db,
+            &self.archival_db,
+        )?;
+
+        let serialized_size_kernel = live_db_batch.size_in_bytes();
```
bkolad, line 181: This is different from https://github.com/Sovereign-Labs/rockbound/blob/main/src/versioned_db.rs#L583. From the code, it looks to me that the previous values (in the link) were wrong. 
We should report live_db_batch.size not live_db_batch.size+archival_db_batch.size

crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -139,10 +157,133 @@ impl FlatStateDb {
...
+            &mut live_db_batch,
+            &state.root_hash_batch,
+            &self.live_db,
+        )?;
+
+        // Write batches.
+        self.archival_db.write_db_batch(archival_db_batch)?;
+        self.kernel.store_committed_archival_version(version);
+        self.user.store_committed_archival_version(version);
+        self.live_db.write_db_batch(live_db_batch)?;
```
bkolad, line 210: See the PR description. `self.kernel :VersionedDB and self.user: VersionedDB` should likely be merged into one object, but I’d prefer to do that after we address the crash-consistency issue.

rocbound doesn’t know about `Kernel/User` namespaces—it’s a lower-level layer. `sov-db` currently assumes those updates are tracked separately. The changes here are enough to ensure correct cache locking policy and atomic db writes. I suggest we leave the remaining refactor for a follow-up

crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -139,10 +157,133 @@ impl FlatStateDb {
...
+        // 4.Update root hash.
+        DB::update_db_batch_with_schema_data(
+            &mut live_db_batch,
+            &state.root_hash_batch,
+            &self.live_db,
+        )?;
+
+        // Write batches.
+        self.archival_db.write_db_batch(archival_db_batch)?;
+        self.kernel.store_committed_archival_version(version);
```
preston-evans98, line 208: nit: Let's add a comment that rockbound relies on `store_committed_archival_version` being called before we write the `live_db_batch` to ensure consistency

preston-evans98 (approved):

crates/full-node/sov-db/src/flat_db.rs
bkolad: Will update it in the next PR.

PR #2270, opened 2025-12-26
Add test for money printing vuln. Fix panic at genesis

preston-evans98:
# Description

This PR adds a test which catches a potential misconfiguration which causes REVM to allow money-printing. This bug has never been present in our codebase, but only because another change (apparently accidentally) removes the default trait implementation that uses this config. 

The PR also makes a tiny change to return an error rather than panicking when bad data is provided at genesis.

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/genesis.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/genesis.rs
@@ -2,6 +2,7 @@ use alloy_consensus::constants::KECCAK_EMPTY;
 use alloy_consensus::{EMPTY_OMMER_ROOT_HASH, EMPTY_ROOT_HASH};
 use alloy_primitives::{Address, Bloom, B256, B64, U256};
 use alloy_primitives::{BlockNumber, Bytes};
+use anyhow::bail;
 use revm::primitives::hardfork::SpecId;
 use revm::state::AccountInfo;
 use sov_address::{EthereumAddress, FromVmAddress};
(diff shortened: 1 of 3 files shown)


crates/module-system/module-implementations/sov-evm/src/genesis.rs
```diff
@@ -141,7 +141,7 @@ fn init_spec(config: &EvmGenesisConfig) -> anyhow::Result<Vec<(BlockNumber, Spec
     if spec.is_empty() {
         spec.push((0, SpecId::CANCUN));
     } else if spec[0].0 != 0u64 {
-        panic!("EVM spec must start from block 0");
+        anyhow::bail!("EVM spec must start from block 0");
```
LogvinovLeon: Let's import `anyhow::bail!` and use `bail!` here

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
```diff
@@ -37,6 +37,36 @@ fn test_simple_transfer() {
...
+        &to,
+        TEST_DEFAULT_USER_BALANCE.0.checked_add(1).unwrap(),
+    )
+    .tx;
+
+    let evm = Evm::<S>::default();
+    runner.execute_transaction(TransactionTestCase {
+        input: transfer_tx,
+        assert: Box::new(move |ctx, state| {
+            // assert!(!ctx.tx_receipt.is_successful(), "Transaction should not be successful: {:?}", ctx.tx_receipt);
```
LogvinovLeon: Remove?

LogvinovLeon (approved):

crates/module-system/module-implementations/sov-evm/tests/integration/transactions.rs
preston-evans98: Good catch. Removed

PR #2272, opened 2025-12-29
sov-db: Track DB commit status

### LINKED ISSUE #2267: Prevent database corruption on ungraceful shutdown

bkolad:
# Description

This PR adds tracking for DB commit status.

Previously, the `CommitFlag` mechanism saved the `CommitStatus` to a file, but it had two major limitations:
1. It did not cover all possible database shutdown scenarios.
2. The flag was written after committing a specific DB. It is now written before the commit, which simplifies error handling.

The `CommitStatus` is now written to a file in the `DbGroup::commit` method, and recovery is performed in `NomtStateDb::`validate_commit_flag_and_rollback_if_necssesary.

Unit tests will be included in a follow-up PR.

## Linked Issues
- Related to #2267

### DIFF
--- a/crates/full-node/sov-db/src/commit_flag.rs
+++ b/crates/full-node/sov-db/src/commit_flag.rs
@@ -18,9 +18,9 @@ pub enum CommitStatus {
     CommittingUserNomt([u8; 32]),
     /// A commit of **archival user and kernel state** in the flat-db is in progress.
     CommittingArchivalUserAndKernel,
-    /// A commit of **live live user and kernel state** in the flat-db is in progress.
+    /// A commit of **live user and kernel state** in the flat-db is in progress.
     CommittingLiveUserAndKernel,
-    /// The commit suceeded.
+    /// The commit succeeded.
     Success,
 }
 
--- a/crates/full-node/sov-db/src/commit_flag.rs
+++ b/crates/full-node/sov-db/src/commit_flag.rs
@@ -59,13 +59,13 @@ impl CommitFlag {
                 Ok(status) => Ok(status),
                 Err(err) => {
                     tracing::warn!(error = ?err, "Commit flag file is corrupted, defaulting to completed");
-                    self.write_status(CommitStatus::Completed)?;
+                    self.write_status(&CommitStatus::Completed)?;
                     Ok(CommitStatus::Completed)
                 }
             },
             Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
                 // File not found, create it with COMPLETED status
-                self.write_status(CommitStatus::Completed)?;
+                self.write_status(&CommitStatus::Completed)?;
                 Ok(CommitStatus::Completed)
             }
             Err(e) => Err(anyhow::Error::from(e).context("Failed to read commit flag file")),
@@ -87,7 +87,7 @@ impl CommitFlag {
     /// Returns `anyhow::Result<()>` which is `Ok(())` on successful write, or an error
     /// if any step of the atomic write process fails (e.g., I/O errors, permission issues,
     /// disk full).
-    pub fn write_status(&self, status: CommitStatus) -> anyhow::Result<()> {
+    pub fn write_status(&self, status: &CommitStatus) -> anyhow::Result<()> {
         let message = borsh::to_vec(&status)?;
 
         // Write to a temporary file first
@@ -120,6 +120,7 @@ impl CommitFlag {
         Ok(())
     }
 
+    /// TODO
     pub fn log_reset_instruction(&self) {
         tracing::error!(
             "To reset commit flag, please remove commit flag file: `rm {}`",
@@ -155,7 +156,7 @@ mod tests {
         let in_progress_msg = CommitStatus::InProgress(root_hash);
 
         // 2. Write InProgress
-        flag.write_status(in_progress_msg).unwrap();
+        flag.write_status(&in_progress_msg).unwrap();
         assert_eq!(flag.read_status().unwrap(), in_progress_msg);
 
         let mut f = File::open(dir.path().join(FLAG_FILE_NAME)).unwrap();
@@ -166,7 +167,7 @@ mod tests {
         drop(f);
 
         // 3. Write Completed
-        flag.write_status(CommitStatus::Completed).unwrap();
+        flag.write_status(&CommitStatus::Completed).unwrap();
         assert_eq!(flag.read_status().unwrap(), CommitStatus::Completed);
 
         let mut f = File::open(dir.path().join(FLAG_FILE_NAME)).unwrap();
--- a/crates/full-node/sov-db/src/commit_flag.rs
+++ b/crates/full-node/sov-db/src/commit_flag.rs
@@ -7,14 +7,21 @@ use borsh::{BorshDeserialize, BorshSerialize};
 
 const FLAG_FILE_NAME: &str = "commit_status.flag";
 
-/// Represents the status of a two-phase commit operation.
+/// Represents the current state of a commit operation.
 #[derive(Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, Clone, Copy)]
 pub enum CommitStatus {
-    /// Indicates that the first phase of a commit is done, but the second is pending.
-    /// Write root hash that has been written while in progress.
-    InProgress([u8; 32]),
-    /// Indicates that a commit operation is fully completed, or no operation is in progress.
-    Completed,
+    /// A commit of **kernel state** is in progress.
+    /// Stores the previous `kernel` root hash before the commit began.
+    CommittingKernelNomt([u8; 32]),
+    /// A commit of **user state** is in progress.
+    /// Stores the previous `user` root hash before the commit began.
+    CommittingUserNomt([u8; 32]),
+    /// A commit of **archival user and kernel state** in the flat-db is in progress.
+    CommittingArchivalUserAndKernel,
+    /// A commit of **live live user and kernel state** in the flat-db is in progress.
+    CommittingLiveUserAndKernel,
+    /// The commit suceeded.
+    Success,
 }
 
 /// Manages a persistent flag file to track the state of two-phase commits.
@@ -59,14 +66,14 @@ impl CommitFlag {
                 Ok(status) => Ok(status),
                 Err(err) => {
                     tracing::warn!(error = ?err, "Commit flag file is corrupted, defaulting to completed");
-                    self.write_status(&CommitStatus::Completed)?;
-                    Ok(CommitStatus::Completed)
+                    self.write_status(&CommitStatus::Success)?;
+                    Ok(CommitStatus::Success)
                 }
             },
             Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
                 // File not found, create it with COMPLETED status
-                self.write_status(&CommitStatus::Completed)?;
-                Ok(CommitStatus::Completed)
+                self.write_status(&CommitStatus::Success)?;
+                Ok(CommitStatus::Success)
             }
             Err(e) => Err(anyhow::Error::from(e).context("Failed to read commit flag file")),
         }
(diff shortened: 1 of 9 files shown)


crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -544,8 +538,6 @@ mod tests {
     // Then we create 2 more overlays on top of those 2, let's call them "test overlays".
     // Then the "base" kernel overlay is committed, but the "user" base overlay is kept in memory.
     // This creates a precondition for `StateOverlay::commit` to fail on the user overlay phase.
-    // After this error is confirmed, we commit the "base user" overlay and re-open the NOMT database.
```
bkolad, line 547: This test is not relevant anymore, and will be replaced in the next PR.

crates/full-node/sov-db/src/lib.rs
```diff
@@ -35,7 +35,7 @@ pub mod accessory_db;
 pub mod namespaces;
 
 /// Implements commit flag logic for state_db_nomt.
-pub(crate) mod commit_flag;
+pub mod commit_flag;
```
citizen-stig, line 39: Why should it become public? if it is only for `sov-test-utils`, then let's make it public only if `test-utils` is enabled. 

I don't want to expose too much of internals outside of the crate

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -350,20 +376,6 @@ where
             }
         }
     }
-
-    pub(crate) fn verify_and_fix_commited_root_hashes(&self) -> anyhow::Result<()> {
```
citizen-stig, line 354: Why we don't do this check anymore?

crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
```diff
@@ -138,8 +138,6 @@ where
             "Pruner versions to keep should be at least 1, got {pruner_versions_to_keep}",
         );
         let db_group = DbGroup::new(config)?;
-
-        db_group.verify_and_fix_commited_root_hashes()?;
```
citizen-stig, line 142: Why we don't do this check anymore?

crates/full-node/sov-db/src/commit_flag.rs
```diff
@@ -120,80 +127,12 @@ impl CommitFlag {
         Ok(())
     }
 
-    pub fn log_reset_instruction(&self) {
-        tracing::error!(
-            "To reset commit flag, please remove commit flag file: `rm {}`",
```
citizen-stig, line 125: This was useful if user knows that it just wants to remove flag, they can just copy command and execute it. Why we removed it?

crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -203,10 +208,14 @@ impl FlatStateDb {
             &self.live_db,
         )?;
 
-        // Write batches.
+        // Write archival db batches.
+        commit_flag.save_commit_status(&CommitStatus::CommittingArchivalUserAndKernel)?;
```
citizen-stig, line 212: Nit: what do you think about debug_assert that previous state is matching expected state of NOMT being commited (without actually checking root hash)

crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -19,142 +20,117 @@ const BOTH: &str = "user_and_kernel_state";
...
-            return Err(flag_write_err).with_context(|| {
-                self.commit_flag.log_reset_instruction();
-                "Failed to write IN-PROGRESS status after kernel commit; kernel was rolled back, but flag does not match. Manual intervention required."
-            });
-        }
-
-        debug_assert_eq!(self.commit_flag.read_status()?, in_progress_commit_status);
+        commit_flag.save_commit_status(&CommitStatus::CommittingKernelNomt(
+            self.kernel.root().into_inner(),
+        ))?;
```
citizen-stig, line 119: What happens if NOMT has commited successfully, but writing flag has failed?

crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -19,142 +20,117 @@ const BOTH: &str = "user_and_kernel_state";
...
-        let user = {
-            let opts = config.get_user_options();
-            Nomt::<BinaryHasher<H>>::open(opts)?
-        };
+            CommitStatus::CommittingArchivalUserAndKernel
+            | CommitStatus::CommittingLiveUserAndKernel => {
+                tracing::warn!(
+                "Detected in-progress commit {commit_status:?}. Rolling back kernel & user DBs."
+            );
+                // User & Kernel commit was sucefull but we don't see `Success`. We rollback both User & Kernek.
```
citizen-stig:
```suggestion
                // User & Kernel commit was successful but we don't see `Success`. We rollback both User & Kernel.
```

crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -19,142 +20,117 @@ const BOTH: &str = "user_and_kernel_state";
...
-        }
 
-        let user = {
-            let opts = config.get_user_options();
-            Nomt::<BinaryHasher<H>>::open(opts)?
-        };
+            CommitStatus::CommittingArchivalUserAndKernel
+            | CommitStatus::CommittingLiveUserAndKernel => {
+                tracing::warn!(
+                "Detected in-progress commit {commit_status:?}. Rolling back kernel & user DBs."
```
citizen-stig: nit: Can we make `commit_status` as an attribute?

citizen-stig (changes requested):
Overall is good, left couple of comments

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -47,13 +49,27 @@ where
...
+        // Validate root hashes.
+        Self::are_root_hashes_match(&merklized_state, &flat_state)?;
+
         let accessory_rocksdb =
             AccessoryDb::get_rockbound_options().default_setup_db_in_path(&path)?;
         let ledger_rocksdb = LedgerDb::get_rockbound_options().default_setup_db_in_path(&path)?;
-        let flat_state = FlatStateDb::new(path, state_cache_size, separate_archival_state)?;
+
+        // After validation override the commit flag to Success.
+        commit_flag.write_status(&CommitStatus::Success)?;
```
citizen-stig: What happens if we commited live-db successfully, but failed to write commit status `Success` ?

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: We do it in`DbGroup::new`

crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
bkolad: We do it in `DbGroup::new`

crates/full-node/sov-db/src/commit_flag.rs
bkolad: Why would we ever want to reset the flag manually? The flow has changed and now we save the file before committing the nomt.

crates/full-node/sov-db/src/state_db_nomt.rs
bkolad: We set the flag before committing to Nomt, so if saving the flag fails, we won’t attempt the Nomt commit

crates/full-node/sov-db/src/lib.rs
bkolad: fixed

crates/full-node/sov-db/src/state_db_nomt.rs
bkolad: fixed

crates/full-node/sov-db/src/state_db_nomt.rs
bkolad: fixed

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: We discussed this with Preston when you ware on holiday:

1. We will rollback both the `user` and `kernel` so they remain in a consistent state. 
2. The flat-db and ledger will temporarily be ahead of Nomt, 
3. On startup, the rollup will start from the latest height committed to nomt. After processing the first block, we will override the flat-db/ledger

crates/full-node/sov-db/src/flat_db.rs
bkolad: Inside `save_commit_status` there is a check that the status was saved. 
```
   pub(crate) fn save_commit_status(&self, commit_status: &CommitStatus) -> anyhow::Result<()> {
        self.write_status(commit_status).with_context(|| {
            format!("Failed to write {commit_status:?} status after successful user commit")
        })?;

        debug_assert_eq!(&self.read_status()?, commit_status);
        Ok(())
    }
```

crates/full-node/sov-db/src/commit_flag.rs
```diff
@@ -7,14 +7,21 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
+    /// Stores the previous `kernel` root hash before the commit began.
+    CommittingKernelNomt([u8; 32]),
+    /// A commit of **user state** is in progress.
+    /// Stores the previous `user` root hash before the commit began.
+    CommittingUserNomt([u8; 32]),
+    /// A commit of **archival user and kernel state** in the flat-db is in progress.
+    CommittingArchivalUserAndKernel,
+    /// A commit of **live live user and kernel state** in the flat-db is in progress.
+    CommittingLiveUserAndKernel,
+    /// The commit suceeded.
```
citizen-stig:
```suggestion
    /// The commit succeeded.
```

crates/full-node/sov-db/src/commit_flag.rs
```diff
@@ -7,14 +7,21 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
-    Completed,
+    /// A commit of **kernel state** is in progress.
+    /// Stores the previous `kernel` root hash before the commit began.
+    CommittingKernelNomt([u8; 32]),
+    /// A commit of **user state** is in progress.
+    /// Stores the previous `user` root hash before the commit began.
+    CommittingUserNomt([u8; 32]),
+    /// A commit of **archival user and kernel state** in the flat-db is in progress.
+    CommittingArchivalUserAndKernel,
+    /// A commit of **live live user and kernel state** in the flat-db is in progress.
```
citizen-stig:
```suggestion
    /// A commit of **live user and kernel state** in the flat-db is in progress.
```

crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -19,142 +20,121 @@ const BOTH: &str = "user_and_kernel_state";
...
-            let opts = config.get_user_options();
-            Nomt::<BinaryHasher<H>>::open(opts)?
-        };
+            CommitStatus::CommittingArchivalUserAndKernel
+            | CommitStatus::CommittingLiveUserAndKernel => {
+                tracing::warn!(
+                    ?commit_status,
+                    "Detected in-progress commit. Rolling back kernel & user DBs."
+                );
+                // User & Kernel commit was sucefull but we don't see `Success`. We rollback both User & Kernel.
```
citizen-stig:
```suggestion
                // User & Kernel commit was successful but we don't see `Success`. We rollback both User & Kernel.
```

crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -19,142 +20,121 @@ const BOTH: &str = "user_and_kernel_state";
...
+                self.kernel.rollback(1)?;
+                self.user.rollback(1)?;
+            }
+            CommitStatus::Success => {}
+        }
```
citizen-stig: Do we need to update flag to success at the end, so it is consistent on the next restart?

citizen-stig: Meaning if it accidentally shut down immediately before starting new block

crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -19,142 +20,121 @@ const BOTH: &str = "user_and_kernel_state";
...
-                        hex::encode(current_kernel_root_hash),
+        let user = {
+            let opts = config.get_user_options();
+            Nomt::<BinaryHasher<H>>::open(opts)?
+        };
+
+        Ok(Self { user, kernel })
+    }
+
+    pub(crate) fn validate_commit_flag_and_rollback_if_necssesary(
```
citizen-stig:
```suggestion
    pub(crate) fn validate_commit_flag_and_rollback_if_necessesary(
```

bkolad: fixed

crates/full-node/sov-db/src/state_db_nomt.rs
bkolad: Added the save in `DbGroup::new`

crates/full-node/sov-db/src/state_db_nomt.rs
bkolad: fixed

crates/full-node/sov-db/src/state_db_nomt.rs
citizen-stig: Discussed offline

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
citizen-stig: Discussed offline. Fixes are going to be done in future PRs

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -62,6 +77,8 @@ where
     }
 
     pub(crate) fn commit(&mut self, group: CommitGroup) -> anyhow::Result<()> {
+        // The last commit had to be successful.
+        debug_assert_eq!(&self.commit_flag.read_status()?, &CommitStatus::Success);
```
citizen-stig, line 81: Non blocking: What if it is going to be normal assert? It will add couple ms to the commit time, which is not great, but give extra safety

citizen-stig (approved):

PR #2276, opened 2025-12-30
StorableMockDa: Bump logging and minor optimizations

citizen-stig:
# Description

This PR promotes several log messages from trace to debug. These are happening once per DA block or per submitted blob, so log verbosity won't be huge.

Following messages are going to be printed.

* `New block has been produced` with `producing_time` and `blobs_count`. It does not include bytes.
* `Submitted batch is saved` and `Submitted proof is saved`. Those have bytes and insertion time.

**Produce block optimizaton**.

Stop querying blob data when new block is produced, because data is not used for block header hash calculation, saving I/O budget.

## Testing

## Docs
No updates

### DIFF
--- a/crates/adapters/mock-da/src/storable/layer.rs
+++ b/crates/adapters/mock-da/src/storable/layer.rs
@@ -101,10 +101,11 @@ impl StorableMockDaLayer {
         &mut self,
         timestamp: sov_rollup_interface::da::Time,
     ) -> anyhow::Result<()> {
-        tracing::trace!(
+        let start = std::time::Instant::now();
+        tracing::debug!(
             next_height = self.next_height,
             ?timestamp,
-            "Start producing a new block at"
+            "Start producing a new block"
         );
         if self.next_height >= i32::MAX as u32 {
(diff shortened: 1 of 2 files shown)


crates/adapters/mock-da/src/storable/layer.rs
```diff
@@ -150,11 +158,12 @@ impl StorableMockDaLayer {
         let block_model = block_headers::ActiveModel::from(new_head.clone());
         block_model.insert(&self.conn).await?;
         let _ = self.head_header_sender.send_replace(new_head);
-        tracing::trace!(
+        tracing::debug!(
             blobs_count,
```
bkolad, line 162: Let's also add total blobs size

bkolad (approved):

crates/adapters/mock-da/src/storable/layer.rs
citizen-stig: This is PR in description. 

We print sizes when we save blob.

If we want to print total blobs sizes here, we have only those options:

1. Query full blob data, just to print it. It will cancel optimization for I/O in this PR.
2. Save blob len as a separate field. Doable, but will require bigger change.
3. I don't know if it is possible, to do len of the field in SQL query, but it will also touch the disk.

Is this a deal breaker for this PR?

PR #2279, opened 2025-12-30
Set base fee in EVM blocks

preston-evans98:
# Description

Set the base fee reported in EVM blocks to (first dimension of) the SDK gas price.

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/genesis.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/genesis.rs
@@ -49,7 +49,7 @@ where
         let spec = init_spec(config)?;
         let chain_cfg = evm_chain_config(config, spec);
 
-        let block = init_block(config, 0);
+        let block = init_block(config);
 
         self.cfg.set(&chain_cfg, state)?;
         self.head.set(&block, state)?;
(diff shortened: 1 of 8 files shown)


crates/module-system/module-implementations/sov-evm/src/genesis.rs
```diff
@@ -49,13 +49,13 @@ where
         let spec = init_spec(config)?;
         let chain_cfg = evm_chain_config(config, spec);
 
-        let block = init_block(config, self.base_fee());
+        let block = init_block(config, 0);
```
LogvinovLeon: If we now always pass 0 - let's remove the param

crates/module-system/module-implementations/sov-evm/src/lib.rs
```diff
@@ -220,8 +221,21 @@ where
 }
 
 impl<S: Spec> Evm<S> {
-    pub(crate) fn base_fee(&self) -> u64 {
-        0
+    pub(crate) fn base_fee<
+        Reader: VersionReader + StateReader<User, Error = E> + StateReader<Kernel, Error = E>,
```
LogvinovLeon: We can remove the bound here as it's already defined in the where clause

LogvinovLeon (approved):
LGTM as soon as CI passes

crates/module-system/module-implementations/sov-evm/src/genesis.rs
preston-evans98: Good catch. Fixed

crates/module-system/module-implementations/sov-evm/src/lib.rs
preston-evans98: Fixed!

LogvinovLeon (approved):

PR #2281, opened 2025-12-31
Add new method `DaService::get_approximate_block_time`

citizen-stig:
# Reason

Currently we have following configuration parameters, that customers have to set manually based on DA block time:

```toml
[runner]
da_polling_interval_ms = 50
da_total_timeout_secs = 300

[sequencer]
blob_processing_timeout_secs = 3000

[sequencer.preferred]
batch_execution_time_limit_millis = 2000
```

Customers need to know, how to choose those values and dont' forget to update them when migrating from MockDa to Celestia or any other production ready DA layer.

This PR makes first step towards removing those parametrs or making them optional

If runner and sequencer have block time, they can set their timeouts and polling accordingly without involvement of the human

**Current PR does not have any impact on any of existing runner or sequencer code**

## Testing

## Docs
New method is documented in rollup-interface crate


crates/adapters/mock-da/src/storable/service.rs
```diff
@@ -95,4 +97,15 @@ impl DaService for StorableMockDaService {
     async fn get_signer(&self) -> Option<<Self::Spec as DaSpec>::Address> {
         Some(self.sequencer_da_address)
     }
+
+    async fn get_approximate_block_time(&self) -> Duration {
```
bkolad, line 101: Shouldn't this be an option, so OnBatchSubmit,OnAnySubmit and Manual could return None

bkolad (approved):

crates/adapters/mock-da/src/storable/service.rs
citizen-stig: I thought about it. But I am not in favour of skewing interface because of test-only use-cases.

PR #2282, opened 2025-12-31
Testing utility for checking rollup crashes

### LINKED ISSUE #2267: Prevent database corruption on ungraceful shutdown

bkolad:
# Description

A small utility and example tests for verifying whether the rollup can recover after a hard crash.
The idea is that in the tests, we set the `CrashLocation` via an environment variable, and then we check whether the rollup is healthy after restarting.


examples/demo-rollup/tests/restart/crash.rs
```diff
@@ -0,0 +1,251 @@
...
+            nb_of_events += 1;
+            assert!(
+                nb_of_events < max_nb_of_txs,
+                "The node didn't crash, but it was expected to."
+            );
+        }
+    }
+
+    // Give the OS time to clean up file handles after the crash.
+    tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
```
bkolad, line 144: Unfortunately, I didn't find a way to remove this delay.

crates/full-node/sov-db/src/test_utils.rs
```diff
@@ -290,3 +290,41 @@ fn is_version_selected_for_key(
...
+#[derive(Debug, Clone, Display, EnumString, Eq, PartialEq)]
+pub enum CrashLocation {
+    /// Rollup crashes before saving `the CommittingKernelNomt` flag.
+    BeforeSavingKernelNomt,
+    /// Rollup crashes before committing the kernel.
+    BeforeCommittingKernelNomt,
+    /// Rollup crashes before saving the `CommittingUserNomt` flag.
+    BeforeSavingUserlNomt,
+    /// Rollup crashes before committing the user nomt.
+    BeforeCommittingUserNomt,
```
citizen-stig, line 309: I think it is very important to add failures after commit is done, but before flag is written.

crates/full-node/sov-db/Cargo.toml
```diff
@@ -44,6 +44,7 @@ serde_with = { workspace = true }
 derivative = { workspace = true }
 digest = { workspace = true }
 sov-db-types = { workspace = true, features = ["native"] }
+strum = { workspace = true, features = ["derive"], optional = true}
```
citizen-stig:
```suggestion
strum = { workspace = true, features = ["derive"], optional = true }
```

examples/demo-rollup/tests/restart/graceful.rs
```diff
@@ -0,0 +1,483 @@
+//! Tests for shutdown/restart cases.
```
citizen-stig, line 1: Just to clarify, `graceful.rs` is moved from mod.rs, right?

bkolad: yep

crates/full-node/sov-db/Cargo.toml
bkolad: fixed

citizen-stig (approved):

PR #2288, opened 2026-01-02
Add support for `LedgerDb::rollback`

### LINKED ISSUE #2267: Prevent database corruption on ungraceful shutdown

bkolad:
# Description

This PR introduces the `LedgerDb::rollback` method along with its corresponding tests.
We chose not to reuse the existing `LedgerStateProvider::get_slots` logic because its complex generic signature would require adding intermediary types just to satisfy the API. Even with that in place, additional glue code would still be necessary to construct a SchemaBatch.

## Linked Issues
- Related to #2267


crates/full-node/sov-db/src/ledger_db/mod.rs
```diff
@@ -457,6 +477,8 @@ impl LedgerDb {
             // TODO: Add a method to the slot data trait allowing additional data to be stored
             extra_data: vec![].into(),
             batches: BatchNumber(first_batch_number)..BatchNumber(last_batch_number),
+            discarded_blobs: DiscardedBlobNumber(first_discarded_blob_number)
+                ..DiscardedBlobNumber(last_discarded_blob_number),
```
bkolad, line 481: We now also store information about discarded blobs in the slot, enabling rollback of discarded blobs as well.

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -119,6 +138,83 @@ where
...
+                {
+                    tracing::warn!(
+                        ?commit_status,
+                        "Detected in-progress commit. Rolling back kernel & user DBs."
+                    );
+                    merkelized_state.kernel.rollback(1)?;
+                }
+            }
+
+            CommitStatus::CommittingLedger(ledger_root_hash) => {
```
bkolad, line 189: Note for the reviewer: Only this part is new

crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -543,145 +484,4 @@ mod tests {
...
-    // This test emulates failure behaviour during a commit.
-    // We want to test a case when a kernel database commits, but the user fails.
-    // First, we write some data to both databases to have a reference point.
-    // Then we create 2 overlays, let's call them "base overlays".
-    // They have data, but are not committed.
-    // Then we create 2 more overlays on top of those 2, let's call them "test overlays".
-    // Then the "base" kernel overlay is committed, but the "user" base overlay is kept in memory.
-    // This creates a precondition for `StateOverlay::commit` to fail on the user overlay phase.
-    #[test]
-    fn test_namespaced_db_stay_in_sync_on_error() {
```
bkolad, line 556: This is now tested by the new tests in /demo-rollup

crates/full-node/sov-db/src/ledger_db/mod.rs
```diff
@@ -618,4 +640,166 @@ impl LedgerDb {
...
+        Ok(())
+    }
+
+    fn create_schema_batch_for_rollback(db: Arc<rockbound::DB>) -> anyhow::Result<SchemaBatch> {
+        let mut schema_batch = SchemaBatch::new();
+        let db = DeltaReader::new(db, Vec::new());
+
+        // Get the current head slot
+        let Some((head_slot_number, head_slot)) = db.get_largest::<SlotByNumber>()? else {
+            tracing::debug!("No slots in database, nothing to rollback");
```
citizen-stig: Should this be a warning, because why would we trigger rollback on empty DB?

crates/full-node/sov-db/src/ledger_db/mod.rs
```diff
@@ -618,4 +640,166 @@ impl LedgerDb {
...
+                            if let Some(event) = db.get::<EventByNumber>(&current_event_number)? {
+                                Self::delete_event(
+                                    &mut schema_batch,
+                                    current_tx_number,
+                                    &event,
+                                    current_event_number,
+                                )?;
+                            }
+                        }
+                        Self::delete_tx(&mut schema_batch, &tx, current_tx_number)?;
```
citizen-stig, line 720: Deleting individual txs will produce a lot of thumbstones, and because we know that all tx numbers in in single batch are consecutive deleting range will be prefferable. 

Non blocking, but highly important issue, if we don't want to deal with slow LedgerDb down the line.

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -91,6 +96,21 @@ where
 
         let merklized_commit = self.merklized_state.commit(state, &self.commit_flag)?;
 
+        #[cfg(feature = "test-utils")]
+        crate::test_utils::CrashLocation::BeforeSavingLedger.crash_if_env_set();
+
+        // Immediately after genesis, the ledger contains no entries, and the root hash is initialized to [0; 64].
+        let root_hash = LedgerDb::get_head_root_hash(self.ledger.clone())?.unwrap_or([0; 64]);
```
citizen-stig: Is this reliable approach? `Storage` trait defines `PRE_GENESIS_ROOT` which might not be [0; 64]

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -119,6 +138,83 @@ where
...
+                        ?commit_status,
+                        "Detected in-progress commit. Rolling back kernel & user DBs."
+                    );
+                    merkelized_state.kernel.rollback(1)?;
+                }
+            }
+
+            CommitStatus::CommittingLedger(ledger_root_hash) => {
+                let current_root_hash = LedgerDb::get_head_root_hash(ledger_db.clone())?
+                    .expect("Error: The ledger database does not contain the state root hash.");
```
citizen-stig, line 192: Could we add a comment, why this is not possible?

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -119,6 +138,83 @@ where
...
+                }
+            }
+
+            CommitStatus::CommittingArchivalUserAndKernel
+            | CommitStatus::CommittingLiveUserAndKernel => {
+                tracing::warn!(
+                    ?commit_status,
+                    "Detected in-progress commit. Rolling back kernel & user DBs."
+                );
+                // TODO: Requires careful consideration.
```
citizen-stig, line 213: Can we replace todo comment with todo! so it is panics if happened

crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -18,8 +18,8 @@ const BOTH: &str = "user_and_kernel_state";
 
 /// Contains all the most recent rollup data.
 pub struct NomtStateDb<H> {
-    user: Nomt<BinaryHasher<H>>,
-    kernel: Nomt<BinaryHasher<H>>,
+    pub(crate) user: Nomt<BinaryHasher<H>>,
+    pub(crate) kernel: Nomt<BinaryHasher<H>>,
```
citizen-stig, line 22: nit: non-blocking. To prevent future convolution between different databases, instead of exposing Nomt directly, can we just add 3 small methods:


* `pub(crate) fn rollback_user`
* `pub(crate) fn rollback_kernel`
* `pub(crate) fn rollback_both` (optional, might just keep 2)

citizen-stig (approved):

crates/full-node/sov-db/src/state_db_nomt.rs
bkolad: I would also need methods for user_root() and kernel_root(). I can add them, but I don’t see much advantage over making the fields pub(crate)

crates/full-node/sov-db/src/ledger_db/mod.rs
bkolad: fixed

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: `PRE_GENESIS_ROOT` is not avaliable in this crate but added the following method. 
```
   fn pre_genesis_root() -> [u8; 64] {
        let mut pre_genesis_root = [0u8; 64];
        pre_genesis_root[..32].copy_from_slice(&nomt::trie::TERMINATOR);
        pre_genesis_root[32..].copy_from_slice(&nomt::trie::TERMINATOR);
        pre_genesis_root
    }
```

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: fixed

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: fixed

crates/full-node/sov-db/src/ledger_db/mod.rs
bkolad: Added a followup here:
https://github.com/Sovereign-Labs/sovereign-sdk/issues/2267#issuecomment-3705233210

PR #2289, opened 2026-01-02
`StorageManager`: Simplify state root validation.

### LINKED ISSUE #2267: Prevent database corruption on ungraceful shutdown

bkolad:
# Description

1. Removes `RootHash` from `CommitState`, as this information was redundant—it already exists in the `live_db`, which is committed as the last database.

2. Adds `AllDBsStateRoots`, used for validating state root hashes across different databases. This change replaces the `are_root_hashes_match` method.

## Linked Issues
- Related to #2267

### DIFF
--- a/crates/full-node/sov-db/src/commit_flag.rs
+++ b/crates/full-node/sov-db/src/commit_flag.rs
@@ -11,13 +11,12 @@ const FLAG_FILE_NAME: &str = "commit_status.flag";
 #[derive(Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, Clone, Copy)]
 pub enum CommitStatus {
     /// A commit of **kernel state** is in progress.
-    /// Stores the previous `kernel` root hash before the commit began.
-    CommittingKernelNomt([u8; 32]),
+    CommittingKernelNomt,
     /// A commit of **user state** is in progress.
     /// Stores the previous `user` root hash before the commit began.
-    CommittingUserNomt([u8; 32]),
+    CommittingUserNomt,
     /// /// A commit of **LedgerDB** is in progress.
-    CommittingLedger([u8; 64]),
+    CommittingLedger,
     /// A commit of **archival user and kernel state** in the flat-db is in progress.
     CommittingArchivalUserAndKernel,
     /// A commit of **live user and kernel state** in the flat-db is in progress.
--- a/crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
+++ b/crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
@@ -194,7 +194,7 @@ where
             CommitStatus::Success => {}
         }
 
-        let state_roots = AllDBsStateRoots::from_dbs(&merkelized_state, ledger_db, flat_state_db)?;
+        let state_roots = AllDBsStateRoots::from_dbs(merkelized_state, ledger_db, flat_state_db)?;
         state_roots.validate_all();
 
         Ok(())
@@ -550,27 +550,27 @@ impl AllDBsStateRoots {
 
     fn validate_all(&self) {
         assert_eq!(
-            hex::encode(&self.root_hash_nomt.user),
+            hex::encode(self.root_hash_nomt.user),
             hex::encode(&self.root_hash_from_live_db[0..32])
         );
 
         assert_eq!(
-            hex::encode(&self.root_hash_nomt.kernel),
+            hex::encode(self.root_hash_nomt.kernel),
             hex::encode(&self.root_hash_from_live_db[32..])
         );
 
         assert_eq!(
             hex::encode(&self.root_hash_from_ledger_db),
-            hex::encode(&self.root_hash_from_live_db)
+            hex::encode(self.root_hash_from_live_db)
         );
     }
 
     fn warning_on_rollback(&self, commit_status: &CommitStatus) {
         tracing::warn!(
-            live_db_kernel_root_hash = hex::encode(&self.root_hash_from_live_db),
-            root_hash_ledger_db = hex::encode(&self.root_hash_from_ledger_db),
-            user_nomt_db_root_hash = hex::encode(&self.root_hash_nomt.user),
-            kernel_nomt_db_root_hash = hex::encode(&self.root_hash_nomt.kernel),
+            live_db_kernel_root_hash = hex::encode(self.root_hash_from_live_db),
+            root_hash_ledger_db = hex::encode(self.root_hash_from_ledger_db),
+            user_nomt_db_root_hash = hex::encode(self.root_hash_nomt.user),
+            kernel_nomt_db_root_hash = hex::encode(self.root_hash_nomt.kernel),
             ?commit_status,
             "Detected in-progress commit. Rolling back DBs"
         );
--- a/crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
+++ b/crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
@@ -562,7 +562,7 @@ impl AllDBsStateRoots {
         );
 
         assert_eq!(
-            hex::encode(&self.root_hash_from_ledger_db),
+            hex::encode(self.root_hash_from_ledger_db),
             hex::encode(self.root_hash_from_live_db)
         );
     }
--- a/crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
+++ b/crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
@@ -23,7 +23,7 @@ use crate::namespaces::{KernelNamespace, UserNamespace};
 use crate::pruner::Pruner;
 use crate::schema::namespace::NomtStateValues;
 use crate::schema::tables::ModuleAccessoryState;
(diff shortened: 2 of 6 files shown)


crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -32,6 +32,74 @@ const GIGABYTE: usize = 1024 * 1024 * 1024;
 // without consuming excessive memory.
 pub(crate) const DEFAULT_MAX_PRUNING_BATCH_SIZE: usize = 300_000;
 
+struct AllDBsStateRoots {
+    root_hash_from_live_db: [u8; 64],
+    root_hash_from_ledger_db: [u8; 64],
+    root_hash_nomt: StateRootHashes,
```
bkolad: We will add here hashes also from the other dbs

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -216,6 +262,10 @@ where
             CommitStatus::Success => {}
         }
 
+        let state_roots =
+            AllDBsStateRoots::from_dbs(merkelized_state.clone(), ledger_db, flat_state_db)?;
+        state_roots.validate_all();
```
bkolad, line 198: After all the rollbacks, validate the state roots.

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -32,6 +32,74 @@ const GIGABYTE: usize = 1024 * 1024 * 1024;
 // without consuming excessive memory.
 pub(crate) const DEFAULT_MAX_PRUNING_BATCH_SIZE: usize = 300_000;
 
+struct AllDBsStateRoots {
```
citizen-stig: nit: I personally prefer adding new structs at the end of the file.

When I open `groups.rs`, the first thing I expect is to have main struct, which is `DbGroup` in this case

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -99,12 +166,8 @@ where
         #[cfg(feature = "test-utils")]
         crate::test_utils::CrashLocation::BeforeSavingLedger.crash_if_env_set();
 
-        // Immediately after genesis, the ledger contains no entries, and the root hash is initialized to [0; 64].
-        let root_hash = LedgerDb::get_head_root_hash(self.ledger.clone())?
-            .unwrap_or_else(|| Self::pre_genesis_root());
-
         self.commit_flag
-            .save_commit_status(&CommitStatus::CommittingLedger(root_hash))?;
+            .save_commit_status(&CommitStatus::CommittingLedger)?;
```
citizen-stig, line 102: Why don't we save root hash for ledger anymore?

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -144,58 +207,41 @@ where
         commit_flag: &CommitFlag,
         merkelized_state: Arc<NomtStateDb<H>>,
```
citizen-stig: Non-related, but why it does need arc, and not just & ? Instead of cloning everywhere we can have .as_ref()

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -144,58 +207,41 @@ where
...
-                        ?commit_status,
-                        "Detected in-progress commit. Rolling back kernel DB."
-                    );
+                if state_roots.is_user_nomt_root_newer() {
                     merkelized_state.kernel.rollback(1)?;
                     merkelized_state.user.rollback(1)?;
                 } else
                 // Only Kernel commit was successful. We rollback only the kernel.
                 {
-                    tracing::warn!(
```
citizen-stig, line 181: Why warnings are going away?

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -551,3 +548,36 @@ impl PrunerJob {
...
+    else {
+        return Ok(None);
+    };
+
+    let state_root_hash = HistoricalStateReader::get_serialized_root_hash_from_reader(
+        &historical_state_delta_reader,
+        last_version,
+    )?
+    .unwrap_or_else(|| {
+        // If `last_version`` is persent we must always have root hash.
```
citizen-stig:
```suggestion
        // If `last_version` is present we must always have root hash.
```

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -551,3 +548,36 @@ impl PrunerJob {
...
+
+// Root hash for empty nomt state.
+fn pre_genesis_root() -> [u8; 64] {
+    let mut pre_genesis_root = [0u8; 64];
+    pre_genesis_root[..32].copy_from_slice(&nomt::trie::TERMINATOR);
+    pre_genesis_root[32..].copy_from_slice(&nomt::trie::TERMINATOR);
+    pre_genesis_root
+}
+
+fn hex(data: &[u8]) -> String {
```
citizen-stig: nit: is this function really helpful?

crates/full-node/sov-db/src/commit_flag.rs
```diff
@@ -11,13 +11,12 @@ const FLAG_FILE_NAME: &str = "commit_status.flag";
 #[derive(Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, Clone, Copy)]
 pub enum CommitStatus {
     /// A commit of **kernel state** is in progress.
-    /// Stores the previous `kernel` root hash before the commit began.
-    CommittingKernelNomt([u8; 32]),
+    CommittingKernelNomt,
```
citizen-stig, line 14: Why we are not writing root hashes into flags anymore?

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: Fixed

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: The reason is in the PR description. It is redundant with the state root hash form 'live_db`

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: Fixed

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: It was extracted to  `state_roots.warning_on_rollback(&commit_status)`  (for example in line 170)

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: fixed

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: removed

crates/full-node/sov-db/src/commit_flag.rs
bkolad: The `live_db` is committed last. This allows us to perform the same checks we previously had (with the root hash saved in a file). If any earlier database contains the same root hash as live_db (and the status is different than Success), it indicates that it was not committed in the previous run.

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -551,3 +479,102 @@ impl PrunerJob {
...
+    pre_genesis_root[32..].copy_from_slice(&nomt::trie::TERMINATOR);
+    pre_genesis_root
+}
+
+struct AllDBsStateRoots {
+    // The `live_db` is committed last. We can use `root_hash_from_live_db` to verify
+    // whether all other databases were committed in the previous run.
+    root_hash_from_live_db: [u8; 64],
+    root_hash_from_ledger_db: [u8; 64],
+    root_hash_nomt: StateRootHashes,
```
bkolad, line 517: In the next PRs we will add  root_hash for accessory & archival

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
citizen-stig: Ok, so just to clarify my understanding:

* Live DB has "canonical" root hash, because it is committed the last.
* If root hash from any other DB does not match it, it is rolled back.

Correct?

citizen-stig (approved):

PR #2292, opened 2026-01-02
Allow updating EVM chain configs

preston-evans98:
# Description

This PR adds support for updating the EVM configs set at genesis. This addresses a finding from the internal audit. 

The config may be updated only by the registered admin, and reasonable limits have been set to prevent the admin from completely shutting down the EVM module (which would be a censorship risk). 


crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -82,15 +97,136 @@ where
...
+    fn apply_chain_spec_update(
+        &mut self,
+        chain_spec_update: ChainSpecUpdate,
+        cfg: &mut EvmRuntimeConfig,
+    ) -> anyhow::Result<()> {
+        // Update the contract size limit
+        if let Some(new_limit) = chain_spec_update.new_limit_contract_code_size {
+            ensure!(
+                new_limit < MAX_CONTRACT_CODE_SIZE,
+                "Contract code size limit must be less than 2MB"
```
citizen-stig: Nit: I would embed constant into the error message, so if it is updated we won't have wrong error message

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -82,15 +97,136 @@ where
...
+                new_block_gas_limit > MIN_BLOCK_GAS_LIMIT,
+                "Block gas limit must be greater than 5M to avoid censorship"
+            );
+            cfg.chain_spec.block_gas_limit = new_block_gas_limit;
+        }
+
+        // Update the tx gas limit
+        if let Some(new_tx_gas_limit) = chain_spec_update.new_tx_gas_limit {
+            ensure!(new_tx_gas_limit <= cfg.chain_spec.block_gas_limit, "Tx gas limit must be less than or equal to the effective block gas limit after applying the update");
+            ensure!(new_tx_gas_limit <= MAX_TX_GAS_LIMIT, "Tx gas limit must be less than or equal to 10B to avoid DOS by the sequencer/operator");
```
citizen-stig: nit: same about constant in the message

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -82,15 +97,136 @@ where
...
+                "Contract code size limit must be less than 2MB"
+            );
+            cfg.chain_spec.limit_contract_code_size = Some(new_limit);
+        }
+
+        // Update the block gas limit
+        if let Some(new_block_gas_limit) = chain_spec_update.new_block_gas_limit {
+            ensure!(
+                new_block_gas_limit > MIN_BLOCK_GAS_LIMIT,
+                "Block gas limit must be greater than 5M to avoid censorship"
```
citizen-stig: nit: same about constant in the message

citizen-stig (approved):

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -82,15 +97,136 @@ where
...
+        };
+        ensure!(
+            context.sender() == &admin,
+            "Only the admin can update the runtime configuration. Got {} but expected {admin}",
+            context.sender()
+        );
+        let mut cfg = self.cfg(state)?;
+
+        // Update admin (no validation required)
+        if let Some(new_admin) = update.new_admin {
```
bkolad: Nit: If we destructure the update at the beginning of the method like this:

```
let EvmRuntimeConfigUpdate {
    new_hardfork,
    new_contract_creation_policy,
    chain_spec_update,
    new_admin,
} = update;
```

Then the Rust linter will warn us if any fields are added but not handled in this method.

bkolad (commented):
Looks like some local files were included 'examples/demo-rollup/tests/resync/data/mock_da.sqlite-shm & wall`

bkolad (approved):

citizen-stig:
> Looks like some local files were included 'examples/demo-rollup/tests/resync/data/mock_da.sqlite-shm & wall`

@bkolad I think this is intentional for resync test if CHAIN_HASH changes

crates/module-system/module-implementations/sov-evm/src/call.rs
preston-evans98: Good idea. Done.

crates/module-system/module-implementations/sov-evm/src/call.rs
preston-evans98: Good idea. Fixed all of these.

PR #2293, opened 2026-01-02
Set memory limit on EVM as fallback if gas limit is raised

preston-evans98:
# Description

This PR sets a reasonable memory limit on the EVM in case the tx gas limit is raised significantly. This addresses an audit finding.

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
@@ -196,7 +196,7 @@ where
         block_number: Option<String>,
         state: &mut ApiStateAccessor<S>,
     ) -> Result<ResultAndState, EthApiError> {
-        let block_env = self.resolve_block_env_for_call(block_number, state)?;
+        let block_env = self.resolve_block_env(block_number, state)?;
         let tx_env = prepare_call_env(&block_env, request.clone())?;
         let caller = tx_env.caller;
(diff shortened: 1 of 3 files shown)


crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -363,21 +363,19 @@ where
...
             .get_sealed_block_by_number(block_number, state)?
             .ok_or(EthApiError::UnknownBlock)?;
-        let mut block_env = match maybe_block {
+
+        Ok(match maybe_blcok {
             MaybeSealedBlock::Pending(_) => self.block_env(state).unwrap_infallible(),
             MaybeSealedBlock::Sealed(sealed_block) => BlockEnv::from(sealed_block),
-        };
-        // Set the base fee to zero for evm execution. Gas is paid for by the sov gas meter instead
-        block_env.basefee = 0;
```
LogvinovLeon, line 379: This seems to be a merge issue. This is unrelated to a memory limit

LogvinovLeon (approved):
Approve except for the merge issue

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
preston-evans98: Indeed, good catch. Fixed!

PR #2296, opened 2026-01-02
Add limits on `RawTx` and `FullyBakedTx` sizes

preston-evans98:
# Description

This PR adds size limits on RawTx and FullyBakedTx that are enforced during deserialization.

### DIFF
--- a/crates/rollup-interface/src/state_machine/stf/batch.rs
+++ b/crates/rollup-interface/src/state_machine/stf/batch.rs
@@ -10,23 +10,15 @@ use serde_with::serde_as;
 use crate::Bytes;
 
 /// Maximum allowed size for a raw transaction in bytes (1MB default)
-pub const MAX_RAW_TX_SIZE: usize = 1 * 1024 * 1024;
+pub const MAX_RAW_TX_SIZE: usize = 1024 * 1024;
 
-/// Maximum allowed size for a fully baked transaction in bytes (2MB default)
-pub const MAX_FULLY_BAKED_TX_SIZE: usize = 2 * 1024 * 1024;
+/// Maximum allowed size for a fully baked transaction in bytes 1.05MB default
+pub const MAX_FULLY_BAKED_TX_SIZE: usize = (1024 + 50) * 1024;
 
 /// `FullyBakedTx` represents a serialized signed rollup transaction that has been encoded with
 /// authentication information and is ready to be placed on the DA layer.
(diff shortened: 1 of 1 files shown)


crates/rollup-interface/src/state_machine/stf/batch.rs
```diff
@@ -9,20 +9,16 @@ use serde_with::serde_as;
 
 use crate::Bytes;
 
+/// Maximum allowed size for a raw transaction in bytes (1MB default)
+pub const MAX_RAW_TX_SIZE: usize = 1024 * 1024;
```
citizen-stig, line 13: Nit, non-blocking: Are we fine and set with these values, or we expecte different clients might want them slightly different? 

Because if they do, maybe putting them in `constants.toml` would give us flexibility

crates/rollup-interface/src/state_machine/stf/batch.rs
```diff
@@ -34,6 +30,85 @@ pub struct FullyBakedTx {
...
+        struct FullyBakedTxHelper {
+            #[serde(with = "serde_with::As::<serde_with::base64::Base64>")]
+            data: Bytes,
+            #[serde(with = "serde_with::As::<Option<serde_with::base64::Base64>>")]
+            sequencing_data: Option<Bytes>,
+        }
+
+        let helper = FullyBakedTxHelper::deserialize(deserializer)?;
+
+        let total_size = helper.data.len() + helper.sequencing_data.as_ref().map_or(0, |d| d.len());
```
citizen-stig, line 97: Could we please add a comment here, why it is fine for serde to first deserialize and then check? 

Is it because serde won't be used between ZK guest host comms?

citizen-stig (approved):

crates/rollup-interface/src/state_machine/stf/batch.rs
preston-evans98: Yeah, exactly - it's because transactions coming in from DA are always borsh encoded. Txs coming from the API can use serde, but they're protected by nginx and axum size limits

crates/rollup-interface/src/state_machine/stf/batch.rs
preston-evans98: Happy to add a constant if anyone ever needs it, but I don't expect these to need changes. Going to hold off for now.

PR #2297, opened 2026-01-04
Add support for `ArchivalDb::rollback`

### LINKED ISSUE #2267: Prevent database corruption on ungraceful shutdown

bkolad:
# Description

See relevant update in https://github.com/Sovereign-Labs/rockbound/pull/43

1. Removes the `separate_archival_state` flag; the archival and live Dbs are now always created in separate locations.
2. Adds a rollback mechanism to ArchivalDb.
3. Adds a test that crashes the rollup during an archival commit and verifies database consistency after restart.

Accesorry state rollback will be implemented in the next PR.

## Linked Issues
- Related to #2267

### DIFF
--- a/crates/full-node/sov-db/src/flat_db.rs
+++ b/crates/full-node/sov-db/src/flat_db.rs
@@ -524,7 +524,7 @@ mod tests {
         test_rollback(CrashLocation::BeforeCommittingLive, 1)
     }
 
-    // This test commits data for version 0 of the rollup state and panics at various points during the commit.
+    // This test commits data for version 0 of the rollup state and panics at various points during the commit for version 1.
     // Afterward, it checks whether the rollback logic correctly reverted the archival state.
     fn test_rollback(crash_location: CrashLocation, archival_version: u64) -> anyhow::Result<()> {
         let tempdir = tempfile::tempdir().unwrap();
--- a/crates/full-node/sov-db/src/flat_db.rs
+++ b/crates/full-node/sov-db/src/flat_db.rs
@@ -2,6 +2,7 @@
 //! used by NOMT.
 
 use crate::commit_flag::{CommitFlag, CommitStatus};
+use crate::historical_state::{HistoricalStateReader, STATE_ROOT_HASH_SINGLETON};
 use crate::metrics::nomt::FlatStateCommitMetric;
 use crate::{
     historical_state::StateChanges,
@@ -11,12 +12,14 @@ use crate::{
 };
 use parking_lot::{MappedRwLockReadGuard, MappedRwLockWriteGuard, RwLock};
 use parking_lot::{RwLockReadGuard, RwLockWriteGuard};
+use rockbound::cache::delta_reader::DeltaReader;
 use rockbound::versioned_db::CacheForVersionedDB;
 use rockbound::versioned_db::SchemaWithVersion;
 use rockbound::{
     default_cf_descriptor, rocksdb::ColumnFamilyDescriptor, versioned_db::VersionedDB,
 };
-use rockbound::{rocksdb, DB};
+use rockbound::{rocksdb, SchemaBatch, DB};
+use sov_rollup_interface::common::SlotNumber;
 use std::sync::Arc;
 
 /// A database to store the flat state of the rollup (i.e. the raw key-value pairs)
@@ -41,6 +44,7 @@ impl FlatStateDb {
     ) -> anyhow::Result<Self> {
         let mut columns: Vec<ColumnFamilyDescriptor> =
             vec![default_cf_descriptor(StateRootHashes::table_name())];
+
         VersionedDB::<NomtStateValues<UserNamespace>, DbCache>::add_column_families(
             &mut columns,
             separate_archival,
@@ -55,6 +59,7 @@ impl FlatStateDb {
         let archival_db = if separate_archival {
             let archival_path = path.join(Self::ARCHIVAL_DB_PATH_SUFFIX);
             let archival_columns = vec![
+                default_cf_descriptor(StateRootHashes::table_name()),
                 default_cf_descriptor(
                     NomtStateValues::<UserNamespace>::HISTORICAL_COLUMN_FAMILY_NAME,
                 ),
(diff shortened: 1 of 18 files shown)


crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -237,6 +334,106 @@ impl FlatStateDb {
...
+    fn rollback_archival(
+        &self,
+        current_version_archival: u64,
+        current_version_live: u64,
+        root_hash_live: [u8; 64],
+    ) -> anyhow::Result<()> {
+        let user_changes = keys_and_values_for_rollback(&self.user, current_version_archival)?;
+        let kernel_changes = keys_and_values_for_rollback(&self.kernel, current_version_archival)?;
+
+        let mut state_changes = HistoricalStateReader::materialize_values(
```
bkolad, line 390: We don’t remove entries from the Prunig table, as they will be overridden in the next DA block. Adding this as an item to the tracking issue.

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -105,12 +104,9 @@ where
         crate::test_utils::CrashLocation::BeforeCommittingLedger.crash_if_env_set();
         let ledger_commit = self.commit_ledger(&ledger)?;
 
-        #[cfg(feature = "test-utils")]
-        crate::test_utils::CrashLocation::BeforeSavingArchival.crash_if_env_set();
```
citizen-stig, line 109: Why we don not crash here anymore?

bkolad: I moved this line to `flat_db.rs` so it is closer to the place where we save the flag.

examples/demo-rollup/tests/restart/crash.rs
```diff
@@ -260,10 +290,14 @@ async fn send_txs(
...
 
         // It's fine not to check the result here — it will be verified later via subscription.
-        let _ = api_client.send_tx_to_sequencer(&tx).await;
+        let res = api_client.send_tx_to_sequencer(&tx).await;
+
+        if res.is_err() && std::env::var(CRASH_ENV_NAME).is_ok() {
+            return;
+        }
 
         // Send transactions continuously every 100ms to maintain steady TX traffic during the test.
```
citizen-stig: Comment is not updated with code-change

examples/demo-rollup/tests/restart/crash.rs
```diff
@@ -260,10 +290,14 @@ async fn send_txs(
         );
 
         // It's fine not to check the result here — it will be verified later via subscription.
-        let _ = api_client.send_tx_to_sequencer(&tx).await;
+        let res = api_client.send_tx_to_sequencer(&tx).await;
+
+        if res.is_err() && std::env::var(CRASH_ENV_NAME).is_ok() {
+            return;
```
citizen-stig, line 296: Should we compare error at least, that it matches roughly the panic from crash env?

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -494,6 +494,7 @@ where
         // Once we're this close to `deferred_slots_count`, we risk crossing the
         // `deferred_slots_count` threshold before the next call to
         // `update_state`. That's no good.
+
```
citizen-stig: Looks like random change, that does not belong to this PR

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -105,12 +104,9 @@ where
...
-
         let flat_metrics = self
             .flat_state
-            .commit(historical_state, &self.commit_flag)?;
+            .commit(historical_state, Some(&self.commit_flag))?;
```
citizen-stig, line 109: Why commit flag became optional?

crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -237,6 +334,106 @@ impl FlatStateDb {
...
+
+    /// Validate consistency between live and archival DBs, rollback archival changes if necessary.
+    pub fn validate_and_rollback_archival(&self) -> anyhow::Result<()> {
+        let live_version_and_root_hash = self.latest_version_and_root_hash_live_db()?;
+
+        let Some((current_version_archival, root_hash_archival)) =
+            self.latest_version_and_root_hash_archival_db()?
+        else {
+            assert!(live_version_and_root_hash.is_none());
+            tracing::info!("Roolup Archival & Live DBs are empty, nothing to rollback");
```
citizen-stig:
```suggestion
            tracing::info!("Rollup Archival & Live DBs are empty, nothing to rollback");
```

crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -296,3 +493,347 @@ impl CacheForVersionedDB<NomtStateValues<KernelNamespace>> for DbCache {
...
+        test_rollback(CrashLocation::BeforeCommittingArchival, 0)
+    }
+
+    #[test]
+    fn test_rollback_crash_before_saving_live() -> anyhow::Result<()> {
+        test_rollback(CrashLocation::BeforeSavingLive, 1)
+    }
+
+    #[test]
+    fn test_rollback_crash_before_comitting_live() -> anyhow::Result<()> {
```
citizen-stig:
```suggestion
    fn test_rollback_crash_before_committing_live() -> anyhow::Result<()> {
```

crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -296,3 +493,347 @@ impl CacheForVersionedDB<NomtStateValues<KernelNamespace>> for DbCache {
...
+        fn change_set(&self, version: u64) -> StateChanges {
+            make_change_set(
+                self.get_user_data_for_version(version),
+                self.get_kernel_data_for_version(version),
+                version,
+            )
+        }
+    }
+
+    fn data_to_insert_per_verson() -> TestData {
```
citizen-stig:
```suggestion
    fn data_to_insert_per_version() -> TestData {
```

crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -140,28 +211,29 @@ impl FlatStateDb {
...
+    ) -> anyhow::Result<FlatStateCommitMetric> {
+        self.commit_internal(state_changes, commit_flag, true)
+    }
+
+    fn commit_internal(
+        &self,
+        state_changes: StateChanges,
+        commit_flag: Option<&CommitFlag>,
+        commit_live_db: bool,
     ) -> anyhow::Result<FlatStateCommitMetric> {
```
citizen-stig, line 228: Now that I am looking into this method more, I would like to have more crash points here.

crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -237,6 +334,106 @@ impl FlatStateDb {
...
+            assert_eq!(current_version_archival, 0);
+            anyhow::bail!("The Live DB is empty; please remove the Archival DB manually.")
+        };
+
+        if current_version_archival == current_version_live {
+            assert_eq!(root_hash_archival, root_hash_live);
+            return Ok(());
+        }
+
+        assert_eq!(current_version_archival, current_version_live + 1);
```
citizen-stig:
```suggestion
        assert_eq!(current_version_archival, current_version_live + 1, "Live and Archival DBs are in unrecoverable state.");
```

I think this comment allow to identify the issue quicker. I hope this issue never happens.

crates/full-node/sov-db/src/flat_db.rs
```diff
@@ -202,21 +274,46 @@ impl FlatStateDb {
             archival_db_batch.size_in_bytes() - archival_serialized_size_kernel;
 
         // 4.Update root hash.
+        DB::update_db_batch_with_schema_data(
```
citizen-stig, line 277: Why do we ignore returned metrics?

crates/full-node/sov-db/src/flat_db.rs
citizen-stig: Nevermind, I thought `:update_db_batch_with_schema_data` touches the disk. 

As long as we have crash before any function that touches disk we are fine

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
bkolad: fixed

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
citizen-stig: Discussed offline

bkolad: We don’t save it during rollback. This will be addressed in the next PR, where we’ll remove the need to persist the flag in a file.

citizen-stig (approved):

crates/full-node/sov-db/src/flat_db.rs
bkolad: fixed

crates/full-node/sov-db/src/flat_db.rs
bkolad: fixed

crates/full-node/sov-db/src/flat_db.rs
bkolad: fixed

crates/full-node/sov-db/src/flat_db.rs
bkolad: fixed

examples/demo-rollup/tests/restart/crash.rs
bkolad: fixed

crates/full-node/sov-db/src/flat_db.rs
bkolad: https://github.com/Sovereign-Labs/sovereign-sdk/issues/2267#issuecomment-3705233210

PR #2299, opened 2026-01-05
Add support for  `AccessoryDb::rollback`

### LINKED ISSUE #2267: Prevent database corruption on ungraceful shutdown

bkolad:
# Description

1. Adds `AccessoryDb::rollback`.
2. Adds a `StateRootHashes` table to `AccessoryDb` to enable tracking of historical root hashes.
3. Because all databases now track root hashes, rollback requirements can be inferred directly from those hashes, allowing removal of the commit flag.


crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -132,69 +120,56 @@ where
...
-                    merkelized_state.kernel.rollback(1)?;
-                }
-            }
-            CommitStatus::CommittingUserNomt => {
-                merkelized_state.kernel.rollback(1)?;
-                if state_roots.is_user_nomt_root_newer() {
-                    merkelized_state.user.rollback(1)?;
-                }
-            }
+        if state_roots.is_kerner_nomt_root_newer() {
```
preston-evans98: nit: 
```suggestion
        if state_roots.is_kernel_nomt_root_newer() {
```

bkolad: fixed

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -556,11 +536,25 @@ impl AllDBsStateRoots {
...
     fn check_all(&self) {
+        assert_eq!(
+            hex::encode(self.root_hash_from_archival_db),
+            hex::encode(self.root_hash_from_live_db)
+        );
+
+        assert_eq!(
+            hex::encode(self.root_hash_from_accessory_db),
+            hex::encode(self.root_hash_from_live_db)
+        );
```
preston-evans98, line 556: nit: it would be helpful to add an error message describing what `left` and `right` are

crates/full-node/sov-db/src/accessory_db.rs
```diff
@@ -57,19 +62,119 @@ impl AccessoryDb {
...
+            &mut accessory_db_batch,
+            accessory_bach,
+            accessory_db,
+        )?;
+
+        rockbound::DB::update_db_batch_with_schema_data(
+            &mut accessory_db_batch,
+            root_hash_batch,
+            accessory_db,
+        )?;
```
preston-evans98, line 106: (Question): Why do we use separate batches for `root_hash_batch` and `accessory_batch` here?

preston-evans98 (approved):

crates/full-node/sov-db/src/accessory_db.rs
bkolad: The higher-level code is structured this way. Inside `DbGroup::commit,` we receive a `SnapshotGroup { accessory, ..historical_state { root_hash_batch } }`, which then has to be merged here. I haven’t changed this yet because it started to spiral in scope, but I’ll open an issue to address it.

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
bkolad: Will fix it in the next PR.

PR #2316, opened 2026-01-09
Error on finalized headers going backwards

citizen-stig:
# Description

Runner state manager is going to check if new head during reorg went below last finalized header.

## Testing

## Docs
No updates

### DIFF
--- a/crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
+++ b/crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
@@ -241,9 +241,11 @@ async fn background_header_fetch_task<Da: DaService>(
                                     last_seen = %previously_seen_finalized_header.display(),
                                     "Critical error in DaService, finalized header when backwards");
                             }
+                        } else {
+                            // Preventing adding rolled back finalized header
+                            last_seen_finalized_header = Some(finalized_header.clone());
+                            recent_headers.insert_new_header(finalized_header);
                         }
-                        last_seen_finalized_header = Some(finalized_header.clone());
-                        recent_headers.insert_new_header(finalized_header);
                     }
                     FutureOrShutdownOutput::Output(Err(error)) => {
                         // DaService should do all retries, so we just stop and fail.
--- a/crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
+++ b/crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
@@ -231,10 +231,10 @@ async fn background_header_fetch_task<Da: DaService>(
                             None => true,
                             Some(prev_seen) => {
                                 if finalized_header.height() < prev_seen.height() {
-                                    tracing::error!(
+                                    tracing::warn!(
                                     received = %finalized_header.display(),
                                     last_seen = %prev_seen.display(),
-                                    "Critical error in DaService, finalized header when backwards");
+                                    "finalized header when backwards in DaService. This update won't be propagated to consumers of `DaServiceWithCachedFinalizedHeaders`");
                                     false
                                 } else {
                                     true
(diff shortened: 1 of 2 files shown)


crates/full-node/sov-stf-runner/src/da/finalized_headers_cache.rs
```diff
@@ -225,11 +227,29 @@ async fn background_header_fetch_task<Da: DaService>(
...
-                            tracing::info!("All DA header receivers dropped, shutting down");
-                            break;
+                        let is_received_header_valid = match last_seen_finalized_header.as_ref() {
+                            None => true,
+                            Some(prev_seen) => {
+                                if finalized_header.height() < prev_seen.height() {
+                                    tracing::error!(
+                                    received = %finalized_header.display(),
+                                    last_seen = %prev_seen.display(),
+                                    "Critical error in DaService, finalized header when backwards");
```
preston-evans98: This is expected behavior from RPCs. They multiplex requests across nodes, so this isn't an error case. Let's log it at debug level so that we can see it but remove the `Critical error` message

preston-evans98: On further thought, I'd even be fine with `warn` level here - but error seems excessive

citizen-stig: Updated!

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -602,6 +605,13 @@ where
                     return Ok(fork_point);
                 }
                 ForkPointSearchResult::HeadChanged(new_head) => {
+                    if new_head.height() < last_finalized_header.height() {
+                        anyhow::bail!(
```
citizen-stig, line 609: Should we bail here or try another iteration of the loop? Bailing will stop runner and the node, continuing will give a couple more rounds, but probably won't help if RPC provider fell behind by a lot

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -581,6 +581,9 @@ where
             .expect("Choosing fork point only possible if some transitions have been seen");
 
         let mut head = da_service.get_head_block_header().await?;
+        let last_finalized_header = self
+            .finalized_headers_provider
+            .get_last_finalized_block_header()?;
```
preston-evans98, line 586: I think we're making the problem (slightly) worse here. We really should not be getting this info from the DA service at all, because the height that it reports can jump around. Instead, we should be tracking the highest finalized header that has been processed by the node *in one central location*. (Note that the latest *processed* finalized header has no guaranteed relationship to the latest finalized header returned by the DA layer.)

So the logic should be something like...
1. Fetch the block at `next_da_height`:
  - Check if we've already applied a block with this hash. If so, increment next_da_height and `continue`
  -  Otherwise, try to find the local fork containing prev_hash. If one exists, apply the block to it. If not, decrement next_da_height and try again. If you reach the `local_finalized_height` without finding a block, panic.
2. Now that we've applied a block, finalize. 
  - Fetch latest finalized block from DA. If its prev_hash is the hash of the previous finalized block and we've already processed this block locally, finalize it. if it's height is less than local_finalized_height, ignore it - continue the main loop. SOtherwise, recursively
    -   Look for a local block matching the `prev_hash` from the DA service's finalized block. If one exists in `state_on_block` and _that_ block is the child of `local_finalized_block`, finalize it. Then finalize the successor.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
preston-evans98: Ideally we shouldn't cause a panic here. This isn't unexpected/unrecoverable behavior, so panicking isn't the right response

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: Discussed offline. All cases will be handled by https://github.com/Sovereign-Labs/sovereign-sdk/pull/2317

I might add single u64 in Cached header provider and track it, just for indication purposes. Backward will probably be logged with info

PR #2317, opened 2026-01-09
More strict handling of finalized headers in `StateManager`

citizen-stig:
# Description

The Problem Before:

When `state_on_block` was empty, `has_reorg_happened()` would call `da_service.get_last_finalized_block_header()` to determine if a block was valid. 
This created a race condition - the DA's finalized block could advance between calls, leading to inconsistent decisions.

The Solution:
 - Store genesis_da_height and last_processed_finalized_header in StateManager
 - For new blocks, require they be direct descendants of last_processed_finalized_header (not just any finalized block from DA)
 - Update last_processed_finalized_header only after transitions are actually processed and finalized locally
 - This makes reorg detection deterministic based on local state, not external DA queries

Main goal:

Stricter invariants - ensure that new blocks are **direct descendants** of the last processed finalized header when state_on_block is empty

* Move `genesis_da_height` from param in `process_slot` to internal state of `StateManager` That allows it to handle continuation of transitions more carefully
* Fixes in MockDa, different variants of genesis block header are not acceptable anymore
* Fix in crash tests. Previously MockDa was restart from genesis everytime (because it was in memory) and node just waits it to progress. But now node makes query early on to block header, which does not have wait mechanism and fails early. 

## Testing

## Docs
Updated docstring, and expanded on comments


crates/adapters/mock-da/src/storable/layer.rs
```diff
@@ -314,18 +323,15 @@ impl StorableMockDaLayer {
     }
 
     pub(crate) async fn get_block_header_at(&self, height: u32) -> anyhow::Result<MockBlockHeader> {
-        if height >= self.next_height {
-            anyhow::bail!("Block at height {} has not been produced yet", height);
-        }
-        if height == 0 {
-            return Ok(GENESIS_HEADER);
-        }
```
citizen-stig, line 322: This check is duplicated, `self.get_header_at` does it too

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -583,6 +598,7 @@ where
         let mut head = da_service.get_head_block_header().await?;
 
         for attempt in 0..MAX_REORG_FINDING_ATTEMPTS {
+            // TODO: new head goes below
```
citizen-stig: Address this

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -602,6 +618,7 @@ where
                     return Ok(fork_point);
                 }
                 ForkPointSearchResult::HeadChanged(new_head) => {
+                    // TODO: new head goes below last processed finalized height. Sleep here? Continue? do not updated head?
```
citizen-stig: Address this

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -941,6 +984,31 @@ where
             finalized_transitions = finalized_transitions.len(),
             "Completed check for finalized transitions"
         );
+
+        // TODO: Verify that finalized transitions connected to last processed finalized header,
```
citizen-stig: Address this

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -795,11 +828,11 @@ where
...
     }
 
-    /// Returns all [`StateTransitionInfo`] which are below finalized height
-    /// and relevant LedgerDb changes.
+    /// Returns all [`StateTransitionInfo`] which are below finalized height at this point.
     async fn process_finalized_state_transitions(
         &mut self,
     ) -> anyhow::Result<Vec<StateOnBlock<Da::Spec, StateRoot>>> {
+        // This is going to be cut off point
         let last_finalized_header = self
```
citizen-stig, line 862: What happens if this is below last processed finalized header?

Then it is probably won't finalize anything

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -941,6 +984,31 @@ where
...
+
+        // TODO: Verify that finalized transitions connected to last processed finalized header,
+        // Thath they are continous, etc.
+
+        if let Some(last_processed_transition) = finalized_transitions.iter().last() {
+            self.last_processed_finalized_header = last_processed_transition.block_header.clone();
+        }
+
+        // Verify invariant: all earliest seen transitions must descend from last_processed_finalized_header
+        #[cfg(debug_assertions)]
```
citizen-stig, line 1033: Should these be normal assertions?

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -941,6 +1009,57 @@ where
...
+            );
+        }
+
+        if let Some(last_processed_transition) = finalized_transitions.iter().last() {
+            self.last_processed_finalized_header = last_processed_transition.block_header.clone();
+        }
+
+        // Verify invariant: all earliest seen transitions must descend from last_processed_finalized_header
+        #[cfg(debug_assertions)]
+        if let Some(earliest_blocks) = self.seen_on_height.first_key_value() {
```
citizen-stig, line 1034: Extract this into a method

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -941,6 +1009,57 @@ where
             finalized_transitions = finalized_transitions.len(),
             "Completed check for finalized transitions"
         );
+
+        // ---
+        // Check that finalized transitions are valid
+        // First is connected to current
+        if let Some(first_processed_finalized_transition) = finalized_transitions.first() {
```
citizen-stig: Extract this into a function

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -469,61 +478,66 @@ where
         Ok(())
     }
 
-    /// Returns true, if passing block_header is not an incremental continuation of the current canonical chain.
+    /// Returns true, if passed `block_header` is not an incremental continuation of the current chain.
     async fn has_reorg_happened(
```
preston-evans98: nit: This no longer needs to be async

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -469,61 +478,66 @@ where
...
-            if block_header.hash() == past_finalized_block.hash() {
-                tracing::trace!("Passed block header has been finalized in the past => no reorg");
-                return Ok(false);
+            if block_header.hash() == self.last_processed_finalized_header.hash() {
+                // We are at genesis
+                if self.last_processed_finalized_header.height() == self.genesis_da_height {
+                    tracing::trace!("empty state_on_block => genesis");
+                    return Ok(false);
+                }
+                debug_assert_eq!(
```
preston-evans98: nit: This looks like it could be a real assertion

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -469,61 +478,66 @@ where
...
+                    block_header.prev_hash(),
+                    self.last_processed_finalized_header.prev_hash(),
+                    "Corrupt DA"
+                );
+                anyhow::bail!("Trying to process same finalized header twice.");
             }
-            tracing::trace!(
-                "This block header has not been finalized in the past => reorg happened"
+            assert!(
+                block_header.height() > self.last_processed_finalized_header.height(),
```
preston-evans98: (Question): Should we assert that block_header.height() == self.last_processed_finalized_header.height() + 1 here? Or is there a scenario where we expect a different value?

preston-evans98: Oh, nevermind - if the block number is further in the future we want to handle that.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
preston-evans98: nit: We might want to consider renaming this because it returns "true" even when the chain hasn't reorged if the block header is not the child of the last finalized block. This could happen if we had even a very temporary p2p issue which caused us to see two blocks at once.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
preston-evans98: On even further digging this should be impossible. `fetch_block_reorg_aware` always returns either the requested DA height or a *lower* height, which means that seeing a future height is a logic error.

preston-evans98: Thinking out loud: fact that the behavior of this method depends on non-local information about the `filtered_block` seems unfortunate; maybe we can modify the control flow of process_next_slot to differentiate the normal case (fetched the next block in sequence) from the unusual case (reorg detected); then we could put all of the short-circuiting cases on a separate path and it would be obvious which assertions were safe/necessary.

citizen-stig: You're right, we can tighten the assert.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -25,7 +25,7 @@ use sov_rollup_interface::zk::StateTransitionWitness;
 use sov_rollup_interface::{ProvableHeightTracker, StateUpdateInfo};
 use tokio::sync::watch;
 
-const MAX_REORG_FINDING_ATTEMPTS: u8 = 30;
+const MAX_REORG_FINDING_ATTEMPTS: u16 = 10_000;
```
preston-evans98, line 28: (Question): Why did we bump the attempts to such a large value? I see the commit is just named "Whoopsie"...

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: Wait. This is for the block has has been passed by the caller. So original idea was not be too strict on how runner calls the `prepare_storage`

So caller can pass a future block, we don't consider that an error, but we consider that a reorg. And future block can be passed.

I see that we make this connection in mind "seen_on_height empty, means instant finality". But it might not always be the case. What if all block that we processed are finalized, but node just slow to get new blocks. Technically runner should not jump forward. Are we ok to put this as panic? Or as error?

> fetch_block_reorg_aware always returns either the requested DA height or a lower height,

Hm, can you elaborate on that? If finality is non instant and we are jumping between bunch of different forks, it can be higher height.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -625,16 +657,33 @@ where
...
         highest_seen_height: u64,
     ) -> anyhow::Result<ForkPointSearchResult<Da, StateRoot>> {
+        let last_processed_finalized_height = self.last_processed_finalized_header.height();
+        if head.height() <= last_processed_finalized_height {
+            tracing::info!(
+                passed = head.height(),
+                last_processed_finalized_height,
+                "Passed height is below or equal of what has been processed"
+            );
+            return Ok(ForkPointSearchResult::DaHeadIsBelowProcessed);
```
preston-evans98: nit: I think `DaHeadIsBelowFinalized` would be slightly clearer

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: That's what Cem requested: do not shutdown earlier, but wait for node to catchup.

Whoopsie comment was about me forgetting update from u8 to u16, original commit is https://github.com/Sovereign-Labs/sovereign-sdk/pull/2317/changes/a3acb17f4bc2c9eccf56ceb6cd51b30db5b8c828

It will include following cases:

 * DaIsBelowHead it will sleep defined amount of time and retry this number of time
 * Underlying DA is forking so aggressively, so state manager cannot find the new candidate fast enough. This was original upper bound, but I think the same logic applies, let node keep trying, sequencer is alive and API is working, but "non ready" alert will trigger, so we can investigate what is happening.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -625,16 +657,33 @@ where
...
+        }
+
         let mut low = earliest_seen_height;
         let mut high = std::cmp::min(highest_seen_height, head.height()).saturating_add(1);
 
+        assert_eq!(
+            last_processed_finalized_height
+                .checked_add(1)
+                .expect("End of chain"),
```
preston-evans98:
```suggestion
                .expect("Slot number overflow: this should be unreachable"),
```

citizen-stig: By the way, the file is non consistent about it.

Should we always do `checked` add or saturation is fine too?

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: I've decided to use `DaHeadIsBelowProcessedFinalized` to be in line with `processed_finalized` concept rather than keeping vague `finalized` which might imply finalized returned by the DA.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: Could you please elaborate?

What does it mean seeing 2 blocks at once?

We only enforce it to be "child of the last finalized block" in case of instant finality. It should be child of last processed, which might be not finalized, if finality is not instant.

preston-evans98 (approved):
@citizen-stig and I made a plan to further improve this logic as a fast-follow. We're going to merge this as-is with one exception: `get_new_head_height_if_roll_back` needs to be improved to ensure it doesn't break during shutdown.

Approving assuming that change is made

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: Will be changed in the future refactoring

citizen-stig:
> @citizen-stig and I made a plan to further improve this logic as a fast-follow. We're going to merge this as-is with one exception: `get_new_head_height_if_roll_back` needs to be improved to ensure it doesn't break during shutdown.
> 
> Approving assuming that change is made

@preston-evans98 change is in https://github.com/Sovereign-Labs/sovereign-sdk/pull/2317/changes/d9e078ca6de9029855c226fb7ce61c9494508ef8

PR #2318, opened 2026-01-12
Ensure replicas shut down gracefully when stop height is reached

theodorebugnet:
# Description

When `--stop-at-rollup-height` is passed to a rollup, replicas mostly work correctly, except that there was no explicit handling in the replica event handler. So if the master sequencer finished the upgrade and opened a batch at a new version, replicas would process the pg event and also try to open a batch, then panic because of the `BatchCreationError`.

This adds logic to make the replica loop a no-op while waiting for the node to shut down the process, avoiding the panic. An alternative way of handling this would be by sending a signal to stop the event receiver task and unsubscribe from postgres, but that would add another channel or adding a complex type and extra semantics to the `ready_to_process_db_events_recv`, and IMO is not worth the plumbing.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/replica/replica_sync_task.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/replica/replica_sync_task.rs
@@ -162,7 +162,14 @@ impl ReplicaSyncTask {
                                     break 'outer;
                                 }
                                 FutureOrShutdownOutput::Output(Some(_)) => {
-                                    // Discard - past stop height
+                                    // Discard - past stop height.
+                                    // As mentioned above, this assumes that we will always shut
+                                    // down after reaching the stop height. The node will process
+                                    // the on-disk state up to the stop height and shut down
+                                    // immediately, so these events only affect the in-memory state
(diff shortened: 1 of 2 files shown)


crates/full-node/sov-sequencer/src/preferred/replica/replica_sync_task.rs
```diff
@@ -137,6 +139,34 @@ impl ReplicaSyncTask {
...
+                        loop {
+                            let fut =
+                                future_or_shutdown(db_data_receiver.recv(), &shutdown_receiver);
+                            match fut.await {
+                                FutureOrShutdownOutput::Shutdown
+                                | FutureOrShutdownOutput::Output(None) => {
+                                    break 'outer;
+                                }
+                                FutureOrShutdownOutput::Output(Some(_)) => {
+                                    // Discard - past stop height
```
preston-evans98: Is it safe to silently drop events? If so, let's add documentation about why

preston-evans98 (approved):

PR #2324, opened 2026-01-14
typescript: Split ledger-solana signers into their own entrypoints to avoid having to bring in ledger dependencies unless necessary

theodorebugnet:
# Description

This was breaking browser builds due to the `node-hid` transport dependency being pulled in dynamically. This change should let bundlers skip building the dependency entirely if it's unused.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


typescript/packages/signers/package.json
```diff
@@ -21,6 +21,16 @@
       "types": "./dist/wasm.d.ts",
       "require": "./dist/wasm.js",
       "import": "./dist/wasm.mjs"
+    },
+    "./ledger-solana/browser": {
```
ross-weir: We can make this into one export `./ledger-solana` and use conditionals so then the right one is automatically selected based on env, so users dont need to try to select the right entrypoint when importing. Like:

```
  "./ledger-solana": {
    "node": {
      "types": "./dist/ledger-solana/node.d.ts",
      "require": "./dist/ledger-solana/node.js",
      "import": "./dist/ledger-solana/node.mjs"
    },
    "default": {
      "types": "./dist/ledger-solana/browser.d.ts",
      "require": "./dist/ledger-solana/browser.js",
      "import": "./dist/ledger-solana/browser.mjs"
    }
  }
  ```
  
 Then we could also rename the entrypoint from `browser` because even if it's only running the browser it can be consider either browser or bundler, etc. So maybe `default` instead of `browser`

File: typescript/packages/signers/src/index.ts [hunk 2]
```diff
@@ -3,4 +3,4 @@ export * from "./snap";
 export * from "./ed25519";
 export * from "./secp256k1";
 export * from "./privy";
-export * from "./ledger-solana";
+export * from "./ledger-solana/types";
```
ross-weir, line 6: nit: We could probably move this export to each of the `ledger-solana` entrypoints so we dont have unneeded types available for users not using ledger-solana

ross-weir (approved):

typescript/packages/signers/src/index.ts
theodorebugnet: The reason it's available here is for the `web3` package's solana-signable-rollup, which has an `"auto"` authenticator that needs to compare the type. But it's up to the user to import the concrete ledger-solana entrypoint so solana-signable-rollup.ts can't depend on either one, hence it's exported from index.ts.

While it's likely not useful it's also tiny and IMO it's not conceptually wrong: the type exported is for checking if something is a ledger signer, you could e.g. use this to implement a and print a message "ledger signing is not supported" without importing the ledger signers. Probably not a particularly important usecase but I think this export isn't necessarily architecturally wrong.

PR #2328, opened 2026-01-14
New randomization behaviour for MockDa: Rewind below last finalized height

citizen-stig:
# Description

Adds new `RandomizationBehaviour` variant, that allows MockDaService to behave like an RPC that is syncing, thus returns head or last finalized header waay before it returned previuosly.

This test already caught and bad invariant in process_finalized_stf changes.

### What it does

Makes get_head_block_header() and get_last_finalized_block_header() randomly return block headers below the actual finalized height. This simulates DA layer inconsistencies where stale data is reported, useful for testing rollup resilience.

### Configuration

```rust
  RandomizationConfig {
      seed: HexHash::new([42; 32]),
      reorg_interval: 5..10,  // Triggers ~10-20% of the time
      behaviour: RandomizationBehaviour::RewindBelowLastFinalized {
          max_depth: 5,  // How far below finalized to go
      },
  }
```

### Behavior

  - Trigger probability: 1 / (reorg_interval.end - reorg_interval.start)
  - When triggered:
    - get_last_finalized_block_header() returns a random height in range [finalized - max_depth, finalized]
    - get_head_block_header() returns a different random height, but always >= finalized (never below what finalized reports)
  - Deterministic: Same seed + block height = same result
  - Read-only effect: Does not affect actual block production or storage

### Example

```toml
  # In MockDaConfig
  [randomization]
  seed = "0x2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a"
  reorg_interval = [5, 10]

  [randomization.behaviour]
  RewindBelowLastFinalized = { max_depth = 5 }
```

### StateManager Compatibility

  StateManager now handles stale finalized headers gracefully - if DA reports a finalized height lower than already processed, it uses the last processed finalized header instead (with a warning log).

## Testing
Adds new tests

## Docs
No updates


crates/adapters/mock-da/src/storable/layer.rs
```diff
@@ -318,9 +326,56 @@ impl StorableMockDaLayer {
...
+            _ => return None,
+        };
+
+        // Need some finalized blocks to rewind below
+        if self.last_finalized_height == 0 {
+            return None;
+        }
+
+        // Derive decision deterministically from seed + current height + discriminator
+        let mut hasher = sha2::Sha256::new();
```
preston-evans98: This logic looks to me like it could loop infinitely, because `seed` is not new randomness (it returns the same 32 bytes on each call). So if you call get_finalized at height 10 and rewind to 9, the exact same sequence can replay on the next time around

crates/adapters/mock-da/src/storable/layer.rs
```diff
@@ -318,9 +326,56 @@ impl StorableMockDaLayer {
...
+        hasher.update(discriminator);
+        let hash: [u8; 32] = hasher.finalize().into();
+
+        // Use first bytes to decide if we should return stale header (based on reorg_interval)
+        let interval_size = randomizer
+            .reorg_interval
+            .end
+            .saturating_sub(randomizer.reorg_interval.start)
+            .max(1);
+        let choice = u32::from_le_bytes([hash[0], hash[1], hash[2], hash[3]]) % interval_size;
```
preston-evans98: nit: you can use `rng.next_u32()` for this rather than the doing it manually with a hasher. That will also solve the problem with the looping

crates/adapters/mock-da/src/storable/layer.rs
```diff
@@ -1980,6 +2037,138 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_rewind_below_last_finalized_triggers() -> anyhow::Result<()> {
```
preston-evans98, line 2065: Does this test actually stand a chance of catching bugs? It looks like we only run a single iteration, but the behavior we're testing has only a relatively small chance of occuring for each choice of params

crates/adapters/mock-da/src/storable/layer.rs
```diff
@@ -1980,6 +2037,138 @@ mod tests {
...
+            da_layer.produce_block().await?;
+        }
+
+        let actual_finalized = da_layer.get_last_finalized_block_header().await?.height();
+        assert_eq!(actual_finalized, 10, "Should have finalized blocks");
+
+        // Enable RewindBelowLastFinalized with a seed that triggers the behavior
+        // Using reorg_interval 1..2 means 100% trigger rate
+        da_layer.set_randomizer(Randomizer::from_config(RandomizationConfig {
+            seed: HexHash::new([42; 32]),
```
preston-evans98: One more nit: I'm not sure if chacha is robust to seeding with very low entropy. It probably doesn't matter too much since we get a very small number of values, but I think it would be safe to seed with sha256([42;32])

crates/adapters/mock-da/src/storable/layer.rs
```diff
@@ -1980,6 +2037,138 @@ mod tests {
...
+        for _ in 0..20 {
+            da_layer.produce_block().await?;
+        }
+
+        let normal_head = da_layer.get_head_block_header().await?;
+        let normal_finalized = da_layer.get_last_finalized_block_header().await?;
+
+        // Enable then disable randomizer
+        da_layer.set_randomizer(Randomizer::from_config(RandomizationConfig {
+            seed: HexHash::new([42; 32]),
```
preston-evans98: nit: same comment about seeding chachaRng with very low entropy

preston-evans98 (approved):
Core change looks good to me; left some nits related to randomness

crates/adapters/mock-da/src/storable/layer.rs
citizen-stig: This is how we triggered a bug in state manager that this PR fixes. 
But I will extend number of iterations or maybe to a full-on proptest

crates/adapters/mock-da/src/storable/layer.rs
citizen-stig: Good point, addressed that. Took a step back and allowed using Rng in non mut context

crates/adapters/mock-da/src/storable/layer.rs
citizen-stig: Oh, actually not this test. Extended it

PR #2329, opened 2026-01-14
NOMT storage by default in TestRollup

citizen-stig:
# Description

NOMT is our main production storage at the moment. More tests run using it, the better.

* `TestSpec` -> `TestJmtSpec`
* `TestNomtSpec` -> `TestSpec`
* `SimpleStorageManager` -> `SimpleJmtStorageManager`
* `SimpleNomtStorageManager` -> `SimpleStorageManager`

Couple more type aliases add in `sov-test-utils` for easier usage.

Implemented `open_proof` which is small change actually. But keep in mind, that it only works for live state, and not historical state. Because of that, majority of the tests that needs proofs have been ignored, because it needs to be redesigned.

## Potential follow ups

* rename `ProverStorage` to `JmtProverStorage` and comment that it is going to be phased out at some point

## Testing

## Docs
No updates


.config/nextest.toml
```diff
@@ -29,6 +29,20 @@ filter = 'package(sov-sequencer) & test(/flaky_test_archival_state_is_immediatel
 priority = 95
 threads-required = 3
 
+[[profile.default.overrides]]
```
citizen-stig, line 32: TODO: With CI update, check if we still need this.

.github/workflows/rust.yml
```diff
@@ -265,6 +265,16 @@ jobs:
       - uses: namespacelabs/nscloud-cache-action@v1
         with:
           cache: rust
+      - name: Debug memlock limits
```
citizen-stig: TODO: remove this

crates/full-node/sov-db/src/config.rs
```diff
@@ -63,29 +63,21 @@ impl RollupDbConfig {
...
-            kernel_preallocate_ht: if cfg!(debug_assertions) {
-                Some(false)
-            } else {
-                None
-            },
-            kernel_page_cache_size: None,
-            kernel_leaf_cache_size: None,
-            pruner_block_interval: Some(100),
+            kernel_preallocate_ht: Some(false),
+            kernel_page_cache_size: Some(4),
```
citizen-stig: Bump to 8 or 16 as well

crates/full-node/sov-db/src/config.rs
```diff
@@ -146,6 +145,10 @@ impl RollupDbConfig {
         if let Some(leaf_cache_size) = self.user_leaf_cache_size {
             opts.leaf_cache_size(leaf_cache_size);
         }
+        if cfg!(debug_assertions) {
+            opts.page_cache_upper_levels(1);
```
citizen-stig: Remove this, or better move to params and do soak check

crates/utils/sov-test-utils/src/test_rollup.rs
```diff
@@ -724,7 +724,7 @@ where
     R: FullNodeBlueprint<Native> + Default + 'static,
 {
     /// Default timeout for polling operations in seconds.
-    pub const POLLING_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(20);
+    pub const POLLING_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(60);
```
citizen-stig, line 740: TODO: Should we rollback this?

crates/module-system/module-implementations/sov-attester-incentives/tests/integration/attestation_processing.rs
```diff
@@ -16,6 +16,7 @@ use crate::helpers::{
 };
 
 #[test]
+#[ignore = "NOMT does not support open_proof yet"]
```
citizen-stig: This

crates/full-node/sov-db/src/config.rs
```diff
@@ -102,13 +94,16 @@ impl RollupDbConfig {
             self.kernel_commit_concurrency
                 .expect("`kernel_commit_concurrency` concurrency must be set"),
         );
-        if cfg!(debug_assertions) {
+        if let Some(hashtable_buckets) = self.kernel_hashtable_buckets {
+            opts.hashtable_buckets(hashtable_buckets);
+        } else if cfg!(debug_assertions) {
             // 9.77MB
             opts.hashtable_buckets(2_500);
```
theodorebugnet: Should this be changed to 500 like the user buckets above?

crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
```diff
@@ -537,7 +550,7 @@ async fn test_tx_ws_submission() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
-#[ignore = "This test covers pruning behavior, which is only relevant for NOMT. Enable it when we switch to NOMT for the sequencer tests."]
+#[ignore = "This test covers pruning behavior, which is currently disable."]
```
theodorebugnet:
```suggestion
#[ignore = "This test covers pruning behavior, which is currently disabled."]
```

theodorebugnet (approved):
I don't have proper context of the NOMT proving (so I can't properly review the ignored tests and the open_proof() implementation) but otherwise changes LGTM.

Would it make sense to also have `sov_test_utils::storage::SimpleNomtStorageManager -> SimpleStoragemanager` and `SimpleStorageManager -> SimpleJmtStorageManager` the same way?

citizen-stig:
> Would it make sense to also have `sov_test_utils::storage::SimpleNomtStorageManager -> SimpleStoragemanager` and `SimpleStorageManager -> SimpleJmtStorageManager` the same way?

Yep, it also should reduce size of the PR

crates/module-system/module-implementations/sov-attester-incentives/tests/integration/challenger.rs
```diff
@@ -257,6 +258,7 @@ fn test_invalid_challenge_helper(
 }
 
 #[test]
+#[ignore = "NOMT does not support opening proof for archival state"]
```
bkolad, line 261: Is it some fundamental limitation?

crates/module-system/sov-state/src/nomt/mod.rs
```diff
@@ -1,5 +1,27 @@
...
+
+/// Wrapper around [`nomt_core::proof::MultiProof`] that implements `PartialEq`/`Eq`
+/// by comparing borsh-serialized representations.
+///
+/// This is necessary because `MultiProof` doesn't derive these traits.
+#[derive(Debug, Clone, Serialize, serde::Deserialize, BorshSerialize, BorshDeserialize)]
+pub struct NomtMultiProof(pub MultiProof);
+
+impl PartialEq for NomtMultiProof {
+    fn eq(&self, other: &Self) -> bool {
```
bkolad, line 73: Is it possible to use a custom method instead of eq?
Very often `eq` is used in loops, so it is better if it is trivially cheap. 
If we use something like `serialize_and_eq`, it will be clear that the method is not trivial.

crates/module-system/module-implementations/sov-attester-incentives/tests/integration/challenger.rs
citizen-stig: In current state of things, yes. NOMT does not track history. 

With JMT we've used same rocksbd mechanism with versioning. But with NOMT perofrmance we have to give it up.

Potentially we can hack around that and manually build proofs from historical data, but I think it would be generally better if clients just query live proof at any given height and then just use it.

crates/module-system/sov-state/src/nomt/mod.rs
citizen-stig: Then it will ripple into the type system, because there are bounds for this particular trait.

I've opened PR in NOMT: https://github.com/thrumdev/nomt/pull/930

So I can create an issue for us to follow up on. 

Considering that proofs are not used at the moment anyway, I think this is the most pragmatic approach.

crates/module-system/sov-state/src/nomt/zk_storage.rs
```diff
@@ -160,10 +161,51 @@ impl<S: MerkleProofSpec> Storage for NomtVerifierStorage<S> {
     fn materialize_changes(self, _state_update: Self::StateUpdate) -> Self::ChangeSet {}
 
     fn open_proof(
```
bkolad, line 163: This looks identical to 
https://github.com/Sovereign-Labs/sovereign-sdk/pull/2329/changes#diff-f9ed009fdaa0c4fcf8514593de47d2ffc54b94fcfff6bb60c6399a7184f0cc93R622

Let's extract the common code.

crates/module-system/sov-state/src/nomt/mod.rs
citizen-stig: Done as discussed in slack

crates/module-system/sov-state/src/nomt/zk_storage.rs
citizen-stig: Done

bkolad (approved):

PR #2334, opened 2026-01-15
Add address calculation for multisigs in web3

ross-weir:
# Description

Adds a method to calculate the credential id that will be used by the multisig configuration to web3 js

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/typescript/packages/multisig/src/index.ts
+++ b/typescript/packages/multisig/src/index.ts
@@ -1,3 +1,4 @@
+import { sha256 } from "@noble/hashes/sha2";
 import type {
   SignatureAndPubKey,
   Transaction,
@@ -5,6 +6,8 @@ import type {
   UnsignedTransaction,
 } from "@sovereign-sdk/types";
 import type { HexString } from "@sovereign-sdk/utils";
+import { bytesToHex, hexToBytes } from "@sovereign-sdk/utils";
+import * as borsh from "borsh";
 
 /**
  * Base error class for multisig-related errors.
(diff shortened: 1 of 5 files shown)


typescript/packages/multisig/src/index.ts
```diff
@@ -223,6 +226,34 @@ export class MultisigTransaction {
...
+  /**
+   * Calculates the multisig address (credential ID) by hashing the threshold and sorted public keys.
+   * This matches the Rust implementation: hash(threshold || borsh(sorted_pubkeys))
+   * @param hasher - The hash algorithm to use (currently only 'sha256' is supported)
+   * @returns The 32-byte multisig address as a Uint8Array
+   */
+  getMultisigAddress(hasher: "sha256" = "sha256"): Uint8Array {
+    if (hasher !== "sha256") {
+      throw new Error(`Unsupported hasher: ${hasher}`);
+    }
```
preston-evans98, line 238: (Question) Why do we accept a hasher param and then throw an error if it's not sha256?

preston-evans98 (approved):

typescript/packages/multisig/src/index.ts
ross-weir: Because claude said so

It's to account for the fact that the rollups crypto spec could have other hashers, at the moment we only use sha256 so throw a error if some customer in the distance future is using something else and tries to use their unsupported hasher

PR #2338, opened 2026-01-15
Refactoring re-org handling

citizen-stig:
# Description

## Changes in architecture

```
Runner                                StateManager
  │                                        │
  ├─ fetch block at height N ──────────────┤
  │                                        │
  ├─ prepare_storage(block) ───────────────┤
  │                                        ├─ has_reorg_happened()?
  │                                        │    │
  │                                        │    └─ LOOP 1: choose_fork_point() try find block between reorgs
  │                                        │       ├── FETCHES BLOCKS ()
  │                                        │       ├── returns different block
  │                                        │       └─ LOOP 2: (rewind) Have seen all blocks in this fork, waiting for the new block
  │ ◄────── (stf_state, possibly_new_block)│
  │                                        │
  ├─ if block changed: update next_height  │
  │                                        │
  └─ process block with STF ───────────────┘
```

```
Runner                                StateManager
  │                                        │
  ├─ fetch block at height N ──────────────┤
  │                                        │
  ├─ check_block_continuity(header) ───────┤
  │                                        ├─ Check passed block header against tree of known transitions
  │ ◄────── Continuation or SwitchToHeight │
  │                                        │
  ├─ if SwitchToHeight:                    │
  │     next_height = suggested_height     │
  │     continue loop (runner fetches)     │
  │                                        │
  ├─ prepare_storage_for_block(header) ────┤
  │ ◄────── stf_state                      │
  │                                        │
  └─ process block with STF ───────────────┘
```  

* `StateManange`r does not have concept of "current state root" anymore. It has 2 sibling params: `last_processed_finalized_state_root` and `last_processed_finalize_block_header`, which always match each other. And for non finalized transitions it has a tree of `StateOnBlock` items, which contain `pre/post state_root` and `block_header`. "current" chain is just height tracked by the runner

## Significant changes in behaviour

* Sequencer and API state is updated only and always after new block has been processed. Previously it has been updated before processing block in case of re-orgs. This trade-off simplifies logic of `StateManager` significantly with slight delay in API status in case of reorg.
* `da_total_timeout` Parameter is gone. Node will continue trying to progress, until sequencer won't shut it down due to blobs it sent not being processed. That's the worst case.

## Testing
Tests have been adjusted and passing.

TODO: Add more tests

## Docs
Documentation and comments are extended


crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -67,10 +80,30 @@ impl<Da: DaSpec, StateRoot: Clone> StateOnBlock<Da, StateRoot> {
     }
 }
 
-enum ForkPointSearchResult<Da: DaService, StateRoot> {
-    Found(ForkPoint<Da, StateRoot>),
+enum ForkPointSearchResult<Da: DaService> {
+    /// Found a valid fork point - the first unprocessed block whose predecessor was seen.
+    Found {
+        block_header: <Da::Spec as DaSpec>::BlockHeader,
```
preston-evans98: It looks like we don't need to return the full header here; we call header.height() on the returned value - so this will let us skip a da fetch in some cases

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -312,10 +366,13 @@ where
                     "Mismatch in block hashes after transition",
                 );
             }
-            assert_eq!(
-                transition_witness.initial_state_root.as_ref(),
-                self.state_root.as_ref(),
-                "Wrong transition, its pre_state_root does not match the current state root of StateManager");
+            // This is not the case anymore
+            // TODO: How to better check that it is not messed up?
+            // Check that initial state root is known and matches block, right?
```
preston-evans98: Why doesnt this assertion hold anymore?

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -1019,21 +1020,65 @@ where
...
+            .expect("Should be called after at least single transition added");
+        let highest_seen = self
+            .get_highest_seen_height()
+            .expect("Should be called after at least single transition added");
+
+        debug_assert!(
+            earliest_seen <= highest_seen,
+            "bug in state manager. earliest and highest transition numbers are calculated incorrectly"
+        );
+        debug_assert_eq!(
```
preston-evans98: nit: These look like they could be real assertions

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: because passed in transition might be not the descendant of finalized, but of some non-finalized block.

Previously StateManager had `.state_root`, which was "current". But now it holds only `last_finalized_processed_state_root`. All other `pre_state` roots are available in `seen_on_block`

I will add this comment.

This assertion can be done if `seen_on_block` is empty or we can do another assertion that pre-decessor block post-state root matches.

I will expand comment and add asserts.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -778,224 +834,169 @@ where
...
-            "Going to extract finalized transitions from previously seen transitions"
+            "Extracting finalized transitions from previously seen transitions"
         );
+
         for height in range {
-            // DaService call # 3 + n
             let finalized_at_that_height = self
                 .finalized_headers_provider
                 .get_block_header_at(height)
                 .await?;
```
preston-evans98: There's an extremely nasty edge case here where we switch RPC nodes during finalization. If the new node we get happens to be on a p2p network partition, it might possibly return a block that's not on the chain to finalize even though a more up to date node believes the height it's seeing is finalized. 

This only affects chains that actually have reorgs (can't happen on Celestia) - but we should probably fix it anyway.  The easy fix is to simply use the prev_root of the previous loop's block header as the finalized hash rather than trusting our finalized block headers cache. (That cache sometimes passes requests through to the DA service, so we can't trust it)
 
```rust,suggestion
        for height in range {
        let mut next_hash_to_finalize = effective_finalized_header.hash();
        for height in range {
           
            let blocks_on_height = self
                .seen_on_height
                .remove(&height)
                .expect("Should be at least one seen transition on each height");

            let mut pushed_for_this_height = false;
            for block_hash in blocks_on_height {
                let transition = self
                    .state_on_block
                    .remove(&block_hash)
                    .expect("Should be there");
                if block_hash == next_hash_to_finalize {
                    assert!(
                        !pushed_for_this_height,
                        "Should be only one finalized transition per height"
                    );
                    next_hash_to_finalize = transition.block_header.prev_hash();
                    finalized_transitions.push(transition);
                    pushed_for_this_height = true;
                }
            }
            assert!(pushed_for_this_height, "Should have pushed at least one transition for this height");
        }
```

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -1019,21 +1020,65 @@ where
...
 
-        if let Some(last_processed_transition) = finalized_transitions.iter().last() {
-            self.last_processed_finalized_header = last_processed_transition.block_header.clone();
+        if let Some(last) = finalized_transitions.last() {
+            self.last_processed_finalized_header = last.block_header.clone();
+            self.last_processed_finalized_state_root = last.post_state_root.clone();
         }
 
-        // Verify invariant: all earliest seen transitions must descend from last_processed_finalized_header
         self.verify_earliest_seen();
```
preston-evans98, line 1065: verify_earliest_seen only runs debug assertions - but nothing it does looks very expensive. Can we make it a regular assertion?

crates/adapters/mock-da/src/storable/local_service.rs
```diff
@@ -357,8 +394,51 @@ impl StorableMockDaService {
...
+                Ok(())
+            }
+            FailureBehavior::ReorgDuringCall {
+                trigger_at_height,
+                triggered,
+            } => {
+                if height == *trigger_at_height && !*triggered {
+                    *triggered = true;
+                    // Drop the mutex before acquiring write lock on DA layer
+                    drop(behavior);
```
preston-evans98: Does this open us up to race conditions? What happens if another thread gets interleaved between the `drop` and the time we acquire the write lock?

preston-evans98 (commented):
This looks to me like it should fix everything. Thanks a ton for the work @citizen-stig!

crates/adapters/mock-da/src/storable/local_service.rs
citizen-stig: I rewrote this section

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: Good point, solved

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -335,6 +407,9 @@ where
             "Processed finalized transitions"
         );
 
+        self.ledger_db.replace_reader(ledger_pre_state);
+        // TODO: We can check state root from ledger and verify its connected
```
preston-evans98: This todo looks non-blocking to me; but let's document why it's not a priority atm so that we don't mistake this for a bug later.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -668,9 +692,11 @@ where
             return Ok(ForkPointSearchResult::DaHeadIsBelowProcessedFinalized);
         }
 
-        let mut low = earliest_seen_height;
-        let mut high = std::cmp::min(highest_seen_height, head.height()).saturating_add(1);
+        let low = earliest_seen_height;
+        let high = std::cmp::min(highest_seen_height, head.height()).saturating_add(1);
 
+        // TODO: Should we derive low just from `last_processed_finalized_height`?, this will simplify this function.
```
preston-evans98: What's the status of this todo? Are there downsides to using `last_processed_finalized_height + 1`?

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: I didn't add it originally because I didn't want to do another db call

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -778,238 +856,181 @@ where
...
 
-        // Handle edge cases where DA layer reports unexpected finalized heights:
-        // 1. If syncing: finalized might be higher than we've seen - cap at highest_seen
-        // 2. If DA reports stale: finalized might be lower than already processed - use last_processed
-        let last_seen_finalized_header = if last_finalized_header.height() > highest_seen_transition
-        {
+        if last_finalized_header.height() > highest_seen_transition {
             // Syncing case: DA is ahead of us
-            self.finalized_headers_provider
+            // TODO: PROBLEM: THIS RETURNED BLOCK CAN BE SOMETHING COMPLETELY WRONG AND NOT FINALIZED.
```
preston-evans98: nit: Let's document that this is only possible on chains with reorgs *AND* it happens only when the RPC node switches abruptly

citizen-stig: Done!

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: Done

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: I've added check, but for perf reasons decided to keep for debug assertions.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -162,96 +192,125 @@ where
...
-            tracing::trace!(
-                "Reorg has happened, updating API and Ledger storage before returning STF state"
-            );
-            // In case if reorg happened, we want to keep ledger and API storages in sync.
-            // Otherwise, the API storage and LedgerDb have been updated in [`Self::update_api_and_ledger_storage`]
-            self.update_channels(stf_pre_state.clone(), ledger_state)
-                .await?;
-        }
+        // Block is a continuation - get pre_state_root and create state
+        let pre_state_root = self.get_pre_state_root_for(block_header);
```
preston-evans98: nit: `get_pre_state_root_for` panics if the block is not a continuation. Maybe we should rename to 
```suggestion
        let pre_state_root = self.get_pre_state_root_for_continuation(block_header);
```

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -162,96 +192,125 @@ where
...
+                }
+                ForkPointSearchResult::HeadChanged(new_head) => {
+                    // DA reorged during fork point search.
+                    // If chain rewound below requested height, retry at new head + 1.
+                    // If chain extended or same level, retry original height.
+                    let retry_height = if new_head.height() < block_header.height() {
+                        new_head.height().saturating_add(1)
+                    } else {
+                        block_header.height()
+                    };
```
preston-evans98, line 244: (Non-blocking: I think this logic will struggle with Solana. We return head changed if the header advances by more than 1, but because we make so many network calls during the binary search, we're likely to get stuck in a  loop of just requesting the same height over and over)

preston-evans98: Let's add a TODO but not fix for now

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -335,11 +407,36 @@ where
             "Processed finalized transitions"
         );
 
+        self.ledger_db.replace_reader(ledger_pre_state);
+        #[cfg(debug_assertions)]
```
preston-evans98: this check doesn't look very expensive to me. Should we make it a regular assertion?

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -668,88 +705,133 @@ where
...
-            let (candidate, this_head) = tokio::try_join!(
-                da_service.get_block_at(mid),
+            let (candidate_header, this_head) = tokio::try_join!(
+                da_service.get_block_header_at(mid),
                 da_service.get_head_block_header()
             )?;
 
             if is_head_changed::<Da::Spec>(&head, &this_head) {
-                return Ok(ForkPointSearchResult::HeadChanged(this_head));
+                return Ok(BinarySearchOutcome::Done(
```
preston-evans98, line 716: non-blocking for this PR: I think the head changed logic is too aggressive here. If we ever get stuck on an RPC with very bad roundtrip time, we'll hit HeadChanged basically every time we try to binary search even though nothing is actually happening. we should think about this more

citizen-stig: So head change logic checks if new head does not equal current or if new head is direct descendant for of current.

So that gives us 2*block_time to find new fork point. 

What can be alternatives?

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: I don't have strong objection, but that's another DB read, which likely hit the disk.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: If we bring back subcriptions TO da service, each binary search can spin head subcription, which will update if head has changed.

preston-evans98 (approved):
Left a few more comments, but the core logic still looks solid with fresh eyes. I think we're ok to merge for now! 

But, let's think very carefully about the HeadChanged case - that looks to me like it could cause us to stall when RPC performance is bad relative to chain performance. That's not a huge concern for Celestia while block times are so slow, but it's still a concern. IMO we should try to follow up on this in the next week or two

citizen-stig:
@preston-evans98 thanks for thorough review!

I've moved all important TODOs from this PR into this issue: 

* https://github.com/Sovereign-Labs/sovereign-sdk/issues/2375

Please feel free to comment there, or add items directly to the issue

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: Added to https://github.com/Sovereign-Labs/sovereign-sdk/issues/2375

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: Added to https://github.com/Sovereign-Labs/sovereign-sdk/issues/2375

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2338#discussion_r2709079043

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: Done in 3a2809815a17b53f195f8809510019352e1b97e4

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: Done in c90379179e6038e821a9a7de0357752b748674e2

PR #2340, opened 2026-01-15
feat: count deliveries in the mailbox

vmmon:
# Description

Counts the number of inbound messages received.  This allows an indexer to use binary search to find all delivered messages in O(log(N)).  This makes Delivery indexing as fast as MerkleTree and Dispatch indexing.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/hyperlane/src/api.rs
+++ b/crates/module-system/hyperlane/src/api.rs
@@ -62,6 +62,7 @@ impl<S: Spec, R: Recipient<S>> HasCustomRestApi for Mailbox<S, R> {
     fn custom_rest_api(&self, state: ApiState<S>) -> axum::Router<()> {
         axum::Router::new()
             .route("/nonce", get(Self::get_nonce))
+            .route("/deliveries", get(Self::get_deliveries))
             .route("/recipient-ism/:address", get(Self::get_recipient_ism))
             .route(
                 "/recipient-ism/:address/validators_and_threshold",
@@ -114,6 +115,18 @@ impl<S: Spec, R: Recipient<S>> Mailbox<S, R> {
             .map(|dispatch_state| dispatch_state.nonce)
             .unwrap_or_default();
         Json(json!({"nonce": nonce}))
+    }
+
+    async fn get_deliveries(
+        state: ApiState<S, Self>,
+        mut accessor: ApiStateAccessor<S>,
+    ) -> impl IntoResponse {
+        let deliveries = state
+            .delivery_count
+            .get(&mut accessor)
+            .unwrap_infallible()
+            .unwrap_or_default();
+        Json(json!({"count": deliveries}))
     }
 
     async fn get_recipient_ism(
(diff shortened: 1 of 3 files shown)


crates/module-system/hyperlane/src/lib.rs
```diff
@@ -85,6 +85,10 @@ pub struct Mailbox<S: Spec, R: Recipient<S>> {
     #[state]
     pub deliveries: StateMap<MessageId, Delivery>,
 
+    /// The number of successful deliveries.
+    #[state]
+    pub delivery_count: StateValue<u64>,
```
citizen-stig: Anything stored in value is proven, which costs.

If there's any data that is not important for the security of the rollup it is better to use AccessoryState.

vmmon: Good point.

This information is not necessary for the integrity of the system but rather for the availability.  A malicious rollup can lie here but that's easily detected by looking at all slots.

citizen-stig: So reducing amount of data needs to be proven is beneficial for throughput.

vmmon: The main obstacle to use an AccessoryStateValue is that TxState doesn't inherit fron StateReader<Accessory, _>.

While one should be able to emulate that via get_value() it seems to be more appropriate to add it to the TxState trait.

What is the best choice here?

preston-evans98: This would be a breaking change! The new state item must be the last #[state] item in the struct for backwards compatibilty; otherwise, the state keys for `validators` will change

preston-evans98: No objection to using regular state though; modifying a single value isn't too expensive

citizen-stig: Ok, that's a good point, since AccessoryState does not have access to this reader. 

Then we should accept this trade off

crates/module-system/hyperlane/src/api.rs
```diff
@@ -116,6 +117,18 @@ impl<S: Spec, R: Recipient<S>> Mailbox<S, R> {
         Json(json!({"nonce": nonce}))
     }
 
+    async fn get_deliveries(
```
citizen-stig: Non-blocking, but worth checking `http://localhost:12346/swagger-ui/`, The REST API endpoitn might be auto-generated already,

crates/module-system/hyperlane/src/lib.rs
vmmon: Good to know. I somehow missed that. This is one place where changing the semantics would make things easier.

crates/module-system/hyperlane/src/api.rs
vmmon: It is. Fixed that.

citizen-stig (approved):

vmmon:

PR #2346, opened 2026-01-16
Fix EIP-712 signer high-s signature handling

cemozerr:
MetaMask's eth_signTypedData_v4 may return high-s signatures for EIP-712 messages (EIP-2 low-s requirement only applies to transactions, not off-chain message signing). This caused ~50% of signature verifications to fail on the server.

Changes:
- Switch from ethers to viem for signature parsing (ethers rejects high-s)
- Add low-s normalization using @noble/secp256k1
- Add comprehensive signing tests including high-s normalization test

# Description

*Summary of the changes...*

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/typescript/packages/signers/src/wasm/eip712.test.ts
+++ b/typescript/packages/signers/src/wasm/eip712.test.ts
@@ -13,38 +13,74 @@ const testAccount = privateKeyToAccount(TEST_PRIVATE_KEY);
 
 const schema = Schema.fromJSON(JSON.stringify(demoRollupSchema));
 
-// Create a sample unsigned transaction
-const sampleCall = {
-  bank: {
-    transfer: {
-      to: {
-        Standard: "sov1lzkjgdaz08su3yevqu6ceywufl35se9f33kztu5cu2spja5hyyf",
-      },
-      coins: {
-        amount: "1000",
-        token_id:
-          "token_1rwrh8gn2py0dl4vv65twgctmlwck6esm2as9dftumcw89kqqn3nqrduss6",
+// Sample unsigned transaction for testing
+const sampleUnsignedTx = {
+  runtime_call: {
+    bank: {
+      transfer: {
+        to: { Standard: "sov1lzkjgdaz08su3yevqu6ceywufl35se9f33kztu5cu2spja5hyyf" },
+        coins: {
+          amount: "1000",
+          token_id: "token_1rwrh8gn2py0dl4vv65twgctmlwck6esm2as9dftumcw89kqqn3nqrduss6",
+        },
       },
     },
   },
+  generation: "12345",
+  details: { max_priority_fee_bips: "1000", max_fee: "10000", gas_limit: null, chain_id: "1" },
 };
 
-const sampleUnsignedTx = {
-  runtime_call: sampleCall,
-  generation: "12345",
-  details: {
-    max_priority_fee_bips: "1000",
-    max_fee: "10000",
-    gas_limit: null,
-    chain_id: "1",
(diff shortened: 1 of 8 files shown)


typescript/packages/signers/src/wasm/eip712.ts
```diff
@@ -2,10 +2,38 @@ import type { MetaMaskInpageProvider } from "@metamask/providers";
...
+  yParity: number;
+}
+
+/**
+ * Parse a 65-byte hex signature string into r, s, and recovery bit.
+ * Format: 0x + r (32 bytes) + s (32 bytes) + v (1 byte)
+ * v is either 27/28 (legacy) or 0/1 (EIP-155)
+ */
+function parseSignatureHex(signatureHex: string): ParsedSignature {
+  const sigBytes = hexToBytes(signatureHex.slice(2));
```
preston-evans98: do we need to handle strings missing the prefix?

cemozerr: Good catch, fixed: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2346/commits/53849f99cefaf081d52ba39e3190d2356e3c2340

typescript/packages/signers/src/wasm/eip712.ts
```diff
@@ -121,11 +154,23 @@ export class Eip712Signer implements Signer {
       );
     }
 
-    // Parse the signature for public key recovery
-    const signature = Signature.from(signatureHex);
+    // Parse the signature (format: 0x + r + s + v, 65 bytes total)
+    const signature = parseSignatureHex(signatureHex);
```
ross-weir: It seems like we have to do a bunch of manipulation in `parseSignatureHex` and the following lines to get the signature in the right format

Is the signature format non-standard? If so, why?
If the signature format is standard are there any library functions we can use instead of manually implementing this

ross-weir (approved):

typescript/packages/signers/src/wasm/eip712.ts
cemozerr: Fixed: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2346/commits/42810f7db99db97ee487e3f93d7c6218fe7d760e

typescript/packages/signers/src/wasm/eip712.test.ts
```diff
@@ -62,14 +136,118 @@ describe("Eip712Signer", () => {
...
+              const r = sigBytes.slice(0, 32);
+              const s = sigBytes.slice(32, 64);
+              const v = sigBytes[64];
+
+              const sBigInt = BigInt(
+                `0x${Array.from(s)
+                  .map((b) => b.toString(16).padStart(2, "0"))
+                  .join("")}`,
+              );
+              const newS = N - sBigInt; // Force high-s
```
preston-evans98: Does this work if the signature is already high-s? I think we probably need a condition here to force high-s only `if S < N/2` to start with

preston-evans98 (approved):
Change lgtm; we may want to document somewhere in the rollup that it requires low-s signatures for EIP-712 to avoid malleability though

typescript/packages/signers/src/wasm/eip712.test.ts
cemozerr: Fixed: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2346/commits/f7099da5dce36a6f2373048726f922d4942e3531

PR #2332, opened 2026-01-15
Add `NodeRole::DbElected` mode that enables automatic leader election.

bkolad:
# Description

Summary:

  Add `NodeRole::DbElected` mode that enables automatic leader election using PostgreSQL. When multiple nodes start with `DbElected` role, they compete for leadership via the database - one becomes the Leader and others become Replicas.

  Key features:
  - Automatic leader election at startup using PostgreSQL `sequencer_leader` table
  - Leader maintains heartbeat to retain leadership
  - Replicas monitor for leadership availability
  - New `/sequencer/role` REST endpoint to query the current node's role

- [x] ~I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.~
- [x] I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. (Are all new dependencies necessary? Is any module dependency leaked into the full-node (hint: it shouldn't)?)

PR #2347, opened 2026-01-17
Leader failover: Add `test_db_elected_leader_failover`

bkolad:
# Description

This PR adds another failover test and fixses todos from #2332.

### DIFF
--- a/examples/demo-rollup/tests/replica/db_elected.rs
+++ b/examples/demo-rollup/tests/replica/db_elected.rs
@@ -110,7 +110,7 @@ async fn test_db_elected_leader_failover() {
     };
 
     let DbElectedTestSetup {
-        postgres,
+        postgres: _postgres,
         leader,
         replica,
         da_shutdown,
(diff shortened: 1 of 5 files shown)


examples/demo-rollup/tests/replica/db_elected.rs
```diff
@@ -63,12 +97,60 @@ async fn test_db_elected_two_nodes_leader_and_replica() {
...
+    assert_eq!(
+        new_role,
+        SequencerRole::Leader,
+        "Expected restarted node to be Leader, got {:?}",
+        new_role
+    );
+
+    let _ = restarted_rollup.shutdown().await;
     let _ = da_shutdown.send(());
+    drop(postgres);
```
theodorebugnet: Out of curiosity, is an explicit drop needed? Since it's the end of the scope anyway

theodorebugnet (approved):

examples/demo-rollup/tests/replica/db_elected.rs
bkolad: Otherwise, Rust complains that the variable is not used, but I want it to be dropped at the end.

theodorebugnet: I think the usual convention is to just name it `_postgres` (e.g. that's how we use `_otel_guard` in the rollup blueprint). Very minor though

bkolad: fixed

Issue #2309, opened 2026-01-07
Celestia background metrics small imporvements


citizen-stig:
- [ ] Add address as a tag, so it is visible in the alerts and everywhere
- [x] Allow configuring interval, with 0 meaning disabled completely. https://github.com/Sovereign-Labs/sovereign-sdk/pull/2349

citizen-stig:
For example background stat can be disabled on backup node, saving API credits and networking cost

PR #2349, opened 2026-01-17
Celestia configurational improvements

citizen-stig:
# Description

Splits high level request timeout from individual API calls made to RPC gRPC. This allows to have more fine grained controls on errors. For instance, blobs will be resubmitted sooner.

Also adjusts tx-status polling to 2s (default in celestia-client is 500ms), allowing us to cut QuickNode cost almost 4x and reduce load on their nodes. 

Adds another optional config for changing how often background stat is collected, or completely disable it.

## Linked Issues
- Related to #2309

## Testing
All existing tests are passed, but several days soak is required and results will be posted in the comment

## Docs
Updated config docstring and reference toml

### DIFF
--- a/crates/adapters/celestia/src/config.rs
+++ b/crates/adapters/celestia/src/config.rs
@@ -37,12 +37,27 @@ pub struct CelestiaConfig {
     /// `cel-key export key-name --unarmored-hex --unsafe --node.type light --p2p.network=mocha`
     #[serde(default = "default_signer_private_key")]
     pub signer_private_key: Option<String>,
-    /// The timeout for a Celestia RPC request, in seconds.
+    /// High-level timeout for Celestia RPC operations that may include multiple requests (in seconds).
+    /// Default: 38 (6 blocks × 6 seconds + 2 seconds polling buffer).
     #[serde(
         default = "default_request_timeout_seconds",
         alias = "celestia_rpc_timeout_seconds"
     )]
     pub request_timeout_secs: NonZero<u64>,
+    /// Timeout for individual API requests to the Celestia node (in seconds).
+    /// This is passed to the underlying celestia-client for each API call.
+    /// Default: 6 (one block time).
+    #[serde(default = "default_api_request_timeout_secs")]
+    pub api_request_timeout_secs: NonZero<u64>,
+    /// Interval for polling transaction status confirmation (in milliseconds).
+    /// Default: 2000.
+    #[serde(default = "default_tx_status_polling_millis")]
+    pub tx_status_polling_millis: u64,
+    /// Interval for background statistics collection (in seconds).
+    /// Set to 0 to disable the background stat collection task.
+    /// Default: 30.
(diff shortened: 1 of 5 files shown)


crates/adapters/celestia/src/config.rs
```diff
@@ -232,6 +257,18 @@ pub(crate) fn default_factor() -> f32 {
...
 
 pub(crate) fn default_request_timeout_seconds() -> NonZero<u64> {
-    // 6 blocks + 1 second for jitter
-    NonZero::new(37).unwrap()
+    // 6 blocks × 6 seconds + 2 seconds polling buffer
+    NonZero::new(38).unwrap()
+}
+
+pub(crate) fn default_api_request_timeout_secs() -> NonZero<u64> {
+    NonZero::new(6).unwrap()
```
citizen-stig: Six is too low for getting next header, considering network gitter. Need to add 2 seconds for safety

citizen-stig:
Changes on quicknode

Screenshot 2026-01-20 at 14 56 54

There are various improvements, looks like rollup node becomes snappier:

Screenshot 2026-01-20 at 15 00 53

citizen-stig:
It does not make TPS worse and reduces QuickNode usage, directly cutting cost of RPC provider a little less than 2x
Screenshot 2026-01-20 at 15 02 10

citizen-stig:
Screenshot 2026-01-20 at 20 20 52

crates/adapters/celestia/src/config.rs
```diff
@@ -37,12 +37,27 @@ pub struct CelestiaConfig {
     /// `cel-key export key-name --unarmored-hex --unsafe --node.type light --p2p.network=mocha`
     #[serde(default = "default_signer_private_key")]
     pub signer_private_key: Option<String>,
-    /// The timeout for a Celestia RPC request, in seconds.
+    /// High-level timeout for Celestia RPC operations that may include multiple requests (in seconds).
+    /// Default: 38 (6 blocks × 6 seconds + 2 seconds polling buffer).
```
theodorebugnet, line 41: Minor comment: I feel like adding the defaults in the comment here invites comment rot if the `default_...()` methods ever get changed.
On the other hand having this in the documentation is nice, so I'm not sure what's the better tradeoff.

theodorebugnet (approved):

crates/adapters/celestia/src/config.rs
citizen-stig: I think having it in the comment help the consumers of this struct, comment rot is our problem, which we can solve with LLM checking our docs periodically

theodorebugnet: Good point, what used to be "we need a special tool for linting this (and if it doesn't exist we need to build our own script)" can just be an LLM prompt now. Good thought

citizen-stig:
If old behaviour is needed, set this:

```toml
request_timeout_secs = 37
api_request_timeout_secs = 37
tx_status_polling_millis = 500
```

PR #2353, opened 2026-01-17
Make relayer address optional in quote-dispatch with admin fallback

ross-weir:
# Description

Add an admin field to the IGP module that serves as the fallback relayer when no relayer is specified in quote-dispatch requests. This allows clients to query gas prices without needing to know a specific relayer address.

Changes:
- Add IGPConfig with required admin field for genesis configuration
- Add admin state field to InterchainGasPaymaster
- Add SetAdmin call message for updating the admin (requires auth)
- Add AdminSet event for tracking admin changes
- Add quote_gas_price_with_admin_fallback method for API usage
- Update query_quote_dispatch to use admin fallback when no relayer provided
- Remove recipient_address from QuoteParams (no longer needed)
- Update all test files to provide IGPConfig with admin address

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/hyperlane/src/api.rs
```diff
@@ -43,8 +42,6 @@ pub struct QuoteParams {
     /// Gas limit.
     #[serde(deserialize_with = "deserialize_number_from_str")]
     pub gas_limit: u128,
-    /// Recipient address.
-    pub recipient_address: HexHash,
```
bkolad, line 47: This still exists in:
`crates/module-system/hyperlane/src/openapi-v3.yaml`

crates/module-system/hyperlane/src/igp/mod.rs
```diff
@@ -64,24 +82,29 @@ pub struct InterchainGasPaymaster<S: Spec> {
...
     type CallMessage = call::CallMessage<S>;
     type Event = Event<S>;
     type Error = anyhow::Error;
 
     fn genesis(
         &mut self,
         _genesis_rollup_header: &<<S as Spec>::Da as DaSpec>::BlockHeader,
-        _config: &Self::Config,
-        _state: &mut impl GenesisState<S>,
+        config: &Self::Config,
```
bkolad, line 104: This will only apply to new deployments, so existing clients with an already running chain won’t see the admin setting. Is that okay?

bkolad: Ahh, ok there is a CallMessage for this  as well.

crates/module-system/hyperlane/src/igp/mod.rs
```diff
@@ -126,6 +149,10 @@ impl<S: Spec> Module for InterchainGasPaymaster<S> {
                 tracing::debug!(?domain, ?oracle_value, "Updating oracle data");
                 self.update_oracle_value(domain, oracle_value, context, state)?;
             }
+            CallMessage::SetAdmin { new_admin } => {
```
bkolad, line 152: Let's add a test for it.

crates/module-system/hyperlane/src/igp/mod.rs
ross-weir: yeah but the call message requires the admin to already be set, so I guess it needs genesis currently

Any suggestions on how to handle this?

crates/module-system/hyperlane/src/api.rs
ross-weir: fixed

crates/module-system/hyperlane/src/igp/mod.rs
ross-weir: added

crates/module-system/hyperlane/src/igp/mod.rs
bkolad: Thinking out loud: would it be acceptable to add a dependency on the chain-state and the reuse admin_address?

Issue #1524, opened 2025-08-13
[TRACKER] Replication & Failover


bkolad:
### REPLICATION

 - [x] Sequencer cleanup
         - #1482
         - #1488
         - #1496
 - [x] Make `BlobSender` optional #1506 
 - [x] Make `PreferredSequencerDB optional` #1519
 - [x] Test simple replication setup #1523
 - [x] Implement backfill logic from PostgreSQL #1532
 - [x] Add tests for Postgres notifications #1535
 - [x] Add functionality that allows the `Replica Task` to propagate PostgreSQL updates to the Sequencer state.
 - [x] Add RPC client and server for StorableMockDa. (https://github.com/Sovereign-Labs/sovereign-sdk/pull/1848)
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1866
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1868
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1879
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1883
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1910
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1915
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1929
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1935
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1938
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1948
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1961
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1965
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1986
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1988
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/1990
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/2000
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/2001
 - [x] https://github.com/Sovereign-Labs/sovereign-sdk/pull/2002
 

Still TODO

Add the following unit tests:
- [x] Test replica restart behavior
[...]

PR #2355, opened 2026-01-17
Move Axum tcp listener binding earlier in rollup startup.

bkolad:
# Description

Problem: The sequencer currently cannot access the Axum server’s bound socket address, which is a requirement for the failover workstream.

This PR changes when the TCP listener is bound during rollup startup. Instead of creating the `TcpListener` inside `start_http_server`, it is now initialized earlier and passed to the runner.

This way, we will be able to pass the axum address to the sequencer  (to be implemented in a follow-up PR).

Related issue #1524

### DIFF
--- a/crates/full-node/sov-stf-runner/src/runner.rs
+++ b/crates/full-node/sov-stf-runner/src/runner.rs
@@ -31,6 +31,7 @@ use tracing::{debug, info, trace};
 use crate::da::{DaServiceWithCachedFinalizedHeaders, FinalizedBlocksBulkFetcher};
 use crate::processes::{new_stf_info_channel, Receiver};
 use crate::state_manager::StateManager;
+use tokio::net::TcpListener;
 
 type GenesisParams<ST, InnerVm, OuterVm, Da> =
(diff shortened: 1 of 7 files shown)


crates/full-node/sov-stf-runner/src/runner.rs
```diff
@@ -251,9 +247,23 @@ where
...
             finalized_headers_provider: da_service_with_cached_finalized_headers,
+            axum_tcp: Some(axum_tcp),
         })
     }
 
+    /// Returns the socket address of the Axum server.
+    ///
+    /// This method must be called before [`StateTransitionRunner::run_in_process`], as the TCP listener
+    /// is consumed when the HTTP server starts. Calling this method after `run_in_process` will return
+    /// an error.
```
bkolad, line 258: This is ok because `Rollup::run` consumes self, meaning `axum_socket_address` will no longer be accessible once the rollup starts. It is only needed during rollup construction.

citizen-stig: I have the similiar issue with stateManager require starup.

We can chat later today about implmeneting something like builder to get around that in a type-safe way

citizen-stig (approved):

PR #2369, opened 2026-01-19
Add state layout introspection test

cemozerr:
## Summary\n- expose module state layout metadata via ModuleInfo::state_items\n- emit state item descriptors from the ModuleInfo derive macro\n- generate build-time state layout output as JSON via sov-build\n- add scripts/verify_state_layout.py to enforce append-only changes in CI\n\n## Testing\n- not run (build-time generation change)


crates/module-system/sov-modules-api/src/module/mod.rs
```diff
@@ -136,6 +136,11 @@ pub trait ModuleInfo {
     /// Execution context.
     type Spec: Spec;
 
+    /// Returns the module's state items in declaration order.
+    fn state_items(&self) -> &'static [StateItemDescriptor] {
+        &[]
```
bkolad, line 141: Why do we have a default impl here?

PR #2371, opened 2026-01-19
Remove duplicate sorting test case

majtte:
This PR removes a duplicated `+100` case from `sorting::tests::ok_cases`, keeping the test set concise and focused on distinct inputs.


majtte:

PR #2386, opened 2026-01-21
Replica: Skip `state_update` when a DA batch arrives before the PostgreSQL notification

bkolad:
# Description
The replica receives batches from the master via PostgreSQL notifications and via DA. Under normal conditions, PostgreSQL notifications are faster than DA, so after the initial sync, the replica stays ahead of DA.

However, when database performance degrades, the replica may receive a batch via DA first and only later receive the same batch via PostgreSQL. In this case, we will get the `visible slot number calculation was incorrect` error. 

This PR fixes the issue by skipping the problematic `update_state` call.

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/update_state.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/update_state.rs
@@ -10,6 +10,8 @@ use crate::preferred::{
     PreferredBatchToReplay, PreferredSequencer, ProcessFinalCatchupData, RollupBlockExecutor,
     StateUpdateInfo,
 };
+use crate::SequencerRole;
+use tracing::error;
 
 impl<S, Rt, Da> PreferredSequencer<S, Rt, Da>
 where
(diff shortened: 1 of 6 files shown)


crates/full-node/sov-sequencer/src/preferred/update_state.rs
```diff
@@ -222,32 +246,88 @@ where
...
+                // If the database slowdown is temporary, the issue will be resolved when the next `update_state` call succeeds.
+                // If the situation persists, the replica will eventually enter sync mode in that case that the database
+                // setup needs to be examined.
+                error!(seq_nr_in_of_in_progress_batch, next_sequence_number_according_to_node, "The replica has an in-progress batch whose sequence number is lower than the next_sequence_number expected by the node. 
+                    This indicate that Postgres notifications are delayed. In this case, the update from the node is ignored. 
+                    If this error occurs repeatedly, investigate the database stack in the deployment.");
+                return Err(SequenceNumberMismatchError::SkipStateUpdate);
+            }
+            SequencerRole::Leader | SequencerRole::ReplicaNoLeaderSync => {
+                // For roles other than replicas, we should never observe in-progress batches with sequence numbers lower than what the node expects (ReplicaNoLeaderSync don't create batches).
```
bkolad, line 274: Note to reviewer: This is a new condition for the Leader. I can’t envision a scenario in which we would accept replaying an old, in-progress batch on a newer state.

theodorebugnet (approved):

Issue #2271, opened 2025-12-26
[Tracker] EVM internal audit


preston-evans98:
## Description

This issue tracks internal audit progress. 

## Areas to Audit
- [x] EVM module initialization
- [x] EVM authenticator
- [x] EVM execute call
- [x] EVM RPC handlers

### Critical 
 - [x] `StorageInspector` rebates gas costs as they happen (since these accesses are charged by the SDK later). This means that a caller can run a (roughly) infinite loop of SSTOREs, causing the client to hang. Instead, costs should be rebated after the EVM returns. #2291 

### High
 - [x] No practical gas limit is enforced on EVM txs when the paymaster is active. We should set a configurable hard limit on EVM tx consumption, defaulting to 30M gas. #2256

### Moderate:
- [ ] No way to update `EvmRuntimeConfig` after genesis: #2292
	- [x] Support add/remove allowed contract deployers
	- [ ] Support modifying `hardforks`, `limit_contract_code_size` and `block_gas_limit`. 
	- [ ] Determine if `coinbase` address is used. If so, determine whether to allow changes.
- [x] EVM Authenticator accepts all tx types and coerces them to EIP-1559. Only EIP-1559 txs should be accepted #2294
- [x] `CfgEnv::memory_limit` should be set since we've raised the gas limit significantly. 50MB seems reasonable. #2293
- [x] Our use of `CfgEnv.disable_balance_checks` is incorrect. This option should be left with its default value #2256
- [x] Rollup may panic if exactly the right amount of funds is sent to cover the prechecks. #2270 

### Low
- [x] `validate_chain_id` doesn't respect EIP-7702 (should accept `0`) #2295
- [ ] We allow the code_hash to be set to an unrelated value for accounts set at genesis. We should compute the hash from the code instead.
- [x] `RawTx`, `FullyBakedTx`, and `RlpEvmTransaction` are all missing length limits. Set `RawTx` and `FullyBakedTx` to 1MB. Leave `RlpEvmTransaction` as is. #2296 
- [ ] Re-examine TODOS in EVM DB; should be fixed, removed, or marked as non-security critical
  - [ ] `TODO move to new_raw_with_hash for better performance`
  - [ ] `TODO: Here we generate a default account and set the balance from bank`
[...]

PR #2391, opened 2026-01-23
EVM: Add test cases for `eth_getBlockBy{Hash,Number}`

citizen-stig:
# Description

Adds a set of tests for RPC methods `eth_getBlockByNumber` and `eth_getBlockByHash`

It has only 1 functional change:

* RPC endpoint returns truly finalized rollup height when queried for `Finalized` block by number. This change is not affecting any of existing customers, because all rollups are on instant finality DAs

## Follow up for later (optimization for StdSequencer)

> crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs:R384-396
>
> **finalized_block_number makes an additional archival state construction per RPC call**
> Every call to `finalized_block_number` constructs a full archival state accessor via `build_finalized_state()`, which for `UseExplicit` calls `new_archival_with_true_slot_number` → `build_archival_state`. This includes kernel lookups for slot mappings, base fee, visible slot number, and a test storage query (see `build_archival_state` at `crates/module-system/sov-modules-api/src/state/accessors/http_api.rs:786-862`).
> 
> This is called from `block_tag_to_pending_or_block`, `resolve_block_number`, and `block_id_to_pending_or_block`, meaning a single `eth_getBlockByNumber("finalized")` or `eth_feeHistory(..., "finalized")` call triggers a full archival state build. For the `TrackStorageHead` policy (preferred sequencer), it just clones, so it's cheap. But for `UseExplicit` (standard sequencer), the overhead could be noticeable under high RPC load. Consider caching the finalized block number in the checkpoint or computing it lazily.

## Linked Issues
- Related to #2271  

## Testing
All tests have been adjusted and/or manually verified.

## Docs
Adds documentation in `docs/` about method and test cases in human readable form


crates/module-system/sov-modules-api/src/state/accessors/concurrent_state_checkpoint.rs
```diff
@@ -13,11 +13,30 @@ pub struct ConcurrentStateCheckpoint<S: Spec> {
...
 impl<S: Spec> ConcurrentStateCheckpoint<S> {
     /// Create a `ConcurrentStateCheckpoint` containing the same changes as the given `StateCheckpoint`.
-    pub fn from_state_checkpoint(mut state_checkpoint: StateCheckpoint<S>) -> Self {
+    ///
+    /// Note: this defaults the latest finalized slot to the most recent slot
+    /// available in storage. That effectively treats all known slots as
+    /// finalized. Call `from_state_checkpoint_with_finalized_slot` if you need
+    /// to preserve the node's true finalized slot semantics.
+    pub fn from_state_checkpoint(state_checkpoint: StateCheckpoint<S>) -> Self {
+        let latest_finalized_slot_number = state_checkpoint.delta.inner.latest_version();
```
preston-evans98, line 43: I'm not sure that the semantics of this are correct. IIRC, "latest_version" doesn't correspond to any actual slot. It's really "number of latest finalized slot" plus number of unfinalized *rollup blocks*.  (It's defined in this awkward way because we don't know how many slots will elapse before the rollup block contents actually land on DA).

preston-evans98: I'm not sure what the best way to conceptualize this is, but it seems like trying to directly expose a "finalized slot" abstraction will be tricky with the preferred sequencer. 

Maybe it's cleaner to make the abstraction something like `slots_are_unfinalized_after() -> Option<SlotNumber>`. Then we can have the std sequencer return a value (which is just the last finalized slot number actually processed by the node)  and have the preferred sequencer return None

docs/eth_getBlockBy_test_cases.md
```diff
@@ -0,0 +1,726 @@
...
+
+### Rollup-specific behavior (intended)
+- `latest` resolves to `pending` (both return the pending block being constructed).
+- `safe` and `finalized` both resolve to the **latest finalized** block (may lag `eth_blockNumber` when `finalization_blocks > 0`).
+- `requireCanonical` in EIP-1898 is accepted but ignored (no reorgs in this rollup).
+- If there are no pending txs, `pending`/`latest` fall back to the latest sealed block.
+- Pending blocks use a **synthetic hash** (non-zero) and are resolvable via `eth_getBlockByHash`.
+- Pending block fields (logsBloom, roots, size, gasUsed) are computed from pending txs/receipts.
+
+### Known deviations vs Ethereum L1 (bugs to track)
```
preston-evans98: nit: These aren't really bugs. this language might attract spam PRs from airdrop farmers

docs/eth_getBlockBy_test_cases.md
```diff
@@ -0,0 +1,726 @@
...
+
+### eth_getBlockByHash
+
+| ID | Priority | Description | Expected values |
+|----|----------|-------------|-----------------|
+| TC32 | P0 | Valid sealed hash returns block | Block matches `get_block_by_number(N)` |
+| TC33 | P0 | Non-existent hash returns `null` | Random hash -> `None` |
+| TC34 | P0 | Zero hash returns `null` | `get_block_by_hash(0x0...) == None` |
+| TC35 | P1 | `details` param works same as by_number | Hashes vs full txs |
+| TC36 | P1 | Round-trip: get block, extract hash, get by hash | All fields identical |
```
preston-evans98, line 226: non-blocking: We should probably also add checks on getting blocks by synthetic hash and make sure that the returned tx details are correct. In particular, we should confirm that the tx.hash returned matches the containing block

crates/module-system/sov-modules-api/src/state/accessors/concurrent_state_checkpoint.rs
citizen-stig: > I'm not sure that the semantics of this are correct. IIRC, "latest_version" doesn't correspond to any actual slot. It's really "number of latest finalized slot" plus number of unfinalized rollup blocks.

Is the problem that EVM block hash that returned by this endpoint might have a different height after its content is land on DA (probably larger, not smaller)?

What then we expect to return when EVM client ask for BlockBy(finalized)?




> I'm not sure what the best way to conceptualize this is, but it seems like trying to directly expose a "finalized slot" abstraction will be tricky with the preferred sequencer.

I think previous consensus on this point was that for preferred sequencer last slot number is treated as finalized. And Standard sequencer just returns data from the node.

As a solution, I can remove this code and comment out all tests that rely on finalized semantics and call it as known issue that will be tackled later.

docs/eth_getBlockBy_test_cases.md
citizen-stig: Added 1 special test for that and additional checks for it in another test in 153eac22b9a1432818f9396a0b9b2638b5fae0de

docs/eth_getBlockBy_test_cases.md
citizen-stig: Removed mention of bugs, Don't why it does not show it.

crates/module-system/sov-modules-api/src/state/accessors/concurrent_state_checkpoint.rs
preston-evans98: > I think previous consensus on this point was that for preferred sequencer last slot number is treated as finalized. And Standard sequencer just returns data from the node.

I thought we had agreed that the latest *rollup height* is treated as finalized. We don't know what slot number that will correspond to, though.

citizen-stig: So we still need this slot number in the checkpoint

But blocks are fetched differently now: 105181255d7fcc946f6eccc5cbabe75ab6336739

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -369,18 +381,34 @@ where
...
 
+    fn finalized_block_number(&self, state: &mut ApiStateAccessor<S>) -> u64 {
+        let current_end = *self.block_numbers(state).end();
+        match state.build_finalized_state() {
+            Ok(mut archival) => *self.block_numbers(&mut archival).end(),
+            Err(error) => {
+                tracing::debug!(
+                    ?error,
+                    "Failed to build finalized state, using current head"
+                );
```
preston-evans98: Is this correct? It seems like this is probably a bug if we ever hit it, and we should probably prefer to return 5xx rather (i.e. panic) rather than returning data from (potentially) the wrong block

preston-evans98 (approved):

preston-evans98 (approved):

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
citizen-stig: used panic!

PR #2397, opened 2026-01-23
Replace "Pending" block with the synthetic "latest" block

preston-evans98:
# Description

This PR makes a significant change to the way we handle the active block while the sequencer is accepting transactions for it. Previously, we treated the block as "pending", but this caused some standard tooling (notably foundry) to choke. Now, we treat the open block as being sealed by creating a fake ("synthetic") block header for it, and then doing a pretend reorg of depth one when a new tx is added. 

To avoid overwhelming subscribers, we notify `new_heads` subscribers about these synthetic blocks at most once every 200ms.

Synthetic blocks look exactly like real blocks except that their block hashes are special marker values: `[SENTINEL || BLOCK_NUMBER || NUM_TXS]`. Because the "hashes" contain all the information needed to reconstruct a synthetic block, we don't have to store these blocks anywhere and we can still serve requests for block by hash. 

However, there is one limitation to synthetic blocks. Because the EVM shares state with the sov-modules system, we can't reconstruct the rollup state as of a particular synthetic block just by replaying EVM txs. Since we don't store sov-modules tx bodies anywhere, this creates a problem - there's no way to go back to the state as of a particular synthetic block. To fix this issue, we create a cache of API State accessors which stores concread::ReadTxns for all of the recent states. 

## Linked Issues
- Related to # (issue) 

## Testing
I've manually tested with forge deploy to ensure this doesn't regress foundry support.

## Docs

### DIFF
--- a/crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
+++ b/crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
@@ -120,7 +120,7 @@ impl SyntheticBlockWatermark {
         if synthetic_block.num_transactions() != 0
             && self
                 .tx_index_of_last_notification_if_known
-                .map_or(true, |idx| idx < synthetic_block.last_tx_index())
+                .is_none_or(|idx| idx < synthetic_block.last_tx_index())
         {
             return self.advance_and_emit_synthetic_block_notification(synthetic_block);
         }
@@ -143,7 +143,7 @@ impl SyntheticBlockWatermark {
         if synthetic_block.num_transactions() != 0
             && self
                 .tx_index_of_last_notification_if_known
-                .map_or(true, |idx| idx < synthetic_block.last_tx_index())
+                .is_none_or(|idx| idx < synthetic_block.last_tx_index())
         {
             return SyntheticBlockWatermarkAdvanceResult::NewSyntheticBlock;
         }
@@ -257,7 +257,7 @@ where
                     // Send all of the notifications for new real blocks.
                     while let SyntheticBlockWatermarkAdvanceResult::NewRealBlock(block_number) = watermark.peek(&pending_block) {
                         let sealed = self.evm.blocks.get(&block_number, &mut state).unwrap_infallible().expect("Block was notified but did not exist. This is a bug!");
-                        let rpc_header = Header::from_consensus(sealed.header.into(), None, Some(U256::from(sealed.rlp_size)));
+                        let rpc_header = Header::from_consensus(sealed.header, None, Some(U256::from(sealed.rlp_size)));
                         self.send(&rpc_header).await?;
                         watermark.advance(&pending_block);
                         sent = true;
@@ -396,8 +396,10 @@ mod tests {
         tx_start: u64,
         tx_end: u64,
     ) -> SyntheticBlockWithoutRootsAndBloom {
-        let mut header = Header::default();
-        header.number = block_number;
+        let header = Header {
+            number: block_number,
+            ..Default::default()
+        };
         SyntheticBlockWithoutRootsAndBloom::new(header, tx_start..tx_end)
     }
 
--- a/crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
+++ b/crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
@@ -21,6 +21,9 @@ use std::fmt::Debug;
 use std::sync::Arc;
 use thiserror::Error;
 
+/// Don't send new heads notifcations for synthetic blocks more frequently than this.
+const SYNTHETIC_NEW_HEADS_MAX_FREQUENCY_MS: u64 = 200;
+
 #[derive(Error, Debug)]
 pub enum Error {
     #[error("Block does not exist")]
@@ -43,6 +46,67 @@ where
     evm: Evm<S>,
 }
 
+struct SyntheticBlockWatermark {
+    block_number_of_last_notification: u64,
+    // The tx index of the last notification, if known. 
+    // It's not known to use when we send a notification for a *real* block, 
+    // so we have special handling in that case
+    tx_index_of_last_notification_if_known: Option<u64>,
+}
+
+enum SyntheticBlockWatermarkAdvanceResult {
+    NewRealBlock(u64),
+    NewSyntheticBlock,
+    NoChange,
+}
+
+impl SyntheticBlockWatermark {
+    fn from_synthetic_block(synthetic_block: &SyntheticBlockWithoutRootsAndBloom) -> Self {
+        Self {
+            block_number_of_last_notification: synthetic_block.header.number,
+            tx_index_of_last_notification_if_known: synthetic_block.transactions.end,
+        }
+    }
+
+    fn advance_and_emit_synthetic_block_notification(&mut self, synthetic_block: &SyntheticBlockWithoutRootsAndBloom) -> SyntheticBlockWatermarkAdvanceResult {
+        self.block_number_of_last_notification = synthetic_block.header.number;
+        self.tx_index_of_last_notification_if_known = synthetic_block.transactions.end.saturating_sub(1);
+        SyntheticBlockWatermarkAdvanceResult::NewSyntheticBlock
+    }
+
+
+    fn advance_and_emit_real_block_notification(&mut self) -> SyntheticBlockWatermarkAdvanceResult {
+        // Per https://www.quicknode.com/docs/ethereum/eth_subscribe - we should send a notification each time a new header is appended
+        // Even if that means sending multiple notifications in succession. This means that we should increment the block number by one
+        // rather than jumping to the height of the synthetic block.
+        self.block_number_of_last_notification += 1;
+        // There's no guaranteed relationship between the first tx index of the synthetic block and the last tx index of the *current* real block that we need to notify for.
+        // We might have missed some synthetic block notifications due to tokio's nondeterminism. That's fine; just set it to None to reflect that we don't know for sure.
+        self.tx_index_of_last_notification_if_known = None;
+        SyntheticBlockWatermarkAdvanceResult::NewRealBlock(self.block_number_of_last_notification)
+    }
+    
+
+    // We want to send a notification if...
+    // There's been a new real block
+    // There's been a new synthetic block *and* it's been more than 200ms since the last notification
+    fn advance(&mut self, synthetic_block: &SyntheticBlockWithoutRootsAndBloom) -> SyntheticBlockWatermarkAdvanceResult {
+        if self.block_number_of_last_notification < synthetic_block.header.number {
+            return self.advance_and_emit_real_block_notification();
+        }
+
+        // If the synthetic block is empty, don't notify. 
+        // Similarly if we've already sent a notification for this synthetic block, don't notify again.
+        if synthetic_block.num_transactions() != 0 && || self.tx_index_of_last_notification_if_known.unwrap_or(0) < synthetic_block.last_tx_index() {
+            return self.advance_and_emit_synthetic_block_notification(synthetic_block);
+        }
+
+        SyntheticBlockWatermarkAdvanceResult::NoChange
+    }
+
+}
+
+// TODO: Refactor this into a long-running background task to reduce overhead. Right now, we do duplicate fetching for each subscription.
 impl<S, Seq> Streamer<S, Seq>
 where
     S: Spec,
@@ -61,7 +125,8 @@ where
     /// Stream new logs matching the provided filter to the subscriber.
     pub async fn logs(&self, filter: Box<Filter>) -> Result<(), Error> {
         let mut state = self.ethereum.api_state_accessor();
-        let pending_block = self.evm.pending_block(&mut state);
+        let pending_block = self.evm.pending_block(None, &mut state);
+        let synthetic_block_watermark = SyntheticBlockWatermark::from_synthetic_block(pending_block);
         let mut tx_watermark = Watermark::new(..pending_block.transactions.end);
 
         // Fetch the initial block. If it's stale, it will be replaced below.
@@ -79,7 +144,7 @@ where
                     }
 
                     let mut state = self.ethereum.api_state_accessor();
-                    let pending_block = self.evm.pending_block(&mut state);
+                    let pending_block = self.evm.pending_block(None, &mut state);
 
                     for tx_idx in tx_watermark.advance(..pending_block.transactions.end) {
                         let (receipt, time) = self.get_receipt(tx_idx, &mut state)?;
@@ -101,14 +166,18 @@ where
     }
 
     /// Stream new block headers to the subscriber.
-    pub async fn blocks(&self) -> Result<(), Error> {
+    pub async fn new_heads(&self) -> Result<(), Error> {
+        // Pick up here
+        // Long term todo: Refactor this into a long-running background task 
+
         let mut state = self.ethereum.api_state_accessor();
-        let latest_block = self.evm.latest_block(&mut state);
-        let mut block_watermark = Watermark::new(..latest_block.header.number + 1);
+        let pending_block = self.evm.pending_block(None, &mut state);
+        let mut watermark = SyntheticBlockWatermark::from_synthetic_block(&pending_block);
 
         let mut state_updates = self.ethereum.sequencer.api_state().checkpoint_receiver();
         let mut shutdown_receiver = self.ethereum.shutdown_receiver.clone();
 
+        let mut last_send_time = Instant::now();
         loop {
             tokio::select! {
                 result = state_updates.changed() => {
@@ -117,13 +186,20 @@ where
                         break;
                     }
 
+                    // TODO(@preston-evans98) - I think this checkpoint change notification might be problematic on resync/restart. Figure that out.
+                    // TODO: Wait - if it's been less than 200ms since the last update, skip the pending block check.
                     let mut state = self.ethereum.api_state_accessor();
                     let latest_block = self.evm.latest_block(&mut state);
 
+                    let mut sent = false;
                     for block_number in block_watermark.advance(..latest_block.header.number + 1) {
                         let block = self.get_block(block_number, &mut state)?;
-                        let rpc_header = self.evm.get_rpc_header(block)?;
+                        let rpc_header = self.evm.get_rpc_header(block, &mut state)?;
+                        sent = true;
                         self.send(&rpc_header).await?;
+                    }
+                    if sent {
+                        last_update_time = Instant::now();
                     }
                 }
                 _ = shutdown_receiver.changed() => {
--- a/crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
+++ b/crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
@@ -1,6 +1,7 @@
 use super::watermark::Watermark;
 use crate::Ethereum;
-use alloy_rpc_types::Filter;
+use alloy_primitives::U256;
+use alloy_rpc_types::{Filter, Header};
 use jsonrpsee::DisconnectError;
 use jsonrpsee::SubscriptionMessage;
 use jsonrpsee::SubscriptionSink;
@@ -10,13 +11,16 @@ pub use sov_evm::EthereumAuthenticator;
 use sov_evm::Evm;
 use sov_evm::MaybeSealedBlock;
 use sov_evm::Receipt;
+use sov_evm::SyntheticBlockWithoutRootsAndBloom;
 use sov_modules_api::capabilities::HasKernel;
 use sov_modules_api::da::Time;
 use sov_modules_api::ApiStateAccessor;
 use sov_modules_api::Spec;
+use sov_modules_api::prelude::UnwrapInfallible;
 use sov_rpc_eth_types::EthApiError;
 use sov_rpc_eth_types::LogWithExecutionTimestamp;
 use sov_sequencer::Sequencer;
+use std::time::Duration;
 use std::fmt::Debug;
 use std::sync::Arc;
 use thiserror::Error;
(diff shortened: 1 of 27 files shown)


crates/module-system/module-implementations/sov-evm/src/rpc/trace.rs
```diff
@@ -57,14 +57,15 @@ where
         EthApiError,
     > {
         // Get the block - could be pending or sealed
+        // TODO: Skip fetching the whole block; we just need the block number
```
citizen-stig: Is it planned for follow up, or something left in this PR?

crates/module-system/module-implementations/sov-evm/src/evm/primitive_types.rs
```diff
@@ -208,7 +374,7 @@ pub enum MaybeSealedBlock {
     /// SealedBlock
     Sealed(SealedBlock),
     /// Pending
-    Pending(crate::Block),
+    PartialSynthetic(SyntheticBlockWithoutRootsAndBloom),
```
citizen-stig: Is the comment up to date? How PartialSynthetic different SealedSynthetic?

crates/module-system/module-implementations/sov-evm/src/evm/primitive_types.rs
```diff
@@ -238,52 +404,62 @@ impl MaybeSealedBlock {
...
+    pub fn header_if_sealed(&self) -> Option<&Header> {
+        match self {
+            Self::Sealed(block) => Some(block.header.inner()),
+            Self::PartialSynthetic(_) => None,
+        }
     }
 }
 
+// #[cfg(feature = "native")]
+// impl From<MaybeSealedBlock> for Sealed<Header> {
```
citizen-stig: Do we still need this commented code?

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -316,16 +353,26 @@ where
...
     }
 
-    /// Retrieves the pending block.
-    pub fn pending_block(&self, state: &mut ApiStateAccessor<S>) -> crate::Block {
+    /// Retrieves the pending block. We maintain the invariant that the pending block always has number
+    /// latest_sealed_block.number + 1. (Both are updated during the finalize_hook, so they're atomic).
+    ///
+    /// Note that values in the block_env (including the block number there!) are updated during the begin_rollup_block_hook, so the values here may be stale
+    /// if this function is called while no rollup block is in progress. In that case, the number of transactions will be zero.
+    // TODO: Have this function return None if no block is pending!
```
citizen-stig: What is the status of this TODO?

citizen-stig (approved):
I haven't found anything blocking, so feel free to address left comments in the most reasonable way.

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -217,51 +272,39 @@ where
         block_number: u64,
         state: &mut ApiStateAccessor<S>,
     ) -> Option<MaybeSealedBlock> {
+        // Check if the block is already sealed. Important: We must do this before checking the block env, because blocks are sealed in the finalize_hook.
+        // but the block env is set during the begin_rollup_block_hook.
```
cemozerr, line 353: Does this mean it would be better to also set the block env for the new block in the finalize hook?

There being a gap between the two doesn't make much sense to me. Altough, we don't need to block this PR for this.

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -316,16 +353,26 @@ where
...
 
-    /// Retrieves the pending block.
-    pub fn pending_block(&self, state: &mut ApiStateAccessor<S>) -> crate::Block {
+    /// Retrieves the pending block. We maintain the invariant that the pending block always has number
+    /// latest_sealed_block.number + 1. (Both are updated during the finalize_hook, so they're atomic).
+    ///
+    /// Note that values in the block_env (including the block number there!) are updated during the begin_rollup_block_hook, so the values here may be stale
+    /// if this function is called while no rollup block is in progress. In that case, the number of transactions will be zero.
+    // TODO: Have this function return None if no block is pending!
+    pub fn pending_block(
```
cemozerr, line 535: Shall we rename this to synthetic block? or does keeping it as pending block still serve a purpose?

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -43,6 +50,101 @@ where
     evm: Evm<S>,
 }
 
+#[derive(Debug, Clone)]
+struct SyntheticBlockWatermark {
+    block_number_of_last_notification: u64,
+    // The tx index of the last notification, if known.
+    // It's not known to use when we send a notification for a *real* block,
```
cemozerr:
```suggestion
    // It's not known to us when we send a notification for a *real* block,
```

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -43,6 +50,101 @@ where
...
+    NewRealBlock(u64),
+    NewSyntheticBlock,
+    NoChange,
+}
+
+impl SyntheticBlockWatermark {
+    fn from_synthetic_block(synthetic_block: &SyntheticBlockWithoutRootsAndBloom) -> Self {
+        Self {
+            block_number_of_last_notification: synthetic_block.header().number,
+            tx_index_of_last_notification_if_known: Some(synthetic_block.last_tx_index()),
```
cemozerr: An extremely narrow (and minor) bug here that Devin review caught on its own. 

A subscriber who connects right at genesis misses exactly one notification for the very first transaction. After that, everything works normally.

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -43,6 +50,101 @@ where
     evm: Evm<S>,
 }
 
+#[derive(Debug, Clone)]
+struct SyntheticBlockWatermark {
```
cemozerr, line 57:
```suggestion
// The `SyntheticBlockWatermark` tracks what we've already notified subscribers 
// about, so we don't send duplicate notifications
struct SyntheticBlockWatermark {
```

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -43,6 +50,101 @@ where
...
+
+    // We want to send a notification if...
+    // There's been a new real block
+    // There's been a new synthetic block *and* it's been more than 200ms since the last notification
+    fn advance(
+        &mut self,
+        synthetic_block: &SyntheticBlockWithoutRootsAndBloom,
+    ) -> SyntheticBlockWatermarkAdvanceResult {
+        if self.block_number_of_last_notification
+            < synthetic_block.header().number.saturating_sub(1)
```
cemozerr: How this condition actually checks that there has been a new real block definitely requires some explanation in code here haha
```suggestion
        // Synthetic block number is always one greater than the sealed block number, 
        // so if the last notified block number is less than the synthetic block number - 1, 
        // we need to advance and emit real block notification.
        if self.block_number_of_last_notification
            < synthetic_block.header().number.saturating_sub(1)
```

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -101,29 +203,94 @@ where
...
 
-                    for block_number in block_watermark.advance(..latest_block.header.number + 1) {
-                        let block = self.get_block(block_number, &mut state)?;
-                        let rpc_header = self.evm.get_rpc_header(block)?;
+                    // Send all of the notifications for new real blocks.
+                    while let SyntheticBlockWatermarkAdvanceResult::NewRealBlock(block_number) = watermark.peek(&pending_block) {
+                        watermark.advance(&pending_block);
+                        let sealed = self.evm.blocks.get(&block_number, &mut state).unwrap_infallible().expect("Block was notified but did not exist. This is a bug!");
+                        let rpc_header = Header::from_consensus(sealed.header.into(), None, Some(U256::from(sealed.rlp_size)));
                         self.send(&rpc_header).await?;
```
cemozerr: Would be better to send the notification and then advance the watermark:


```suggestion
                        let sealed = self.evm.blocks.get(&block_number, &mut state).unwrap_infallible().expect("Block was notified but did not exist. This is a bug!");
                        let rpc_header = Header::from_consensus(sealed.header.into(), None, Some(U256::from(sealed.rlp_size)));
self.send(&rpc_header).await?;
                        watermark.advance(&pending_block);

```

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -101,29 +203,94 @@ where
...
+                        sent = true;
+                    }
+
+                    // If the state change only created a synthetic block, send it only if we haven't sent one too recently
+                    if let SyntheticBlockWatermarkAdvanceResult::NewSyntheticBlock = watermark.peek(&pending_block) {
```
cemozerr, line 271: It seems like we could send an update for a new head, and then immediately after send an update for a new synthetic block, (if the pending block we got with the state update above was a new pending block with transaction in it. 

Is it desired to send real block and synthetic block notification one after another?

cemozerr: this is because we update last send time in `if sent {` below, and not inside the `let` statements.

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -101,29 +203,94 @@ where
...
+                        if last_send_time.elapsed() >= Duration::from_millis(SYNTHETIC_NEW_HEADS_MAX_FREQUENCY_MS) {
+                            watermark.advance(&pending_block);
+                            let (rpc_header, _txs) = self.evm.get_synthetic_block_contents_slow(pending_block, &mut state)?;
+                            self.send(&rpc_header).await?;
+                            sent = true;
+                        } else {
+                            // If we don't send the notification now, scheudle a wakeup to try again later. This handles the edge case
```
cemozerr:
```suggestion
                            // If we don't send the notification now, schedule a wakeup to try again later. This handles the edge case
```

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -101,29 +203,94 @@ where
...
+                            // where we have a bunch of new synethitic blocks in rapid succession followed by a gap; in that case,
```
cemozerr:
```suggestion
                            // where we have a bunch of new synthetic blocks in rapid succession followed by a gap; in that case,
```

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -101,29 +203,94 @@ where
...
+                            // we'd never notify for the newer txs.
+                            // Spawn a wakeup for later to handle the edge case. Note that we don't try to dedup wakeups; duplicates are handled in the wakeup_receiver case.
+                            let wakeup_sender = wakeup_sender.clone();
+                            tokio::spawn(async move {
+                                tokio::time::sleep(Duration::from_millis(SYNTHETIC_NEW_HEADS_MAX_FREQUENCY_MS) ).await;
```
cemozerr: assuming we run somewhere close to max capacity, will we be spawning thousands of threads a second per connection? and is that okay?

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -101,29 +203,94 @@ where
...
+                    // In the SyntheticBlockWatermarkAdvanceResult::None case, do nothing.
+                }
+                _ = wakeup_receiver.changed() => {
+                    // If we've sent a notification too recently, ignore the wakeup. Either we've sent all notifications already or - we'll get woken up again in no more than 200ms
+                    if last_send_time.elapsed() >= Duration::from_millis(SYNTHETIC_NEW_HEADS_MAX_FREQUENCY_MS) {
+                        let mut state = self.ethereum.api_state_accessor();
+                        let pending_block = self.evm.pending_block(None, &mut state);
+                        // Only send the notification if it's for a synthetic block. Real blocks are guaranteed to be handeld by the main state_change watcher.
+                        if let SyntheticBlockWatermarkAdvanceResult::NewSyntheticBlock = watermark.peek(&pending_block) {
+                            watermark.advance(&pending_block);
```
cemozerr, line 303: again advancing order.

cemozerr (approved):
LGTM!

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
preston-evans98: We can't. It's a user state value, but you can only touch accessory state during the finalize hook

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: Good catch. Fixed

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: Added

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: Good catch. Added one more update to the timer to fix this.

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: We'll be spawning tasks not threads (tasks are *much* cheaper) - but it's not ideal. Let me think about this one

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: We can't do this because sending consumes the header. it shouldn't matter though - if the send fails, we'll exist this loop because of the `?` so the watermark won't be used again.

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: Added a simple flag to check if there's already a wakeup scheduled

crates/module-system/module-implementations/sov-evm/src/rpc/trace.rs
preston-evans98: This is left from the old implementation. Going to leave it out of scope for now since this PR is pretty big already

crates/module-system/module-implementations/sov-evm/src/evm/primitive_types.rs
preston-evans98: Nope, removed!

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
preston-evans98: Done

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
preston-evans98: This returns the "pending" version of the synthetic block. Happy to discuss more, but planning to leave it for now

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: Done

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: Fixed

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: Same here - can't be helped due to ownership rules but shouldn't be able to cause problems.

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -242,48 +251,53 @@ where
...
                         let sealed = self.evm.blocks.get(&block_number, &mut state).unwrap_infallible().expect("Block was notified but did not exist. This is a bug!");
-                        let rpc_header = Header::from_consensus(sealed.header.into(), None, Some(U256::from(sealed.rlp_size)));
+                        let rpc_header = Header::from_consensus(sealed.header, None, Some(U256::from(sealed.rlp_size)));
                         self.send(&rpc_header).await?;
+                        watermark.advance(&pending_block);
                         sent = true;
                     }
 
+                    if sent {
+                        last_send_time = std::time::Instant::now();
```
cemozerr, line 267: nit: why not just update last send time inside the while above?

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
```diff
@@ -242,48 +251,53 @@ where
...
+                            // where we have a bunch of new synthetic blocks in rapid succession followed by a gap; in that case,
                             // we'd never notify for the newer txs.
-                            // Spawn a wakeup for later to handle the edge case. Note that we don't try to dedup wakeups; duplicates are handled in the wakeup_receiver case.
-                            let wakeup_sender = wakeup_sender.clone();
-                            tokio::spawn(async move {
-                                tokio::time::sleep(Duration::from_millis(SYNTHETIC_NEW_HEADS_MAX_FREQUENCY_MS) ).await;
-                                let _ = wakeup_sender.send(());
-                            });
+                            // Spawn a wakeup for later to handle the edge case. Note that we schedule at most one wakeup every 200ms
+                            if !has_waker_task.swap(true, Ordering::SeqCst) {
```
cemozerr, line 282: nit: Since now waker task handles deduplication, i dont think we need the time check in the wake up receiver, but might be defensive to keep it.

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -29,6 +37,68 @@ use sov_modules_api::{ApiStateAccessor, Spec, VersionReader};
 use sov_rollup_interface::common::RollupHeight;
 use sov_rpc_eth_types::{EthApiError, LogWithExecutionTimestamp, RpcInvalidTransactionError};
 
+// Prune synthetic blocks more than this number of blocks away from the latest block.
+const SYNTHETIC_BLOCKS_CACHE_PRUNE_INTERVAL: u64 = 20;
+
+/// Cache of synthetic block states by hash.
+///
+/// Currently, there's no way to recreate the state of a synthetic block once it's gone. This is because
+/// EVM state is shared with the sov-modles system, so any sov txs can impact the state of the synthetic block - but only
```
cemozerr, line 46:
```suggestion
/// EVM state is shared with the sov-modules system, so any sov txs can impact the state of the synthetic block - but only
```

crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
```diff
@@ -857,6 +861,27 @@ impl<S: Spec + 'static> ApiStateAccessor<S> {
         Ok(state)
     }
 
+    /// Clones the accessor for use in RPC contexts, rese
```
cemozerr, line 864: nit: comment left midway

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -294,16 +447,70 @@ where
                 let pending_or_block = self.block_tag_to_pending_or_block(tag, state);
                 match pending_or_block {
                     PendingOrBlock::Number(number) => self.get_maybe_sealed_block(number, state),
-                    PendingOrBlock::Pending => {
-                        Some(MaybeSealedBlock::Pending(self.pending_block(state)))
+                    PendingOrBlock::Pending => match self.pending_block(None, state) {
+                        Some(pending) => Some(MaybeSealedBlock::PendingSynthetic(pending)),
+                        None => {
```
cemozerr, line 452:
```suggestion
                        // pending_block() returns None when there are no pending txs, and so fall back to the latest sealed block in those cases
                        None => {
```

crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -345,15 +634,17 @@ where
                 .map(|blob_gas| blob_gas.excess_blob_gas),
             base_fee_per_gas: Some(current_block_env.basefee),
             gas_limit: current_block_env.gas_limit,
-            // Default values
-            ommers_hash: EMPTY_OMMER_ROOT_HASH,
-            beneficiary: Address::ZERO,
+            // Values that will be filled in later
+            gas_used: 0,
```
cemozerr, line 638: in a future PR, i hope we can refactor it such that these values are not set to mistakenly usable values like this by the callers but instead just none's or something, but no need to do it in this PR or prioritize for now.

crates/module-system/sov-modules-api/src/module/mod.rs
```diff
@@ -169,7 +170,12 @@ pub trait EventEmitter: ModuleInfo {
     /// Execution context.
     type Spec: Spec;
     /// Module defined event resulting from a call method.
-    type Event: Debug + BorshSerialize + BorshDeserialize + 'static + core::marker::Send;
+    type Event: Debug
+        + BorshSerialize
+        + BorshDeserialize
+        + 'static
```
cemozerr, line 176: do we need to add this to changelog as it might break customer modules?

cemozerr (approved):
LGTM again

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: Nbd, but it requires a syscall to update the time so this is a bit more efficient

crates/full-node/sov-ethereum/src/handlers/subscribe/service.rs
preston-evans98: Going to leave it defensively, but I think you're correct

preston-evans98:
Merged as-is without fixing some nits on comments to avoid waiting on CI - will include those fixes in the next PR

PR #2401, opened 2026-01-26
`Proxy` tool: Save the leader and followers to a file after receiving a notification from `Postgres`.

bkolad:
# Description

This PR adds a node-discovery subscription system that allows external services (such as a load-balancing proxy) to receive real-time updates on sequencer cluster membership: 

1. Add `subscribe_cluster_info_loop` to `NodeDiscovery` that listens for PostgreSQL notifications and writes cluster info to a file                                                                                                                           
2. Fix node address resolution when binding to `0.0.0.0` by detecting the actual local IP                                                                                                                                                                     
3. Add validation and error handling for cluster info parsing   

## Linked Issues
- Related to #1524

### DIFF
--- a/crates/utils/sov-proxy-utils/Cargo.toml
+++ b/crates/utils/sov-proxy-utils/Cargo.toml
@@ -8,10 +8,11 @@ homepage.workspace = true
 publish = false
 repository.workspace = true
 
+
 [dependencies]
 anyhow = { workspace = true }
-sqlx = { version = "0.8", features = ["runtime-tokio", "postgres"] }
-tokio = { workspace = true }
+sqlx = { version = "0.8", features = ["runtime-tokio", "tls-rustls-ring-webpki", "postgres"] }
+tokio = { workspace = true, features = ["time", "sync"] }
 tracing = { workspace = true }
 
 [lints]
--- a/crates/utils/sov-proxy-utils/Cargo.toml
+++ b/crates/utils/sov-proxy-utils/Cargo.toml
@@ -11,9 +11,10 @@ repository.workspace = true
 
 [dependencies]
 anyhow = { workspace = true }
-sqlx = { version = "0.8", features = ["runtime-tokio", "tls-rustls-ring-webpki", "postgres"] }
-tokio = { workspace = true, features = ["time", "sync"] }
+sqlx = { version = "0.8", features = ["runtime-tokio", "tls-rustls-ring-webpki", "postgres", "derive"], default-features = false }
+tokio = { workspace = true, features = ["time", "sync"], default-features = false }
 tracing = { workspace = true }
+
 
 [lints]
 workspace = true
--- a/crates/utils/sov-proxy-utils/src/lib.rs
+++ b/crates/utils/sov-proxy-utils/src/lib.rs
@@ -156,7 +156,7 @@ impl NodeDiscovery {
         match self.get_cluster_info().await {
             Ok(info) => {
                 if let Err(error) = write_to_file(path, info.to_file_content()).await {
-                    tracing::warn!(?error, "Failed to update the cluster info file.");
+                    tracing::warn!(?error, ?path, "Failed to update the cluster info file.");
                 } else {
                     // Notify watchers that the file was saved successfully.
                     let _ = self.file_saved_sender.send(());
--- a/crates/utils/sov-proxy-utils/src/lib.rs
+++ b/crates/utils/sov-proxy-utils/src/lib.rs
@@ -3,11 +3,16 @@
 //! This crate provides the [`Proxy`] struct to retrieve leader and follower
 //! IP addresses from the PostgreSQL database atomically.
 
+use std::collections::HashSet;
 use std::net::SocketAddr;
+use std::time::Duration;
 
-use anyhow::Result;
-use sqlx::postgres::PgPool;
+use anyhow::{Context, Result};
+use sqlx::postgres::{PgListener, PgPool};
 use sqlx::FromRow;
+use tokio::sync::watch;
+
+const MAX_DB_ERRORS_ALLOWED: u32 = 10;
 
 /// Information about a registered node.
 #[derive(Debug, Clone, FromRow, PartialEq, Eq)]
@@ -16,6 +21,12 @@ pub struct NodeInfo {
     pub node_id: String,
     /// The address (ip:port) where the node can be reached.
     pub address: SocketAddr,
+}
+
+impl NodeInfo {
+    fn str(&self, role: &str) -> String {
+        format!("{role}={}", self.address)
+    }
 }
 
 /// Result of querying cluster node information.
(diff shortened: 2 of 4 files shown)


crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -63,9 +119,95 @@ impl NodeDiscovery {
...
+        }
+
         Ok(ClusterInfo { leader, followers })
     }
 
+    /// Subscribes to PostgreSQL notifications for cluster changes and writes
+    /// cluster info to a file whenever the cluster state changes.
+    ///
+    /// Listens on `nodes_changes` and `leader_changes` channels.
+    pub async fn subscribe_cluster_info_loop(
```
bkolad, line 142: `Nodes` table pruning is out of scope for this PR and will be added in a follow-up. Proxy restart logic will also be handled here in the future.

crates/utils/sov-proxy-utils/Cargo.toml
```diff
@@ -8,10 +8,11 @@ homepage.workspace = true
 publish = false
 repository.workspace = true
 
+
 [dependencies]
 anyhow = { workspace = true }
-sqlx = { version = "0.8", features = ["runtime-tokio", "postgres"] }
-tokio = { workspace = true }
+sqlx = { version = "0.8", features = ["runtime-tokio", "tls-rustls-ring-webpki", "postgres"] }
```
bkolad: We had to add `tls-rustls-ring-webpki` because it’s required by our AWS setup. 
I couldn’t get it working with the `default-feature` disabled, but I’ll try disabling it again. Compilation time looks fine on the proxy server.

bkolad: Ok `default-feature` removed.

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -63,9 +119,95 @@ impl NodeDiscovery {
...
+            .listen_all(["nodes_changes", "leader_changes"])
+            .await?;
+
+        tracing::info!("Subscribed to nodes_changes and leader_changes channels");
+
+        // Fetch initial cluster info.
+        match self.get_cluster_info().await {
+            Ok(info) => {
+                if let Err(error) = write_to_file(path, info.to_file_content()).await {
+                    tracing::warn!(?error, "Failed to update the cluster info file.");
```
citizen-stig: Can we add path here as well for faster debugging.

bkolad: fixed

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -82,6 +224,106 @@ impl NodeDiscovery {
...
         tx.commit().await?;
+
         Ok((leader_id, all_nodes))
     }
 }
+
+async fn write_to_file(path: impl AsRef<std::path::Path>, content: String) -> anyhow::Result<()> {
+    tokio::fs::write(path, content)
+        .await
+        .context("Failed to write cluster info to file")?;
```
citizen-stig: nit: with_context + path might help to easily see which file is being writen

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -63,9 +119,95 @@ impl NodeDiscovery {
...
+                    tracing::warn!(?error, "Failed to update the cluster info file.");
+                } else {
+                    // Notify watchers that the file was saved successfully.
+                    let _ = self.file_saved_sender.send(());
+                }
+            }
+            Err(error) => {
+                tracing::warn!(?error, "Failed to fetch initial cluster info");
+            }
+        }
```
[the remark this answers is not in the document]
citizen-stig: initial fetch will return error immediately, won't it?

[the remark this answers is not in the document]
citizen-stig: No, actually it will be just a warning, good point

crates/utils/sov-proxy-utils/src/lib.rs
bkolad: fixed

crates/utils/sov-proxy-utils/Cargo.toml
```diff
@@ -8,11 +8,13 @@ homepage.workspace = true
 publish = false
 repository.workspace = true
 
+
 [dependencies]
 anyhow = { workspace = true }
-sqlx = { version = "0.8", features = ["runtime-tokio", "postgres"] }
-tokio = { workspace = true }
+sqlx = { version = "0.8", features = ["runtime-tokio", "tls-rustls-ring-webpki", "postgres", "derive"], default-features = false }
```
citizen-stig, line 14: Seems like a lot of intersections with sequencer, maybe worth extracting common parts into root cargo.toml

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -63,9 +119,95 @@ impl NodeDiscovery {
...
+        // Drain any additional pending notifications.
+        while listener.next_buffered().is_some() {}
+
+        let info = self.get_cluster_info().await?;
+        write_to_file(path, info.to_file_content()).await?;
+        // Notify watchers that the file was saved successfully.
+        let _ = self.file_saved_sender.send(());
+        Ok(())
+    }
+
```
citizen-stig, line 211: I have stupid question about `get_cluster_info_from_db`

Why does it need ORDER BY?

And can we add comment that we start transaction to get single view on both tables.

Also, why not join? Tables seems to be small...

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -27,33 +38,78 @@ pub struct ClusterInfo {
...
+    file_saved_sender: watch::Sender<()>,
 }
 
 impl NodeDiscovery {
     /// Creates a new NodeDiscovery with a connection pool.
-    pub async fn new(connection_string: &str) -> Result<Self> {
+    ///
+    /// Returns the NodeDiscovery instance and a receiver that gets notified
+    /// whenever the cluster info file is successfully saved.
+    pub async fn new(connection_string: &str) -> Result<(Self, watch::Receiver<()>)> {
```
citizen-stig, line 72: nit: stylistic choice, but if caller accidentally drops receiver, sender won't work. 

Aleternative approach might be: spawn initial receiver into the background blocked task, return only Self and expose subscribe method.

citizen-stig (approved):

crates/utils/sov-proxy-utils/src/lib.rs
bkolad: The ` ORDER BY` is to make the file more deterministic. So if we ever need to look at 2 versions of the file, it will be easier to compare them. The table will have only a couple of entries, so it won't hurt performance. 
The join also would work, but I found it easier to do validation in rust than SQL

crates/utils/sov-proxy-utils/Cargo.toml
bkolad: I will see the final version of the tool, but I think you are right.

crates/utils/sov-proxy-utils/src/lib.rs
bkolad: In the final binary, it’s slightly easier if there is no spawning. The daemon just blocks on the loop, so I will leave it as is for now

crates/utils/sov-proxy-utils/src/lib.rs
citizen-stig: I just wonder, if starting transactions and doing 2 queries will hold lock over them for longer than doing join.

But probably impact on the rollup nodes is minimal, but just wanted to confirm it.

PR #2407, opened 2026-01-26
Improve error handling during nonce buffer wipes and shutdowns

### LINKED ISSUE #2388: Sequencer nonce queue, improve error messages
There are errors like these:

```
Tx bad nonce for credential id: 0x3df5219504adb60e394468c46ddd275e66f2cf135d58799e17b4356a26e82eee, expected: 2, but found: 2. The sequencer queued the transaction for reordering 0 ms ago, when the user's nonce was 2. It was now dropped from the queue for an unknown reason. This should normally only happen when the sequencer is shutting down

```

### TX eviction edge case
In `nonce_buffer_task.rs::wipe()`, we have, at the time of writing:
```rust
        while let Ok(m) = self.buffer_input.try_recv() {
            match m {
                exec @ NonceBufferInput::TxExecuted { .. } => keep_messages.push(exec),
                NonceBufferInput::NewTx { .. }
                | NonceBufferInput::TxPersisted { .. }
                | NonceBufferInput::TxTimedOut { .. } => (),
            }
        }
```
The above "dropped from the queue for an unknown reason" message happens because the TX is stored as a NewTx message in the mpsc input to the buffer task, and hasn't been processed yet. All buffered TXs get an explicit rejection message but `NewTx` messages in the mpsc are silently dropped.
 * `NewTx` messages should be processed here, returning an explicit rejection just like TXs in the buffer get
 * `TxPersisted` messages should be added to `keep_messages` - we need to return a confirmation, rather than a rejection
   * This means this loop should possibly run before the buffers are cleared; at the time of writing it runs after.
 * `TxTimedOut` is safe to drop: the TX was in the buffer while waiting for timeout, so will get a rejection error alongside every other TX in the buffer

### Error message to users
Even with the above edge case fixed, there are some edge cases where we might drop a TX from the queue despite it having the correct nonce. Showing `expected: 2, but found: 2` looks like a bug to users, so we should special-case the error messages in those cases.

theodorebugnet:
# Description

This PR adds shutdown handling to the sequencer's nonce buffer, so the buffer is explicitly wiped on shutdown and sends rejection error messages to users rather than the unclear `"It was now dropped from the queue for an unknown reason. This should normally only happen when the sequencer is shutting down."` message.

In the process it also improves the logic for error messages returned when the buffer is wiped, by adding handling to the transactions currently stuck in the task's input mpsc queue. This closes #2388

- [ ] ~~I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.~~ Internal change.

## Linked Issues
- Fixes #2388 
- Related to # (issue) 

## Testing

## Docs

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2407">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
@@ -264,17 +264,33 @@ struct TimeoutQueueTask<S: Spec, Rt: Runtime<S>> {
     output: mpsc::Sender<NonceBufferInput<S, Rt>>,
     /// The timeout duration for all transactions.
     timeout_duration: Duration,
+    /// Shutdown notification.
+    shutdown_receiver: watch::Receiver<()>,
 }
 
 impl<S: Spec, Rt: Runtime<S>> TimeoutQueueTask<S, Rt> {
     async fn run(mut self) {
-        while let Some(req) = self.input.recv().await {
+        loop {
+            let req = tokio::select! {
+                _ = self.shutdown_receiver.changed() => {
+                    return;
+                }
+                req = self.input.recv() => req,
+            };
+            let Some(req) = req else {
+                return;
+            };
             let elapsed = req.queued_at.elapsed();
             let remaining = self.timeout_duration.saturating_sub(elapsed);
 
             // Sleep until timeout is due
             if !remaining.is_zero() {
-                tokio::time::sleep(remaining).await;
+                tokio::select! {
+                    _ = self.shutdown_receiver.changed() => {
+                        return;
+                    }
+                    _ = tokio::time::sleep(remaining) => {}
+                }
             }
 
             // Send timeout notification to main task
@@ -810,6 +826,7 @@ impl<E: TxExecutionBackend<S, Rt> + Clone + Send + Sync + 'static, S: Spec, Rt: 
             input: timeout_receiver,
             output: buffer_sender_channel.clone(),
             timeout_duration: Duration::from_millis(future_nonce_transaction_timeout_millis),
+            shutdown_receiver: shutdown_receiver.clone(),
         };
 
         let input_sender = NonceBufferInputSender {
--- a/crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
@@ -1416,6 +1416,98 @@ mod tests {
         (backend, sender, shutdown_sender)
     }
 
+    fn build_task_with_channels(
+        backend: MockTxExecutionBackend,
+    ) -> (
+        NonceBufferTask<MockTxExecutionBackend, TestSpec, TestRuntime>,
+        mpsc::Sender<NonceBufferInput<TestSpec, TestRuntime>>,
+        watch::Sender<()>,
+        watch::Receiver<()>,
+    ) {
+        let (buffer_sender_channel, buffer_input) = mpsc::channel(MAX_BUFFER_INPUT_QUEUE);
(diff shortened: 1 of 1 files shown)


crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -264,17 +264,33 @@ struct TimeoutQueueTask<S: Spec, Rt: Runtime<S>> {
...
     timeout_duration: Duration,
+    /// Shutdown notification.
+    shutdown_receiver: watch::Receiver<()>,
 }
 
 impl<S: Spec, Rt: Runtime<S>> TimeoutQueueTask<S, Rt> {
     async fn run(mut self) {
-        while let Some(req) = self.input.recv().await {
+        loop {
+            let req = tokio::select! {
```
citizen-stig: Nit: we commonly use `FutureOrShutdownOutput` to be less macros'y

citizen-stig: I think we can even collapse it into a single match:

```rust
let fut = future_or_shutdown(self.input.recv(), &self.shutdown_receiver);
let req = match fut.await {
    FutureOrShutdownOutput::Shutdown | FutureOrShutdownOutput::Output(None) => return,
    FutureOrShutdownOutput::Output(Some(r)) => r,
};
```

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -264,17 +264,33 @@ struct TimeoutQueueTask<S: Spec, Rt: Runtime<S>> {
...
+                return;
+            };
             let elapsed = req.queued_at.elapsed();
             let remaining = self.timeout_duration.saturating_sub(elapsed);
 
             // Sleep until timeout is due
             if !remaining.is_zero() {
-                tokio::time::sleep(remaining).await;
+                tokio::select! {
+                    _ = self.shutdown_receiver.changed() => {
```
citizen-stig: The same about using enum

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: I forgot we had that, thanks for the suggestion!

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -751,11 +793,11 @@ impl<E: TxExecutionBackend<S, Rt> + Clone + Send + Sync + 'static, S: Spec, Rt:
                     match forced_tx_batch {
                         Ok(notification) => {
                             tracing::info!(?notification, "Wiping nonce buffer after forced batch execution");
-                            self.wipe().await;
+                            self.drain_and_reject_all_txs(wipe_reject_error::<S, Rt>);
```
preston-evans98, line 796: nit: This isn't an overloaded error. We should probably add a new variant for this

preston-evans98 (approved):

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
theodorebugnet: Good idea but I think this should go into a (small) follow-up PR, as there's no good existing error variant I can see that would fit. So it'd be better to review adding a new tx error as a standalone PR, and the current PR just keeps already existing behaviour for now

PR #2410, opened 2026-01-27
Rename` SequencerRole` and `NodeRole` enums for clarity.

bkolad:
# Description

1. Renamed `SequencerRole` variants to `BatchProducer, PgSyncReplica, and DaOnlyReplica`, updating logic, logs, and tests accordingly.
2. Renamed `NodeRole` to `NodeStartingRole` across configs, schemas, and tests while keeping node_role field and enum string values unchanged in rollup_config.toml.

## Linked Issues
- Related to #1524

### DIFF
--- a/crates/full-node/full-node-configs/src/sequencer.rs
+++ b/crates/full-node/full-node-configs/src/sequencer.rs
@@ -130,7 +130,7 @@ pub enum RecoveryStrategy {
 }
 
 #[derive(Debug, Copy, Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, JsonSchema)]
-pub enum NodeStartingRole {
+pub enum ConfiguredNodeRole {
     /// This node runs as the leader. The leader is responsible for producing new batches.
     Leader,
     /// This node runs as a replica, syncing its state from the leader.
@@ -157,7 +157,7 @@ pub struct PostgresConfig {
     /// Id of the node.
     pub node_id: String,
     #[allow(missing_docs)]
-    pub node_role: NodeStartingRole,
+    pub node_role: ConfiguredNodeRole,
 }
 
 /// Configuration for [`PreferredSequencer`].
--- a/crates/full-node/full-node-configs/src/sequencer.rs
+++ b/crates/full-node/full-node-configs/src/sequencer.rs
@@ -130,7 +130,7 @@ pub enum RecoveryStrategy {
 }
 
 #[derive(Debug, Copy, Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, JsonSchema)]
-pub enum NodeRole {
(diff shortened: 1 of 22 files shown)


crates/module-system/module-schemas/rollup-config.json
```diff
@@ -488,7 +488,7 @@
           "type": "string"
         },
         "node_role": {
-          "$ref": "#/definitions/NodeRole"
+          "$ref": "#/definitions/NodeStartingRole"
```
bkolad: This is the only backward-incompatible change, since the type name appears only in this configuration. 
However, this config is generated, and I couldn’t find anywhere it is actually used.
The `rollup_config.yaml` cares only about enum variant names.

crates/full-node/sov-sequencer/src/preferred/replica/event_receiver.rs
```diff
@@ -68,7 +68,7 @@ impl EventReceiverStartNotifier {
     pub(crate) fn check_replica_status_or_ok_for_leader(
         &self,
     ) -> Result<(), SequencerNotReadyDetails> {
-        if self.seq_role == SequencerRole::Leader {
+        if self.seq_role == SequencerRole::BatchProducer {
```
bkolad, line 71: The `DaOnlyReplica` is not handled here (it was already like that before the renaming). I’ll create an issue and fix it in another PR. This PR should be for renaming only.

crates/full-node/full-node-configs/src/sequencer.rs
```diff
@@ -130,7 +130,7 @@ pub enum RecoveryStrategy {
 }
 
 #[derive(Debug, Copy, Clone, serde::Serialize, serde::Deserialize, Eq, PartialEq, JsonSchema)]
-pub enum NodeRole {
+pub enum NodeStartingRole {
```
theodorebugnet: I don't fully agree with this rename: it sets the actual role for the duration of the node's lifetime, not just its role at startup.

If we want to make the name clearer than just NodeRole, maybe `NodeReplicationRole`?

bkolad: In `PreferredSequencerDb::new,` there is a mapping from `NodeStartingRole` to `SequencerRole`, where `SequencerRole` represents the sequencer’s permanent state. `NodeStartingRole` is read from the configuration (at startup) and is not a one-to-one mapping—for example, `NodeStartingRole::DbElected` can map to either `SequencerRole::BatchProducer` or `SequencerRole::PgSyncReplica`

I’m fine with moving back to NodeRole, but Nicolai suggested renaming it offline, and I think that makes sense.

theodorebugnet (approved):

citizen-stig (approved):

PR #2412, opened 2026-01-27
Add support for gzipping messages in ws subscriptions

preston-evans98:
# Description

This PR adds support for gzipping messages on the events and transactions websockets using the `?compression=gzip` query param. When gzip is enabled...
- Messages are chunked and encoded into a json array with at most 128 messages before being compressed and sent
- Errors json encoded, gzipped, and sent immediately

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2412">
  <picture>
    Open with Devin
  </picture>
</a>


crates/utils/sov-rest-utils/src/lib.rs
```diff
@@ -272,31 +312,67 @@ pub async fn serve_generic_ws_subscription<S, M, E>(
...
+                                            }
+                                            batch.clear();
+                                        }
+
+                                        // Send compressed error
+                                        if send_compressed_bytes(&mut socket, err.to_json().as_bytes()).await.is_err() {
+                                            break 'outer;
+                                        }
+
+                                        if !err.is_recoverable() {
```
bkolad, line 343: `send_compressed_bytes` calls `socket.feed(..)`, so this error might not be sent in some cases (in the uncompressed branch, we call `socket.send`) is that ok? Let's rename `send_compressed_bytes` to `feed_compressed_bytes`?

crates/utils/sov-rest-utils/src/lib.rs
bkolad: Here we send an uncompressed error even if ` config.compress == true`

bkolad (approved):
Approving, provided my comments are addressed.

crates/utils/sov-rest-utils/src/lib.rs
preston-evans98: It will get sent by the `close` in the outer loop. Added a comment to clarify

crates/utils/sov-rest-utils/src/lib.rs
preston-evans98: Good catch. Added handling on that path

crates/utils/sov-rest-utils/src/lib.rs
preston-evans98: (And also renamed the function)

Issue #2263, opened 2025-12-22
Failover Design


bkolad:
# Sequencer Role:

In the sequencer, the following components depend on the Sequencer role:

- `BlobSender`
- `SeqDB`
- `CacheWarmUp`
- `TimeOracle`
- `Inner/sync_state`
- `ReplicaSyncTask`

## Goal

The goal of this issue is to define the steps required to transition the Sequencer from the **Replica** role to the **Leader** role.

### 1. Atomicity
The transition must be atomic from the perspective of the external world. The Sequencer must never appear to be in a partially transitioned state—it should be either fully in the Replica role or fully in the Leader role.

### 2. Ordering
A strict ordering for switching components must be established. For example, switching `Inner/sync_state` to `Leader` mode before switching `BlobSender` would result in an invalid sequencer state. During the transition, the replica could start producing empty batches while `BlobSender` is still inactive and unable to send them, pushing the sequencer into a bad state.

### 3. Lifecycle of async tasks
Many of the components above (e.g., `BlobSender`, `CacheWarmUp`, `TimeOracle`) spawn asynchronous actors that are registered when the sequencer starts. Replicas do not start these actors. However, the system currently does not support starting these tasks after the sequencer has already been initialized.

If 2 and 3 are well designed, atomicity should follow automatically.

### 2. Ordering should be enforced by a dedicated component, ReplicaToLeaderTransitionManager.

During the **Replica → Leader** transition, the following sequence should occur:

1. `ReplicaToLeaderTransitionManager` receives a database notification indicating that the node has become the new `Leader`.
2. It sends a shutdown request to `ReplicaSyncTask` and waits for confirmation that the task has fully stopped.
3. It sends a request to `Inner/sync_state` to transition to Leader mode.
4. `Inner/sync_state`:
   - Sends a request via `SideEffectsTask` instructing `BlobSender` and `SeqDB` to upgrade to the Leader role.
   - Updates its local role flag to `Leader`.
   - To support this, we will introduce a new `ExecutorEvent::UpgradeToLeader` message in `SideEffectsTask`.

5. `ReplicaToLeaderTransitionManager` must provide a mechanism to wait until step 4 has fully completed.
6. `ReplicaToLeaderTransitionManager` sends a request to `CacheWarmUp`, notifying it to start.
7. `ReplicaToLeaderTransitionManager` sends a request to `TimeOracle`, notifying it to start.

PR #2414, opened 2026-01-28
 Add `heartbeat` task for node registration and leadership management.

bkolad:
# Description
Replace the old `leadership‑election.rs` task with a new `HeartBeatTask` that handles leader heartbeats, DbElected replica elections, and non‑electing replica registration in one place.
                                                                                                                                                                                                                                                               
 Changes                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                               
1. New `HeartBeatTask`:                                                                                        
    - Leaders: maintain leadership, shutdown if lost                                                                                                                                                                                                           
    - `DbElected` replicas: compete for leadership, restart as leader if acquired                                                                                                                                                                                
    - Static replicas: register only, never compete for leadership     
    -                                                                                                                                                                                         
2. Cluster info improvements:                                                                                                                                                                                                                                 
    - File format now includes node_id: `role=address,timestamp,node_id`                                                                                                                                                                                    

## Linked Issues
- Related to https://github.com/Sovereign-Labs/sovereign-sdk/issues/1524


crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
```diff
@@ -0,0 +1,207 @@
...
+        let configures_node_role = self.postgres_config.node_role;
+        match seq_role {
+            SequencerRole::BatchProducer => self.spawn_leader_heartbeat_task(),
+            SequencerRole::PgSyncReplica => {
+                if configures_node_role == ConfiguredNodeRole::DbElected {
+                    self.spawn_replica_heartbeat_task()
+                } else {
+                    assert_eq!(configures_node_role, ConfiguredNodeRole::Replica);
+                    self.spawn_node_registration_task()
+                }
```
bkolad, line 70: Here we state that when the Sequencer role is SequencerRole::PgSyncReplica, the node will participate in leader election only if the configured node role is ConfiguredNodeRole::DbElected.

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
```diff
@@ -0,0 +1,207 @@
...
+            }
+        })
+    }
+
+    // Spawns a task that only maintains node registration.
+    //
+    // Periodically updates the node's entry in the `nodes` table without
+    // competing for leadership. Failures are logged but don't cause shutdown.
+    fn spawn_node_registration_task(self) -> JoinHandle<()> {
```
bkolad, line 185: This is the new method. 
The `spawn_leader_heartbeat_task and spawn_replica_heartbeat_task` methods already existed in the codebase, but were previously located in leadership_election.rs.

examples/demo-rollup/tests/replica/db_elected.rs
```diff
@@ -215,106 +150,71 @@ async fn test_db_elected_leader_failover() {
...
 
     let _ = restarted_rollup.shutdown().await;
-    let _ = da_shutdown.send(());
+    let _ = setup.shutdown().await;
 }
 
 /// Test that `subscribe_cluster_info_loop` receives PostgreSQL notifications
 /// and writes updated cluster info to file when nodes register and leadership changes.
 #[tokio::test(flavor = "multi_thread")]
 async fn test_subscribe_cluster_info_receives_notifications() {
```
bkolad, line 159: The tests are mostly unchanged (with a few assertions added for the new functionality), but they now use common setup logic in NodeDiscoveryTestSetup.

crates/full-node/sov-sequencer/src/preferred/db/postgres/tests/mod.rs
```diff
@@ -1,8 +1,18 @@
 mod db_operations;
 mod leader_election;
 
+use std::net::SocketAddr;
+
 use super::*;
 use sov_full_node_configs::sequencer::ConfiguredNodeRole;
+
+pub(super) fn test_bind_addr(node_id: &str) -> SocketAddr {
```
citizen-stig: What is the point of this? To have deterministic port? But this might fail if several tests run in parallel.

Why not use port "0" and get actual port after it is bound?

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
```diff
@@ -0,0 +1,207 @@
...
+    fn spawn_leader_heartbeat_task(self) -> JoinHandle<()> {
+        tokio::spawn(async move {
+            info!(node_id = %self.node_id, address = %self.backend.node_address, "Starting leader heartbeat task");
+
+            let mut interval = tokio::time::interval(HEARTBEAT_INTERVAL);
+
+            loop {
+                interval.tick().await;
+
+                if self.shutdown_receiver.has_changed().unwrap_or(true) {
```
citizen-stig: Why we don't use tokio::select here. This means that shutdown will be delayed by interval. Also we can use FutureOrShutdown

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
```diff
@@ -0,0 +1,207 @@
...
+                    Ok(true) => {
+                        // Successfully refreshed leadership and node registration
+                        tracing::trace!("Leadership heartbeat successful.");
+                    }
+                    Ok(false) => {
+                        error!(
+                            node_id = %self.node_id,
+                            "Leadership lost! Another node has taken over. Initiating graceful shutdown."
+                        );
+                        exit_rollup(&self.shutdown_sender).await;
```
citizen-stig: And with graceful shutdown, we just expect it to be restarted by systemd, and then it is going be regular replica, right?

What are major blockers to switch to replica mode without full restart?

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
```diff
@@ -0,0 +1,207 @@
...
+                    return;
+                }
+
+                match self.register_node().await {
+                    Ok(_) => {}
+                    Err(e) => {
+                        warn!(
+                            node_id = %self.node_id,
+                            error = ?e,
+                            "Node registration attempt failed, will retry."
```
citizen-stig: So there's no limit in attempts?

crates/full-node/sov-sequencer/src/preferred/db/postgres/tests/mod.rs
bkolad: fixed

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
bkolad: fixed

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
bkolad: We could exit the node, but systemd would just restart it. I think it’s fine to leave it as is unless you have any counterarguments.

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
bkolad: In the sequencer constructor, we spawn all tasks that differ between a replica and a leader (blob sender, cache warmer etc.. ). On restart, the sequencer constructor runs, so if a replica restarts as a leader, all the appropriate tasks are initialized correctly.

If we wanted to handle this in-process, we would need to move the task-initialization logic outside of the sequencer constructor, which is tricky. We consider this as a follow-up if the current approach turns out to be insufficient.

citizen-stig: That sounds good as follow up. I think it would be good to summarize it in the tracking task while our memory of this is fresh. 

I agree that using simple approach with systemd restart is good and pragmatic now, but we might want to get back to it eventually

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
```diff
@@ -66,39 +66,36 @@ macro_rules! run_with_retries {
...
-        let node_address = node_address(bind_addr)?;
-        let backend = Self::connect_internal(config, node_address).await?;
-        backend.upsert_node_registration_max_timestamp().await?;
+        let backend = Self::connect(config, bind_addr).await?;
+        let _maybe_leader = backend.heartbeat(None).await?;
         Ok(backend)
     }
 
-    async fn connect_internal(config: &PostgresConfig, node_address: String) -> Result<Self> {
+    /// // Connects to Postgres db.
```
citizen-stig:
```suggestion
    /// Connects to Postgres db.
```

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
```diff
@@ -66,39 +66,36 @@ macro_rules! run_with_retries {
...
+    pub async fn connect(config: &PostgresConfig, bind_addr: SocketAddr) -> Result<Self> {
```
citizen-stig, line 96: How connect is different from other `connect_` methods? Should we document that?

citizen-stig: Also, do we need all 3 of them:

* `connect_as_maybe_leader`
* `connect_as_replica`
* `connect`

Wouldn't first 2 be enough?

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
```diff
@@ -0,0 +1,210 @@
...
+                            Ok(true) => {
+                                // Successfully refreshed leadership and node registration
+                                tracing::trace!("Leadership heartbeat successful.");
+                            }
+                            Ok(false) => {
+                                error!(
+                                    node_id = %self.node_id,
+                                    "Leadership lost! Another node has taken over. Initiating graceful shutdown."
+                                );
+                                exit_rollup(&self.shutdown_sender).await;
```
citizen-stig, line 117: Stupid question, but I am not sure, that systemd will restart service if exited with exit code 0. 

It might just remain stopped. Have you checked that?

citizen-stig (approved):
Looks good all remaining comments are non-blocking, but would be great to address them if we have capacity

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
bkolad: We already have it as a sub-issue for #1524
https://github.com/Sovereign-Labs/sovereign-sdk/issues/2263

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
bkolad: The first two methods re-use the third one. But I could rearrange the code to make it private.  Will leave to as a follow-up.

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
bkolad: Discussed offline, this will be tested during test deployments.

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
citizen-stig: Sounds good, please make sure it won't fall through the cracks.

PR #2415, opened 2026-01-29
Fix `s/finalization/finalization_blocks` in MockDaConfig in examples

citizen-stig:
# Description

And other outdated comments.

## Testing

## Docs
No updates

### DIFF
--- a/crates/adapters/mock-da/src/config.rs
+++ b/crates/adapters/mock-da/src/config.rs
@@ -325,6 +325,7 @@ pub fn seed_for_test(small_seed: u8) -> HexHash {
 
 /// The configuration for Mock Da.
 #[derive(Debug, Clone, serde::Deserialize, serde::Serialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct MockDaConfig {
     /// Connection string to the database for storing Da Data.
     ///   - "sqlite://demo_data/da.sqlite?mode=rwc"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2026-01-29
+- #2415 Non-breaking, but **important**. Config for MockDa rollup now stricter and prevents unknown fields. Previously there was incorrect `finalization` field instead of `finalization_blocks`
+
 # 2026-01-17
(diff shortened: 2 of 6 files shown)


crates/adapters/mock-da/src/config.rs
```diff
@@ -325,6 +325,7 @@ pub fn seed_for_test(small_seed: u8) -> HexHash {
 
 /// The configuration for Mock Da.
 #[derive(Debug, Clone, serde::Deserialize, serde::Serialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
```
bkolad, line 328: Devin found an issue here with the `finalization_blocks` field.
The problem is on rollup_config.toml and mock_rollup_config.toml.

```
Config file with incorrect field name will fail to parse after deny_unknown_fields is added

The PR adds #[serde(deny_unknown_fields)] to MockDaConfig to prevent misconfiguration, but there's still a config file that uses the old incorrect field name finalization instead of finalization_blocks.

The Problem

The file still uses the incorrect field:

[da]connection_string = "sqlite://mock_da.sqlite?mode=rwc"sender_address = "0000000000000000000000000000000000000000000000000000000000000000"finalization = 10  # <-- This is wrong, should be finalization_blocks

Actual vs Expected

    Before this PR: The config would silently ignore the finalization field (it would use the default value of 0 for finalization_blocks).
    After this PR: The config will fail to parse with a serde error because finalization is an unknown field.

Impact

The soak testing example will fail at startup when trying to parse the rollup config, preventing any soak tests from running.

```

citizen-stig: Which `mock_rollup_config.toml` does it mean?

There's literally this change:

https://github.com/Sovereign-Labs/sovereign-sdk/pull/2415/changes#diff-72169386a933d7c8b3e725f5fb49c5c848d9cec9bc79171525038f3f393392c9R11

There was one `rollup_config.toml` from soak test, which I just updated in 3b05394de1a76ee139e0343278d497eb528c82b6

bkolad: Oh I greped for this on my local branch that's why I found it in `mock_rollup_config.toml` all good.

bkolad (approved):

PR #2422, opened 2026-02-01
Add `LeaderElectionConfig` to full-node config 

bkolad:
This PR depends on https://github.com/Sovereign-Labs/sovereign-sdk/pull/2414
# Description

1. Add `LeaderElectionConfig` to full-node config — allows specifying both leader timeout and grace period for elections.
2. Update heartbeat & Postgres backend to use the new config values instead of fixed constants.
3. Refactor `try_update_leader_inner` to take into account `grace_period` (This ). This prevents rapid leader flapping when a replica restarts to become a leader, and in the meantime, another replica registers as the leader.

No breaking changes to existing APIs, and configuration defaults preserve current behavior.

## Linked Issues
- Related to #1524


crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
```diff
@@ -19,11 +19,11 @@ use tracing::{error, info, warn};
 
 use super::SequencerRole;
 
-use super::postgres::{PostgresBackend, LEADER_TIMEOUT};
+use super::postgres::PostgresBackend;
 use crate::preferred::exit_rollup;
 
 /// Periodic interval for heartbeat tasks.
-const HEARTBEAT_INTERVAL: Duration = Duration::from_millis(200);
+const HEARTBEAT_INTERVAL: Duration = Duration::from_millis(100);
```
citizen-stig: Why this has changed? Should it be also in the config?

citizen-stig (approved):

crates/full-node/sov-sequencer/src/preferred/db/heartbeat_task.rs
bkolad: The `leader_timeout_millis` is set to 500ms; with this change, we allow 5 heartbeat misses before triggering leader election. With 200ms, we would allow only 2.5.

citizen-stig: but if user sets 100ms timeout, heartbit might not be sent and user has no control over it.

Should it be also exposed as configuration?

bkolad: fixed

PR #2425, opened 2026-02-02
Add `max_age` filtering to NodeDiscovery and simplify PostgresBackend API.

bkolad:
# Description

1. Simplifies the Postgres connection API by removing `connect_as_replica()` and `connect_as_maybe_leader() `helper methods. Callers now use `connect()` directly and call heartbeat() separately when needed. 
2. Adds stale node filtering to `NodeDiscovery` so that nodes whose `last_updated` timestamp exceeds a configurable max_age are filtered out of cluster info (except the leader, who is always included).
                                                                                                                                                        

## Linked Issues
- Related to #1524

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
@@ -444,30 +444,25 @@ impl PreferredSequencerDb {
     ) -> anyhow::Result<(Self, SequencerRole)> {
         let (backend, role): (Option<Box<dyn DbBackend>>, _) = {
             if let Some(postgres_config) = &postgres_config {
+                let backend = PostgresBackend::connect(postgres_config, bind_addr).await?;
+
                 match postgres_config.node_role {
                     ConfiguredNodeRole::ReplicaNoLeaderSync => {
-                        // Connect and register the node without attempting leader election.
-                        // The backend is dropped after registration since replicas don't need it.
-                        PostgresBackend::connect_as_replica(postgres_config, bind_addr).await?;
+                        PostgresBackend::connect(postgres_config, bind_addr).await?;
                         (None, SequencerRole::DaOnlyReplica)
                     }
                     ConfiguredNodeRole::Replica => {
-                        // Connect and register the node without attempting leader election.
-                        // The backend is dropped after registration since replicas don't need it.
-                        PostgresBackend::connect_as_replica(postgres_config, bind_addr).await?;
+                        PostgresBackend::connect(postgres_config, bind_addr).await?;
                         (None, SequencerRole::PgSyncReplica)
                     }
                     ConfiguredNodeRole::Leader => {
-                        let (backend, _) =
-                            PostgresBackend::connect_as_maybe_leader(postgres_config, bind_addr)
-                                .await?;
+                        let backend = PostgresBackend::connect(postgres_config, bind_addr).await?;
                         (Some(Box::new(backend)), SequencerRole::BatchProducer)
                     }
                     ConfiguredNodeRole::DbElected => {
-                        // Connect and attempt to acquire leadership
-                        let (backend, maybe_leader) =
-                            PostgresBackend::connect_as_maybe_leader(postgres_config, bind_addr)
-                                .await?;
+                        let maybe_leader = backend
+                            .heartbeat(Some(postgres_config.leader_election))
+                            .await?;
 
                         let is_leader = maybe_leader
                             .map(|leader| leader.node_id == postgres_config.node_id)
--- a/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
@@ -444,22 +444,19 @@ impl PreferredSequencerDb {
     ) -> anyhow::Result<(Self, SequencerRole)> {
         let (backend, role): (Option<Box<dyn DbBackend>>, _) = {
             if let Some(postgres_config) = &postgres_config {
-                let backend = PostgresBackend::connect(postgres_config, bind_addr).await?;
-
                 match postgres_config.node_role {
-                    ConfiguredNodeRole::ReplicaNoLeaderSync => {
-                        PostgresBackend::connect(postgres_config, bind_addr).await?;
-                        (None, SequencerRole::DaOnlyReplica)
-                    }
-                    ConfiguredNodeRole::Replica => {
-                        PostgresBackend::connect(postgres_config, bind_addr).await?;
-                        (None, SequencerRole::PgSyncReplica)
-                    }
+                    ConfiguredNodeRole::ReplicaNoLeaderSync => (None, SequencerRole::DaOnlyReplica),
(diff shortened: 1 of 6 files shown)


crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -445,29 +445,21 @@ impl PreferredSequencerDb {
         let (backend, role): (Option<Box<dyn DbBackend>>, _) = {
             if let Some(postgres_config) = &postgres_config {
                 match postgres_config.node_role {
-                    ConfiguredNodeRole::ReplicaNoLeaderSync => {
```
bkolad, line 448: This is a follow-up to my pervious pr.

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -268,21 +287,30 @@ impl NodeDiscovery {
...
+        let max_age_secs = self.max_age.as_secs() as i64;
+
+        // Fetch nodes updated within max_age, always including the leader regardless of age.
+        // The leader_id is included in each row via LEFT JOIN, allowing us to get it from the results.
+        let rows: Vec<(String, String, OffsetDateTime, Option<String>)> = sqlx::query_as(
+            "SELECT n.node_id, n.address, n.last_updated, l.node_id as leader_id \
+             FROM nodes n \
+             LEFT JOIN sequencer_leader l ON l.singleton = 1 \
+             WHERE n.last_updated > NOW() - $1 * INTERVAL '1 second' \
+                OR n.node_id = l.node_id \
```
bkolad, line 299: We always return a leader node, even if it exceeds the maximum age, because we can still attempt to send transactions to the last known leader on a best-effort basis.

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -445,29 +445,21 @@ impl PreferredSequencerDb {
...
-                        (None, SequencerRole::PgSyncReplica)
-                    }
+                    ConfiguredNodeRole::ReplicaNoLeaderSync => (None, SequencerRole::DaOnlyReplica),
+                    ConfiguredNodeRole::Replica => (None, SequencerRole::PgSyncReplica),
                     ConfiguredNodeRole::Leader => {
-                        let (backend, _) =
-                            PostgresBackend::connect_as_maybe_leader(postgres_config, bind_addr)
-                                .await?;
+                        let backend = PostgresBackend::connect(postgres_config, bind_addr).await?;
+                        let _ = backend
```
citizen-stig, line 452: Is this join handle to the background task?

I know this was this way before, but restart tests might be flaky, because we don't wait for shutdown of this task

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -268,21 +287,30 @@ impl NodeDiscovery {
...
-                .fetch_all(&mut *tx)
-                .await?;
-
-        tx.commit().await?;
+        let max_age_secs = self.max_age.as_secs() as i64;
+
+        // Fetch nodes updated within max_age, always including the leader regardless of age.
+        // The leader_id is included in each row via LEFT JOIN, allowing us to get it from the results.
+        let rows: Vec<(String, String, OffsetDateTime, Option<String>)> = sqlx::query_as(
+            "SELECT n.node_id, n.address, n.last_updated, l.node_id as leader_id \
```
citizen-stig, line 295: How big we expect table nodes to be? 

Could it be that fetch time will grow above pull interval and will be hammering db?

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -268,21 +287,30 @@ impl NodeDiscovery {
...
+             FROM nodes n \
+             LEFT JOIN sequencer_leader l ON l.singleton = 1 \
```
citizen-stig, line 297: Why does it need to be a left join and not separate query?

I believe 2 simple queries is faster than 1 join. Especially if we call them concurrently with tokio::join

citizen-stig (changes requested):
Sequencer simplifications are really good, but proxy with join query actually looks more complex and might have performance penalty

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
bkolad: No—it sends heartbeat(Some(..)), which attempts to acquire leadership on startup. This is a one-time action. The long running taksas are in `heartbeat_task.rs`

crates/utils/sov-proxy-utils/src/lib.rs
bkolad: The table size equals the number of nodes with unique IDs in the rollup. Since spawning a node with a new ID is a rare operation, the table size won’t grow quickly. For example, if we have a system with 5 nodes and we don’t introduce any new nodes with new IDs, the table size will always remain 5. On top of that, we only select nodes that are younger than max_age, which I plan to set to 1s.

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
citizen-stig: Oh, so '_' is indication if leadership has been aqcuired or not, right?

If so, the its fine

crates/utils/sov-proxy-utils/src/lib.rs
citizen-stig: And their timestamp is "update" or insert operation? Just to confirm, at some point I thought that timestamp are inserted.

> On top of that, we only select nodes that are younger than max_age, which I plan to set to 1s.

The problem is not how much data we fetch over the network, but how much work we give postgresql to do.

crates/utils/sov-proxy-utils/src/lib.rs
bkolad: The query returns only a few rows from very small tables (size = number of nodes unique seen in the cluster), and the join is against a table with a single row, so query performance should not be an issue. Additionally, we only make a single network request.

For me, the more important factor was simplicity;

Specifically:

1. We need to return both the leader and its followers.
2. The `max_age` constraint should apply only to followers, not to the leader. This ensures that even if the database stops receiving heartbeats from leaders or followers, we still return the last known leader and attempt to connect on a best-effort basis.

This can all be handled in a single query. Doing the same in Rust would require two separate queries plus additional reconciliation logic at the Rust layer. And the join does not look complicated.

citizen-stig: Discussed on the call

citizen-stig (approved):

crates/utils/sov-proxy-utils/src/lib.rs
bkolad: Output:
1. The table will be small
2. We do the join on a table that hash only one row
3. We have only one networ request.

Given this we agreed to leave the join. In subsequent PRs I will add metrics to monitor the performance of this select.

crates/utils/sov-proxy-utils/src/lib.rs
bkolad: This is an upsert operation. So we replace rows with the same node id.

PR #2430, opened 2026-02-02
Add rollup-height endpoint

preston-evans98:
# Description

This tiny PR adds a new endpoint to fetch the latest rollup-height from the sequencer

- [ ] I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.
- [ ] I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. (Are all new dependencies necessary? Is any module dependency leaked into the full-node (hint: it shouldn't)?)

## Linked Issues
- Fixes # (issue, if applicable)
- Related to # (issue) 

## Testing
Describe how these changes were tested. If you've added new features, have you added unit tests?

## Docs
Describe where this code is documented. If it changes a documented interface, have the docs been updated?

Rendered docs are available at `https://sovlabs-ci-rustdoc-artifacts.us-east-1.amazonaws.com/<BRANCH_NAME>/index.html`

PR #2418, opened 2026-01-29
Feature: Allow multiple chain hashes for easier upgrades

preston-evans98:
# Description

This PR allows developers to specify chain hash overrides in constants.toml. During the specified block range, the provided hash will be used instead of the "actual" hash compiled into the rollup binary. For `grace_period` additional blocks after `end_height`, both the provided `chain_hash` and the next hash will be accepted. 
```
CHAIN_HASH_OVERRIDES = [{ start_height = 0, end_height = 1000, chain_hash = "0xabc...32_bytes_hex", grace_period = 100 }]
```

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2418">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #2431, opened 2026-02-02
Add script to programmatically update chain hash overrides

preston-evans98:
# Description

Adds a typescript cli to auto-update the chain hash override list. The body of the script works well, but we should settle on organization (does it currently live in the right place?) before merging this PR.  Depends on #2418  and #2430 

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


ross-weir:
Regarding organization, what do you think about:

- We create a new `packages/cli` package called `@sovereign-sdk/cli`
- Add this script as a subcommand (something like `cli upgrade add-chainhash-override`)
  - I think it makes sense to have it under a general `upgrade` command in case we need more upgrade related ops
- Then it could be ran like `npx @sovereign-sdk/cli upgrade add-chainhash-override`

Can use `apps/indexer` as an example of how to distribute the cli (currently users would need to clone this repo and npm run the script)

PR #2441, opened 2026-02-04
`NodeDiscovery`: Update the cluster_info file only when cluster membership changes.

bkolad:
# Description

This PR optimizes cluster state change detection in `NodeDiscovery`. Previously, every database notification (including heartbeat updates) triggered a file write and notified watchers. Now, the system tracks the previous state and only acts when membership or leadership actually changes.

## Linked Issues
- Related to #1524 
- https://github.com/Sovereign-Labs/rollup-starter/pull/121
- https://github.com/Sovereign-Labs/sov-rollup-cdk-starter/pull/69
- https://github.com/Sovereign-Labs/ubuntu-evm-starter-script/pull/28

### DIFF
--- a/crates/utils/sov-proxy-utils/src/lib.rs
+++ b/crates/utils/sov-proxy-utils/src/lib.rs
@@ -271,24 +271,12 @@ impl NodeDiscovery {
             .listen_all(["nodes_changes", "leader_changes"])
             .await?;
 
-        // Fetch initial cluster info.
-        match self.get_cluster_info().await {
-            Ok(info) => {
-                if let Err(error) = write_to_file_atomically(path, info.to_file_content()).await {
-                    tracing::warn!(?error, ?path, "Failed to update the cluster info file.");
-                } else {
-                    // Notify watchers that the file was saved successfully.
-                    self.notifier.on_cluster_update(&info).await;
-                }
-            }
-            Err(error) => {
-                tracing::warn!(?error, "Failed to fetch initial cluster info");
-            }
-        }
-
         tracing::info!("Subscribed to nodes_changes and leader_changes channels");
 
         let mut consecutive_errors: u32 = 0;
+
+        // On startup, write an empty file. If the cluster is not empty, the file will be populated on the first call to `handle_cluster_update`.
+        write_to_file_atomically(path, "".to_string()).await?;
 
         loop {
             match self.handle_cluster_update(&mut listener, path).await {
@@ -316,12 +304,6 @@ impl NodeDiscovery {
         listener: &mut PgListener,
         path: &std::path::Path,
     ) -> anyhow::Result<()> {
-        // Wait for at least one notification.
-        listener.recv().await?;
-
-        // Drain any additional pending notifications.
-        while listener.next_buffered().is_some() {}
-
         let info = self.get_cluster_info().await?;
         let leader_id = info.leader_id();
 
@@ -355,6 +337,12 @@ impl NodeDiscovery {
         }
 
         tracing::debug!(info = ?info, "Last cluster info");
+
+        // Wait for at least one notification.
+        listener.recv().await?;
+
+        // Drain any additional pending notifications.
+        while listener.next_buffered().is_some() {}
 
         Ok(())
     }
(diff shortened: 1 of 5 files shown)


crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -111,51 +125,78 @@ impl NodeInfo {
...
+impl SimpleClusterUpdateNotifier {
+    /// Creates a new notifier and its corresponding receiver.
+    pub fn new() -> (Self, watch::Receiver<()>) {
+        let (sender, receiver) = watch::channel(());
+        (Self { sender }, receiver)
+    }
+}
+
+#[async_trait]
+impl ClusterUpdateNotifier for SimpleClusterUpdateNotifier {
```
bkolad, line 156: This is a useful default implementation of `ClusterUpdateNotifier`. In `sov-rollup-starter`, we will also provide an implementation that reloads OpenResty by invoking:
```
Command::new("systemctl")
    .arg("reload")
    .arg("openresty")
```

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -316,10 +379,34 @@ impl NodeDiscovery {
...
+    let temp_path = dir.join(format!(".{}.tmp", std::process::id()));
+
+    // Write content to temp file.
+    let mut file = tokio::fs::File::create(&temp_path)
+        .await
+        .with_context(|| format!("Failed to create temp file at {temp_path:?}"))?;
+
+    file.write_all(content.as_bytes())
+        .await
+        .with_context(|| format!("Failed to write to temp file at {temp_path:?}"))?;
```
bkolad, line 402: If the .tmp file is corrupted, it will be overwritten on the next DB notification.

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -111,51 +125,78 @@ impl NodeInfo {
...
+    }
+}
+
 /// Client for querying cluster information from the database.
 pub struct NodeDiscovery {
     max_age: Duration,
     pool: PgPool,
     connection_string: String,
-    file_saved_sender: watch::Sender<()>,
+    prev_members: BTreeSet<String>,
```
citizen-stig, line 167: If members are written with prefix, then we don't need to have separate condition for prev leader. What do you think?

crates/utils/sov-proxy-utils/src/lib.rs
```diff
@@ -316,10 +379,34 @@ impl NodeDiscovery {
...
-    tokio::fs::write(path, content)
+/// Atomically writes content to a file.
+///
+/// Uses write-to-temp-then-rename pattern to ensure the file is never
+/// partially written. The data is synced to disk before renaming.
+async fn write_to_file_atomically(path: &Path, content: String) -> anyhow::Result<()> {
+    let dir = path.parent().context("Path has no parent directory")?;
+
+    // Create temp file in same directory to ensure same filesystem for atomic rename.
+    let temp_path = dir.join(format!(".{}.tmp", std::process::id()));
```
citizen-stig: So there's a valid comment from devin, about tempfile remaining if function fails.

Why not just use `tempfile` crate? This way tempfile will be removed anyway, and even if process crashes, it will be in the tempfolder without impact on the logic

citizen-stig (approved):

crates/utils/sov-proxy-utils/src/lib.rs
bkolad: Would we write it with a prefix to the file as well? Or would it be only for members?

crates/utils/sov-proxy-utils/src/lib.rs
bkolad: I addressed it below:
https://github.com/Sovereign-Labs/sovereign-sdk/pull/2441/changes#r2767974961
The _tmp file will be overridden, but now I think we don't need the  std::process::id() in the name, as the node-discovery runs in a single process.  I will remove that part.

crates/utils/sov-proxy-utils/src/lib.rs
citizen-stig: I see, cluster info works with it.

I thought about it differently:

* BTreeSet only used for checking if something has changed.
* Output of get_cluster_info is used for writing the file, always deterministic.

PR #2443, opened 2026-02-05
Proof manager separate db

### LINKED ISSUE #1945: Add Separate database for ProofManager
# Current 

Currently ProofManager uses LedgerDb to store its data:

Image

It has several issues

1. The Last submitted height is saved only after the next iteration of the runner loop, as data is saved only at the end. Proofs can be submitted twice, if stop happened at modulo proof jump.
2. Individual block proofs are stored only in memory and lost during restart.
3. Tests got broken when finalization order changed to state update: https://github.com/Sovereign-Labs/sovereign-sdk/pull/1894

# Suggested update

With separate database:

Image

This issues will be addressed, at the cost of supporting separate database instnace

citizen-stig:
## What this PR does

Extracts the proof-manager's state out of `LedgerDb` into a dedicated `ProofManagerDb` (separate
RocksDB), and — following review — reworks it to persist **only the re-provable
`StateTransitionInfo` queue**, keyed by `(slot, DA block hash)`. All cursor metadata is in-memory
and recomputed on startup. This decouples proof-manager persistence from the ledger commit loop and
minimizes the cross-DB invariants that must hold after a crash.

> This supersedes the original approach in this PR (a separate DB that *persisted*
> `next_height_to_receive` / `write_height` / `oldest` to prevent duplicate proof submissions). See
> "Design change" below.

## Why

The node always runs with `--start-fresh-outer-proof-on-resync` (the network prover can't catch up
on unproven blocks after a restart), so **re-proving the latest STF info after a restart is
acceptable** — a fresh outer proof replaces the previous one. Once re-proving is acceptable,
persisting the receive cursor buys nothing; removing it (and the other persisted cursors) removes
the reconciliation/validation machinery and the cross-DB inconsistency windows it guarded. Fewer
persisted values means fewer invariants to assert on startup and fewer ways the two DBs can diverge
after a crash. (Re #1945: duplicate submission is now tolerated-and-replaced rather than prevented
via persistence.)

## Design change vs the original PR

| | Original | Reworked (this PR) |
|---|---|---|
| Persisted in ProofManagerDb | StfInfo queue **+** `write_height`, `next_height_to_receive`, `oldest` | **only** the StfInfo queue |
| `next_height_to_receive` | persisted immediately after each DA post | in-memory; recomputed as `latest_proof_final_slot + 1` on startup |
| finalized-visible cutoff (`write_height`) | persisted | in-memory; recomputed by `recompute_visible_state` |
| Startup | `validate_and_recover_write_height` reconciles 3 persisted values | recompute the cutoff from the ledger view + stored key range |
| StfInfo key | `SlotNumber` | `(SlotNumber, DA block hash)` (forks distinct) |

Net diff: **+775 / −1299** — the rework removes more than it adds.

## Key changes

- **`ProofManagerDb`** (`sov-db/src/proof_manager_db.rs`): persists one CF,
  `StfInfoByNumber: (SlotNumber, DbHash) => StoredStfInfo`. No metadata table. Adds
  `recompute_visible_state` (recomputes the finalized-visible cutoff from the ledger view and drops
  staged rows above `ledger_head`), seek-first/last helpers, and a height-prefix `prune` that sweeps
  orphan forks.
- **In-memory cursors** (`stf_info_manager.rs`): `next_height_to_receive` and the finalized-visible
  cutoff are in-memory, recomputed on startup, advanced during normal operation, never persisted.
- **Fork keying**: forks at the same slot coexist as distinct rows. The consumer reads the
  **finalized** fork using the canonical DA hash, resolved on the **producer** side
  (`LedgerDb::da_block_hash_for_slot` → `SlotByNumber[slot].hash`) and carried on the notification
  channel `(SlotNumber, DbHash)` — so the receiver stays `ProofManagerDb`-only. Pruning sweeps
  orphans below the cutoff via a height-prefix range delete.
- **Resync-flag gate**: `new_stf_info_channel_for_runner` errors at startup if resuming from an
  existing outer proof without `--start-fresh-outer-proof-on-resync` (the in-memory cursor is only
  crash-safe with it on).
- **Removed**: the immediate-cursor-persistence path, `validate_and_recover_write_height`, and the
  associated crash-injection hook.
- **LedgerDb**: added `da_block_hash_for_slot`; `StfInfoByNumber` / `StfInfoMetadata` CFs are kept in
  `LEDGER_TABLES` for backward compat (TODO #1945 to drop them via migration).

## Impact / migration

- **Module-state / schema breaking-change check: none.** No `#[state]` / `ModuleInfo` changes, no
  stored-struct (Borsh) changes, no `demo-rollup-schema.json` change.
- **Node DB format**: the proof-manager DB's on-disk format changed (composite key; dropped metadata
  CF). The proof-manager DB is new on this branch, unreleased, and is a **disposable re-provable
  queue** (proofs aren't enabled in production) → effectively non-breaking. A dev node that ran an
  *earlier commit of this branch* should delete its `proof-manager/` DB directory on upgrade.
- **On restart**, the node re-proves from `latest_proof_final_slot + 1` (tolerated under the resync
  flag).

## Reviewer notes

- This **inverts** the original PR's mechanism: cursors are no longer persisted, and re-proving on
  restart is intended (not a regression of #1945).
- Fork selection re-couples the **producer** (not the consumer) to LedgerDb via the hash resolver.
- The resync flag is now a hard requirement for the zk pipeline when resuming from a proof.

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2443" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/full-node/sov-stf-runner/src/runner.rs
```diff
@@ -151,6 +157,7 @@ where
         sync_state: Arc<DaSyncState>,
         da_service_with_cached_finalized_headers: DaServiceWithCachedFinalizedHeaders<Da>,
         genesis_da_height: u64,
+        storage_path: Option<PathBuf>,
```
citizen-stig: Should it become part of `ProofManagerConfig`? Why separate param?

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
```diff
@@ -467,6 +477,13 @@ where
 
         let sending_to_prover_start = std::time::Instant::now();
         if let Some(stf_info_sender) = &mut self.stf_info_sender {
+            // Only advance write_height for slots actually finalized to LedgerDb disk.
+            // Staged STF data for non-finalized slots is recovered by
+            // validate_and_recover_write_height on restart.
+            if let Some(finalized_slot) = last_finalized_slot_number {
```
bkolad, line 503: Is it possible that validate_and_recover_write_height sets write_height to a value higher than finalized_slot, and then stf_info_sender.commit_stf_info(finalized_slot).await? reduces the stored value?

crates/full-node/sov-db/src/proof_manager_db.rs
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    /// Get the write height (latest STF info written).
+    pub fn get_write_height(&self) -> anyhow::Result<Option<SlotNumber>> {
+        self.db.get::<StfInfoMetadata>(&WRITE_ROLLUP_HEIGHT_ID)
+    }
+
+    /// Set next_height_to_receive. Writes immediately to disk.
+    ///
+    /// This is the critical method that fixes the duplicate proof submission bug.
```
bkolad: If the bug is fixed, the comment is no longer helpful since it refers to something that no longer exists.

crates/full-node/sov-db/src/proof_manager_db.rs
```diff
@@ -0,0 +1,573 @@
...
+    /// - If ProofManagerDb already has STF state/metadata, startup is allowed.
+    /// - If `ledger_head` is above genesis and ProofManagerDb is empty, startup fails fast.
+    ///
+    /// By design this does not migrate STF state from legacy LedgerDb tables.
+    pub fn ensure_initialized_for_ledger_head(
+        &self,
+        ledger_db: &LedgerDb,
+        ledger_head: SlotNumber,
+    ) -> anyhow::Result<()> {
+        if ledger_head <= SlotNumber::GENESIS {
```
bkolad: How could this happen?

crates/full-node/sov-db/src/proof_manager_db.rs
```diff
@@ -0,0 +1,573 @@
...
+    /// Validate ProofManagerDb state against the ledger head on startup and
+    /// recover `write_height` if metadata lagged behind actual stored STF info.
+    ///
+    /// Enforces the invariant: proof_manager_write_height <= ledger_head.
+    /// If the invariant is violated, truncate ProofManagerDb to `ledger_head`.
+    pub fn validate_and_recover_write_height(&self, ledger_head: SlotNumber) -> anyhow::Result<()> {
+        let maybe_write_height = self.get_write_height()?;
+        let mut write_height = maybe_write_height.unwrap_or(SlotNumber::GENESIS);
+
+        if write_height > ledger_head {
```
bkolad: What is the scenario that could lead to ` write_height > ledger_head`?

crates/full-node/sov-stf-runner/src/processes/stf_info_manager.rs
```diff
@@ -293,69 +308,81 @@ where
...
+            %write_rollup_height,
+            "Done materializing stf_info data"
+        );
+        Ok(schema)
+    }
 
-        // Update the write rollup height.
-        schema.merge(ledger_db.materialize_stf_info_write_slot_number(write_rollup_height)?);
+    /// Stage STF info data in ProofManagerDb immediately.
+    pub async fn stage_stf_info(
```
bkolad, line 350: Let me know if I’m understanding this correctly.

Before this PR, STF info and ledger data were stored in the same DB, so their commits were always consistent. Now that `STFInfo` uses a separate DB, inconsistencies can occur if one commits and the other doesn’t (e.g., during shutdown). To mitigate this, we introduced the concept of `staging` and `validate_and_recover_write_height.` Does that help resolve inconsistencies?  

If the above is correct, let's maybe add a short doc on why we need the staging.

citizen-stig: I've added crash tests to check this

crates/full-node/sov-db/src/proof_manager_db.rs
```diff
@@ -0,0 +1,620 @@
...
+    }
+
+    /// Create a new [`ProofManagerDb`] from an existing RocksDB instance.
+    pub fn new(db: Arc<DB>) -> Self {
+        Self { db }
+    }
+
+    /// Open the database at the given path.
+    pub fn open(path: impl AsRef<std::path::Path>) -> anyhow::Result<Self> {
+        let db = Self::get_rockbound_options().default_setup_db(path)?;
```
bkolad: `default_setup_db` ignores `DB_PATH_SUFFIX`. We need to `call default_setup_db_as_subdir`

citizen-stig: Good catch — fixed in `ea6fabba8`. `open()` now calls `default_setup_db_as_subdir`, so the DB is created under the `proof-manager/` subdirectory again (the `default_setup_db` call had slipped in via the earlier WIP commit).

crates/full-node/sov-stf-runner/src/runner.rs
citizen-stig: Done — `storage_path` is now a field on `ProofManagerConfig` rather than a separate `new()` param. It defaults to `rollup_config.storage.path` when unset (resolved in the blueprint), and `StateTransitionRunner::new` reads it from `config.storage_path`. Resolving.

crates/full-node/sov-stf-runner/src/state_manager/mod.rs
citizen-stig: No, that sequence can't happen:

- `validate_and_recover_write_height` only clamps `write_height` *down* to ≤ `latest_finalized_slot_number` (and the forward-recovery loop also stops at `latest_finalized_slot_number`); it never sets it above the finalized view.
- After startup, the first `commit_stf_info(finalized_slot)` gets a `finalized_slot` ≥ the startup finalized slot (finality is monotonic — see the safety note in `state_manager/mod.rs`), so it's ≥ the recovered `write_height`.
- `commit_stf_info` also asserts `current_write_rollup_height <= write_rollup_height`, so a reduction would be caught, not silently applied.

So `commit_stf_info` can only advance/keep `write_height`, never reduce it.

crates/full-node/sov-db/src/proof_manager_db.rs
citizen-stig: Reworded — the doc no longer mentions "the bug". It now describes the mechanism: immediate persistence avoids re-submitting the same aggregated proof if the node crashes after proof posting but before the next ledger commit.

crates/full-node/sov-db/src/proof_manager_db.rs
citizen-stig: Obsolete — `ensure_initialized_for_ledger_head` (and `bootstrap_future_only`) were removed during cleanup, so this GENESIS check no longer exists; startup now goes straight through `validate_and_recover_write_height`. Resolving. (Note: the empty-DB-on-non-genesis-ledger fail-fast is gone too — if we want to keep that guard, let's track it separately.)

crates/full-node/sov-db/src/proof_manager_db.rs
citizen-stig: It comes from `ProofManagerDb` and `LedgerDb` being separate, non-atomic RocksDB instances. `write_height` is committed to ProofManagerDb in `commit_stf_info` (immediate write) after the ledger change set is saved/finalized, but the two writes aren't a single transaction. If the node crashes in the window where the ProofManagerDb write durably landed but the ledger's slot/finalized record didn't flush, restart observes `write_height > ledger_head`. `validate_and_recover_write_height` handles it: it truncates staged STF rows above `ledger_head` and clamps `write_height` to `min(latest_finalized, ledger_head)`. The staged-rows-above-head path is exercised by `test_proof_manager_crash_after_staging_hides_uncommitted_slot`.

crates/full-node/sov-stf-runner/src/processes/stf_info_manager.rs
citizen-stig: Added — `stage_stf_info` now has a "# Two-phase commit pattern" doc block explaining that ProofManagerDb and LedgerDb can't commit atomically, so STF data is staged before the ledger commit and `write_height` is only advanced after it succeeds. Also added crash tests: `test_proof_manager_crash_after_staging_hides_uncommitted_slot` and `test_proof_manager_restart_recovers_after_ledger_finalize_crash`.

crates/full-node/sov-db/src/proof_manager_db.rs
```diff
@@ -0,0 +1,620 @@
...
+            if self.get_stf_info(current)?.is_some() {
+                return Ok(Some(current));
+            }
+
+            let Some(previous) = current.checked_sub(1) else {
+                return Ok(None);
+            };
+            current = previous;
+        }
+    }
```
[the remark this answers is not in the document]
citizen-stig: Fixed — `find_largest_stored_slot_at_or_below` now does a single reverse seek (`get_prev::<StfInfoByNumber>(&upper_bound)`) instead of the per-slot downward scan, so it's O(log n) regardless of ledger height.

citizen-stig:
## Follow-up: avoid re-proving the posted-but-unaccepted window on restart (double-proving)

The restart SIGABRT in `restart::proofs::test_aggregated_proofs_after_restart_external_da` (slot-number discontinuity in `verify_proof_chain`, `circuit.rs:173`) is **fixed in this PR** with dev-like resume behavior:

- In `load_next_height_to_receive` (`stf_info_manager.rs`), when a predecessor aggregated proof exists, resume the receive cursor **exactly** at `latest_proof_final_slot + 1` instead of `max(persisted_cursor, …)`; when there is no accepted proof yet, keep resuming from the persisted cursor (so the no-predecessor restart path is preserved). This keeps the cursor in lockstep with the outer-aggregation host, which is restored from the latest *accepted* proof.

### What this leaves for follow-up
The receive cursor is persisted at **DA-post time** (`zk_manager/mod.rs:415`, "to avoid duplicate proof submission after restart"), but the outer-host aggregation state is restored from the latest **accepted** proof. With non-zero-finality DA, a proof can be posted to DA yet not read back/accepted before a restart. The dev-like fix correctly resumes from the accepted point — but that means any window posted-but-not-accepted is **re-proved and re-aggregated** on restart, and its already-posted proof blob in DA is re-read. So the "avoid duplicate proof submission" optimization no longer holds across that window: in production this wastes proving work and can post a duplicate aggregated proof for those slots.

(This matches `dev`'s behavior and is correct — just not optimal.)

### Suggested approach
Make both restart-resume reference points share the *posted* checkpoint rather than the *accepted* one, so the cursor and the outer host stay contiguous **without** re-proving:

- Persist the last-**posted** aggregated proof (or enough to reconstruct its witness) in ProofManagerDb alongside the cursor, and restore the outer-host aggregation state from it on startup — both then resume from the last posted proof.
- Alternative (simpler but weaker): persist the receive cursor at **accept-time** instead of post-time so it never leads the accepted proof. This avoids the divergence but still re-posts on a crash inside the post-but-not-accept window, so it doesn't fully eliminate double-proving.

Touches: outer-host restore (`native_only/mod.rs` `create_prover_service`), ProofManagerDb schema, and the cursor-resume wiring in `stf_info_manager.rs`.

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
```diff
@@ -106,7 +106,9 @@ where
 
             let aggregator_shutdown = self.shutdown_receiver.clone();
             let aggregator_handle = tokio::spawn(async move {
-                run_task_until_shutdown("aggregator", aggregator.run(), &aggregator_shutdown).await;
```
bkolad: The previous way of handling shutdowns looks more concise. In the new commit, we moved the check into the method, and it runs at every await point.

crates/full-node/sov-stf-runner/src/processes/stf_info_manager.rs
```diff
@@ -54,16 +54,87 @@ impl<StateRoot, Witness, Da: DaSpec> StateTransitionInfo<StateRoot, Witness, Da>
...
 }
 
+/// Shared `next_height_to_receive` cursor.
+///
+/// A single mutex serializes all access. The persistent advance is a
+/// load-write-store sequence (an in-memory update plus a ProofManagerDb write),
+/// not a single atomic operation, so writers must be serialized; reads share the
+/// same lock.
+struct Cursor {
+    value: Mutex<SlotNumber>,
```
bkolad, line 61: Let's use u64 and atomic operations instead of a mutex. This still can be made typesafe because the accessors can accept/return SlotNumber

PR #2454, opened 2026-02-06
Add raw access APIs (native-only) for DB migrations

preston-evans98:
# Description

This PR adds state access APIs for DB migrations which give access to the raw bytes of the value. It also adds a (native-only) API for iterating over keys in a statemap and getting the raw bytes.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/sov-modules-api/src/containers/map.rs
```diff
@@ -348,6 +350,93 @@ where
...
+    }
+
+    /// Iterates raw values for caller-provided keys.
+    ///
+    /// This is a backend-agnostic fallback when prefix iteration is unavailable.
+    pub fn iter_raw_from_keys<I, Kq, Reader>(
+        &self,
+        keys: I,
+        state: &mut Reader,
+    ) -> Result<impl Iterator<Item = (Kq, Option<Vec<u8>>)>, Reader::Error>
```
bkolad: Let’s return the `entries` Vec here. The allocation is already done, and we can do more with a vector than with an iterator (and the caller can always convert the Vec into an iterator if needed).

crates/module-system/sov-modules-api/src/containers/map.rs
```diff
@@ -404,6 +493,70 @@ where {
...
+    pub fn iter_raw<'a, S>(
+        &'a self,
+        storage: &'a S,
+    ) -> anyhow::Result<Option<impl Iterator<Item = RawMapEntry<K>> + 'a>>
+    where
+        S: NativeStorage,
+    {
+        let prefix = SlotKey::singleton(self.prefix());
+        let maybe_iter = storage.maybe_iter_kernel_values_with_prefix(prefix)?;
+        Ok(maybe_iter.map(move |iter| {
```
bkolad: Let's amke a helper in `NamespacedStateMap<N, K, V, Codec>` for the common code in `NamespacedStateMap::iter_raw` for Kernel and User namespace.

crates/module-system/sov-modules-api/src/containers/map.rs
```diff
@@ -442,3 +595,149 @@ where
...
+        let mut map = StateMap::<u64, u32>::with_codec(prefix, BorshCodec);
+        map.set(&1_u64, &10_u32, &mut state).unwrap_infallible();
+        map.set(&2_u64, &20_u32, &mut state).unwrap_infallible();
+
+        // JMT storage does not support prefix iteration.
+        assert!(map.iter_raw(&storage).unwrap().is_none());
+    }
+
+    #[test]
+    fn state_map_iter_raw_kernel_namespace_on_nomt() {
```
bkolad, line 686: The same test for user namespace is missing, but if we make the iteration code commin then I belive we are ok

bkolad (approved):

crates/module-system/sov-modules-api/src/containers/map.rs
preston-evans98: Done.

crates/module-system/sov-modules-api/src/containers/map.rs
preston-evans98: Good idea. Factored into a helper

crates/module-system/sov-modules-api/src/containers/map.rs
preston-evans98: Makes sense. Done.

PR #2456, opened 2026-02-07
Simplify `NodeDiscovery` tool.

bkolad:
This PR is prerequisite for https://github.com/Sovereign-Labs/sovereign-sdk/pull/2457

# Description

This PR refactors the sov-proxy-utils` crate, which monitors a PostgreSQL database for cluster node membership and leadership changes, writing cluster information to a file for consumers.

Three changes:

1. ClusterInfo.followers: Vec → BTreeMap
The separate `members: BTreeSet<String>` field is removed — membership is derived from the map's keys. A special case where the leader was auto-added to followers when it was the only node is also removed.

2. NodeDiscovery API redesigned
`new() → connect()` (eagerly creates PgListener and writes initial file). `subscribe_cluster_info_loop` now consumes self and returns a NodeDiscoverySubscriber with a built-in `watch::Receiver<ClusterInfo>`, replacing the external SimpleClusterUpdateNotifier pattern.

3. Module split: `lib.rs` split into `file_writer.rs` and `node_discovery.rs.`

## Linked Issues
- Related to https://github.com/Sovereign-Labs/sovereign-sdk/pull/2457


crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -0,0 +1,432 @@
...
+            let node = NodeInfo {
+                node_id,
+                address,
+                last_updated,
+            };
+
+            if Some(&node.node_id) == leader_id.as_ref() {
+                leader = Some(node);
+            } else {
+                match followers.entry(node.node_id.clone()) {
```
bkolad, line 192: This PR is primarily code cleanup and refactoring, breaking the logic into smaller pieces etc.

The only substantive change from the previous version is that if a cluster contains only a leader, we no longer insert it as a follower. This case will be handled in Nginx. Handling it here introduces a number of edge cases that won’t exist once the logic is moved to Nginx.

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -0,0 +1,432 @@
...
+                    Entry::Occupied(occupied) => {
+                        anyhow::bail!("Duplicate node id found in Nodes table: {}", occupied.key());
+                    }
+                }
+            }
+        }
+
+        if let Some(leader_id) = &leader_id {
+            if leader.is_none() {
+                anyhow::bail!("Leader is missing from the Nodes table. leader_id: {leader_id}");
```
citizen-stig: Nit: non-blocking, but it might help if we are able to print whole nodes `all_nodes` vec. It is consumed above, but maybe it is possible to refactor it.

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -0,0 +1,432 @@
...
+        }
+
+        Ok(ClusterInfo { leader, followers })
+    }
+
+    /// Subscribes to PostgreSQL notifications for cluster changes and writes
+    /// cluster info to a file whenever the cluster state changes.
+    ///
+    /// Listens on `nodes_changes` and `leader_changes` channels.
+    pub async fn spawn(mut self) -> NodeDiscoveryTask {
```
citizen-stig: Does this method needs to ba async?

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -0,0 +1,432 @@
...
+    ) -> anyhow::Result<()> {
+        let info = self.get_cluster_info().await?;
+        let leader_id = info.leader_id();
+
+        let followers: BTreeSet<_> = info.followers.keys().cloned().collect();
+
+        let membership_changed = self.prev_followers != followers;
+        let leader_changed = self.prev_leader_id != leader_id;
+
+        tracing::trace!(info = ?info, "Last cluster info");
```
citizen-stig: nit: logging `membership_changed` and `leader_changed` might be helpfuil

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -0,0 +1,432 @@
...
+            self.prev_leader_id = leader_id;
+
+            // Notify watchers that the cluster was updated.
+            if let Some(notifier) = &self.notifier {
+                notifier.on_cluster_update(&info).await;
+            }
+            let _ = sender.send(info);
+        }
+
+        // Wait for at least one notification.
```
citizen-stig: Stupid question, but why notifications are handled here, and after we've processed cluster ifno

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -0,0 +1,432 @@
...
+        // Drain any additional pending notifications.
+        while self.listener.next_buffered().is_some() {}
+
+        Ok(())
+    }
+
+    async fn get_cluster_info_from_db(
+        &self,
+    ) -> Result<(Option<String>, Vec<(String, String, OffsetDateTime)>)> {
+        let max_age_secs = self.max_age.as_secs() as i64;
```
citizen-stig: Non-blocking: but maybe do `try_into` with ? so we 100% sure that incorrect arg won't blow up things

citizen-stig (approved):

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -0,0 +1,432 @@
...
+        let followers: BTreeSet<_> = info.followers.keys().cloned().collect();
+
+        let membership_changed = self.prev_followers != followers;
+        let leader_changed = self.prev_leader_id != leader_id;
+
+        tracing::trace!(info = ?info, "Last cluster info");
+
+        if membership_changed || leader_changed {
+            // Log membership changes
+            for node_id in followers.difference(&self.prev_followers) {
```
citizen-stig: From devin:

> Misleading join/leave log messages when nodes change roles (follower↔leader)

The refactoring changed prev_members (tracking ALL node IDs: leader + followers) to prev_followers (tracking only follower IDs). The "Node joined/left the cluster" log messages now operate on the follower set only, which produces factually incorrect logs during leader changes.

The old code (, removed lines 315-320) tracked membership as all node IDs (leader + followers) via prev_members: BTreeSet<String> and info.members. When a node changed roles (e.g. follower→leader), it remained in members, so no spurious join/leave was logged.

The new code tracks only prev_followers (line 118, 237), so when a follower is promoted to leader:

    It disappears from the followers set
    self.prev_followers.difference(&followers) includes the promoted node
    Log emits: "Node left the cluster" — but the node is still in the cluster as the new leader!

Conversely, when a leader is demoted to follower, the log emits "Node joined the cluster" even though it was already present.

Concrete scenario (leader failover):

    Before: leader=A, followers={B} → prev_followers={B}
    A dies, B promoted: leader=B, followers={} → followers={}
    prev_followers.difference(&followers) = {B} → logs "Node B left the cluster"
    But B is actually the new leader, still very much in the cluster.

This could cause confusion during operational incident response when engineers read the logs.

bkolad: Devin didn't take into account line 253 in this file, which logs the leader change, but the message can still be improved. I will fix it.

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: And not at the beginning of the loop? This is because we want to call `handle_cluster_update` the first time without checking for notifications, so that if there’s already data in the database, we fetch it immediately instead of waiting for a notification.

citizen-stig: My question, what if notification has arrived while we we're parsing info, then we will skip the update until next notification.

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: I will add `followers` to the message.

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: Good point. Fixing it.

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: will add

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: fixed

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: fixed

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: fixed

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: fixed

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: fixed

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -0,0 +1,432 @@
...
+        let leader_id = info.leader_id();
+
+        let followers: BTreeSet<_> = info.followers.keys().cloned().collect();
+
+        let membership_changed = self.prev_followers != followers;
+        let leader_changed = self.prev_leader_id != leader_id;
+
+        tracing::trace!(info = ?info, membership_changed, leader_changed, "Last cluster info");
+
+        if membership_changed || leader_changed {
```
citizen-stig: Another nit is we can do return early:

```rust

if !(membership_changed || leader_changed) {
   return Ok(())
}
```

and the rest of the method is one level intendation less.

crates/utils/sov-proxy-utils/src/node_discovery.rs
citizen-stig: Discussed offline

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: Ohh I think I have to fix it again

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: fixed

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: Ok now it is fixed properly

PR #2457, opened 2026-02-08
Add root hash_checker to sov-proxy-utils.

bkolad:
# Description

This PR adds a background service in  `sov-proxy-utils` that periodically verifies all nodes in a cluster agree on the same state root hash, catching leader/replica divergence early.

The core new component is `ClusterRootHashChecker`, which every second:
1. Queries the leader's finalized slot
2. If the slot advanced ≥5 since the last check, queries every node's HTTP API (GET /ledger/slots/{n}) in parallel
3. Compares the returned state_root values and publishes a RootHashCheck via a watch channel. 

A new `ClusterInfoService` wraps the existing `NodeDiscovery` and the new root checker.


crates/utils/sov-proxy-utils/src/root_hash_checker.rs
```diff
@@ -0,0 +1,410 @@
...
+use std::time::Duration;
+use tokio::sync::watch;
+use tokio::task::JoinHandle;
+
+/// Default timeout for HTTP requests to node APIs.
+const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+/// Minimum slot distance between consecutive root-hash checks.
+const SLOT_QUERY_STEP: u64 = 5;
+/// Period between iterations of the background root-hash checker task.
+const DEFAULT_ROOT_HASH_CHECK_INTERVAL: Duration = Duration::from_secs(1);
```
bkolad: These are good defaults; for now, we’ll keep them here and pass them as command-line arguments.

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
```diff
@@ -0,0 +1,410 @@
...
+
+                if !root_hash_check.failed_nodes.is_empty() {
+                    tracing::warn!(
+                        slot_number = root_hash_check.slot_number,
+                        failed_nodes = ?root_hash_check.failed_nodes,
+                        "Failed to fetch root hash from some cluster nodes"
+                    );
+                }
+
+                match root_hash_check.all_match() {
```
bkolad: We will add metrics here in the next PR

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
```diff
@@ -0,0 +1,410 @@
...
+        if slot_number < last_checked_slot_number {
+            // The reference slot can decrease when leadership changes.
+            return true;
+        }
+
+        slot_number - last_checked_slot_number >= SLOT_QUERY_STEP
+    }
+
+    async fn get_finalized_slot(&self, node: &NodeInfo) -> Result<types::Slot> {
+        let url = format!("http://{}/ledger/slots/finalized", node.address);
```
bkolad: Right now, we compare root hashes for the finalized slot, but once the full solution is working, I’ll consider querying latest or latest-1.

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
```diff
@@ -0,0 +1,405 @@
...
+impl RootHashCheck {
+    /// Evaluates whether the observed root hashes are consistent.
+    /// Returns `AllMatch` even when only one node responded.
+    pub fn all_match(&self) -> RootHashConsistency {
+        let mut values = self.node_results.values();
+        let Some(first) = values.next() else {
+            return RootHashConsistency::NoData;
+        };
+
+        if values.all(|root_hash| root_hash == first) {
```
bkolad, line 58: Devin complains that we don’t consider failed nodes here, but this is intentional. If a node is not ready yet or is failing for any reason, we don’t compare its root hash.

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
citizen-stig: I wonder, why do we need pulling slots, but not subscribe to slot notifications. This way it will be independend from different block/slot times

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// Per-node results: node_id -> state_root.
+    pub node_results: HashMap<String, String>,
+    /// Nodes that failed to return a root hash `(node_id, error_message)`.
+    pub failed_nodes: Vec<(String, String)>,
+}
+
+impl RootHashCheck {
+    /// Evaluates whether the observed root hashes are consistent.
+    /// Returns `AllMatch` even when only one node responded.
+    pub fn all_match(&self) -> RootHashConsistency {
```
citizen-stig: nit: I find method name a little bit confusing.

I propose following options:

* `check_consistency`
* `result`

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
```diff
@@ -0,0 +1,405 @@
...
+        } else {
+            RootHashConsistency::Mismatch
+        }
+    }
+}
+
+/// Task returned when spawning the root-hash checker.
+pub struct ClusterRootHashCheckerTask {
+    /// Subscription receiver for root-hash check results.
+    pub receiver: watch::Receiver<RootHashCheck>,
```
citizen-stig, line 69: Are those attributes needs to be public?

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
```diff
@@ -0,0 +1,405 @@
...
+
+                match root_hash_check.all_match() {
+                    RootHashConsistency::AllMatch => {
+                        tracing::info!(
+                            slot_number = root_hash_check.slot_number,
+                            "Cluster root hashes are consistent"
+                        );
+                    }
+                    RootHashConsistency::Mismatch => {
+                        tracing::warn!(
```
citizen-stig: Should it be error?

We print warning in many other places, so I wonder what is expected alert behaviour?

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
citizen-stig: It is interesting question.

For Celestia we don't really care, because finalized will be pretty close to the latest (0-2 blocks, depending on how RPC background task pulling)

But if there's non-instant finality, it might be possible that replica diverge, because they might get response from different forks of DA. 

But that's broader discussion, my point that we probably want to stick for finalized slot for longer.

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
bkolad: Nodes can join and leave, and the subscription is stateful. This requires additional logic for error handling and cleanup when a node joins or leaves the cluster. The http query is more stateless, which I prefer in this scenario.

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
```diff
@@ -0,0 +1,405 @@
...
+    async fn get_slot_root_hashes_in_parallel(
+        &self,
+        nodes: Vec<NodeInfo>,
+        slot_hash: &str,
+        slot_number: u64,
+    ) -> RootHashCheck {
+        let mut query_tasks = Vec::with_capacity(nodes.len());
+
+        for node in nodes {
+            let http_client = self.http_client.clone();
```
citizen-stig, line 248: nit: I think the body of loop can be extracted into method of NodeInfo with some params, or separate function, so all error handling is done there and makes it compact.

Also about timeouts. 
Why do we put request timeout on tokio future, and not setting on the reqwest client? This can simplify code a lot

citizen-stig: Also reqwest can have more granular timeouts:

for example connect timeout might be smaller, so we faster detect unavailable nodes, and full request timeout larger giving rollup node some room to response

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
bkolad: fixed

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
citizen-stig: But quering more often will put extra load on disk

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
bkolad: Fixed. The receiver remains public to allow subscriptions to RootHashCheck, while the handle is now pub(crate) with abort and join methods exposed for better encapsulation. (same for node-discovery)

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
citizen-stig: Can we add this comment as TODO or item in the tracking issue, so it won't be lost?

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
bkolad: That's correct, changed to error.

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
bkolad: Currently, we query only the reference node every second, and I can safely increase this interval to 5 seconds (which I’ll do). The follower nodes are queried at most every 10 seconds.

The key point is that the database currently determines the cluster and acts as the source of truth. While this information can occasionally be outdated, that’s ok because the system self-corrects on the next database notification.

Introducing subscriptions would require maintaining a second place where we track connected nodes, which adds complexity and additional edge cases. In effect, the tool would maintain a second view of the cluster state, and that view would need to remain consistent with the database. But this is not easy as sometimes the db notifications can arrive before the subscriptions errors and sometime the opposite is true

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
bkolad: Added a TODO. I’ll address it in the next PR, since I first need to verify what needs to be done on the proxy server before it can submit metrics.

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
bkolad: Added 
1. `DEFAULT_CONNECTION_TIMEOUT`
2. `get_slot` method that simplified the loop (this was great suggestion).

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
```diff
@@ -0,0 +1,401 @@
...
+
+                let root_hash_check = self
+                    .get_slot_root_hashes_in_parallel(nodes, slot_hash, slot_number)
+                    .await;
+
+                if !root_hash_check.failed_nodes.is_empty() {
+                    tracing::warn!(
+                        slot_number = root_hash_check.slot_number,
+                        failed_nodes = ?root_hash_check.failed_nodes,
+                        "Failed to fetch root hash from some cluster nodes"
```
citizen-stig:
```suggestion
                        "Failed to fetch root hash from some cluster nodes, will compare root hash from the remaining"
```

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
```diff
@@ -0,0 +1,401 @@
...
+                    );
+                }
+
+                // TODO: Add a metric for root hash consistency check.
+                match root_hash_check.check_consistency() {
+                    RootHashConsistency::AllMatch => {
+                        tracing::info!(
```
citizen-stig: So that will be info log every second or 2. Not big volume, but will add up.

Would it worth it to have a debug level?

citizen-stig (approved):

crates/utils/sov-proxy-utils/src/root_hash_checker.rs
bkolad: fixed

PR #2461, opened 2026-02-09
Prevent background poller from filling finalized header cache during sync

preston-evans98:
# Description

This PR adds a heuristic to prevent the background task which watches for finalized headers from filling up the finalized headers cache with blocks that are far from the current sync site. If there are any old blocks in cache ("old" meaning, "more than max_cache_size from the chain tip"), the background task won't insert into the cache. 

Previously, the background task would fill up the cache within a few minutes, prevent the bulk fetcher from caching any headers. This made the cache useless after the first few minutes of sync, which would slow sync from 60-100 blocks/sec to 3-4. 

Note that the new heuristic only looks at the cache, so we must ensure that the bulk fetcher always keeps at least one header in cache while it's running - otherwise the heuristic won't detect that we're syncing and the cache will fill up again.


citizen-stig:
Approved (click on approve fails)

preston-evans98:
Merged manually from the command line since github is down.

PR #2462, opened 2026-02-10
EVM: BaseFee fix 1: pass it to TransactionResponse

citizen-stig:
# Description

> Prior to this PR, receipts already computed effective_gas_price correctly using block.maybe_partial_header().base_fee_per_gas at, but the transaction response from from_recovered_with_block_context always passed base_fee: None in TransactionInfo. When base_fee is None, alloy's from_transaction falls back to using max_fee_per_gas as the effective gas price, which would disagree with the receipt's value for any EIP-1559 transaction where max_fee_per_gas > base_fee + max_priority_fee_per_gas. This PR fixes that inconsistency. The new integration test eip1559_tx_gas_price_matches_receipt_effective_gas_price explicitly validates that both paths now agree.

## Linked Issues
- Related to #2237

## Testing
Added new tests

## Docs
No updates

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/helpers.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/helpers.rs
@@ -55,6 +55,7 @@ pub(crate) fn from_recovered_with_block_context(
     block_hash: Option<B256>,
     block_number: BlockNumber,
     tx_index: u64,
+    base_fee: Option<u64>,
 ) -> alloy_rpc_types::Transaction {
     let tx_info = TransactionInfo {
         block_hash,
@@ -62,14 +63,16 @@ pub(crate) fn from_recovered_with_block_context(
         index: Some(tx_index),
         // Default values
         hash: None,
-        base_fee: None,
+        base_fee,
     };
     alloy_rpc_types::Transaction::from_transaction(tx.convert(), tx_info)
 }
 
 #[cfg(test)]
 mod tests {
-    use alloy_primitives::{Address, U256};
+    use alloy_consensus::{EthereumTxEnvelope, Signed, TxEip1559};
+    use alloy_primitives::Signature;
+    use alloy_primitives::{Address, B256, U256};
     use revm::context::TransactTo;
 
     use super::*;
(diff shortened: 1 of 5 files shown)


crates/module-system/module-implementations/sov-evm/src/helpers.rs
```diff
@@ -55,21 +55,24 @@ pub(crate) fn from_recovered_with_block_context(
...
     tx_index: u64,
+    base_fee: Option<u64>,
 ) -> alloy_rpc_types::Transaction {
     let tx_info = TransactionInfo {
+        base_fee,
         block_hash,
         block_number: Some(block_number),
         index: Some(tx_index),
-        // Default values
+        // Default value. Why hash is None? Where does it actually set?
```
citizen-stig: TODO: Address this.

citizen-stig:
It says `any EIP-1559 transaction where max_fee_per_gas > base_fee + max_priority_fee_per_gas` 

But is this actually possible? If so, in which cases?

> Yes, this is not only possible but extremely common. Let me explain why.
EIP-1559 Transaction Fields
>
> An EIP-1559 transaction specifies two gas price parameters:
>
>    max_fee_per_gas: The absolute maximum the sender is willing to pay per unit of gas (inclusive of everything).
    max_priority_fee_per_gas: The tip the sender wants to give to the block producer.
> 
> The base_fee is set by the protocol per block and fluctuates based on network congestion.
When does max_fee_per_gas > base_fee + max_priority_fee_per_gas?
>
> This happens whenever the sender sets max_fee_per_gas with headroom above the current base_fee + tip. This is the normal, recommended practice because:
>
>    Base fee volatility: The base fee changes every block. If you set max_fee_per_gas to exactly base_fee + max_priority_fee_per_gas, your transaction may become unexecutable if the base fee rises even slightly before inclusion.
>
>    Wallets overshoot intentionally: Wallets like MetaMask typically set max_fee_per_gas to something like 2 * base_fee + max_priority_fee_per_gas to ensure the transaction gets included even if the base fee doubles.

preston-evans98 (approved):

PR #2469, opened 2026-02-11
  Decouple sov-metrics from sov-rollup-interface 

bkolad:
# Description

This PR removes the sov-rollup-interface dependency from sov-metrics to keep it lightweight, as sov-rollup-interface pulls in heavy dependencies (e.g., rockbound).

We replaced its types with basic Rust types in sov-metrics (e.g., `ExecutionContext → &'static str`, `VisibleSlotNumber → u64`).

This enables using `sov-metrics` in `sov-proxy-utils`, which must compile on relatively low-spec machines and should avoid heavy dependencies.

The relevant graphs continue to work:
http://37.27.235.101:3000/dashboard/snapshot/ouCi7qGu133I5mtvHFy9W8oj59lK3u5q
http://37.27.235.101:3000/dashboard/snapshot/CaGd98IKjgbhCBrqouRe0STvMCOcqfsx

## Linked Issues
- Related to # (issue) 

## Testing
Checked the relevant graphs on graphana.


crates/full-node/sov-metrics/src/influx_db_nonnative.rs
```diff
@@ -243,17 +241,27 @@ impl StateMetrics {
 }
 
 /// Metrics for `auth_and_process_tx`, and the tx hash
-#[derive(Debug)]
 pub struct AuthAndProcessMetrics {
     /// The transaction hash
-    pub tx_hash: HexHash,
+    pub tx_hash: [u8; 32],
```
bkolad, line 246: `tx_hash` field is not included in the metric only  in the Dbug implemntation of `AuthAndProcessMetrics`

crates/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -732,7 +732,7 @@ where
                     {
                         save_elapsed!(processing_time SINCE start_batch_processing);
                         let transactions_count = batch_receipt.tx_receipts.len();
-                        let ignored_transactions_count = batch_receipt.tx_receipts.len();
+                        let ignored_transactions_count = batch_receipt.ignored_tx_receipts.len();
```
bkolad, line 735: Previously, we had a bug here.

crates/full-node/sov-metrics/src/influxdb/publisher.rs
```diff
@@ -12,6 +13,25 @@ use crate::{Metric, TelegrafSocketConfig};
 
 const SHUTDOWN_DRAINING_LIMIT: std::time::Duration = std::time::Duration::from_secs(1);
 
+enum FutureOrShutdownOutput<O> {
```
bkolad, line 16: This is the only logic I had to bring from sov-rollup-interface

bkolad:
Devin found the following pre-existing problem:
```
SlotProcessingMetrics and UserSpaceSlotProcessingMetrics share the same measurement_name

Both SlotProcessingMetrics and UserSpaceSlotProcessingMetrics at and return "sov_rollup_slot_execution_time_us" from measurement_name(). This is a pre-existing pattern (not introduced by this PR), but it means both metric types write to the same Telegraf measurement. They do have different field sets (one has blobs_selection/finalization/da_height, the other begin_hooks/blobs_processing/end_hooks/rollup_height), so InfluxDB can distinguish them by fields present. Still, this could be confusing for dashboards or queries that don't filter by field existence.
```

I will log it as an issue.

citizen-stig (approved):

PR #2476, opened 2026-02-11
EVM: fix `eth_blockNumber` pending and latest semantics

citizen-stig:
# Description

In this PR endpoint `eth_blockNumber` changed to reflect Sovereign Rollup semantic, where latest == pending

## Linked Issues
- Related to #2271 

## Testing
Updated and added tests

## Docs
No updates


examples/demo-rollup/tests/evm/evm_block_by_number_hash.rs
```diff
@@ -130,7 +130,12 @@ async fn test_block_tags_earliest_safe_finalized() -> anyhow::Result<()> {
     let rollup = setup_paused_rollup(0, 2).await;
     let client = alloy_client(rollup.http_addr);
 
-    let sealed_head_number = client.get_block_number().await?;
+    let sealed_head_number = client
+        .get_block_by_number(Finalized)
```
preston-evans98, line 135: Here we rely on the semantics that `Finalized` returns the latest sealed soft confirmed block. I don't think we plan to change that, but maybe we should document that we have that assumption here

crates/module-system/module-implementations/sov-evm/src/rpc/handlers.rs
```diff
@@ -282,8 +282,9 @@ where
     #[rpc_method(name = "eth_blockNumber")]
     pub fn block_number(&self, state: &mut ApiStateAccessor<S>) -> RpcResult<U256> {
         trace!(method = "eth_blockNumber", "EVM module JSON-RPC request");
-        let block_number_range = self.block_numbers(state);
-        Ok(U256::from(*block_number_range.end()))
+        Ok(U256::from(
+            self.resolve_block_number(BlockNumberOrTag::Latest, state),
```
preston-evans98, line 288: This is definitely more correct than our current behavior... but I know that a lot of RPC tooling relies on this endpoint heavily and none of it seems to be broken currently so there's definitely some risk in making the change. I think we've got coverage of foundry's behavior in our acceptance tests, but let's make sure that we do before we merge this to nightly.

citizen-stig: Yeah, trade off is cross RPC calls inconsistency.

Here is from original report.

I will manually run foundry tests before merging.

```
Impact:
 - ethers.js/viem: provider.getBlockNumber() returns N, then provider.getBlock("latest") returns block N+1. Code that does getBlock(getBlockNumber()) gets a different block than getBlock("latest").
 - Polling loops: Tools that poll eth_blockNumber and then fetch that block by number will always be one block behind what "latest" returns.
 - Note: per repo semantics, latest == pending is acceptable (documented in scope). But eth_blockNumber returning a different value than eth_getBlockByNumber("latest").number is an internal inconsistency independent of that design
 choice.
```

citizen-stig:
```
==========================

ONCHAIN EXECUTION COMPLETE & SUCCESSFUL.
```

Just checked Foundry

examples/demo-rollup/tests/evm/evm_block_by_number_hash.rs
citizen-stig: Added a comment

crates/module-system/module-implementations/sov-evm/src/rpc/handlers.rs
```diff
@@ -463,17 +466,32 @@ where
...
+        //     Some(BlockId::Hash(block_hash.into())),
+        //     false.into(),
+        //     state,
+        // ) {
+        //     Ok(block) => block,
+        //     // ByHash count endpoints return null for not found blocks.
+        //     Err(EthApiError::HeaderNotFound(_)) => None,
+        //     Err(err) => return Err(err.into()),
+        // };
+        // Ok(block.map(|b| U64::from(b.transactions.len())))
```
preston-evans98: nit: Commented code

preston-evans98 (approved):

PR #2479, opened 2026-02-12
Unflaky archival state is immeidately available and fix API state update gap

citizen-stig:
# Description

This PR has 2 major changes:

* stabilizing `test_archival_state_is_immediately_available` by preventing automatic batch production, which could create empty batches and break expectations
* fix API state update, where notification via web-socket can arrive a little later than state checkpoint is constructed.

## Problem

  The sequencer had a race condition where clients could receive a WebSocket slot notification before the corresponding API state was queryable. The sync path enqueued a checkpoint update (via mpsc to the side-effects task) and then
  immediately broadcast slot notifications — without waiting for the checkpoint to be applied to the watch channel that the REST API reads from. This caused transient 404 HeightNotAccessible errors for clients that queried right after
  receiving a notification.

## Fix

  Added a oneshot acknowledgment to ForceUpdateApiState and UpdateStateForRecovery events (following the existing FlushTransactionsCache pattern). The sync path now awaits the ack — confirming the checkpoint is applied — before calling
  update_api_ledger which sends notifications. This covers all three affected code paths: normal state updates, recovery overwrites, and node resync.

## Test changes

  - Unflaked test_archival_state_is_immediately_available: disabled automatic_batch_production to prevent empty batches from breaking the test's height-to-value mapping (the actual root cause of the flakiness). Reverted Codex's
  retry/probing workaround back to the original clean index-based assertions.
  - Added regression test test_archival_state_is_accessible_even_with_delayed_checkpoint_update: injects a 750ms artificial delay before checkpoint application and verifies that no 404s occur — proving the ordering guarantee holds even
  under adverse timing.

## Observed failure modes

- Value mismatch due to empty preferred batches advancing rollup height without changing state
- Intermittent 404 invalid rollup height when test queried heights not yet archival-accessible

## What changed

- Removed brittle index-based expectation (expected = j + 1)
- Added per-height expected-value map:
  - First successful read for a height becomes canonical
  - Subsequent reads must match (immutability check)
- Before historical checks, wait/retry until latest value reflects the just-submitted tx
- Replaced tip height loop bound with highest accessible archival height probe
- Added targeted retry wrappers for archival reads on HeightNotAccessible-style errors (invalid rollup height, 404, related message)
- Refactored state query helper to expose a reusable raw query function

## Validation

- Repeated runs of:
  - SKIP_GUEST_BUILD=1 cargo nextest run -p sov-sequencer flaky_test_archival_state_is_immediately_available --retries 0
- Result after final patch: 10/10 passes locally
- SKIP_GUEST_BUILD=1 make lint passes locally (with cargo-dylint skipped by Makefile when not installed)

### More details

In the old logic, expected value was derived from position, not from actual execution:

- Conceptually: expected_value = j + 1 while iterating past heights
- So height/value mapping depended on loop index, not on what happened in that height

Why this breaks in preferred sequencer:

1. The sequencer can produce empty preferred batches (0 tx).
2. Empty batches still advance rollup height.
3. Value doesn’t change on those heights.
4. So the first empty height shifts all later index-based expectations.

Concrete mismatch pattern:

- Real chain can look like:
    - h=10 -> value=6
    - h=11 -> value=6 (empty batch)
    - h=12 -> value=7 (next tx applies)
- Index model expects:
    - h=10 -> 6
    - h=11 -> 7
    - h=12 -> 8
- Result: deterministic off-by-one failures.

There is a second brittleness: using tip height as if all <= tip are archival-queryable. In this system, tip can move ahead of archival accessibility briefly, causing 404 invalid rollup height for some intermediate heights.

That’s why the fix moved to:

- per-height observed expectations (not index-derived), and
- querying only heights that are actually archival-accessible.

----

## Testing
Last 3 runs this test never failed. Added special case of adding an explicit pause to the state checkpoint update

## Docs
No updates


crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -714,9 +706,14 @@ where
...
-        inner
+        let ack = inner
             .force_overwrite_state(info.clone(), recovery_executor)
             .await;
+        // Wait for the side-effects task to apply the checkpoint before sending
+        // slot notifications.
+        if ack.await.is_err() {
+            tracing::debug!("Checkpoint ack dropped; the side-effects task has likely shut down");
+        }
         inner.update_api_ledger(&info).await;
```
bkolad: The current fix will work, but it introduces tighter coupling between the `SideEffect` task and inner.rs.

I would suggest the following:
1. Move the `update_api_ledger` method to `SideEffects` (we will also need to move `api_ledger_db` there).
2. Add a new `ExecutorEvent::UpdateApiLedger` message.

This approach avoids introducing an additional oneshot channel and eliminates the need for two-way communication between the tasks.

citizen-stig: Addressed this in 1037953b8a38bc9e24acad922dc5dfdf3201ca47

But GitHub for some reason does not show me this commit at the moment

citizen-stig: Ok, pushed empty commit, it helped

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -717,7 +713,10 @@ where
         inner
             .force_overwrite_state(info.clone(), recovery_executor)
             .await;
-        inner.update_api_ledger(&info).await;
+        inner
```
citizen-stig, line 707: From devin:

```
process_force_overwrite_state_for_recovery does not send ForceUpdateApiState before UpdateApiLedger
```

citizen-stig: > However, this appears to match the pre-existing behavior — the old code also called force_overwrite_state without an explicit checkpoint event, then called update_api_ledger directly. This path is specifically for recovery entry and the sequencer is in a not ready state, so clients shouldn't be making queries at this point. Still, worth verifying that force_overwrite_state either sends its own checkpoint event or that the recovery state makes this ordering safe.

bkolad: To me this looks ok, `force_overwrite_state` calls `force_update_api_state` whcih sends `ForceUpdateApiState`

bkolad (approved):

PR #2482, opened 2026-02-12
Add timestamps to tx confirmations when possible

preston-evans98:
# Description

This PR adds the sequencer HDTimestamp to the `Confirmation` struct returned by the sequencer whenever possible. The timestamp is set to None if...
1. The tx was not from the preferred sequencer OR
2. The tx is no longer in cache *and* the tx body was not saved in the DB. This makes it impossible to reconstruct the timestamp. 

To ensure that tx bodies are always available, set `save_tx_bodies = true` in your rollup_config.toml

```
[runner]
save_tx_bodies = true # Whether to save transaction bodies to the database
```


crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -487,7 +490,17 @@ impl<S: Spec, Rt: Runtime<S>> RollupBlockExecutor<S, Rt> {
         );
 
         let tx = FullyBakedTxWithMaybeChangeSet::new(tx);
-        match self.apply_tx_to_in_progress_batch(tx).await {
+        let maybe_timestamp = tx.tx.sequencing_data.as_ref().and_then(|data| {
```
bkolad: Is there a reason we’re performing the conversion in `replay_tx` instead of in `apply_tx_to_in_progress_batch`? If we move it to `apply_tx_to_in_progress_batch`, we wouldn’t need to pass the redundant `maybe_timestamp` argument.

crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
```diff
@@ -251,6 +254,18 @@ impl<S: Spec, Rt: Runtime<S>> TransactionCache<S, Rt> {
         else {
             return Ok(None);
         };
+        let maybe_timestamp = tx.body.as_ref().and_then(|body| {
+            body.sequencing_data.as_ref().and_then(|data| {
+                let data = <Rt as HasCapabilities<S>>::SequencingData::try_from_slice(
+                    data.as_ref(),
+                )
+                .expect(
+                    "Invalid sequencing data. This is a bug in the sequencer, please report it.",
```
bkolad: Here we allow invalid sequencing data (which is the right thing)
https://github.com/Sovereign-Labs/sovereign-sdk/pull/2482/changes#diff-2883d9d7db0b3bea82fa48757f59265fc32fcc014889a9c519092331a843f7d2R133
But the get_tx_by_hash will fail if such a tx appears.
Should we just return None if we can't decode the seq data?

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -671,8 +673,20 @@ where
...
 
+        // Extract the timestamp from the sequencing data. We do this even though we could pass the timestamp directly from the place where it is generated
+        // for symmetry with the replicas. Replicas have to extract the timestamp from the sequencing data, but they can only do so if the runtime is using the standard
+        // sequencing data handler. Doing it the same way here ensures that the replica and the master agree on the timestamp in all cases.
+        let maybe_timestamp = baked_tx.sequencing_data.as_ref().and_then(|data| {
+            let data = <Rt as HasCapabilities<S>>::SequencingData::try_from_slice(data).expect(
+                "Invalid sequencing data. This is a bug in the sequencer, please report it.",
+            );
+            (&data as &dyn Any).downcast_ref::<HDTimestamp>().cloned()
+        });
```
bkolad: This block is repeated 4 times — let’s extract it into a helper.
We can add a `SequencingDataHandler::get_maybe_timestamp` method and call it here (and other places). That would mean that some  `SequencingData` implementations can optionally provide a timestamp.

Because inside `SequencingDataHandler`, we know the concrete `SequencingData` (which will be HDTimestamp in this case), we won't need `downcast_ref`.

crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
```diff
@@ -454,7 +470,13 @@ impl<S: Spec, Rt: Runtime<S>> AcceptedTxStream<S, Rt> {
             .into_iter()
             .flatten()
             .enumerate()
-            .map(|(idx, tx)| ApiAcceptedTx {
+            .map(|(idx, tx)| {
+                let timestamp_nanos = tx.body.as_ref().and_then(|body| body.sequencing_data.as_ref().and_then(|data| {
+                    let data = <Rt as HasCapabilities<S>>::SequencingData::try_from_slice(data.as_ref())
+                        .expect("Invalid sequencing data. This is a bug in the sequencer, please report it.");
```
bkolad: Same here lt's only log the error and return `Confirmation` with `timestamp_nanos: None`

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -868,7 +868,10 @@ where
             .map_err(|err| AcceptTxError::RateLimiter(err))?;
 
         let mut baked_tx = baked_tx;
-        baked_tx.set_sequencing_metadata(&get_hd_timestamp_with_maybe_override());
+        let timestamp = get_hd_timestamp_with_maybe_override();
+        // Important: we read the timestamp from the baked tx inside do_new_tx (motivation explained there)
+        // so this must not be moved without updating do_new_tx.
+        baked_tx.set_sequencing_metadata(&timestamp);
```
bkolad: This can be done in another PR/issue
Let's  add  `SequencingDataHandler:: create_sequencing_data() -> Vec<u8>`
then
```
let metadata = <Rt<S> as SequencingDataHandler>::create_sequencing_data()
baked_tx.set_sequencing_metadata(&metadata);
```

This way the `SequencingData` construction won't leak to the sequencer. 

If we also implemnt this:
https://github.com/Sovereign-Labs/sovereign-sdk/pull/2482#discussion_r2876964998
 then all the timestamp `serialisation/creation/handling` logic would be in a single place.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
preston-evans98: This was trickier than I thought, because we expect runtimes to be able to override capabilities. That meant that relying on the particular `SequencingDataHandler` was tricky. But I found a way to do it by adding an extra trait for the SequencingData associated type

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
preston-evans98: Done

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
preston-evans98: No reason anymore! I think this was leftover from an earlier draft. Moved.

crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
preston-evans98: Good catch. I think it's safer to log an error here

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -824,6 +810,10 @@ where
         ip_and_credential: IpAndCredentialId<S::Address>,
         reason: &'static str,
     ) -> Result<oneshot::Receiver<AcceptedTx<Confirmation<S, Rt>>>, AcceptTxError<S>> {
+        let sequencing_data = self
```
bkolad, line 813: Let;s move it below `self.get_inner_with_timing(reason)` so the time metrics show all the operations in process_accept_tx,

bkolad (approved):

bkolad (approved):

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
preston-evans98: Unfortunately we can't, since it has to borrow self.runtime

PR #2487, opened 2026-02-15
Additional metics in cluster-service

bkolad:
# Description

This PR refactors `sov-proxy-utils` to replace the old `root-hash-checker` with a unified `node-checker` that performs both root-hash and latest-height checks.

Before: `ClusterRootHashChecker` only verified that all nodes agree on the same state root at a finalized slot.

After: The checker is renamed to `NodeChecker` and now performs two checks — root-hash check and latest rollup height spread between the slowest and fastest node in the cluster.

The Grafana panels are under construction. 

`ERROR REPORTING/metrics`  in #1524
  
## Linked Issues
- Related to https://github.com/Sovereign-Labs/sovereign-sdk/issues/1524


crates/utils/sov-proxy-utils/src/node_check_metric.rs
```diff
@@ -44,9 +57,8 @@ impl sov_metrics::Metric for RootHashCheckMetric {
     fn serialize_for_telegraf(&self, buffer: &mut Vec<u8>) -> std::io::Result<()> {
         write!(
             buffer,
-            "{},outcome={} slot_number={},nodes_ok={},nodes_failed={},unique_state_roots={}",
+            "{} slot_number={},nodes_ok={},nodes_failed={},unique_state_roots={}",
```
bkolad, line 60: The `outcome` is redundant, as the same information can be derived from `unique_state_roots`.

crates/utils/sov-proxy-utils/src/node_checker.rs
```diff
@@ -0,0 +1,709 @@
...
+            ?self.height_diff,
+            ?self.spread,
+            "Cluster latest heights."
+        );
+    }
+}
+
+/// Result of one cluster-wide latest-height collection round.
+#[derive(Debug, Default, Clone)]
+pub struct LatestHeightCheck {
```
bkolad, line 92: Most of the code in this file was moved from `root_hash_checker.rs`. The new addition is LatestHeightCheck.

crates/utils/sov-proxy-utils/src/node_checker.rs
```diff
@@ -0,0 +1,709 @@
...
+                {
+                    Ok(slot) => slot,
+                    Err(error) => {
+                        tracing::warn!(?error, "Failed to get comparison slot for root-hash check");
+                        continue;
+                    }
+                };
+
+                let slot_number = reference_slot.number;
+                if !Self::should_query_slot(last_checked_slot_number, slot_number) {
```
bkolad, line 346: This returns true when the slot changes, so we won’t emit metrics more than once per slot.
The height may change more frequently than once per slot, but that’s fine—we’re only concerned with cases where the height spread between nodes remains large for an extended period of time.

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -178,7 +196,12 @@ impl NodeDiscovery {
         let handle = tokio::spawn(async move {
             loop {
                 if let Err(error) = self.handle_cluster_update().await {
-                    tracing::warn!(?error, "Cluster update failed");
+                    let stage = error.stage();
+
+                    tracing::warn!(stage, ?error, "Cluster update failed");
+                    sov_metrics::track_metrics(|tracker| {
+                        tracker.submit(ClusterUpdateFailureMetric { stage });
```
bkolad, line 203: On any failure we submit metric.

crates/utils/sov-proxy-utils/src/node_check_metric.rs
citizen-stig: basically alert will shoot if value > 1, right?

crates/utils/sov-proxy-utils/src/node_checker.rs
```diff
@@ -0,0 +1,709 @@
...
+    }
+
+    fn update_min(&mut self, node_id: String, height: u64) {
+        if height < self.min.1 {
+            self.min = (node_id, height);
+        }
+    }
+
+    fn diff(&self) -> u64 {
+        self.max.1 - self.min.1
```
citizen-stig: nit: and it should be impossible in current code, but safe math can give extra safety

crates/utils/sov-proxy-utils/src/node_check_metric.rs
bkolad: yep

citizen-stig (approved):

crates/utils/sov-proxy-utils/src/node_checker.rs
bkolad: fixed

PR #2489, opened 2026-02-16
Celestia: Enable verification of each celestia block on fetch

citizen-stig:
# Description

We've observed that buggy Celestia RPC node can cause silent state diveregence in operator mode or in zk mode with disabled proving.

This PR adds 

## Linked Issues
- Related to #1630 

## Testing
Existing tests are passing. No new tests have been added, because given functions are tested separately and having e2e test will be complicated to setup

## Docs
No updates

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2026-02-16
+- #2489 Non-breaking, but **important**: Celestia adapter now by default validates blocks fetched from RPC node.
+  This adds performance overhead. It can be disabled by setting `verify_on_fetch = false` in `da` section of rollup_config.toml. Do on your own risk.
+
 # 2026-02-09
 - #2459 Fixes in EVM RPC `eth_estimateGas` and `eth_getStorageAt`
 # 2026-01-29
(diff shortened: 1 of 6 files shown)


theodorebugnet (approved):
The main downside I see is that this would have to be implemented and configured per-DA, while having it in the runner would allow for a nice global switch. But I don't have full context on the runner and whether it would be worth the effort and/or the extra clone you mentioned on Slack. (And practically speaking Celestia is our only production DA right now.)

As far as this change goes it LGTM, but feel free to wait for a second opinion.

citizen-stig:
> The main downside I see is that this would have to be implemented and configured per-DA, while having it in the runner would allow for a nice global switch. But I don't have full context on the runner and whether it would be worth the effort and/or the extra clone you mentioned on Slack. (And practically speaking Celestia is our only production DA right now.)
> 
> As far as this change goes it LGTM, but feel free to wait for a second opinion.

Yes, probably better option would be to plug verifier directly in the runner, but this will increase the scope massively

preston-evans98 (approved):

PR #2494, opened 2026-02-17
Short-circuit transaction submission retries on a stop-height rejection

theodorebugnet:
# Description

Some time ago we added a special case to `sov-soak-testing-lib` to shutdown a soak worker if a "sequencer has reached the stop height and is shutting down" error is returned. However if a transaction is rejected for another reason (e.g. batch is full), and `use_retries == true`, the soak test calls `client.send_tx_to_sequencer_with_retry()`, and if this happened near the end of the soak test this would then swallow the "stop height reached" error.

Moving this check to the actual client code has implications for everyone else using the client as well. Do we believe upgrades will eventually be so low downtime, that continuing to retry with backoff across an upgrade will be a good strategy? If so then the error should *not* be special-cased, but otherwise IMO this change would makes sense: we float the error to the user's code which is then responsible for polling until the rollup goes back up (and e.g. fetching the new chain hash, if needed).

As far as the soak tests go, this can be worked around (by ignoring errors near the expected end of the test) so it's not strictly necessary. So the main question is what behaviour we want to expose from the client.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-api-spec/src/client.rs
+++ b/crates/full-node/sov-api-spec/src/client.rs
@@ -29,13 +29,24 @@ pub extern crate tokio_tungstenite;
 pub type WsSubscription<T> = Result<BoxStream<'static, anyhow::Result<T>>, WsError>;
 
 /// Message substring returned by the sequencer when it reached the configured stop height.
-const STOP_HEIGHT_ERROR_MARKER: &str = "The preferred sequencer has reached the stop height ";
+const STOP_HEIGHT_ERROR_MARKER: &[u8] = b"The preferred sequencer has reached the stop height ";
+const HTTP_4XX_STATUS_MARKER: &[u8] = b"\"status\":4";
+
+fn contains_subslice(haystack: &[u8], needle: &[u8]) -> bool {
+    if needle.is_empty() {
+        return true;
+    }
+    haystack
+        .windows(needle.len())
+        .any(|window| window == needle)
+}
 
 /// Checks whether an API client error indicates that the preferred sequencer reached stop height.
 pub fn is_stop_height_error(err: &Error<types::ApiError>) -> bool {
     matches!(
         err,
-        Error::ErrorResponse(response) if response.message.contains(STOP_HEIGHT_ERROR_MARKER)
+        Error::ErrorResponse(response)
+            if contains_subslice(response.message.as_bytes(), STOP_HEIGHT_ERROR_MARKER)
     )
 }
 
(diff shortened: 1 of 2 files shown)


crates/full-node/sov-api-spec/src/client.rs
```diff
@@ -142,12 +153,17 @@ impl Client {
...
                 match err {
                     Error::InvalidRequest(_) | Error::InvalidUpgrade(_) | Error::PreHookError(_) => false,
-                    Error::CommunicationError(_) | Error::ErrorResponse(_) | Error::ResponseBodyError(_) | Error::UnexpectedResponse(_) => true,
+                    Error::ErrorResponse(_) => !is_stop_height_error(err),
+                    Error::CommunicationError(_) | Error::ResponseBodyError(_) | Error::UnexpectedResponse(_) => true,
                     // This needs further improvement on the generated client.
                     // Details in https://github.com/Sovereign-Labs/sovereign-sdk-wip/pull/2799
                     Error::InvalidResponsePayload(bytes, _error) => {
+                        let response_body = std::str::from_utf8(bytes.as_ref()).unwrap_or("");
+                        if response_body.contains(STOP_HEIGHT_ERROR_MARKER) {
```
citizen-stig: Nit: The marker is already ASCII, so we can store it as byte: = b"The pre.." and then compare just bytes inclusion. Less allocations..

crates/full-node/sov-api-spec/src/client.rs
```diff
@@ -142,12 +153,17 @@ impl Client {
...
+                            return false;
+                        }
                         // All non-HTTP 4** are retried.
-                        !std::str::from_utf8(bytes.as_ref()).unwrap_or("").contains("\"status\":4")
+                        !response_body.contains("\"status\":4")
```
citizen-stig: Same about bytes comparison. Also it was this way, but if server decised to pretty print, it will fail

citizen-stig (approved):

crates/full-node/sov-api-spec/src/client.rs
```diff
@@ -142,12 +164,17 @@ impl Client {
...
                     // This needs further improvement on the generated client.
                     // Details in https://github.com/Sovereign-Labs/sovereign-sdk-wip/pull/2799
                     Error::InvalidResponsePayload(bytes, _error) => {
+                        let response_body = bytes.as_ref();
+                        if contains_subslice(response_body, STOP_HEIGHT_ERROR_MARKER) {
+                            return false;
+                        }
                         // All non-HTTP 4** are retried.
-                        !std::str::from_utf8(bytes.as_ref()).unwrap_or("").contains("\"status\":4")
+                        !contains_subslice(response_body, HTTP_4XX_STATUS_MARKER)
```
preston-evans98: Do we want to enforce that this is a 4xx error? It seems to me that searching for the `"The preferred sequencer has reached the stop height"` substring is the important check

theodorebugnet: This was existing logic here - 4xx errors abort retries immediately. I've just added the "reached the stop height" as another abort condition.
But your comment made me realise this isn't structured very clearly - I simplified it to make it more readable at a glance!

PR #2497, opened 2026-02-18
Add GET /ledger/events/counts endpoint for per-event-key counts

ross-weir:
# Description

Introduces an EventCountByKey table that tracks how many times each event key has been emitted. Counts are incremented during slot materialization and decremented on rollback. The new REST endpoint returns a JSON object mapping event keys to their total counts.

Will require a resync to produce accurate values.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-db/src/ledger_db/mod.rs
+++ b/crates/full-node/sov-db/src/ledger_db/mod.rs
@@ -864,7 +864,12 @@ impl LedgerDb {
                 // Decrement EventCountByKey entries for rolled-back events
                 for (key, deleted_count) in &deleted_events_per_key {
                     let current_count = db.get::<EventCountByKey>(key)?.map(|n| n.0).unwrap_or(0);
-                    let new_count = current_count.saturating_sub(*deleted_count);
+                    let new_count = current_count.checked_sub(*deleted_count).ok_or_else(|| {
+                        anyhow::anyhow!(
+                            "Ledger DB corruption during rollback: event count underflow for key {:?} (current_count={}, deleted_count={})",
+                            key, current_count, deleted_count,
+                        )
+                    })?;
                     if new_count == 0 {
                         schema_batch.delete::<EventCountByKey>(key)?;
                     } else {
--- a/crates/full-node/sov-db/src/ledger_db/mod.rs
+++ b/crates/full-node/sov-db/src/ledger_db/mod.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::fmt::Debug;
 use std::sync::{Arc, Mutex, RwLock};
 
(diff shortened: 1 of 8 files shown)


crates/full-node/sov-db/src/ledger_db/mod.rs
```diff
@@ -390,10 +393,26 @@ impl LedgerDb {
         event: &StoredEvent,
         event_number: &EventNumber,
         tx_number: TxNumber,
+        event_key_count: u64,
         schema_batch: &mut SchemaBatch,
     ) -> anyhow::Result<()> {
         schema_batch.put::<EventByNumber>(event_number, event)?;
-        schema_batch.put::<EventByKey>(&(event.key().clone(), tx_number, *event_number), &())
+        schema_batch.put::<EventByKey>(&(event.key().clone(), tx_number, *event_number), &())?;
```
citizen-stig, line 400: Does it handle ledger db rollback? The one that happens during recovery from ungraceful shutdown?

ross-weir: yep: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2497/changes#diff-e0b85c3ad4ead56e441cc1a683023048212b1e21a54607ac128f76eff21a8c7eR865 - unless there's a different code path this can happen on that I'm missing

crates/full-node/sov-db/src/ledger_db/rpc.rs
```diff
@@ -199,6 +199,24 @@ impl LedgerRpcReader {
         }
     }
 
+    async fn get_event_key_counts(&self) -> anyhow::Result<Vec<(String, u64)>> {
+        use sov_rollup_interface::stf::EventKey;
+
+        let range = EventKey::new(&[])..EventKey::new(&[255u8; 4096]);
```
citizen-stig: Comment from devin:

Devin
Range scan upper bound in `get_event_key_counts` silently excludes event keys longer than 4096 bytes

The get_event_key_counts method at uses an artificial upper bound EventKey::new(&[255u8; 4096]) for its range scan. Because EventCountByKey uses define_table_with_seek_key_codec!, keys are encoded via bincode with fixint_encoding + big_endian, meaning a Vec<u8> is serialized as a u64 big-endian length prefix followed by the raw bytes.

The encoded form of the upper bound starts with [0,0,0,0,0,0,16,0] (length 4096 in big-endian u64) followed by 4096 0xFF bytes. Any event key with more than 4096 bytes would have a length prefix larger than 0x1000, causing its encoded form to sort after the upper bound. Since the range is exclusive at the end (..), these keys would be silently excluded from the scan — the endpoint would return incomplete results with no error indication.

Additionally, the exact key [255u8; 4096] itself is excluded by the exclusive range bound.

While event keys in practice tend to be short module-name strings, this is an arbitrary invisible limit that could silently produce wrong results if longer keys are ever used.

ross-weir: Event keys anywhere close to 4096 bytes aren't practical so not worried about it

citizen-stig: Can we at least make a warn or something, because if you're not worried, does not mean that in 6 months I won't be worried when debugging some obscure bug.

Also, why not use open range `..` ? Or first query for max length event and then build range based on that?

ross-weir: updated to query for the max length first

citizen-stig (approved):

crates/full-node/sov-db/src/ledger_db/mod.rs
```diff
@@ -831,6 +861,17 @@ impl LedgerDb {
                     schema_batch.delete::<TxByHash>(&(tx.hash, current_tx_number))?;
                 }
 
+                // Decrement EventCountByKey entries for rolled-back events
+                for (key, deleted_count) in &deleted_events_per_key {
+                    let current_count = db.get::<EventCountByKey>(key)?.map(|n| n.0).unwrap_or(0);
+                    let new_count = current_count.saturating_sub(*deleted_count);
```
citizen-stig: Another comment from devin:

> During rollback at line 867, current_count.saturating_sub(*deleted_count) is used to compute the new event count. If the DB is inconsistent (e.g., deleted_count > current_count due to a bug or corruption), saturating_sub silently clamps to 0 and the entry gets deleted. This prevents a panic but hides a data integrity issue. The rest of the rollback code uses anyhow::bail! for corruption detection (e.g., missing batches/txs/events). Consider whether a count mismatch should also be treated as corruption and produce an error, rather than silently clamping.

In which case we do expect deleted_count to be more than equal? 

I think we can do pattern match with checked sub and probably return an error, because that will indicate some db inconsistency.

Or at least warning.

ross-weir: done, returning a error if underflow

PR #2500, opened 2026-02-18
EVM: `eth_call` / `eth_estimateGas` use correct default nonce for simulation

citizen-stig:
## Summary

Fixes `eth_call` and `eth_estimateGas` to use the caller's current account nonce (instead of 0) when the request omits the nonce field. Also adds upfront cost validation so fee-bearing simulation requests fail early with proper error codes when the caller cannot afford the gas.

## Problem

When `nonce` is omitted in `eth_call`/`eth_estimateGas` requests, `prepare_call_env` defaulted it to 0. Real transaction execution uses the account's current nonce. This divergence causes:

- **CREATE address mismatch**: Contract deployment simulations compute `CREATE` addresses from `(sender, nonce)`. After the first deployment, simulations at nonce=0 produce a different address than the real tx, leading to "cannot estimate gas" errors in wallets/SDKs.
- **Silent behavior divergence**: Any nonce-dependent EVM logic (e.g. deployer factories) behaves differently in simulation vs execution.

Additionally, simulation calls with fee fields (`gasPrice`, `maxFeePerGas`) but insufficient caller balance would enter the EVM and fail with opaque internal errors instead of returning standard `-32003` insufficient funds errors.

### `sov-evm/src/rpc/mod.rs`

**Nonce defaulting** (in `call()`):
- When `request.nonce` is `None` and `request.from` is set, the caller's current EVM account nonce is looked up and injected before building the tx env.

**Upfront cost validation** (new `call_upfront_cost()` helper):
- When fee fields are present (`gas_price` or `max_fee_per_gas`), computes `gas_limit * fee_per_gas + value` and checks caller balance.
- When `gas` is also omitted, caps the gas limit to what the caller can actually afford (instead of using the full block gas limit), with a floor of 21,000 (intrinsic tx cost).
- Validates fee field consistency: rejects conflicting `gasPrice` + `maxFeePerGas`, rejects `maxPriorityFeePerGas > maxFeePerGas`.
- Returns `None` when no fee fields are set (free simulation, existing behavior unchanged).

**Unit tests** (7 tests in `mod tests`):
- `call_upfront_cost_caps_omitted_gas_by_caller_balance`
- `call_upfront_cost_keeps_explicit_gas_requirements`
- `call_upfront_cost_reserves_value_before_affordable_gas`
- `call_upfront_cost_rejects_omitted_gas_when_allowance_below_intrinsic_cost`
- `call_upfront_cost_rejects_conflicting_fee_fields`
- `call_upfront_cost_rejects_tip_above_fee_cap`
- `call_upfront_cost_returns_none_without_fee_fields`

### New integration tests

**`evm_gas_estimation.rs`** (2 new tests):
- `eth_estimate_gas_omitted_nonce_differs_from_explicit_zero_after_nonce_advance` — Deploys a contract (consuming nonce 0), then verifies that `eth_estimateGas` with omitted nonce produces a different result than explicit nonce=0.
- `eth_call_omitted_nonce_matches_explicit_nonce_for_create` — After advancing nonce, verifies `eth_call` with omitted nonce returns the same result as `eth_call` with explicit current nonce.

**`evm_call_fee_fields.rs`** (6 new tests):
- `eth_call_rejects_unfunded_caller_with_gas_price`
- `eth_estimate_gas_rejects_unfunded_caller_with_omitted_gas_and_fee`
- `eth_estimate_gas_rejects_unfunded_caller_with_omitted_gas_and_gas_price`
- `eth_estimate_gas_rejects_missing_from_with_gas_price`
- `eth_estimate_gas_rejects_missing_from_with_omitted_gas_and_gas_price`
- `eth_estimate_gas_rejects_unfunded_caller_with_max_fee_per_gas`

## Key Review Points

1. **Nonce lookup placement**: The nonce is resolved *after* `resolve_state_for_block_id` so it reads from the correct historical state when a block ID is specified.
2. **Affordability-capped gas**: When fee fields are present but `gas` is omitted, the gas limit is capped to `(balance - value) / fee_per_gas` instead of using the block gas limit. This prevents false "insufficient funds" errors for callers who can afford 21k gas but not 30M.
3. **No-fee passthrough**: When neither `gasPrice` nor `maxFeePerGas` is set, the upfront cost check returns `None` and behavior is unchanged (free simulation with block gas limit).

## Testing

- 7 unit tests for `call_upfront_cost` covering all edge cases
- 8 integration tests across `evm_gas_estimation` and `evm_call_fee_fields`


crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -389,17 +458,48 @@ where
...
+                .accounts
+                .get(&from, maybe_archival_state.deref_mut())
+                .unwrap_infallible()
+                .map(|acc| acc.nonce)
+                .unwrap_or_default();
+            request.nonce = Some(account_nonce);
+        }
+
         let mut evm_db: EvmDb<_, S> = self.db(maybe_archival_state.deref_mut());
+        let caller = request.from.unwrap_or_default();
```
bkolad: If `request.from` is None then we will get default address which I assume is `0s` is that right here?

citizen-stig: Just did some research with codex and it confirmed that this is expected behaviour. Adding a comment

bkolad (approved):

PR #2503, opened 2026-02-19
fix: docs for default_uniqueness error behavior

operagxoksana:


crates/web3/src/schema.rs
citizen-stig: cc @theodorebugnet this is not related to this PR, but should we use `try_into()? ` instead of silent conversion?

citizen-stig:
```suggestion
   Ok(UniquenessData::Generation(now.try_into::<u64>()?)
```

crates/web3/src/schema.rs
```diff
@@ -256,15 +256,17 @@ pub const DEFAULT_MAX_FEE: u128 = 100000000;
...
+/// # Errors
 ///
-/// Panics if the system clock is set to a time before the Unix epoch.
+/// Returns [`TransactionBuilderError::TimeError`] if the system clock
+/// is set to a time before the Unix epoch.
 pub fn default_uniqueness() -> Result<UniquenessData, TransactionBuilderError> {
     use std::time::{SystemTime, UNIX_EPOCH};
     let now = SystemTime::now().duration_since(UNIX_EPOCH)?.as_millis();
-    Ok(UniquenessData::Generation(now as u64))
+   Ok(UniquenessData::Generation(now.try_into::<u64>()?)
```
citizen-stig:
```suggestion
    Ok(UniquenessData::Generation(now.try_into::<u64>().expect(
        "system clock is far beyond Unix epoch range; milliseconds since 1970 do not fit in u64",
    )))
```

crates/web3/src/schema.rs
```diff
@@ -256,15 +256,19 @@ pub const DEFAULT_MAX_FEE: u128 = 100000000;
...
 ///
-/// Panics if the system clock is set to a time before the Unix epoch.
+/// Returns [`TransactionBuilderError::TimeError`] if the system clock
+/// is set to a time before the Unix epoch.
 pub fn default_uniqueness() -> Result<UniquenessData, TransactionBuilderError> {
     use std::time::{SystemTime, UNIX_EPOCH};
     let now = SystemTime::now().duration_since(UNIX_EPOCH)?.as_millis();
-    Ok(UniquenessData::Generation(now as u64))
+    Ok(UniquenessData::Generation(now.try_into::<u64>().expect(
+        "system clock is far beyond Unix epoch range; milliseconds since 1970 do not fit in u64",
```
citizen-stig, line 274: With the most recent change, now it also panics, if system clock is foo. 

So documentation needs an update.

operagxoksana: Updated docs to reflect panic behavior

citizen-stig (approved):

PR #2505, opened 2026-02-19
EVM: Add support for block and state override

citizen-stig:
## Summary

Implements `stateOverrides` and `blockOverrides` parameters for `eth_call` and `eth_estimateGas`, which were previously accepted but ignored.

## What changed

**Core logic** (`sov-evm/src/rpc/mod.rs`):
- `apply_state_overrides()` — overrides account code, balance, nonce, and storage (supports both full `state` and partial `state_diff`)
- `apply_block_overrides()` — overrides block number, timestamp, basefee, difficulty, coinbase, gas limit, and block hashes
- `apply_call_overrides()` — orchestrates both; wraps `EvmDb` in `revm::State` only when overrides are present (fast path for the common case)
- Validation: rejects basefee/block number overflow, precompile address overrides, and simultaneous `state` + `state_diff`

**Handler plumbing** (`handlers.rs`, `sov-ethereum/handlers/mod.rs`):
- `eth_call` and `eth_estimate_gas` now accept and forward override params instead of discarding them

**Tests** (`rpc_call_overrides.rs`, +574 lines):
- State overrides: code injection, balance, nonce, storage (`state` vs `state_diff`)
- Block overrides: basefee, number, timestamp, block hash
- Validation: overflow rejection, precompile rejection, conflicting state+state_diff
- Hardfork spec stability under block number overrides

## Review attention

- **`apply_state_overrides`** — storage slot construction uses `EvmStorageSlot::new_changed(!value, value, 0)` (inverted original to force dirty-marking). This matches the alloy-evm reference impl but is worth understanding.
- **`apply_block_overrides`** — strict overflow rejection (returns error) vs alloy-evm's saturation. Intentional choice for stricter validation.
- **Fast path** — when no overrides are provided, the existing `EvmDb` is used directly without wrapping in `revm::State`.

## Linked Issues
- Related to #2271

## Testing
New integration test module with 11 tests covering happy paths and error cases.

### DIFF
--- a/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
+++ b/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
@@ -849,7 +849,7 @@ where
         apply_state_overrides(db, state_overrides)?;
     }
     if let Some(block_overrides) = block_overrides {
-        apply_block_overrides(db, block_env, *block_overrides);
+        apply_block_overrides(db, block_env, *block_overrides)?;
     }
     Ok(())
 }
@@ -858,7 +858,7 @@ fn apply_block_overrides<DB: Database>(
     db: &mut RevmState<DB>,
     block_env: &mut BlockEnv,
     block_overrides: BlockOverrides,
-) {
+) -> Result<(), EthApiError> {
     let BlockOverrides {
         number,
         difficulty,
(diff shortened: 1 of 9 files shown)


crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs
```diff
@@ -391,15 +398,38 @@ where
...
+            let cfg_env = get_cfg_env(&block_env, &cfg, Some(get_cfg_env_template()));
+            let mut evm_db: EvmDb<_, S> = self.db(maybe_archival_state.deref_mut());
+            let result = executor::transact(&mut evm_db, &block_env, tx_env, cfg_env)?;
+            verify_contract_creation_allowlist(&result.state, &caller, &cfg, &mut evm_db)
+                .map_err(|e| EthApiError::other(into_rpc_error(e)))?;
+            return Ok(result);
+        }
+
+        let db: EvmDb<_, S> = self.db(maybe_archival_state.deref_mut());
+        let mut evm_db = RevmState::builder().with_database(db).build();
```
bkolad: Let's call the variable in line 422 evm_db and in 423 state (so it is consistent wit the naming in line 415)

bkolad (approved):

PR #2510, opened 2026-02-20
Hyperlane test flakey pt500

ross-weir:
# Description

Added a time based polling, waiting for slots like we currently do seems to be the cause of the flakey-ness and this can be influenced by a few different things. Using purely timed based avoids the issues and I'm hoping should be more reliable. `DEFAULT_FINALIZATION_BLOCKS * 15` gives us a slot budget that the event must fall into, using time we avoid this constraint.

These changes also make each of the tests run 20 seconds faster on my local machine.

AI written summary:

```
  The inbound EVM→Sovereign transfer step in test_warp_transfer_back_and_forth consistently failed with Warp/TokenTransferReceived event not found when DEFAULT_FINALIZATION_BLOCKS was lowered to 3 and block time was reduced to 250ms.

  The root cause was a two-phase wait pattern:

  1. Phase 1 — wait_for_messages_processed() blocks for up to 60s polling relayer metrics, without consuming from the WebSocket slot subscription.
  2. Phase 2 — A for loop iterates up to DEFAULT_FINALIZATION_BLOCKS * 15 = 45 times, consuming one slot per iteration via next_slot_events().

  During Phase 1, slots continue to be produced (60s / 250ms = ~240 slots) and buffer in the WebSocket stream. When Phase 2 starts, it must drain all those buffered slots before reaching the slot that actually contains the event. With a budget of only 45 iterations, the loop exhausts before ever seeing the target slot.

  Fix

  Replace the two-phase pattern with a single time-based deadline loop (120s) that continuously consumes slots while also periodically checking relayer metrics for critical errors. This eliminates the unconsumed gap entirely — every slot is consumed as it arrives, so the event is found as soon as it appears regardless of how long the relayer
  takes to process the message.

  The other two loops in this test (Loop 1: Warp/RouteRegistered, Loop 3: Warp/TokenTransferredRemote) don't have this problem because they have no wait_for_messages_processed gap before them. Their iteration budgets were bumped from *15 to *30 as a minor safety margin increase.
```

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/hyperlane/tests/integration/with_agent/helpers/mod.rs
+++ b/crates/module-system/hyperlane/tests/integration/with_agent/helpers/mod.rs
@@ -39,7 +39,6 @@ type Container = ContainerAsync<GenericImage>;
 
 pub const FINALIZED_BLOCKS_AT_START: usize = 3;
 pub const DEFAULT_FINALIZATION_BLOCKS: u32 = 3;
-pub const MOCK_DA_BLOCK_TIME_MS: u64 = 250;
 /// Use `container.get_host_port_ipv4(RELAYER_METRICS_PORT)` to get metrics
 pub const RELAYER_METRICS_PORT: u16 = 9091;
 pub const VALIDATOR_METRICS_PORT: u16 = 9097;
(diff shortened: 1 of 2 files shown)


crates/module-system/hyperlane/tests/integration/with_agent/helpers/mod.rs
```diff
@@ -38,7 +38,8 @@ pub type PrivateKey = <<TestSpec as Spec>::CryptoSpec as CryptoSpec>::PrivateKey
 type Container = ContainerAsync<GenericImage>;
 
 pub const FINALIZED_BLOCKS_AT_START: usize = 3;
-pub const DEFAULT_FINALIZATION_BLOCKS: u32 = 10;
+pub const DEFAULT_FINALIZATION_BLOCKS: u32 = 3;
+pub const MOCK_DA_BLOCK_TIME_MS: u64 = 250;
```
citizen-stig: That's might be too much for the rollup. Please use `TEST_DEFAULT_MOCK_DA_BLOCK_TIME_MS` or `TEST_DEFAULT_MOCK_DA_PERIODIC_PRODUCING` from `sov-test-utils`, those are blessed values, controlled centrally

citizen-stig (approved):

citizen-stig (approved):

PR #2512, opened 2026-02-20
Add support for MockDa -> Celestia migrations

preston-evans98:
# Description

This PR exposes functionality in the DB to make migration across DA layers possible. It also adds a migration script to demo-rollup as an example. 

## What Needs to Be Migrated
  - sequencer_registry.known_sequencers keys/values remapped from MockDA -> Celestia addresses.
  - sequencer_registry.preferred_sequencer remapped.
  - chain_state.slots re-encoded for new DA slot hash type.
  - chain_state.genesis_da_height calibrated from cutover inputs.
  - blob_storage.deferred_blobs cleared.
  - paymaster.sequencer_to_payer cleared.
  - paymaster.payers non-AuthorizedSequencers::All policies optionally deleted with force flag.
  - Ledger historical batch receipts rewritten to use the new DA address type

## Migration Instructions

First, stop your rollup and run the migration script. Then, you'll need to update some params in your config before restarting with the Celestia binary. 

 ### Preconditions

  1. Stop the rollup at your chosen cutover rollup height.
  2. Build migration binary with migration-script feature.
  3. Decide and record:

  - celestia_height_at_cutover (the first Celestia block number that will be considered to contain rollup data)
  - sequencer mapping (--new-sequencer-da-address for single sequencer, or --sequencer-map OLD=NEW for multi sequencer)

4. Update your `rollup_config.toml` file with Celestia params
5. Update `constants.toml` with your Celestia namespaces. 
6. Build the new celestia-compatible rollup binary

 ### Initial migration (first instance)

  Use --cutover-rollup-slot-from-db for the first migrated instance if you do not yet know the exact cutover rollup slot number.

Required params for this mode:

  - --rollup-config-path <path>
  - one of:
      - --new-sequencer-da-address <celestia1...> (single sequencer)
      - --sequencer-map <old_mockda_hex>=<new_celestia_bech32> (repeat per sequencer)
  - `--cutover-rollup-slot-from-db`
  - `--celestia-height-at-cutover <u64>`

  Recommended first run:

  - add --dry-run
  - add --report-out <file>

  Example:

```
  cargo run -p sov-demo-rollup --features migration-script --bin mockda-to-celestia-migrate -- \
    --rollup-config-path examples/demo-rollup/configs/mock_rollup_config.toml \
    --new-sequencer-da-address celestia1... \
    --cutover-rollup-slot-from-db \
    --celestia-height-at-cutover 10072492 \
    --dry-run \
    --report-out migration-report.json
```

  After the first successful run, record:

  - `resolved_rollup_slot_at_cutover` from the report.

  ### Subsequent migrations (other instances)

  Do not use --cutover-rollup-slot-from-db.
  Use explicit:

  - `--cutover-rollup-slot <resolved_rollup_slot_at_cutover>`
  - same `--celestia-height-at-cutover` used in first migration

  Example:

```
  cargo run -p sov-demo-rollup --features migration-script --bin mockda-to-celestia-migrate -- \
    --rollup-config-path examples/demo-rollup/configs/mock_rollup_config.toml \
    --new-sequencer-da-address celestia1... \
    --cutover-rollup-slot 48 \
    --celestia-height-at-cutover 10072492 \
    --report-out migration-report.json
```

## Starting Up again

1. Update your *celestia* genesis json to set `chain_state.genesis_da_height` to the appropriate Celestia block number, which you can find in the migration report as `genesis_da_height_after`
2. Delete the blob_sender db: `rm -r <path_to_rollup_data_dir>/blob_sender
3. Start the rollup as normal

## Known Issues
The script as written does *not* attempt to migrate events. If any of your event types contain DA-specific information, those events will not be deserializable after migration which can cause panics in the RPC code. If you have any such events, you'll want to extend the migration script to handle them


examples/demo-rollup/src/main.rs
```diff
@@ -349,8 +349,8 @@ async fn new_rollup_with_mock_da_and_nomt(
             rt_genesis_paths,
             rollup_config,
             prover_config,
-            stop_at_rollup_height,
```
preston-evans98: This was a bug; the correct param ordering is `start_at_height` before `stop_at_height`

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -123,6 +125,62 @@ where
...
+        let ledger_commit = self.commit_ledger(&ledger)?;
+
+        tracing::trace!("Commiting Accessory DB...");
+        let accessory_commit =
+            self.commit_accessory(&accessory, &historical_state.root_hash_batch)?;
+
+        tracing::trace!("Committing Flat DB at latest version...");
+        let flat_metrics = self
+            .flat_state
+            .commit_at_latest_checked(historical_state, expected_latest)?;
```
bkolad: `fn commit_at_latest_checked and fn commit` differ only by the `expected_latest` parameter and how they handle `flat_state`. Whatabout having a single method where `expected_latest` is an Option? If it’s Some, we call `self.flat_state.commit_at_latest_checked(...)`; otherwise we fall back to `self.flat_state.commit(...)`.

bkolad: On second thought, having two separate methods actually makes the intent clearer. I’d introduce a third private method, `fn commit_helper`, that takes the optional parameter. Then both `commit_at_latest_checked` and `commit` could delegate to it.

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
```diff
@@ -0,0 +1,1447 @@
...
+        ledger_change_set.merge(batch_receipt_ledger_patch);
+        storage_manager
+            .commit_migration_change_set_at_head(head_slot_number, change_set, ledger_change_set)
+            .context("failed to commit migration changeset at head version")?;
+    }
+
+    let post_state_root = if args.dry_run {
+        next_state_root
+    } else {
+        let (post_storage, post_ledger_reader) = storage_manager
```
bkolad, line 398: It looks like the code block at line 400 can be moved inside this else block, since we continue operating on storage_manager here.

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
```diff
@@ -0,0 +1,1447 @@
...
+    rollup_slot_at_cutover: Option<u64>,
+
+    /// Resolve the rollup slot at cutover from the DB head slot.
+    /// Mutually exclusive with `--rollup-slot-at-cutover`.
+    #[arg(
+        long = "rollup-slot-at-cutover-from-db",
+        visible_alias = "cutover-rollup-slot-from-db",
+        default_value_t = false
+    )]
+    rollup_slot_at_cutover_from_db: bool,
```
bkolad: Let's remove this parameter from the command arguments and implicitly set it to true when `rollup_slot_at_cutover.is_none()`? That way, it will be impossible to have `rollup_slot_at_cutover: Some(_)` while also setting `rollup_slot_at_cutover_from_db` to true.

This will make some error handling easier, as `failed to resolve rollup slot at cutover; provide --rollup-slot-at-cutover...` will become impossible.

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
```diff
@@ -0,0 +1,1447 @@
...
+            )
+        })?;
+    notes.push(format!(
+        "calibrated chain_state.genesis_da_height to {calibrated_genesis_da_height} \
+         from celestia_height_at_cutover={celestia_height_at_cutover} \
+         and rollup_slot_at_cutover={rollup_slot_at_cutover}"
+    ));
+
+    Ok(CutoverCalibration {
+        resolved_rollup_slot_at_cutover,
```
bkolad: `resolve_cutover_calibration` constructs a `CutoverCalibration` and ensures that none of its fields are None. Given that, it seems the struct:

```
struct CutoverCalibration {
    resolved_rollup_slot_at_cutover: Option<u64>,
    rollup_slot_at_cutover_source: Option<String>,
    calibrated_genesis_da_height: Option<u64>,
}
```
should instead be:
```
struct CutoverCalibration {
    resolved_rollup_slot_at_cutover: u64,
    rollup_slot_at_cutover_source: String,
    calibrated_genesis_da_height: u64,
}
```
This would also simplify the consumers of `CutoverCalibration`, such as `add_genesis_da_height_notes and apply_genesis_da_height_update`, which currently perform redundant checks like verifying whether `cutover.calibrated_genesis_da_height` is None, even though those validations have already been done.

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
```diff
@@ -0,0 +1,1447 @@
...
+    storage: &S,
+    module_discriminants: ModuleDiscriminants,
+    notes: &mut Vec<String>,
+) -> anyhow::Result<BlobStorageInspection> {
+    let deferred_blobs_prefix = Prefix::new(
+        module_discriminants.blob_storage,
+        OldBlobStorage::DEFERRED_BLOBS_ITEM_DISCRIMINANT,
+    );
+    let deferred_blobs_entries = collect_kernel_entries(storage, deferred_blobs_prefix)?;
+    notes.push(format!(
```
bkolad, line 696: Why do we collect `notes`? Isn't it just `tracing::info!/dbug!`?

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
```diff
@@ -0,0 +1,1447 @@
...
+) {
+    match (genesis_da_height_before, calibrated_genesis_da_height) {
+        (Some(before), Some(after)) if before == after => notes.push(format!(
+            "chain_state.genesis_da_height remains unchanged at {after}"
+        )),
+        (Some(before), Some(after)) => notes.push(format!(
+            "chain_state.genesis_da_height will be updated from {before} to {after}"
+        )),
+        (None, Some(after)) => notes.push(format!(
+            "chain_state.genesis_da_height was missing and will be set to {after}"
```
bkolad: We already guarantee that `calibrated_genesis_da_height` is Some, so some of these code paths here are redundant. This method could accept `calibrated_genesis_da_height: u64`.

It also looks like `genesis_da_height_before` should be validated and logged earlier in the flow. If we do that, we could potentially remove this method entirely.

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
```diff
@@ -0,0 +1,1447 @@
...
+                anyhow::anyhow!(
+                "--rollup-slot-at-cutover-from-db was requested, but DB head slot is unavailable"
+            )
+            })?;
+            notes.push(format!(
+                "resolved cutover rollup slot from DB head slot: {resolved_slot}"
+            ));
+            (Some(resolved_slot), Some("db-head".to_string()))
+        } else {
+            (None, None)
```
bkolad: This is unreachable (we check it in line 584)

bkolad (commented):
Question: 
1. If we migrate from mock-da to Celestia and later want to resync the rollup from scratch, would that require two steps? First, resync the mock-da portion and run the script, and then continue the resync from Celestia?

2. It seems like the tools could potentially work between any two DAs if we wrapit in a generic struct (over the 2 specs). I’m not sure whether it’s worth the effort, but it could enable unit testing. For example, we could simulate a mock DA → mock DA migration within a unit test and verify that all the expected data is present.

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
preston-evans98: Good idea. Done

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
preston-evans98: Good catch; fixed

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
preston-evans98: Cleaned this up as well. On closer inspection, it didn't need to be optional

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
preston-evans98: Yep, this is a good call

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
preston-evans98: That was Codex's idea, but it gives you a nice copy-pastable json report that includes a record of how the outputs were generated which I thought was helpful.

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
```diff
@@ -0,0 +1,1406 @@
...
+        db_head_rollup_slot: Some(head_slot_number.get()),
+    })
+}
+
+fn resolve_cutover_calibration(
+    args: &Args,
+    db_head_rollup_slot: Option<u64>,
+    notes: &mut Vec<String>,
+) -> anyhow::Result<CutoverCalibration> {
+    let Some(celestia_height_at_cutover) = args.celestia_height_at_cutover else {
```
bkolad: Looks like celestia_height_at_cutover is always required. Let's make it non optional?

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
```diff
@@ -0,0 +1,1406 @@
...
+        bail!("--celestia-height-at-cutover is required for migration calibration");
+    };
+
+    let (rollup_slot_at_cutover, rollup_slot_at_cutover_source) =
+        if let Some(explicit_slot) = args.rollup_slot_at_cutover {
+            (explicit_slot, "explicit-arg".to_string())
+        } else {
+            let resolved_slot = db_head_rollup_slot.ok_or_else(|| {
+                anyhow::anyhow!(
+                "--rollup-slot-at-cutover-from-db was requested, but DB head slot is unavailable"
```
bkolad: There is no more rollup-slot-at-cutover-from-db  so the error message is outdated.

bkolad (approved):

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
preston-evans98: Fixed!

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
preston-evans98: Updated

PR #2518, opened 2026-02-23
perf(storage): eliminate unnecessary heap allocations in DB key handling

VolodymyrBg:
Previously, every DB read/write operation was allocating a new Vec<u8> via .to_vec() calls when passing keys and values to rockbound APIs. This was caused by API mismatch: SlotKey provides &[u8] via AsRef, but SchemaKey (Vec<u8>) was required by DB methods.
 
This commit:
- Changes get_value_option_by_key to accept &[u8] directly
- Adds generic KeyEncoder/SeekKeyEncoder impls for types with AsRef<[u8]>
- Updates materialize_values to accept generic K: AsRef<[u8]>
- Removes all .to_vec() calls

### DIFF
--- a/crates/full-node/sov-db/src/accessory_db.rs
+++ b/crates/full-node/sov-db/src/accessory_db.rs
@@ -1,3 +1,4 @@
+use std::fmt::Debug;
 use std::sync::Arc;
 
 use rockbound::cache::delta_reader::DeltaReader;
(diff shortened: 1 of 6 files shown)


citizen-stig:
Hey @VolodymyrBg !

This PR looks promising, but CI is failing. We enforce cargo fmt and cargo clippy. There's make mini-ci target that runs the whole thing.

Tagging @preston-evans98 as he originally envisioned this change.

PR #2404, opened 2026-01-26
Celestia: More tests with different blobs layout

citizen-stig:
# Description

This PR hardens native Celestia extraction-proof construction and verification by replacing panic-based proof-building paths with fallible helpers and explicit error handling.

## What changed

- Added `ExtractionProofError` for proof-construction failures such as invalid blob ranges, out-of-bounds access, malformed share metadata, invalid padding, and proof self-check failures.
- Split extraction-proof construction into:
  - `build_extraction_proof(...) -> Result<...>`
  - `get_extraction_proof(...)`, which preserves the existing panic-on-error wrapper for current callers
- Renamed and hardened proof helpers so they return `Result` instead of relying on `panic!`, `expect`, `assert!`, or unchecked indexing:
  - `build_inclusion_proof(...)`
  - `build_ranges_to_prove_for_skipped_blobs(...)`
  - `split_blob_range_by_rows(...)`
  - `NamespaceBoundaryProof::build_from_namespace_data(...)`
  - upfront blob-range validation during proof construction
- Replaced several verifier-side panic paths with checked arithmetic, bounds-safe lookups, and explicit validation errors.
- Preserved the current fail-closed behavior for ambiguous empty namespaces with multiple candidate rows and added coverage that makes that behavior explicit.

## Test coverage

- Reorganized Celestia DA tests into focused modules for:
  - positive verification
  - rejection cases
  - adversarial tampering
  - multi-row absence/spec behavior
  - fixture generation
  - submission/integration flows
- Added property tests covering:
  - multi-row blob layouts
  - mid-row and single-row-mid-row layouts
  - single-row absence and zero-candidate-row absence
  - mixed v0/v1 fixture verification
  - proof-namespace fixture verification
  - proof mutation and tampering cases that must return errors without panicking
- Added shared native test support helpers for synthetic ODS/block construction and proof-index assertions.
- Added the zero-length payload regression test to confirm it still occupies one share.

## Minor cleanup

- Replaced the old `error::*` glob re-export with explicit re-exports.
- Removed a stale commented import.
- Simplified deterministic payload seed packing with `to_le_bytes()`.

----

- [x] ~~I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.~~
[...]

PR #2525, opened 2026-02-24
Celestia: fix verification of namespaces with unsupported blobs and signer-aware share counting

citizen-stig:
Fixes multiple Celestia adapter verification bugs discovered on Mocha/testnet and expands regression coverage. The main observed symptom was false rejection of valid blocks (availability). This PR also adds completeness hardening for boundary checks and ambiguous absence cases.

## Bug 1: Namespaces containing only unsupported blobs could fail verification pathing

When a namespace contains shares but all blobs within it are v0 (unsigned/unsupported version), the extractor skips them and returns an empty blob list. Proof handling could then enter the wrong path, treating this as an absence case rather than proceeding with full verification.

### Fix

- `verifier/mod.rs` (prevalidate_blobs):
    - If blobs is empty and inclusion proofs are present, continue full verification (don't force absence-proof path).
    - Keep fail-closed behavior for ambiguous empty-blob/multi-row-root absence case.
- `verifier/proofs.rs` (new_inclusion_proof):
    - Build skip proofs for unsupported shares when needed.
    - Explicitly error when input is inconsistent (supported shares exist but blob list is empty).

## Bug 2: Share counting ignored signer overhead for v1 signed blobs

Old counting used first-share capacity 478 unconditionally. Signed v1 blobs include a 20-byte signer field, reducing usable first-share capacity from 478 to 458 bytes. Payloads in the 459–478 byte range would fit in a single share unsigned but require two shares when signed, causing miscounts.

### Impact

Near first-share boundary, verifier/proof builder could undercount occupied shares and misalign continuity/boundary logic, causing false rejections.

### Fix

- shares.rs: introduce signer-aware helper:
    - `shares_needed_for_bytes_with_signer`(payload_bytes, has_signer)
- Update signer-aware counting in verifier/proof builder callsites:
    - `verifier/mod.rs`: `authenticate_blob_data`, `verify_skipped_blob` occupancy checks
    - `verifier/proofs.rs`: relevant/full proof range sizing and skipped-range traversal

## Bug 3: Continuity tracking used extraction range end instead of computed blob occupancy

prev_range_end previously tracked extraction range end, which can diverge from computed occupancy under signer-aware logic.

### Fix

- verifier/proofs.rs:
    - Track continuity via computed full_blob_end
    - Keep guard full_blob_end <= range.end as consistency check

## Bug 4: Boundary verification assumed proof always belongs to last candidate row root

Old boundary logic hardcoded last row-root expectations and could reject valid cases where last proven share/proof alignment did not match that assumption.

### Fix

- `verifier/mod.rs` (check_namespace_end_boundary):
    - Derive boundary row using trusted index arithmetic:
        - `delta = last_proven_share_idx - proof_start_in_row`
        - `row_idx = delta / row_len, requiring delta % row_len == 0`
    - Enforce security invariant: boundary proof must terminate namespace in the last candidate row root

## Hardening: ambiguous multi-row absence proof remains fail-closed

Multiple row roots may "contain" a namespace by NMT range semantics without all containing real namespace leaves. A single-row absence proof is not sufficient to prove namespace-wide absence across multiple candidate rows.

### Current decision

- Keep `namespace_row_roots.len() > 1` empty-blob absence case as `MissingBlobs` (fail closed) until stronger absence-proof shape is introduced.

## Diagnostic tracing

Added `tracing::error!` and `tracing::warn!` calls at failure paths in both `verifier/mod.rs` and `verifier/proofs.rs` for production debugging of false rejections.

## Additional internal proof-generation robustness changes

- `da_service/mod.rs`:
    - `get_extraction_proof(...)` now returns `anyhow::Result<...>` internally
    - trait-level `DaService::get_extraction_proof` (non-Result signature) documents and uses fail-fast panic for inconsistent caller input
- `test_helper/serialization.rs` updated for new internal Result return path

### Updated / added

- verification_fails_if_tx_missing (expects MoreProofsThanBlobs)
- extraction_proof_fails_for_empty_blob_list_when_supported_shares_exist
- verification_fails_if_supported_namespace_has_empty_blob_list
- prevalidate_rejects_ambiguous_empty_namespace_without_inclusion_proofs
- prevalidate_continues_when_inclusion_proofs_are_present_even_if_no_blobs
- prevalidate_single_row_requires_boundary_proof
- boundary_proof_for_non_last_row_is_rejected
- bytes_for_shares_accounts_for_signer_overhead
- test_multi_sender_multi_namespace_full_verification_roundtrip

### Test intent note

The multi-sender roundtrip test now uses signer-aware payload sizing assumptions (all active senders are configured with signer keys), so share-boundary coverage corresponds to signed-v1 behavior.

## Known follow-up

This PR focuses on correctness/hardening for the discovered bugs and false-rejection cases. Additional panic-elimination work for all adversarial/untrusted paths is tracked separately
in audit follow-up.

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2525" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


citizen-stig:
### Code review

Found 3 issues:

1. Assertion argument order reversed (`.claude/rules/coding-guidelines.md` says "Place `actual` on the left, `expected` on the right: `assert_eq!(actual, expected)`")

`expected_batch`/`expected_proof` (what was submitted) are on the left; `observed_batch`/`observed_proof` (what was extracted from the blockchain) are on the right. Should be swapped.

https://github.com/Sovereign-Labs/sovereign-sdk/blob/4fa3c02ebd3e25c587b0a848f7b08fb68d605c6a/crates/adapters/celestia/src/da_service/tests.rs#L648-L658

2. Test helper `bytes_for_shares` ignores v1 signer field, weakening boundary coverage

`bytes_for_shares(1)` returns 478 (v0 first-share capacity), but all 4 test senders submit v1 signed blobs where first-share capacity is 458 bytes. A 478-byte v1 blob actually needs 2 shares, not 1. The test intends to test "fits in 1 share" but accidentally tests "fits in 2 shares". 16 of 32 payloads (478 and 479 bytes) fall in the 459-479 byte zone where the signer-aware fix matters. The test still passes because production code counts correctly at runtime, but if the signer-aware fix were broken, this test would not catch it.

Fix: make `bytes_for_shares` accept `has_signer: bool` and subtract `SIGNER_SIZE` (20) from first-share capacity when true.

https://github.com/Sovereign-Labs/sovereign-sdk/blob/4fa3c02ebd3e25c587b0a848f7b08fb68d605c6a/crates/adapters/celestia/src/da_service/tests.rs#L72-L81

3. `check_namespace_end_boundary` function-level doc comment is outdated after rewrite

The doc comment still describes the old logic. It doesn't mention the new row derivation from `last_proven_share_idx` via delta-based arithmetic, the alignment validation (`delta % row_len == 0`), or the security invariant that the boundary proof must terminate in the last candidate row root. These are only in inline comments (lines 532-534).

https://github.com/Sovereign-Labs/sovereign-sdk/blob/4fa3c02ebd3e25c587b0a848f7b08fb68d605c6a/crates/adapters/celestia/src/verifier/mod.rs#L500-L508

Note: the initial concern about a coordinate system mismatch in `check_namespace_end_boundary` was investigated and confirmed to be a false positive. `last_proven_share_idx = R * row_len + C` by construction (via `split_blob_range_by_rows` adding `first_row_offset`), so `delta = (R * row_len + C) - C = R * row_len`, yielding the correct `row_idx = R`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

crates/adapters/celestia/src/verifier/mod.rs
```diff
@@ -275,18 +274,30 @@ fn prevalidate_blobs(
     } else if namespace_row_roots.is_empty() && !blobs.is_empty() {
         return Err(InvalidBlobData(BlobDataError::UnexpectedBlobs));
     } else if blobs.is_empty() && !namespace_row_roots.is_empty() {
-        // We get a list of all row roots that "contain" our namespace (i.e. MIN <= NAMESPACE <= MAX)
-        // it's possible that there's a row whose root "contains" our namespace even though no shares from our namespace are actually present in that row.
-        // For that to be the case, the row needs to contain shares from at least two different namespaces,
-        // one, which is less than our namespace, and one which is greater.
-        // Because shares are ordered by namespace, there can only be at most one such row.
-        // The row before that one will only have shares that are strictly less than our namespace, and the row after will only have shares that are strictly greater.
```
citizen-stig, line 283: It was removed because that comment encoded an assumption that turned out to be false in the real EDS/NMT setting.

Specifically, this line in the old comment was wrong:

- “there can only be at most one such row.”

Why it is wrong:

- prevalidate_blobs filters row roots by root.contains(namespace) (range check: min_ns <= ns <= max_ns), not by “row has an actual leaf in this namespace”.
- In Celestia EDS, parity shares use the max namespace (0xFF...FF), which widens many row-root ranges.
- So multiple rows can “contain” your namespace by range even when they have zero actual shares for it.

That old comment justified single-row absence reasoning. After we discovered the ambiguity/security risk, we changed logic to fail closed for namespace_row_roots.len() > 1 when blobs/proofs are empty (MissingBlobs) in crates/adapters/
celestia/src/verifier/mod.rs:289.

So removal was intentional: the old explanation was misleading and would suggest a weaker security model than current behavior.

crates/adapters/celestia/src/verifier/proofs.rs
```diff
@@ -168,13 +194,20 @@ pub(crate) fn new_inclusion_proof(
...
     for blob in blobs.iter() {
         let range = blob.range_in_namespace.clone();
         let relevant_len = blob.blob.accumulator().len();
-        let relevant_end =
-            range.start + crate::shares::shares_needed_for_bytes(relevant_len).max(1);
-        // Guard in depth - should never be false unless either shares_needed_for_bytes() is bugged
+        let start_share = flat_shares[range.start];
+        let has_signer = start_share.signer().is_some();
+        let relevant_end = range.start
+            + crate::shares::shares_needed_for_bytes_with_signer(relevant_len, has_signer).max(1);
```
citizen-stig: Should we use save math here?

crates/adapters/celestia/src/verifier/mod.rs
```diff
@@ -509,31 +537,51 @@ fn check_namespace_end_boundary(
...
-            }
-            Ordering::Equal => {}
-            Ordering::Greater => {
-                return Err(IncompleteNamespace(
-                    IncompleteNamespaceError::corrupted_proof(),
-                ));
-            }
+        // Determine the row of the boundary proof from the trusted `last_proven_share_idx`.
+        // Security invariant: the boundary proof must terminate the namespace in the last
+        // candidate row root. Otherwise, rows after `row_idx` could still contain this namespace.
```
citizen-stig, line 552: Bring back comment about why it is upsized to u64

crates/adapters/celestia/src/verifier/mod.rs
preston-evans98: Just to ensure I'm understanding correctly, is it the case that lots of rows will contain some real data and some parity shares with the namespace `0xFFF...`. So for example, this would be an unsurprising list of rows:
```
Row 1: 0x02A1... through 0xFFF...
Row 2: 0x0337... through 0xFFF...
Row 3: 0x05BE... through 0xFFF...
```

crates/adapters/celestia/src/verifier/mod.rs
```diff
@@ -472,20 +489,31 @@ fn verify_skipped_blob(
...
+    );
 
     Ok(shares_occupied_total)
 }
 
-// After all blobs have been verified, we need to check that there are no more blobs in the namespace.
-// It does it by explicitly checking proof of the last share. For this proof, the leaf on the right must be from another namespace.
+// After all blobs have been verified, check namespace right boundary for completeness/censorship resistance.
+// This is done with a proof for the last share of the namespace. For a valid boundary,
+// the sibling on the right must belong to a strictly greater namespace.
```
preston-evans98, line 505: Isn't it also the case that the sibling on the right must not be a parity share (namespace `0xFFF`)? Or rather, if it is a parity share we need to check the first share of the next row to decide if completeness is satisfied

crates/adapters/celestia/src/verifier/mod.rs
```diff
@@ -472,20 +489,31 @@ fn verify_skipped_blob(
...
+//
+// Security-critical behavior:
+// * Derive boundary row as:
+//   `delta = last_proven_share_idx - proof_start_in_row`, `row_idx = delta / row_len`.
```
preston-evans98, line 509: this took me some time to parse. Am I correct in understanding that `last_proven_share_idx` is a *global* share index while `proof_start_in_row` is an offset into a row.  (So for example, in a 10x10 square it would make sense that `last_proven_share_idx` is 72 and `proof_start_in_row` is 3? 

Do we need to worry about off by one and/or overflow? It seems to me that this small square might cause problems: 

```
# X represents already proven shares, O represents unproven
X X X X
X X X X
O O O O
O O O O 
```


In this case, `last_proven_share_idx` will be 7 and `proof_start_in_row` will be zero. That will give us a delta of `7`, which gives a `row_idx` of 1. But in fact, the row_idx should be 2. (This would also fail the `delta % row_len == 0` check). What am I missing? Is `last_proven_share_idx` one-indexed while `proof_start_in_row` is zero-indexed, or something?

preston-evans98: Ah, I see. `proof_start_in_row` refers to the last share of the proven range (i.e. share index 7 in this case). Since it's relative to a row, we get `3`.  7 - 3 gives `4`. 

when we divide by `row_len`, that yields `1` - and indeed. Row `1` is the correct one to check

crates/adapters/celestia/src/verifier/proofs.rs
```diff
@@ -196,11 +234,38 @@ pub(crate) fn new_inclusion_proof(
...
+            "Invariant violation: supported shares exist but extracted blob list is empty"
+        );
+        panic!(
+            "supported shares exist in namespace {:?}, but extracted blobs are empty (namespace share count: {})",
+            rollup_data.namespace,
+            end_of_ns
+        );
+    }
+
+    if blobs.is_empty() && end_of_ns > 0 && !namespace_has_supported_shares {
```
preston-evans98: I think we can unify these two blocks: 
```
if blobs.is_empty() && end_of_ns { 

  if namespace_has_supported_shares { panic! }  else {
     let skipped_blob_ranges =
            build_ranges_to_prove_for_skipped_blobs(0..end_of_ns, &flat_shares);
        needed_share_ranges.extend(skipped_blob_ranges);
 }
// ...
```

preston-evans98:
Bug fixes look good to me! But, I think there might be one more edge case we haven't considered. In particular, I think parity shares can throw off the namespace completeness check; when the last_share of the proof is the last non-parity share in its row, we need to check that the next row doesn't contain the namespace. That might already be handled by other checks on the contiguity, but I'm not 100% sure

crates/adapters/celestia/src/verifier/mod.rs
citizen-stig: I've double  checked, and it is not the case.

citizen-stig: The problematic line in this comment was:

> // Because shares are ordered by namespace, there can only be at most one such row.

row with 2 different NS is not the only one. There might be many v0 shares, but there will be no "blobs", because we don't support non-authored blobs and we won't extract them.

crates/adapters/celestia/src/verifier/mod.rs
citizen-stig: Preliminary clarification from Codex:


- Boundary enforces right sibling strictly greater than namespace:
    - `if *namespace >= lrs.min_namespace() -> MissingBlobs`
    - [crates/adapters/celestia/src/verifier/mod.rs:603](https://github.com/Sovereign-Labs/sovereign-sdk/blob/cfe1849e89e83c04d7e675247d3c3932a080129e/crates/adapters/celestia/src/verifier/mod.rs#L603)
- Boundary also enforces proof row must be the last candidate row:
    - if `row_idx != last_row_idx` -> MissingBlobs
    - [crates/adapters/celestia/src/verifier/mod.rs:578](https://github.com/Sovereign-Labs/sovereign-sdk/blob/cfe1849e89e83c04d7e675247d3c3932a080129e/crates/adapters/celestia/src/verifier/mod.rs#L578)

  So if right sibling is parity, that is acceptable by itself, and completeness across rows is handled by the last-candidate-row invariant (not by reading first share of next row).

Supporting source for candidate rows:

- Candidate rows are from DAH row roots via contains(namespace):
    - crates/adapters/celestia/src/celestia.rs:195
- Boundary function explicitly documents this invariant:
    - crates/adapters/celestia/src/verifier/mod.rs:504

----

But I will post my verdict later

citizen-stig:
I just want to add, that I keep working on more comprehensive test suite in this PR: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2404 which contains learning from last couple bugs

citizen-stig:
> Bug fixes look good to me! But, I think there might be one more edge case we haven't considered. In particular, I think parity shares can throw off the namespace completeness check; when the last_share of the proof is the last non-parity share in its row, we need to check that the next row doesn't contain the namespace. That might already be handled by other checks on the contiguity, but I'm not 100% sure

@preston-evans98 
I've added couple more synthetically generated blocks to the suite to cover exactly those cases.

crates/adapters/celestia/src/verifier/mod.rs
citizen-stig: I think main confusion here comes that this function only checks per row. 

And that caller checks if the checked row is the last one.

> if it is truly the last candidate row, then a right sibling of PARITY_SHARE is enough (with the boundary proof) to conclude completeness.

I think this is correct, because we know that it is the last row by checking namespace boundary against DAH

citizen-stig: Codex can phrase it way better:

> So the critical protection is: right-sibling check is only row-local; DAH candidate-row position check gives global completeness.

crates/adapters/celestia/src/shares.rs
```diff
@@ -342,16 +338,37 @@ pub(crate) fn is_tail_padding(share: &celestia_types::Share) -> bool {
...
+/// Converts a target share count into payload bytes, accounting for v1 signer bytes.
+/// `share_count <= 1` maps to the first-share payload capacity.
+#[cfg(test)]
+pub(crate) fn payload_bytes_for_shares_with_signer(share_count: usize, has_signer: bool) -> usize {
+    let first_share_content_size = payload_bytes_in_first_share(has_signer);
+    if share_count <= 1 {
+        return first_share_content_size;
+    }
+    first_share_content_size
+        + (share_count - 1).saturating_mul(appconsts::CONTINUATION_SPARSE_SHARE_CONTENT_SIZE)
```
citizen-stig, line 359: todo: safe math here

crates/adapters/celestia/src/verifier/proofs.rs
```diff
@@ -100,10 +98,9 @@ impl BlobProof {
             }));
         }
 
-        // 3. If the first blob starts not from the beginning of the row (start_idx() == 0),
-        // we need to prove that there are no skipped shares from the namespace.
-        // Otherwise, we safely know that the first column of the related row contains this proof,
-        // so there are blobs that have been skipped.
+        // 3. If the first blob starts after the beginning of the row (`start_idx() > 0`),
+        // prove that the immediate left sibling belongs to a strictly smaller namespace.
```
citizen-stig, line 102: todo: explaine why: so there are no censored blobs

preston-evans98 (approved):

crates/adapters/celestia/src/shares.rs
citizen-stig: Not needed, there's a check above

PR #2526, opened 2026-02-24
Tests(replica): add toxiproxy failover coverage for DB-partitioned replica.

bkolad:
# Description

This PR adds test for the scenario when replica stops receiving data from the sequencer-db for a few blocks.

  - Add a new integration test: `test_replica_catches_up_via_da_after_postgres_partition` that verifies a replica catches up via DA when Postgres connectivity is partitioned.
  - Add `toxi_proxy_helper` test infrastructure using `testcontainers` + Toxiproxy to:
    - proxy Postgres and DA traffic for the replica,
    - toggle Postgres partitioning,
    - inject extreme DA latency.
  


examples/demo-rollup/tests/replica/toxi_proxy_helper.rs
```diff
@@ -0,0 +1,416 @@
+use super::*;
+use serde_json::json;
+use std::net::ToSocketAddrs;
+use testcontainers::core::{Host, IntoContainerPort};
+use testcontainers::runners::AsyncRunner;
+use testcontainers::{ContainerAsync, GenericImage, ImageExt};
+
+const TOXIPROXY_IMAGE: &str = "ghcr.io/shopify/toxiproxy";
```
citizen-stig: we have `crates/utils/sov-test-utils/src/docker.rs` and I think this container will be useful for other crates outside of demo-rollup.

Could we place it there?

examples/demo-rollup/tests/replica/toxi_proxy_helper.rs
```diff
@@ -0,0 +1,416 @@
...
+        assert_eq!(role, expected_role);
+        node
+    }
+
+    /// Starts toxiproxy, creates Postgres/DA proxies, and returns proxied endpoints.
+    async fn start_toxiproxy_for_postgres_and_da(
+        postgres_connection_string: &str,
+        da_upstream_port: u16,
+    ) -> (
+        ContainerAsync<GenericImage>,
```
citizen-stig: Can we extract this into a single type, that has all helper methods

examples/demo-rollup/tests/replica/toxi_proxy_helper.rs
```diff
@@ -0,0 +1,416 @@
...
+
+    addrs
+        .find(|addr| addr.is_ipv4())
+        .or_else(|| addrs.next())
+        .map(|addr| addr.ip())
+        .expect("toxiproxy host resolved to no addresses")
+}
+
+/// Polls toxiproxy's `/version` endpoint until it responds or the timeout is reached.
+async fn wait_for_toxiproxy_ready(client: &reqwest::Client, api_base_url: &str) {
```
citizen-stig: If we define toxiproxy as new struct that implements `testconstainers::Image` we can define readyness conditions easier.

For example see how it is done for Celestia:

https://github.com/Sovereign-Labs/sovereign-sdk/blob/16842c73c92b07d52c1ae7848510dbd96034e0f5/crates/adapters/celestia/src/test_helper/docker.rs#L33

examples/demo-rollup/tests/replica/toxi_proxy_helper.rs
```diff
@@ -0,0 +1,416 @@
...
+        client,
+        format!("{api_base_url}/proxies"),
+        create_proxy_body,
+        "Failed to create toxiproxy postgres proxy",
+    )
+    .await;
+}
+
+/// Creates the toxiproxy listener that forwards DA traffic to the host DA service.
+async fn create_da_proxy(client: &reqwest::Client, api_base_url: &str, da_port: u16) {
```
citizen-stig: This can be method on `SovToxiProxiImage`, and `NodeTestSetup` will only track which ports maps to which service and provide connection strings.

examples/demo-rollup/tests/replica/toxi_proxy_helper.rs
```diff
@@ -0,0 +1,416 @@
...
+async fn post_json(
+    client: &reqwest::Client,
+    url: String,
+    body: serde_json::Value,
+    send_error_context: &str,
+) {
+    let response = client
+        .post(url)
+        .header("content-type", "application/json")
+        .body(body.to_string())
```
citizen-stig: Why it converts it to string? 

I though if anythign supports serde::serialize you can post directly?

citizen-stig: Yes, with `.json()` method

examples/demo-rollup/tests/replica/toxi_proxy_helper.rs
bkolad: Moved `sov_toxi_proxi_image.rs` to `crates/utils/sov-test-utils/src/`

examples/demo-rollup/tests/replica/toxi_proxy_helper.rs
bkolad: Now we have `ToxiProxySetup`

examples/demo-rollup/tests/replica/toxi_proxy_helper.rs
bkolad: Added `SovToxiProxiImage` type

examples/demo-rollup/tests/replica/toxi_proxy_helper.rs
bkolad: fixed

examples/demo-rollup/tests/replica/toxi_proxy_helper.rs
bkolad: Now this is a method on `ToxiProxySetup`

examples/demo-rollup/tests/replica/replica_partitioned_db.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    let collector = LogCollector::new(Level::ERROR);
+    let subscriber = registry().with(collector.clone());
+    subscriber.init();
+
+    // Simulate a partitioned sequencer DB. Now replica is disconnected and the DA is very slow.
+    {
+        setup.set_postgres_partition(true).await;
+        setup.set_replica_da_slow(true).await;
```
citizen-stig, line 70: Why we test 2 conditions at the same time? Why it is not enough to just partition postgres?

citizen-stig: ideally we should have 3 or more tests, with each combination of things

citizen-stig: And even better would be to paramterize test based on toxics for da and postgres, and run all bunch of them. Because there's a bunch of them:

https://github.com/Shopify/toxiproxy?tab=readme-ov-file#toxics

Slicers, upstream, downstream. We can stress the whole system

bkolad: I wanted to simulate a scenario where the replica remains in an outdated state for a couple of blocks. If I don’t enable set_replica_da_slow, the replica will receive the updated state via DA in the very next block. To make the test more comprehensive, I decided to introduce a DA delay so that the replica and the leader stay diverged for more than one block.

citizen-stig: Ok, I see. Can we add it as comment, so it clearer, why we set DA slow?

citizen-stig (approved):
Approved, it is a good first step, I hope we will keep expanding toxiproxy based tests

examples/demo-rollup/tests/replica/replica_partitioned_db.rs
bkolad: fixed

PR #2539, opened 2026-02-26
Fix flaky tests in `blob_sender.rs`

bkolad:
# Description

The tests in `blob_sender.rs` previously used sleep(Duration::from_secs(1)) to wait for async blob operations to complete — a pattern that's inherently flaky. This PR replaces all sleeps with event-driven waiting via a broadcast::Receiver<BlobExecutionStatus>.

### DIFF
--- a/crates/full-node/sov-blob-sender/tests/blob_sender.rs
+++ b/crates/full-node/sov-blob-sender/tests/blob_sender.rs
@@ -357,7 +357,7 @@ async fn blobs_with_seq_nr_too_low_are_not_resubmitted() -> anyhow::Result<()> {
             matches!(
                 status.blob_submission_status,
                 BlobSubmissionStatus::Finalized { .. }
-            ) & matches!(
+            ) && matches!(
                 status.blob_selector_status,
                 Some(BlobSelectorStatus::Discarded(
                     BlobDiscardReason::SequenceNumberTooLow
--- a/crates/full-node/sov-blob-sender/tests/blob_sender.rs
+++ b/crates/full-node/sov-blob-sender/tests/blob_sender.rs
@@ -4,7 +4,7 @@ use std::time::Duration;
 use async_trait::async_trait;
 use sov_blob_sender::{
     BlobExecutionStatus, BlobInternalId, BlobSelectorStatus, BlobSender, BlobSenderHooks,
-    FinalizationManager,
+    BlobSubmissionStatus, FinalizationManager,
 };
 use sov_mock_da::storable::layer::StorableMockDaLayer;
 use sov_mock_da::storable::StorableMockDaService;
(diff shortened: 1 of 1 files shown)


crates/full-node/sov-blob-sender/tests/blob_sender.rs
```diff
@@ -116,19 +117,37 @@ async fn blob_sender_posts_data_to_da() -> anyhow::Result<()> {
...
+
+    wait_for_submission_statuses(
+        &mut status_receiver,
+        |status| {
+            matches!(
+                status.blob_submission_status,
+                BlobSubmissionStatus::Finalized { .. }
+            )
+        },
+        2,
```
citizen-stig: Can we move it to constant, and explain there, why it is only 2. 

Also, is it possible that we receive several "Published" if Blob sender retries?

citizen-stig (approved):

crates/full-node/sov-blob-sender/tests/blob_sender.rs
```diff
@@ -220,13 +240,29 @@ async fn blob_sender_resubmits_blobs_in_progress_after_restart() -> anyhow::Resu
...
+            },
+            2,
+        )
+        .await;
         deps.da.produce_block_now().await?;
-        // We have to wait a littele bit for the async task in blob sender.
-        sleep(Duration::from_secs(1)).await;
-
-        let submissions = blob_sender.nb_of_concurrent_blob_submissions();
-        assert_eq!(submissions, 0);
```
citizen-stig, line 229: Devin
Removed assertion on `nb_of_concurrent_blob_submissions` after finalization in `blob_sender_posts_data_to_da`

The old test at lines 129-131 asserted submissions == 0 after blob finalization, verifying that all in-flight tasks completed. The new version removes this assertion (the old lines are deleted and no equivalent assertion is added after waiting for Finalized statuses). While waiting for 2 Finalized statuses does confirm the blobs reached that state, it doesn't verify the tasks actually cleaned up and decremented the counter. This is a minor reduction in test coverage — the Finalized notification is sent before the task fully exits (see vs 317), so the counter might not yet be decremented when the test ends.

citizen-stig: Is this intentional?

bkolad: Yes, what I really want to test here is that the status changed. using `nb_of_concurrent_blob_submissions` was a way to do this before we had the subscription mechanism.

crates/full-node/sov-blob-sender/tests/blob_sender.rs
bkolad: fixed

PR #2540, opened 2026-02-26
[ZKP] Panic if witness generation is enabled and pin cache is in use

### LINKED ISSUE #2534: [ZKP] Handling pinned cache (probably panic if used, since it breaks the witness)

ross-weir:
# Description

Closes #2534 

- Decouples the consistency check & witness generation flag so they can be enabled independently
- Passes a flag that indicates if witnesses should be generated (currently based on the presence of a prover confg)
- If witness generation is enabled and a pin cache is present then panic in the storage constructor

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
+++ b/crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
@@ -46,7 +46,7 @@ impl<Cache> WitnessMode<Cache> {
     }
 
     /// Returns `true` if witness generation is enabled.
-    pub fn with_witness(&self) -> bool {
+    pub fn is_witness_enabled(&self) -> bool {
         matches!(self, Self::On)
     }
 
--- a/crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
+++ b/crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
@@ -121,7 +121,7 @@ pub struct NomtStorageManager<Da: DaSpec, H, S: InitializableNativeNomtStorage<H
     pruner_max_batch_size: usize,
 
     /// When true, `create_state_for` will generate witness hints for ZK proving.
-    /// This disables pinned cache since it bypasses witness recording. See #2514.
+    /// This disables pinned cache since it bypasses witness recording.
     witness_generation_enabled: bool,
 
     _phantom_s: PhantomData<S>,
@@ -137,7 +137,7 @@ where
     ///
     /// `witness_generation` controls whether `create_state_for` will generate witness hints
     /// for ZK proving. When enabled, pinned cache is disabled since it bypasses witness
-    /// recording. See #2514.
+    /// recording.
     pub fn new(config: RollupDbConfig, witness_generation: bool) -> anyhow::Result<Self> {
         let pruner_block_interval = config.get_pruner_interval();
         let pruner_versions_to_keep = config.get_pruner_versions_to_keep();
@@ -265,7 +265,8 @@ where
         // and we expect a change set from this storage to be saved.
         // That's why it is created in strict mode.
         // If witness generation is enabled, NomtProverStorage::create will panic
-        // if a pinned cache is present — this is intentional, see #2514.
+        // if a pinned cache is present — this is intentional because pinned cache is not
+        // compatible with witness generation. 
         let pinned_cache = self.pinned_caches.remove(&block_header.prev_hash());
         let state = self.create_state_up_to(
             block_header.prev_hash(),
--- a/crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
+++ b/crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
@@ -45,6 +45,20 @@ impl<Cache> WitnessMode<Cache> {
         Self::Off { pinned_cache: None }
     }
 
+    /// Returns `true` if witness generation is enabled.
+    pub fn with_witness(&self) -> bool {
+        matches!(self, Self::On)
+    }
+
+    /// Takes the pinned cache out of the `Off` variant, leaving `None` in its place.
+    /// Returns `None` if witness mode is `On` or no cache is present.
+    pub fn take_pinned_cache(&mut self) -> Option<Cache> {
+        match self {
+            Self::Off { pinned_cache } => pinned_cache.take(),
+            Self::On => None,
+        }
+    }
+
     /// Constructs a [`WitnessMode`] from separate `with_witness` and `pinned_cache` values.
     ///
     /// # Panics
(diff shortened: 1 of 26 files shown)


crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -78,21 +81,28 @@ where
...
         state_session_builder: NomtSessionBuilder<S::Hasher, K>,
         historical_state: HistoricalStateReader,
         accessory: AccessoryDb,
-        strict_with_witness: bool,
+        strict_mode: bool,
+        with_witness: bool,
         pinned_cache: Option<PinnedCache>,
     ) -> Self {
+        assert!(
+            !(with_witness && pinned_cache.is_some()),
```
citizen-stig: Wouldn't be more ergonomic to use single enum with explicit variants described, and derive deeper params from there, rather than having runtime checks? Basically compile time checks vs. runtime checks

ross-weir: added a `WitnessMode` enum

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -436,7 +449,7 @@ where
     ) -> Option<SlotValue> {
         match self.read_value::<N>(key, None) {
             Ok(val) => {
-                if self.strict_with_witness {
+                if self.with_witness {
```
citizen-stig: Comment from Devin about `get_leaf`

```
crates/module-system/sov-state/src/nomt/prover_storage.rs:R431-443

**get_leaf unconditionally writes witness hints, inconsistent with get**

The `get` method at `crates/module-system/sov-state/src/nomt/prover_storage.rs:452` gates witness hint recording with `if self.with_witness`, but `get_leaf` at line 436 always passes `Some(witness)` to `do_get_leaf`, which unconditionally calls `witness.add_hint(...)` at line 274-275. Similarly, `put_in_witness` at line 428 always writes to the witness without checking `with_witness`.

This is pre-existing behavior (predates this PR — the asymmetry existed since commit `5f46bfa7a` which only gated `get` but not `get_leaf`). The impact is minor: on non-prover nodes (`with_witness=false`), `get_leaf` and `put_in_witness` add unnecessary `NodeLeafAndMaybeValue` hints to the witness object. These hints are never consumed because `compute_state_update_namespace` at line 344 is correctly gated by `write_witness`. So it's purely unnecessary memory allocation, not a correctness issue.

Now that `with_witness` is an explicit, well-documented field separate from `strict_mode`, it would be cleaner to gate `get_leaf` and `put_in_witness` the same way `get` is gated.
```

crates/module-system/sov-modules-rollup-blueprint/src/native_only/mod.rs
```diff
@@ -356,7 +361,8 @@ pub trait FullNodeBlueprint<M: ExecutionMode>: RollupBlueprint<M> {
         .await?;
         let current_finalized_header = da_service.get_last_finalized_block_header().await?;
 
-        let mut storage_manager = self.create_storage_manager(&rollup_config)?;
+        let mut storage_manager =
+            self.create_storage_manager(&rollup_config, prover_config.is_some())?;
```
citizen-stig: Is it intended to generate witness even if ` RollupProverConfig` is `Skip` ?

CC @bkolad, this `Option` has always been confusing to me. Why do we need Option<RollupProverConfig> and not to resolve it to `::Skip` where we currently get None?

crates/module-system/sov-state/src/prover_storage.rs
```diff
@@ -378,8 +378,12 @@ impl<S: MerkleProofSpec> Storage for ProverStorage<S> {
         state_accesses: StateAccesses,
         witness: &Self::Witness,
         prev_state_root: Self::Root,
-        _pinned_cache: Option<PinnedCache>,
+        pinned_cache: Option<PinnedCache>,
     ) -> anyhow::Result<(Self::Root, Self::StateUpdate)> {
+        assert!(
```
citizen-stig, line 386: nit: Would result work better here? 

Do you think it is worth adding docstring about panic, for better discoverability

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -38,10 +38,12 @@ where
...
-    /// If set to true, witness will be populated in all necessary places.
-    /// Also, will do consistency check between NOMT and rocksdb in some cases.
-    /// Please check [`NomtProverStorage::should_check_dbs_sync`] for more details.
-    strict_with_witness: bool,
+    /// Enable staleness/consistency checks (NOMT vs rocksdb, root hash comparison).
+    /// Should be true for all node-context block processing.
+    strict_mode: bool,
+    /// Enable witness hint generation for ZK proofs.
+    /// Should only be true for proving nodes.
+    with_witness: bool,
```
citizen-stig: It might be good idea to reuse WitnessMode here what do you think?

crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
```diff
@@ -25,6 +25,46 @@ use crate::storage_manager::nomt_based::groups::{CommitGroup, DbGroup, PrunerJob
...
+    /// Creates a [`WitnessMode::Off`] variant with no pinned cache.
+    pub fn off() -> Self {
+        Self::Off { pinned_cache: None }
+    }
+
+    /// Constructs a [`WitnessMode`] from separate `with_witness` and `pinned_cache` values.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `with_witness` is true and `pinned_cache` is `Some`, since pinned cache
```
citizen-stig, line 66: Where is codepath where the panic is possible? Just wondering if panic can be eliminated completely. 

It is already good that we will catch it earlier

crates/module-system/sov-state/src/nomt/prover_storage.rs
ross-weir: addressed

crates/module-system/sov-modules-rollup-blueprint/src/native_only/mod.rs
ross-weir: disabled for `Skip` now too

crates/module-system/sov-state/src/prover_storage.rs
ross-weir: Added docstring

I like the assert because this should be a hard error & it removes the need for callers to know how to handle the Err variant

crates/module-system/sov-state/src/nomt/prover_storage.rs
ross-weir: done

crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
ross-weir: Mostly to protect from `save_change_set` accumulating pinned caches when witness generation is enabled because it seems like `save_change_set` can be called from multiple places

citizen-stig (approved):

PR #2541, opened 2026-02-27
EVM: exact match of gas fee in receipt to balance change

citizen-stig:
# Description

- Receipt `gas_used * effective_gas_price` now exactly matches the sender's balance delta
- After `EVM_RECEIPT_ACTUAL_FEE_HEIGHT` (in `constants.toml`), receipt `gas_used` is projected from the actual fee charged by the Sovereign gas meter: `projected_gas = fee_paid / uniform_gas_price`
- Non-uniform gas-price dimensions are treated as a hard error (fail-closed)
- RPC-008 compliance test un-ignored
- It was very hard to tighten other tests, because `AccessPattern` module in demo-rollup has post-tx hook, which affects gas. There are several solutions to this: move access pattern tests to TestRollup, move Evm tests to Testrollup. Made access pattern hooks to use unmetered gas. But this will increase scope of this PR. Original commit in this PR verified the change with unmetered tx hook.

## Follow up

Known limitation: receipt-implied gas cost can be lower than final sender balance delta when runtime-level metered work occurs outside the EVM call path (for example, cross-module tx-hook state access). For this PR we relaxed strict equality checks to lower-bound checks (actual_spent >= receipt-implied). Follow-up issue: reconcile receipt fee projection at the full runtime transaction boundary after all module post-dispatch hooks have executed.

Since there are not post tx hooks for production deployments, this will be addressed later.

----

## Linked Issues
- Related to #2271 

## Testing
Ignored tests enabled, other tests have been updated

## Docs
No updates


crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -496,6 +559,47 @@ where
     }
 }
 
+fn derive_receipt_gas_used_from_actual_fee(
```
bkolad: This is sepecial to sov rollups let's extract all the related functionality to a new file for better separation. 
Same for `project_receipt_gas_from_actual_fee`

crates/module-system/module-implementations/sov-evm/src/call.rs
```diff
@@ -496,6 +559,47 @@ where
...
+    ensure!(
+        gas_price_per_dimension
+            .iter()
+            .all(|price| price == uniform_gas_price),
+        "EVM: cannot reconcile receipt from actual fee with non-uniform gas prices: {gas_price_per_dimension:?}"
+    );
+
+    ensure!(
+        tx_fee_paid.0 % uniform_gas_price.0 == 0,
+        "EVM: tx fee {tx_fee_paid} is not divisible by uniform gas price {uniform_gas_price}"
```
bkolad: This will only work under the following conditions:
1. All gas constants in constants.toml have identical values across every dimension.
2. The price is the same in every dimension.

Otherwise, prices across dimensions will start to diverge due to `ChainState::compute_base_fee_per_gas`.

We need to address this in a way that allows different prices across dimensions.
One possible approach is to compute:

`projected_gas_used = roundup(gas * price / price[0])`
The roundup operation can be implemented using integer arithmetic only.

crates/module-system/module-implementations/sov-evm/src/call.rs
citizen-stig: extracted to separate file

crates/module-system/module-implementations/sov-evm/src/call.rs
citizen-stig: Done in https://github.com/Sovereign-Labs/sovereign-sdk/pull/2541/commits/e4cf44cf3da578a70f553c1ac9db8e557d00eb27

bkolad (approved):

PR #2542, opened 2026-03-02
Support separate ledger db path via config

preston-evans98:
# Description

Add a config to control the ledger-db location.

### DIFF
--- a/examples/demo-rollup/configs/mock_rollup_config.toml
+++ b/examples/demo-rollup/configs/mock_rollup_config.toml
@@ -28,6 +28,7 @@ block_time_ms = 1_000
 [storage]
 # The path to the rollup's data directory. Paths that do not begin with `/` are interpreted as relative paths.
 path = "demo_data"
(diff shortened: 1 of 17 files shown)


examples/demo-rollup/configs/mock_rollup_config.toml
```diff
@@ -28,6 +28,7 @@ block_time_ms = 1_000
 [storage]
 # The path to the rollup's data directory. Paths that do not begin with `/` are interpreted as relative paths.
 path = "demo_data"
+# ledger_db_path = "demo_data/ledger-db" # A path override for the ledger db; by default, it is stored at {storage_path}/ledger-db
```
bkolad, line 31: The change lgtm. What about `archival db` in flat_db.rs. Should we allow separate path as well (for now we just hardcode `ARCHIVAL_DB_PATH_SUFFIX`)?

bkolad (approved):

examples/demo-rollup/configs/mock_rollup_config.toml
preston-evans98: Going to leave this for now since it wasn't requested. Will be an easy follow-up if we want it

PR #2547, opened 2026-03-03
EVM tests unflakification

citizen-stig:
# Description

## Why

- We had a race in flaky EVM tests: pause_preferred_batches() sets a flag, but tests could still submit txs before the update-state loop actually observed that pause.

## What

- Added an explicit pause-ack signal from the preferred sequencer update-state loop, reusing the existing state-update test notification stream (no sleep/poll hacks).

### How

- Extended StateUpdateNotification with update_skipped_due_to_pause (test-utils only, serde-defaulted for compatibility):
  `common.rs:265`
- In preferred sequencer update_state_task_inner, when SOV_TEST_PAUSE_SEQUENCER_UPDATE_STATE=1 is observed, it now emits a notification with update_skipped_due_to_pause: true before returning early: `preferred/mod.rs:640`
- Normal state-update notifications now explicitly set update_skipped_due_to_pause: false so tests can reliably distinguish normal updates from pause acknowledgments:
    `sync_state.rs:303`
- Wired the notification sender into PreferredSequencerFields behind test-utils to keep blast radius minimal:
    `preferred/initialization.rs:174`, `preferred/mod.rs:131`

Net effect: deterministic, signal-based pause acknowledgment in tests; no production behavior change.

## Testing
Run tests 50 times on sov-dev-box

## Docs
No updates

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/initialization.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/initialization.rs
@@ -174,6 +174,10 @@ where
         let test_only_state_update_notification_receiver = synchronized_state
             .test_only_state_update_notification_sender
             .subscribe();
+        #[cfg(feature = "test-utils")]
+        let test_only_state_update_notification_sender = synchronized_state
+            .test_only_state_update_notification_sender
+            .clone();
         let synchronized_state_task = synchronized_state.start().await;
         handles.push(synchronized_state_task);
 
@@ -219,6 +223,8 @@ where
             tx_queue_id,
             stop_at_rollup_height,
             test_only_state_update_notification_receiver,
+            #[cfg(feature = "test-utils")]
+            test_only_state_update_notification_sender,
             test_only_forced_tx_batch_notification_receiver: forced_tx_batch_notifier.subscribe(),
             runtime: Rt::default(),
         }));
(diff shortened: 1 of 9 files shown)


crates/full-node/sov-sequencer/src/preferred/mod.rs
```diff
@@ -640,6 +643,15 @@ where
         let skip_flag = std::env::var("SOV_TEST_PAUSE_SEQUENCER_UPDATE_STATE");
         if skip_flag == Ok("1".to_string()) {
             tracing::warn!("skipping state update due to env var flag");
+            #[cfg(feature = "test-utils")]
+            let _ = seq
+                .test_only_state_update_notification_sender
+                .send(StateUpdateNotification {
+                    slot_number: info.slot_number,
+                    finalized_slot_number: info.latest_finalized_slot_number,
+                    #[cfg(feature = "test-utils")]
```
preston-evans98: I think this gate is redundant

preston-evans98 (approved):

crates/full-node/sov-sequencer/src/preferred/initialization.rs
```diff
@@ -174,6 +174,10 @@ where
         let test_only_state_update_notification_receiver = synchronized_state
             .test_only_state_update_notification_sender
             .subscribe();
+        #[cfg(feature = "test-utils")]
```
bkolad, line 177: I think we should either remove this feature gating or put it also above `test_only_state_update_notification_receiver` and inside `synchronized_state`.

PR #2550, opened 2026-03-04
Unflaky sequencer upgradability tests

citizen-stig:
# Description

### What was fixed and why:

1. Deterministic readiness for upgradability tests
    - Switched startup prep to paced block production(tenderly_produce_blocks(finalization_blocks + 2)) + explicit readiness wait.
    - Why: avoid transient “node syncing” windows caused by over-advancing helper behavior during setup.
2. TOCTOU(Time-Of-Check to Time-Of-Use)-safe pre-stop tx loop
    - In sequencer_does_not_accept_tx_after_stop, tx rejection right at boundary is now accepted if it matches expected stop-height error.
    - Why: height check and tx submission are not atomic.
3. Stronger shutdown wait path
    - Added non-panicking shutdown waiter in test utils (try_wait_for_rollup_to_shutdown) and used it with bounded retry loop + DA block driving in test.
    - Why: avoid false timeout panic when shutdown needs extra finalized progression.
4. Race-safe assertions during shutdown
    - Height reads in post-stop loops now tolerate RPC unavailability during shutdown race.
    - Why: CI timing can close node between assertion steps.

---

## Linked Issues
- Related to # (issue) 

## Testing
Run tests 5 times on commit with debug https://github.com/Sovereign-Labs/sovereign-sdk/actions/runs/22714995331/job/65922028245 you can check that all 5 runs passed without any flakiness, with only evm pinned cache failure

## Docs
No updates

### DIFF
--- a/crates/full-node/sov-sequencer/tests/integration/upgradability/mod.rs
+++ b/crates/full-node/sov-sequencer/tests/integration/upgradability/mod.rs
@@ -213,8 +213,9 @@ async fn sequencer_does_not_accept_tx_after_stop(finalization_blocks: u32) {
         slot_subscription.next().await;
     }
 
-    let mut shutdown_timeout_error = None;
-    let shutdown_wait_step = Duration::from_millis(250);
+    // We use "standard" MockDa block time to not overwhelm node.
+    let shutdown_wait_step =
+        Duration::from_millis(sov_test_utils::TEST_DEFAULT_MOCK_DA_BLOCK_TIME_MS);
     let shutdown_deadline = tokio::time::Instant::now() + shutdown_timeout;
 
     loop {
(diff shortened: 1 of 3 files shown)


crates/full-node/sov-sequencer/tests/integration/upgradability/mod.rs
```diff
@@ -208,9 +213,37 @@ async fn sequencer_does_not_accept_tx_after_stop(finalization_blocks: u32) {
...
+                if tokio::time::Instant::now() >= shutdown_deadline {
+                    shutdown_timeout_error = Some(error);
+                    break;
+                }
+                test_rollup.da_service.produce_block_now().await.unwrap();
+            }
+        }
+    }
+
+    if let Some(error) = shutdown_timeout_error {
```
bkolad: Let's  remove `shutdown_timeout_error` variable and just panic in line 234?

crates/full-node/sov-sequencer/tests/integration/upgradability/mod.rs
```diff
@@ -208,9 +213,37 @@ async fn sequencer_does_not_accept_tx_after_stop(finalization_blocks: u32) {
...
     }
 
-    test_rollup
-        .wait_for_rollup_to_shutdown(TEST_NORMAL_SHUTDOWN_TIMEOUT)
-        .await;
+    let mut shutdown_timeout_error = None;
+    let shutdown_wait_step = Duration::from_millis(250);
+    let shutdown_deadline = tokio::time::Instant::now() + shutdown_timeout;
+
+    loop {
```
bkolad, line 221: This is a good idea, what about changing the implemntation to:
What about the following 
```
if `rollup.is_sequencer_ready`
    test_rollup.da_service.produce_block_now().await.unwrap();
else{
test_rollup
        .wait_for_rollup_to_shutdown(..)
}
```

bkolad (approved):

crates/full-node/sov-sequencer/tests/integration/upgradability/mod.rs
citizen-stig: This adds dependency on sequencer ready endpoint. It might not return data we expect. 

I prefer simpler approach, which just produce blocks and call it a day.

Issue #2558, opened 2026-03-05
Fix Sequencer Proof Support


preston-evans98:
Currently, state root assertions are disabled in `test_proof_generation_doesnt_break_sequencer`. To re-enable those assertions, we need to fix the following issues:

  1. Proofs are excluded from replay planning.
     executor_events.rs:289 explicitly ignores ReadBlob::Proof in fetch_completed_blobs_by_sequence.
     This feeds resync/replay decisions via inner.rs:309, sync_state.rs:418, and sync_state.rs:467.
  2. Proof events are no-op during catchup/replay.
     update_state.rs:349 handles DbEvent::ProofBlobAccepted(_) by doing nothing.
     That path is used in replay and final catchup loops from mod.rs update-state flow.
  3. Proof ingestion path still has explicit TODO instead of executor processing.
     sync_state.rs:803 says TODO: Process proof blob in the executor; currently it only publishes side effects.
  4. Replica path for proofs is unimplemented.
     replica/event_handler.rs:64 has DbData::NewProof { .. } => todo!().
  5. Resync condition logic is batch-centric, so proof-only pending work is invisible.
     sync_state.rs:534 uses are_there_batches_to_replay.
     sync_state.rs:509 derives oldest unfinalized from batch list only.

preston-evans98:
Blazej adds another item here: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2559#issuecomment-4011391683

```
This is not directly related to this PR but is relevant for the follow-up.

In replica_sync_task::run_handler, we handle two conditions:

DBDataRejected::ExecutorBehind – the replica executor is behind the DB updates (e.g., on startup when DA notifications are ahead). In this case, we wait until the replica catches up via DA updates.

DBDataRejected::ExecutorAhead – the replica executor is ahead, which can happen if DA is faster than DB notifications (e.g., during networking issues). In this case, the sync task skips DA events until they catch up with the executor.

The assert at replica_sync_task.rs:129 assumes the catch-up boundary event is BatchStart. After enabling proof processing, sequence numbers will also advance on proofs, so this case will need to be handled.
```

Note: This was handled by #2571

preston-evans98:
```
This is pre-existing behavior, but it can became a real issue once we started publishing proofs through the blob sender.

[...]

Issue #2514, opened 2026-02-20
[TRACKER]: Re-enable ZK Proofs


preston-evans98:
## Scope
The initial implementation of proving should focus on creating a minimal production-ready implementation that customers can use to get up and running without fear of catastrophic liveness and/or DOS. 

The simplest path forward at this point is to use sp1 hypercube, wince its 64-bit VM eliminates DOS concerns around memory exhaustion which are likely feasible in 32 bits. Sp1 also provides a plug-and-play prover network, which we should prefer over local proving. Whether or not trace generation is performant enough to run locally is currently an open question. If not, we have two options:
1. Use [skip simulation](https://docs.succinct.xyz/docs/sp1/prover-network/advanced-usage#skip-simulation). (Open question: Does this completely eliminate local execution of the program?)
2. Create a small remote cluster in CDK which splits zkvm trace generation across several machines (this can be done with only the witness). Encapsulate this behind the ProofManager abstraction in the SDK

## Checklist
 - [x] (@ross-weir) Updating sp1 integration to hypercube: https://docs.succinct.xyz/docs/sp1/getting-started/quickstart. 
	 - [x] "The default behavior of `sp1up` now installs a V6-compatible 64 bit toolchain" https://docs.succinct.xyz/docs/sp1/getting-started/migration https://github.com/Sovereign-Labs/sovereign-sdk/pull/2548
	 - [x] Update the prover to use https://docs.succinct.xyz/docs/sp1/prover-network/intro https://github.com/Sovereign-Labs/sovereign-sdk/pull/2603
	 - [ ] Update demo-rollup tests to use sp1
 - [x] Fixing proof support in postgres / replicas (see below) https://github.com/Sovereign-Labs/sovereign-sdk/pull/2571
 - [ ] Fixing proof support in preferred sequencer (@preston-evans98)
	 - [ ] Re-enable end-to-end bank test with proof generation in demo-rollup
         - [x] Fix sequence number bugs #2559 
         - [x] #2558. Fixed by #2571 
         - [ ] Add end-to-end tests including
           - [ ] Interleaved txs and proofs
           - [ ] Resync
           - [ ] Recovery
           - [ ] Replica / failover
 - [ ] Auditing proof incentive flows
	 - [ ] Ensure that funds flow correctly, prover is reimbursed for expenses
	 - [ ] Calibrate gas prices for sp1 hypercube
 - [ ] Add sp1 proof aggregation support (@bkolad)
[...]

PR #2559, opened 2026-03-05
Fix sequence numbering bug. Add test

preston-evans98:
# Description

This PR fixes several issues with sequence numbering when zk proofs are enabled and adds a new test asserting that the sequencer can continue to accept txs and process proofs after the first proof blobs are submitted. 

This PR fixes the "Sequence Number Bug" tracked in the #2514 and improves but does not finish off the correctness of the replicas. 

## Remaining Work

This PR does *not* fix all issues related to proving in the sequencer. In particular...
- Replicas still ignore proof blobs
- Proof blobs aren't actually processed by the rollup block executor, so rewards/penalties will not be taken into account for soft confirmations
- Resync and other conditions that cause replay do not see proof blobs. 

These remaining issues are tracked in #2558.


crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -650,7 +665,9 @@ where
             );
         }
 
-        let sequence_number = self.current_sequence_number();
+        let sequence_number = self
+            .sequence_number_of_open_batch
+            .expect("No batch in progress");
```
bkolad: Let’s add the method name to the message like:`.sequence_number_of_open_batch.expect("No batch in progress {location}")` in all the calls `sequence_number_of_open_batch` so if it panics, we know the exact location.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -952,7 +961,10 @@ where
         reason: &'static str,
     ) -> Result<(), ReplicaError<S>> {
         let mut inner = self.get_inner_with_timing(reason).await;
-        let seq_nr_of_current_blob_for_this_executor = inner.current_sequence_number();
+        let seq_nr_of_current_blob_for_this_executor = inner
+            .sequence_number_of_open_batch
+            .expect("No batch in progress");
+        inner.sequence_number_of_open_batch = None;
```
bkolad, line 967: If `validate_db_data_from_replica` returns early, this method (`process_close_current_batch_replica`) will be retried later, but with `sequence_number_of_open_batch=None`, which will cause a panic at line 966.

Moving `inner.sequence_number_of_open_batch = None;` to after the call to `validate_db_data_from_replica` will fix the issue.

bkolad: Actually, we set it to None in `close_current_batch` which is called after `validate_db_data_from_replica` so we can just remove `inner.sequence_number_of_open_batch = None;` here

crates/full-node/sov-sequencer/src/preferred/replica/event_handler.rs
```diff
@@ -61,7 +61,8 @@ where
                 self.close_current_batch_msg_replica(batch_to_store, "replica_close_batch")
                     .await
             }
-            DbData::NewProof => Ok(()),
+            // TODO(@preston-evans98): Handle new proof blobs.
+            DbData::NewProof { .. } => todo!(),
```
bkolad: Let's return Ok(()) and eventually log an error?

crates/full-node/sov-sequencer/src/preferred/replica/db_data.rs
```diff
@@ -174,7 +176,7 @@ pub(crate) fn row_to_event(row: PgRow) -> Result<(DbData, EventType), ParsingErr
             let batch_to_store = parse_serialized_batch(data, sequence_number)?;
             DbData::BatchEnd(batch_to_store)
         }
-        EventType::NewProof => DbData::NewProof,
+        EventType::NewProof => DbData::NewProof(sequence_number),
```
bkolad: `DbData` is later forwarded to `ReplicaEventHandler::on_db_event`, where DB events are handled.
The `DbData::NewProof` case is still TODO; but eventually we will execute the proof on the replica’s `executor`, so let's inculde also the proof data.

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
```diff
@@ -210,12 +210,17 @@ impl PostgresBackend {
             });
         }
 
-        let completed_blobs_metadata: Vec<(i64, Vec<u8>)> =
-            sqlx::query_as::<Postgres, _>(
-                "SELECT sequence_number, data FROM events WHERE event_type IN ('batch_end', 'new_proof') ORDER BY sequence_number",
-            )
-            .fetch_all(&mut *tx)
-            .await?;
+        let completed_blobs_metadata: Vec<(i64, Vec<u8>)> = sqlx::query_as::<Postgres, _>(
```
bkolad, line 213: We have `test_db_operations_leader` test, where we can add more assertions for this change, or create another test in the same way (it just requires postures, so it has to be run locally).

bkolad:
This is not directly related to this PR but is relevant for the follow-up.

In `replica_sync_task::run_handler`, we handle two conditions:

1. `DBDataRejected::ExecutorBehind` – the replica executor is behind the DB updates (e.g., on startup when DA notifications are ahead). In this case, we wait until the replica catches up via DA updates.

2. `DBDataRejected::ExecutorAhead` – the replica executor is ahead, which can happen if DA is faster than DB notifications (e.g., during networking issues). In this case, the sync task skips DA events until they catch up with the executor.

The assert at `replica_sync_task.rs:129` assumes the catch-up boundary event is BatchStart. After enabling proof processing, sequence numbers will also advance on proofs, so this case will need to be handled.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
preston-evans98: Great catch. Fixed

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
preston-evans98: Done

crates/full-node/sov-sequencer/src/preferred/replica/event_handler.rs
preston-evans98: Hmm, good question. I thought the replica path was unreachable for now, but on further reflection there might be a way to trigger it. To avoid panics for now, this makes sense

preston-evans98: Chatted with Blazej in more depth. Since this would potentially cause desync between replicas and master *and* it should be unreachable (the preferred sequencer should never be configured to produce proofs atm), we're going to leave this as a panic

crates/full-node/sov-sequencer/src/preferred/replica/db_data.rs
preston-evans98: Done

crates/full-node/sov-sequencer/src/preferred/db/postgres/mod.rs
preston-evans98: Added additional cases!

bkolad (approved):

PR #2563, opened 2026-03-06
EVM: RPC compatibility improvmenets: week 10

citizen-stig:
# Description

## 1. Fee-cap-too-low error standardization

What: When a transaction's maxFeePerGas is below the current base fee, the RPC now returns the standard Ethereum error message "max fee per gas less than block base fee" (code -32000) instead of the internal "Insufficient
  max_fee_per_gas" message.
Why: Ethereum clients (e.g. MetaMask, ethers.js) expect this specific error code and message to detect underpriced transactions.
How: FatalError::InsufficientMaxFeePerGas is serialized into the sequencer's error details as a structured fatal_error field. The RPC handler layer deserializes it and maps to RpcInvalidTransactionError::FeeCapTooLow. Both
  authenticate_tx (WebSocket path) and map_accept_tx_error (REST path) are covered.

## 2. Fee validation added to eth_call and eth_createAccessList

What: eth_call and eth_createAccessList now reject requests where the specified fee cap is below the block's base fee.
Why: Geth rejects these calls with FeeCapTooLow; our RPC was silently accepting them.
How: Extracted fee field validation into validate_call_fee_request() and called it in eth_call's handler (handlers.rs) alongside the existing eth_estimateGas path.

## ~~3. Post-Cancun withdrawals compatibility~~

~~What: Block responses now include withdrawals_root: EMPTY_ROOT_HASH and withdrawals: [] instead of null for both.~~
~~Why: Post-Cancun Ethereum blocks always include these fields; returning null breaks clients that parse blocks as Cancun-format.~~
~~How: Set withdrawals_root to EMPTY_ROOT_HASH and withdrawals to an empty list in block construction (sealed, pending, synthetic) and in RLP size calculation (primitive_types.rs). Block sizes increased by 1 byte accordingly.~~

## 4. Default tracer support for debug_traceTransaction

What: When debug_traceTransaction is called without specifying a tracer (or with default options), it now returns a DefaultFrame with struct logs instead of erroring with "unsupported tracer".
Why: Ethereum's default tracing mode (opcode-level struct logs) is widely used by debugging tools.
How: Added a fallback branch in trace.rs that creates a TracingInspector with GethDefaultTracingOptions, executes the transaction, and returns the struct log frame. Respects the limit config option for truncating logs.

## 5. eth_feeHistory with block_count=0 returns empty instead of error

What: eth_feeHistory(0, ...) now returns an empty FeeHistory instead of an invalid-params error.
Why: Geth returns an empty response for zero block count; erroring breaks clients that pass 0.
How: Changed the block_count == 0 branch to return FeeHistory::default().

## 6. New RPC compliance test suite (evm_rpc_compliance_validation_2.rs)

What: Added ~724 lines of new integration tests covering: estimate/send fee admission consistency, affordability consistency, eth_gasPrice base-fee floor, gas cap limits, block header field presence (withdrawals root, gas fields, hash/number), debug_traceTransaction default tracer, and more.
Why: Systematic validation of Ethereum RPC spec compliance for fields and behaviors that were previously untested.

## 7. Test helper refactoring

What: Moved shared utilities (rpc_call, hex_u64, tx_count, raw_signed_eip1559, etc.) from evm_rpc_compliance_validation.rs into evm_test_helper.rs. Renamed number_selector → hex_u64. Made raw_signed_eip1559 more flexible (accepts TxKind, Bytes input, configurable fees).
Why: Reduce duplication across test files and support the new compliance test suite.

## 8. Deleted obsolete test-case docs

What: Removed ~1,263 lines of markdown test case documentation (eth_feeHistory_test_cases.md, eth_getBlockBy_test_cases.md, eth_getLogs_test_cases.md, eth_getTransactionCount_test_cases.md).
Why: These were superseded by the actual integration tests.

---

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2563" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/full-node/sov-sequencer/src/common.rs
```diff
@@ -123,6 +125,55 @@ pub(crate) type SequencerTxStream<Confirmation> = Pin<
     >,
 >;
 
+/// Machine-readable error codes that may appear in `accept_tx` REST error details.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum AcceptTxErrorCode {
```
citizen-stig: Why all this is needed and cannot be embedded properly?

crates/full-node/sov-sequencer/src/common.rs
```diff
@@ -123,6 +125,81 @@ pub(crate) type SequencerTxStream<Confirmation> = Pin<
...
+    /// The transaction's `max_fee_per_gas` is lower than the rollup base fee.
+    InsufficientMaxFeePerGas,
+}
+
+/// Structured details for `accept_tx` REST errors.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct AcceptTxErrorDetails {
```
citizen-stig: This + `AcceptTxErrorCode` adds to much code, why it cannot be part of existing error types?

crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -166,6 +166,12 @@ where
     fn authenticate_tx(tx: &FullyBakedTx, ethereum: &Arc<Ethereum<S, Seq>>) -> RpcResult<()> {
         let mut state = ethereum.api_state_accessor().to_provable_reader();
         let _ = <Seq::Rt as Runtime<S>>::Auth::authenticate(tx, &mut state).map_err(|e| {
+            if matches!(
```
bkolad: Here we use `if matches!` and in line 175 if `let AuthenticationError::FatalError(FatalError::DeserializationFailed(err_msg), _) =&e` convention.

Let's us `if let ` everywhere or just `match e {..}`  so the style is uniform.

examples/demo-rollup/tests/evm/evm_rpc_compliance_validation_2.rs
```diff
@@ -0,0 +1,718 @@
...
+    assert!(
+        reward_value > 0,
+        "reward percentile should reflect non-zero tip payments"
+    );
+
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "Knonw issue. To be discussed and prioritized"]
```
bkolad: Typo: Known

examples/demo-rollup/tests/evm/evm_rpc_compliance_validation_2.rs
```diff
@@ -0,0 +1,718 @@
...
+        rollup.http_addr,
+        "eth_sendRawTransaction",
+        json!([raw_tx]),
+    )
+    .await?;
+
+    assert_eq!(
+        estimate_response.get("error").is_some(),
+        send_response.get("error").is_some(),
+        "estimate and send should agree on affordability classification: estimate={estimate_response}, send={send_response}"
```
bkolad: It looks like `eth_estimateFee` does the following check
error if `balance < gas * user_maxFeePerGas + value`

while `eth_sendRawTransaction` does
error if ` balance < gas_limit * current_gas_price`

Which is different. If you print the erros here you would get:
1. `message": "insufficient funds for gas * price + value: have 100000000000000000 want 1786482426334926933182716689016783110123000`
2. vs `400 Bad Request - 'Transaction execution unsuccessful' ({\"CoreModuleError\": Object {\"error_code\": String(\"generic\"), \"message\": String(\"An error occurred while trying to set the value (key 0b880000000000000000) in the state:
  The amount to charge is greater than the funds available in the meter. Amount to charge 5454, remaining_funds  2052, price GasPrice[9, 9], namespace: User\")}})"`

The test is passing because both responses are errors, but I think the second response suggests that the affordability check passed, the transaction began executing, and was then reverted.

The correct thing would be also to fail `affordability check` also for `eth_sendRawTransaction`?

citizen-stig: Ok, this one is tricky.

Original report from Codex:

```
### Item 2. Affordability mismatch: estimate may reject txs that submission accepts

1. Severity: High
2. Endpoints: `eth_estimateGas`, `eth_call`, `eth_sendRawTransaction`
3. Category: Gas/fee coherence; deterministic preflight false negative
4. Code locations:
   - `crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs:172`
   - `crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs:201`
   - `crates/module-system/module-implementations/sov-evm/src/authenticate.rs:118`
   - `crates/module-system/module-implementations/sov-evm/src/authenticate.rs:132`
   - `crates/full-node/sov-ethereum/src/handlers/mod.rs:147`
5. Why wrong in practical tooling terms:
   - Estimation uses request fee cap for affordability checks, while submission uses rollup pricing path, so estimation can reject a sendable tx.
6. Minimal reproduction (RPC + expected vs likely actual):
   - Choose balance `B` where `gas_limit * rollup_gas_price < B < gas_limit * maxFeePerGas`
   - Run `eth_estimateGas` using high `maxFeePerGas`
   - Submit same raw tx
   - Expected: estimate and send agree
   - Likely actual: estimate fails (`InsufficientFunds`), submission succeeds
7. Concrete impact path:
   - Wallets/SDKs that block send on failed estimation can prevent valid transactions.
8. Confidence: Medium-High
```


But then Devin disagreed: https://app.devin.ai/review/Sovereign-Labs/sovereign-sdk/pull/2563#analysis-ANALYSIS_pr-review-job-665cb09008b24cd1afb76fd89ed5eea9_0001

citizen-stig: I need to get to the bottom of this and do my own research

examples/demo-rollup/tests/evm/evm_rpc_compliance_validation_2.rs
```diff
@@ -0,0 +1,723 @@
...
+        "deployment should succeed: {deploy_response}"
+    );
+
+    let deploy_hash: B256 = rpc_result_hex(&deploy_response).parse()?;
+    let deploy_receipt = ws_client.wait_for_receipt(deploy_hash).await;
+    let contract_address = deploy_receipt
+        .contract_address
+        .expect("deployment should return contract address");
+
+    let no_gas_response = rpc_call(
```
bkolad, line 386: In `prepare_call_env`, we set `let gas_limit = gas.unwrap_or(block_env.gas_limit)`;.

In this case, `block_gas_limit = 100_000_000_000`, while in `capped_response` the `gas_limit  =30_000_000`. So the gasleft for the two txs cases will differ.

The test currently passes only because there is a mistake in the `GAS_LEFT_CONTRACT_DEPLOY_CODE` bytecode. Once it is corrected to `0x6009600c60003960096000f35a60005260206000f3,` the test will fail.

bkolad: More info about the bytecode:
```
 The code copies and returns only 0x08 bytes from offset 0x0c, so the deployed runtime becomes 5a60005260206000, which is missing the final f3 return opcode. That
  means the contract does not actually return gasleft(); it falls off the end and eth_call returns empty data ("0x"). Then evm_test_helper.rs#L149 parses empty hex as 0, for both calls.
```

bkolad: We already have `tx_gas_limit = 30_000_000` field in evm `genesis.json`, so it looks that the fix is to set
```
let gas_limit = gas.unwrap_or(spec.tx_gas_limit)
```
in `prepare_call_env`

examples/demo-rollup/tests/evm/evm_rpc_compliance_validation_2.rs
```diff
@@ -0,0 +1,723 @@
...
+        .ws
+        .request("eth_estimateGas", rpc_params![&tx_request, "latest"])
+        .await?;
+
+    let tx_hash = client.set_value(contract, 0x1337).await;
+    let receipt = client.wait_for_receipt(tx_hash).await;
+
+    let estimate_value = estimate.to::<u64>();
+    assert!(
+        estimate_value <= receipt.gas_used.saturating_add(10_000),
```
bkolad: We should assert the following:
`abs(receipt.gas_used - estimate_value) < 10_000`

For example the current assert passes if `estimate_value=1 and receipt.gas_used = 1000000` 

Let's investigate as a followup (separate PR) how to remove `apply_margins`

bkolad: Here is an AI generated attempt to remove it:

https://github.com/Sovereign-Labs/sovereign-sdk/compare/nikolai/evm-rpc-compat-2026-03-05...blaze/remove_apply_margins?expand=1

It will need more thought and cleanup, but I think the direction is correct.

PR #2564, opened 2026-03-06
feat: time-based uniqueness

vmmon:
# Description

*Summary of the changes...*

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/module-implementations/sov-uniqueness/src/capabilities.rs
+++ b/crates/module-system/module-implementations/sov-uniqueness/src/capabilities.rs
@@ -1,7 +1,6 @@
 use sov_modules_api::capabilities::UniquenessData;
 use sov_modules_api::ExecutionContext;
-use sov_modules_api::{CredentialId, Spec, StateAccessor, StateReader, TxHash};
-use sov_state::User;
+use sov_modules_api::{CredentialId, Spec, TimeStateAccessor, TxHash};
 
 use crate::Uniqueness;
 
@@ -19,7 +18,7 @@ impl<S: Spec> Uniqueness<S> {
         transaction_uniqueness: UniquenessData,
         transaction_hash: TxHash,
         execution_context: &ExecutionContext,
-        state: &mut impl StateReader<User>,
+        state: &mut impl TimeStateAccessor,
     ) -> anyhow::Result<()> {
         match transaction_uniqueness {
             UniquenessData::Nonce(nonce) => match execution_context {
@@ -30,6 +29,9 @@ impl<S: Spec> Uniqueness<S> {
             },
             UniquenessData::Generation(generation) => {
                 self.check_generation_uniqueness(credential_id, generation, transaction_hash, state)
+            }
+            UniquenessData::Timestamp(ts) | UniquenessData::TimestampNonce(ts) => {
+                self.check_timestamp_uniqueness(credential_id, ts, transaction_hash, state)
             }
         }
     }
@@ -43,7 +45,7 @@ impl<S: Spec> Uniqueness<S> {
         credential_id: &CredentialId,
         transaction_generation: UniquenessData,
         transaction_hash: TxHash,
-        state: &mut impl StateAccessor,
+        state: &mut impl TimeStateAccessor,
     ) -> anyhow::Result<()> {
         match transaction_generation {
             UniquenessData::Nonce(_) => self.mark_nonce_tx_attempted(credential_id, state),
@@ -53,6 +55,12 @@ impl<S: Spec> Uniqueness<S> {
                 transaction_hash,
                 state,
             ),
+            UniquenessData::Timestamp(ts) => {
+                self.mark_timestamp_tx_attempted(None, ts, transaction_hash, state)
+            }
+            UniquenessData::TimestampNonce(ts) => {
+                self.mark_timestamp_tx_attempted(Some(credential_id), ts, transaction_hash, state)
+            }
         }
     }
 }
--- a/crates/module-system/module-implementations/sov-uniqueness/src/lib.rs
+++ b/crates/module-system/module-implementations/sov-uniqueness/src/lib.rs
@@ -3,6 +3,7 @@
 mod capabilities;
 mod generations;
 mod nonces;
+mod timestamp;
 use std::collections::{BTreeMap, HashSet};
 
 use sov_modules_api::{
@@ -33,6 +34,14 @@ pub struct Uniqueness<S: Spec> {
     /// Mapping from a credential id to a nonce.
     #[state]
     pub(crate) nonces: StateMap<CredentialId, u64>,
+
+    /// Buckets of transactions with their expiry timestamp. The
+    /// buckets are taken from the first bytes of the TxHash.
+    #[state]
+    pub(crate) timestamps: StateMap<u16, Vec<(TxHash, u64)>>,
+
+    #[module]
+    pub(crate) chain_state: sov_chain_state::ChainState<S>,
 
     #[phantom]
     phantom: std::marker::PhantomData<S>,
--- a/crates/module-system/module-implementations/sov-uniqueness/src/timestamp.rs
+++ b/crates/module-system/module-implementations/sov-uniqueness/src/timestamp.rs
@@ -1,0 +1,63 @@
+use sov_modules_api::{CredentialId, Spec, TimeStateAccessor, TxHash};
+
+use crate::Uniqueness;
+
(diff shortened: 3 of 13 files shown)


crates/module-system/module-implementations/sov-uniqueness/src/capabilities.rs
```diff
@@ -53,6 +55,12 @@ impl<S: Spec> Uniqueness<S> {
                 transaction_hash,
                 state,
             ),
+            UniquenessData::Timestamp(ts) => {
+                self.mark_timestamp_tx_attempted(None, ts, transaction_hash, state)
+            }
+            UniquenessData::TimestampNonce(ts) => {
+                self.mark_timestamp_tx_attempted(Some(credential_id), ts, transaction_hash, state)
+            }
```
preston-evans98, line 63: (non-blocking): flagging that this will be a chain hash breaking change

crates/module-system/module-implementations/sov-uniqueness/src/timestamp.rs
```diff
@@ -0,0 +1,63 @@
...
+        let mut data = self.timestamps.get(&bucket, state)?.unwrap_or_default();
+        data.retain(|(_, ts)| *ts as u128 > current_time);
+        data.push((transaction_hash, expires));
+        Ok(self.timestamps.set(&bucket, &data, state)?)
+    }
+
+    /// Use the first two bytes as bucket value.
+    fn hash_to_bucket(transaction_hash: &TxHash) -> u16 {
+        let ptr: &[u8] = transaction_hash.as_ref();
+        unsafe { core::ptr::read_unaligned(ptr.as_ptr() as *const u16) }
```
preston-evans98, line 61: We have a strong norm of avoiding unsafe unless absolutely necessary. It looks to me like we can just `copy_from_slice` and `from_le_bytes` to get the same behavior but with no unsoundness on big-endian platforms (admittedly, those are very rare!) and no headaches around alignment

crates/module-system/module-implementations/sov-uniqueness/src/timestamp.rs
```diff
@@ -0,0 +1,63 @@
...
+        transaction_hash: TxHash,
+        state: &mut impl TimeStateAccessor,
+    ) -> anyhow::Result<()> {
+        let current_time = self.chain_state.get_oracle_time_nanos(state)? / 1000;
+        anyhow::ensure!(
+            expires as u128 > current_time,
+            "Time-expired transaction for credential_id {credential_id}: hash {transaction_hash:}"
+        );
+
+        anyhow::ensure!((expires as u128) < current_time + 60_000_000,
```
preston-evans98, line 19: nit: let's extract this into a const

crates/module-system/module-implementations/sov-uniqueness/src/timestamp.rs
```diff
@@ -0,0 +1,63 @@
...
+            "Future transaction for credential_id {credential_id}: hash {transaction_hash:} needs to wait."
```
preston-evans98, line 20: nit: it's probably helpful to include the two timestamps that lead to the rejection in this message

crates/module-system/module-implementations/sov-uniqueness/src/timestamp.rs
```diff
@@ -0,0 +1,63 @@
...
+        );
+        anyhow::ensure!(
+            expires > self.nonces.get(credential_id, state)?.unwrap_or_default(),
+            "Nonce-expired transaction for credential_id {credential_id}: hash {transaction_hash:} is invalidated"
```
preston-evans98, line 24: nit: it's probably helpful to include the two timestamps that lead to the rejection in this message

crates/module-system/module-implementations/sov-uniqueness/src/timestamp.rs
```diff
@@ -0,0 +1,63 @@
...
+        expires: u64,
+        transaction_hash: TxHash,
+        state: &mut impl TimeStateAccessor,
+    ) -> anyhow::Result<()> {
+        let bucket = Self::hash_to_bucket(&transaction_hash);
+        let current_time = self.chain_state.get_oracle_time_nanos(state)? / 1000;
+
+        if let Some(credential_id) = credential_id {
+            self.nonces.set(credential_id, &expires, state)?;
+        }
```
preston-evans98, line 50: Anyone who uses this functionality is vulnerable to replay attacks. What's the design goal here?

crates/module-system/module-implementations/sov-uniqueness/src/lib.rs
```diff
@@ -34,6 +35,14 @@ pub struct Uniqueness<S: Spec> {
     #[state]
     pub(crate) nonces: StateMap<CredentialId, u64>,
 
+    /// Buckets of transactions with their expiry timestamp. The
+    /// buckets are taken from the first bytes of the TxHash.
+    #[state]
+    pub(crate) timestamps: StateMap<u16, Vec<(TxHash, u64)>>,
```
preston-evans98, line 41: (non-blocking): Noting for other reviewers that this looks non-breaking since it's the last #[state] item in the module

preston-evans98 (commented):
I think this high level approach is probably workable, but there are some scenarios to think.

### Analysis

- What happens if the timing oracle goes offline (i.e. because the preferred sequencer breaks)? In that case, the buckets will fill without limit until the sequencer comes back online. This will either DOS the rollup (if gas charges for storage are zero or too low) or spike gas usage for honest user txs. Honest users can circumvent a gas spike by switching to an alternate uniqueness strategy, but there is no substitute for good gas pricing to prevent DOS.

- What happens if the timing oracle is malicious? The timing oracle could theoretically pull off three attacks. It could rewind the clock, it could freeze or slow time, or it could jump forward in time. 
  - Rewinding the clock would be devastating, since it would allow replay of all historical txs. This is a new attack vector introduced in this PR. To mitigate it, **_we MUST enforce that the timestamp oracle is monotonic_**. 
  - Freezing time was analyzed above. That would degrade the utility of this uniquness strategy, but honest users could switch to a different strategy with only short term impacts. 
  - The oracle could wind the clock forward aggressively. This would essentially brick this uniqueness vector (all timestamps would be expired) but honest users could switch to a different uniqueness vector.

- What can a malicious user do if the timing oracle behaves correctly? They can fill up buckets with as many txs as possible in 60 seconds by mining collisions (with only 16 bits of entropy, this is trivial). 
-   DOS: Since the compute cost of colliding transactions is quadratic, this would allow them to "hog" compute on the chain if the paymaster was active, especially if rate limits were not in place at the API layer. Sophisticated users can probably circumvent the rate limits with sybils. The only effective defense for this attack is to disallow paymaster transactions above a certain gas cost. 
- Griefing: If users rely on standard gas costs, it may be possible to grief some transactions by filling up the bucket that their deduplicator would fall into. This is not very effective because there are `2^16` buckets. 

### Required Mitigations
This PR introduces a few new attack vectors that need to be carefully mitigated. 

1. The Timing Oracle must enforce that timestamps are monotonic. IMO, that should be done before this PR is merged
2. **_Gas costs for deduplication must be set appropriately._** To mitigate DOS concerns, it's probably best to add an **_extra_** gas charge for deduplication beyond the state access charge. This charge should be super-linear in the size of the bucket and must be large enough to (1) make the cost of malicious transactions larger than the cost of any ordinary transaction AND (2) make sustating a DOS attack expensive in real world terms. That will ensure that any coarse-grained paymaster limits can reject malicious txs and that those paymaster restrictions are effective at preventing the attack
3. The paymaster must enforce limits on the gas cost that it will cover to avoid subsidizing DOS attacks through this vector.

### Conclusion
The tools we have available in the SDK to prevent abuse are very limited. With the new DOS vectors added in this PR, I think it's somewhat unlikely that non-experts will be able to set the gas parameters and paymaster limits correctly in this environment.  

We should think very carefully about whether the tradeoffs here are worth the risk. If we do want to go ahead, this PR adds significant urgency to gas constant calibration.

crates/module-system/module-implementations/sov-uniqueness/src/timestamp.rs
preston-evans98: Oh, I see  this is only intended to increment the nonce. this is probably fine then, but let's add documentation explaining why this is safe

vmmon:
Given the uncertainty with oracle time I prefer a different approach based on slot numbers.  Will do a different PR for that.

PR #2571, opened 2026-03-10
Enable proof processing in sequencer

preston-evans98:
# Description

This PR enables proof processing in the prover.

## Key Changes
The main change is that proofs are now processed by the primary *and* by replicas. They are persisted to postgres and stored in the local BlobCache. Unfortunately, proofs do not fit neatly into the batch model of the existing sequencer. Previously, sequence numbers only changed at open/close batch. Now, a proof can be given a sequence number while a batch is in progress. This means that...
- The `fetch_completed_blobs` function on the blob cache which previously returned a contiguous sorted array can now have a gap where the in-progress batch should go. To simplify handling, I've changed the type from `Vec` to `BTreeMap` to make it robust to duplicate inserts and done some strategic renaming to convey the guarantees more precisely.
- All of the call sites which relied on completed blobs being fetched in order had to be updated

There are two more minor changes required for correctness in this PR. First, Proof Receipts (minus the proof) are now stored in the ledger db by hash. This is required to allow the blob sender to decide whether they've been posted correctly. To enable rollback, we add a secondary index listing hashes by slot number. Second, the `Vec<u8>` types which were passed around previously were internally inconsistent. To fix the issue, I've added new types (`PreferredProofDataBytes` and `SerializedProofWithDetailsBytes`) to distinguish the various layers of serialization.

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2571" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/full-node/sov-sequencer/src/lib.rs
```diff
@@ -74,10 +74,22 @@ pub enum SequencerNotReadyDetails {
     /// The replica is waiting for the first batch from master.
     ReplicaNotReady,
 }
+
+/// A serialized SerializeProofWithDetails<S>. We convert to bytes to avoid pasing a Spec generic
```
bkolad: typo `pasing` -> `passing`

crates/full-node/sov-sequencer/src/preferred/update_state.rs
bkolad: Let's rename it to `fetch_completed_blobs_msg`

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -987,6 +1028,42 @@ where
...
+
+    async fn process_new_proof_replica(
+        &mut self,
+        sequence_number_of_proof: u64,
+        proof_bytes: PreferredProofDataBytes,
+        reason: &'static str,
+    ) -> Result<(), ReplicaError<S>> {
+        let mut inner = self.get_inner_with_timing(reason).await;
+
+        let assigned_sequence_number = inner.take_sequence_number_for_proof();
```
bkolad: `take_sequence_number_for_proof` has the side effect of incrementing `inner.next_unassigned_sequence_number`. If the validation on line 1047 returns early, that side effect is still applied.

Let’s do the following:

Rename `take_sequence_number_for_proof` to `take_and_increase_sequence_number_for_proof` to make the side effect explicit.

Increment `inner.next_unassigned_sequence_number` only after the validation passes.

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -243,15 +248,99 @@ impl BlobsCache {
...
+
+    /// Fetch all proofs that need to be played at the start of the batch with the given sequence number.
+    pub fn proofs_for_replay(
+        &self,
+        batch_sequence_number: SequenceNumber,
+    ) -> Vec<PreferredProofToReplay> {
+        self.sanity_check_batch_sequence_number_is_in_range(batch_sequence_number);
+
+        let mut output = Vec::new();
+        // We have alternating runs of proofs and batches: [proof1, batch1, batch2, proof2, proof3, batch3, ...]
```
bkolad: completed_blobs_out_of_order is unordered, so the following scenario is possible:

`[proof1(7), batch2(6), proof2(8), proof3(9), batch3(10), ...]` <- sequence numbers are shown in parentheses

Currently, `proofs_for_replay` returns `proof2(8) and proof3(9)`. Shouldn’t it also return proof1(7)?

In short:
Do we care only about proofs between two batches in `completed_blobs_out_of_order`, or about all proofs with sequence numbers between `batch2` and `batch3`?

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -243,15 +248,99 @@ impl BlobsCache {
...
+        // Using the above example, if the caller provided batch3, we would need to return [proof2, proof3].
+        for blob in self.completed_blobs_out_of_order.iter() {
+            match blob {
+                ReadBlob::Batch(batch) => {
+                    // If it's a batch, either we've hit our target or we need to reset our output.
+                    if batch.sequence_number == batch_sequence_number {
+                        break;
+                    }
+                    output.retain(|p: &PreferredProofToReplay| {
+                        p.sequence_number > batch_sequence_number
```
bkolad, line 313: Here is the condition that causes https://github.com/Sovereign-Labs/sovereign-sdk/pull/2571/changes#r2930700922

crates/full-node/sov-sequencer/src/preferred/update_state.rs
```diff
@@ -107,14 +113,38 @@ where
...
+            //  - Process batch 7, draining 6 from unprocessed_proofs
+            //  - Push batch 8. Since we don't have any completed batches after the last proof, unprocessed_proofs will have an entry when the loop exits.
+            // Note that some "completed" proofs may have sequence numbers greater than the sequence number of the in-progress batch.
+            // We have to be careful to ignore these proofs for now (they'll get processed when we start the *next*batch)
+            for blob in completed_blobs {
+                match blob {
+                    PreferredBlobToReplay::Batch(batch) => {
+                        batches_count += 1;
+                        transactions_count += batch.batch.inner.data.len();
+                        next_sequence_number = batch.batch.inner.sequence_number.saturating_add(1);
```
bkolad: We only advance `next_sequence_number` for batches, and it looks like this can lead to duplicates in `unprocessed_proofs`.

For example, suppose that on the first iteration `fetch_completed_batches_msg` returns `[batch(6), proof(7)]`. After processing both blobs, `next_sequence_number` is `7`.

If, in the meantime, `batch(8)` is completed and the next fetch returns `[proof(7), batch(8)]`, then `proof(7)` ends up being included twice.

bkolad: I think we should dedup on each `unprocessed_proofs.push`

crates/full-node/sov-sequencer/src/lib.rs
preston-evans98: Fixed!

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
preston-evans98: Good catch! I've moved the side effect after validation.

crates/full-node/sov-sequencer/src/preferred/update_state.rs
preston-evans98: Good thinking. I switched to a BTreeMap to get automatic dedup

preston-evans98: > We only advance next_sequence_number for batches, and it looks like this can lead to duplicates in unprocessed_proofs.

I think this is unrelated and that the existing implementation is correct (but confusing). This `next_sequence_number` value is only used to performa a sanity check that the sequence number of the in-progress batch is greater than the largest sequence number implied by the current full-node state. 

Since it's perfectly normal for there to be *proofs* with sequence numbers larger than the in-progress batch, we don't want to increment the counter on proofs. We only want to surface an error if the node has a *batch* with a larger sequence number than the in-progress batch (that would indicate a broken invariant, because batches are always closed before they get published). 

I've renamed the variable to make this clearer.

crates/full-node/sov-sequencer/src/preferred/db/mod.rs
preston-evans98: Very good catch. We would want proof 7 in this instance. Switching the map to be ordered by sequence number fixes this bug

crates/full-node/sov-sequencer/src/preferred/update_state.rs
preston-evans98: Renamed!

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
preston-evans98: I left the name as is because `take_sequence_number` communicates that it has a side effect (to my ear). Happy to change if there's pushback.

bkolad:
Great change, thx!
Once this PR merged let's unflaky this:
https://github.com/Sovereign-Labs/sovereign-sdk/pull/2584

crates/full-node/sov-blob-sender/src/lib.rs
```diff
@@ -437,7 +437,10 @@ impl FinalizationManager for LedgerDb {
                     blob.slot_number,
                     BlobSelectorStatus::Discarded(blob.discarded_blob.reason),
                 ),
-                None => return Ok(None),
+                None => match self.get_proof_receipt_by_hash(blob_hash).await? {
```
bkolad, line 440: This is pre-existing behavior, but it can became a real issue once we started publishing proofs through the blob sender.

Proofs can sit in the blob sender for longer than batches. A proof is published to DA right away, but it is only marked as processed once batch with higher seq nr was published (https://github.com/Sovereign-Labs/sovereign-sdk/blob/dev/crates/module-system/module-implementations/sov-blob-storage/src/capabilities.rs#L994). So if we have open batch 10 and proof 11, we need to wait for batch with `seq_nr>11` to process the proof.

Because of that, proofs may stay pending for longer than batches normally do.
No need to do anything in this PR but we already discussed to increase `blob_processing_timeout` in the past, this will also fix this scenario.

bkolad (approved):

crates/full-node/sov-blob-sender/src/lib.rs
preston-evans98: Good call out. I've added this to the tracking issue here: https://github.com/Sovereign-Labs/sovereign-sdk/issues/2558

PR #2575, opened 2026-03-11
Remove remaining older rust toolchain version references

citizen-stig:
# Description

Clean up 1.88 - 1.93 migration


ross-weir (approved):
thanks!

Issue #2551, opened 2026-03-05
[TRACKER]: Implement proof aggregation


bkolad:
## Description

The goal of this workstream is to implement a circuit for proof aggregation.  
The work will be divided into two parts:

1. Implement proof aggregation as a standalone circuit.
2. Integrate it into the prover-service workflow.

### 1. Implement proof aggregation as a standalone circuit

To develop the aggregated circuit, we will first generate a set of per-block proofs, store them in files, and then use those proofs as inputs for aggregation (only for the development phase).

- [x] Examine `ProverClient::builder().mock()` usage in SP1 adapter (I think this should be controlled by a separate flag)
- [x] Updated the `SP1Host` and `SP1Verifier` implementations to support SP1 v6.
- [x] Update `test_proof_generation` to work with SP1
- [x] Update `test_proof_generation` to save generated proofs to files along with relevant metadata
- [x] Migrate `crates/full-node/sov-aggregated-proof` to SP1
- [x] Update `sov-aggregated-proof` to aggregate the proofs saved by `test_proof_generation`
- [x] Port the following algorithm to SP1:  
      https://github.com/Sovereign-Labs/sovereign-ibc-types/blob/main/src/aggregated_proof_verifier.rs
- [x] Extract the generic part of the circuit to rollup-interface
- [x] Implement the above algorithm in `sov-aggregated-proof`.

### 2. Integrate proof aggregation into the prover-service workflow
Once the standalone proof aggregation is working in `sov-aggregated-proof`, integrate the circuit in `prover-service`
- [ ] Create and integrate the host responsible for proof aggregation.
- [ ] fix https://github.com/Sovereign-Labs/sovereign-sdk/pull/2715/changes#r3063583536
- [ ] fix https://github.com/Sovereign-Labs/sovereign-sdk/pull/2715/changes#r3063599276
- [ ] The proving service currently runs as an internal process within the full node. Consider separating it into its own node. In that setup, the proving service would act as an HTTP client communicating with an external proving server.
- [ ] TODO Create subissues for the above.
- [ ] Resolve the following issue:  
  https://github.com/Sovereign-Labs/sovereign-sdk-wip/issues/316

## Relevant code from the old repository

- https://github.com/Sovereign-Labs/sovereign-sdk-wip/pull/1986
- https://github.com/Sovereign-Labs/sovereign-ibc-types/blob/main/src/aggregated_proof_verifier.rs
[...]

PR #2591, opened 2026-03-16
`proof-aggregation`: verify DA slot hashes in the proof aggregation circuit

bkolad:
# Description

This PR adds `DA` block header verification to the aggregated proof circuit. For each proof, the guest now checks that `da_block_header.hash() == stf_public_data.slot_hash`, and proof fixtures are stored as BlockHeaderWithProof<Da>.

The circuit still hardcodes `MockDa`; support for arbitrary DA will come in follow-up PRs.
This PR mainly ensures the circuit has the data it needs. The actual constraint will be added later.

## Linked Issues
- Related to #2551

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2591" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/full-node/sov-aggregated-proof/program/Cargo.toml
+++ b/crates/full-node/sov-aggregated-proof/program/Cargo.toml
@@ -5,6 +5,12 @@ edition.workspace = true
 publish = false
 
 [dependencies]
-sha2.workspace = true
-sp1-zkvm.workspace = true
-sov-aggregated-proof-shared.workspace = true
+bincode = { workspace = true }
+sha2 = { workspace = true }
+sp1-zkvm = { workspace = true }
+
+sov-aggregated-proof-shared = { workspace = true }
+sov-mock-da = { workspace = true }
+sov-mock-zkvm = { workspace = true }
+sov-modules-api = { workspace = true }
+sov-sp1-adapter = { workspace = true }
--- a/crates/full-node/sov-aggregated-proof/program/Cargo.toml
+++ b/crates/full-node/sov-aggregated-proof/program/Cargo.toml
@@ -6,6 +6,7 @@ publish = false
 
 [dependencies]
 bincode = { workspace = true }
+demo-stf = { path = "../../../../examples/demo-rollup/stf", default-features = false }
 sha2 = { workspace = true }
 sp1-zkvm = { workspace = true }
 
--- a/crates/full-node/sov-aggregated-proof/program/src/main.rs
+++ b/crates/full-node/sov-aggregated-proof/program/src/main.rs
@@ -2,21 +2,20 @@
 
 sp1_zkvm::entrypoint!(main);
 
+use demo_stf::MultiAddressEvmSolana;
 use sha2::{Digest, Sha256};
 use sov_aggregated_proof_shared::DeferredProofInput;
 use sov_mock_da::MockDaSpec;
 use sov_mock_zkvm::MockZkvm;
+use sov_modules_api::configurable_spec::ConfigurableSpec;
 use sov_modules_api::da::BlockHeaderTrait;
-use sov_modules_api::default_spec::ConfigurableSpec;
 use sov_modules_api::execution_mode::Zk;
 use sov_modules_api::Spec;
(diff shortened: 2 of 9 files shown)


crates/full-node/sov-aggregated-proof/program/src/main.rs
```diff
@@ -4,11 +4,31 @@ sp1_zkvm::entrypoint!(main);
...
 
     for (index, proof_input) in proof_inputs.iter().enumerate() {
+        let stf_public_data: StateTransitionPublicData<
+            <S as Spec>::Address,
+            MockDaSpec,
+            <<S as Spec>::Storage as Storage>::Root,
+        > = bincode::deserialize(proof_input.public_values.as_slice()).unwrap();
+
+        let da_block_header = &proof_input.da_block_header;
+        assert_eq!(da_block_header.hash(), stf_public_data.slot_hash);
```
bkolad, line 31: This is just a sanity check. The  remaining constraints will be added in separate PR (that DA forms a chain etc)

crates/full-node/sov-aggregated-proof/program/Cargo.toml
```diff
@@ -5,6 +5,13 @@ edition.workspace = true
 publish = false
 
 [dependencies]
-sha2.workspace = true
-sp1-zkvm.workspace = true
-sov-aggregated-proof-shared.workspace = true
+bincode = { workspace = true }
+demo-stf = { path = "../../../../examples/demo-rollup/stf", default-features = false }
```
bkolad, line 9: In the next PR, I’ll move `sov-aggregated-proof` to `examples/demo-rollup` as a separate change so it doesn’t get mixed in with the logic changes. Then we will be able to remove this import

citizen-stig: Does it mean, that each runtime has to copy/paste `sov-aggregated-proof` and adjust types?

Why `sov-aggregated-proof` cannot be generic over runtime?

bkolad: The public data for the inner proof depends on `demo_stf::MultiAddressEvmSolana`. At the moment, this is the only dependency we need on the demo-stf runtime.

I think the right approach is:

Create a generic aggregation circuit in `rollup-interface` that works with any Spec, DA/Runtime, and related types.

Add a small wrapper in demo-rollup that just sets the concrete spec and uses the generic implementation.

This would mirror how we already handle `per-block` proofs with `run_block`. That said, it will be easier for me to first implement the circuit using mock-da and then extract the generic part afterward, rather than trying to make it generic from the start.

So after thinking about it a bit more, I’m going to leave the code here for now and split out the generic part once the circuit is ready.

citizen-stig: > Create a generic aggregation circuit in rollup-interface that works with any Spec, DA/Runtime, and related types.
> Add a small wrapper in demo-rollup that just sets the concrete spec and uses the generic implementation.

This sounds reasonable, I hope we won't forget it at the end.

crates/full-node/sov-aggregated-proof/script/src/main.rs
```diff
@@ -139,48 +128,21 @@ fn read_saved_inner_vk(file_path: &Path) -> anyhow::Result<SP1VerifyingKey> {
...
+    let block_header_with_proof: BlockHeaderWithProof<MockDaSpec> =
+        serde_json::from_slice(&file_contents).with_context(|| {
             format!(
-                "Failed to deserialize saved SP1 proof bytes from {}",
+                "Failed to deserialize saved proof JSON at {}",
                 file_path.display()
             )
         })?;
 
-    match proof {
```
citizen-stig, line 164: Why don't we do this check anymore?

citizen-stig (approved):

crates/full-node/sov-aggregated-proof/script/src/main.rs
bkolad: We deserialize it in the circuit now.

PR #2593, opened 2026-03-16
Add DA-related constraints to the aggregated proof circuit.

bkolad:
# Description

1. Passes the DA header to the aggregation circuit.
2. Verifies that consecutive headers form a chain.
3. Verifies the inner proof for each DA block.

## Linked Issues
- Related to #2551

### DIFF
--- a/crates/full-node/sov-aggregated-proof/program/src/main.rs
+++ b/crates/full-node/sov-aggregated-proof/program/src/main.rs
@@ -43,11 +43,11 @@ fn verify<S: Spec, Da: DaSpec>(proof_inputs: Vec<DeferredProofInput<Da>>, vkey_h
         let stf_public_data =
             deserialize_pub_data::<S, Da>(proof_input.public_values.as_slice(), index);
 
+        // Check that DA blocks form a chain.
         {
             let da_block_header = &proof_input.da_block_header;
             let current_block_hash = da_block_header.hash();
 
-            // Check that DA blocks form a chain.
             if let Some(expected_prev_hash) = &expected_prev_hash {
                 assert_eq!(
                     expected_prev_hash,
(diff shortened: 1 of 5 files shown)


crates/full-node/sov-aggregated-proof/program/src/main.rs
```diff
@@ -4,35 +4,88 @@ sp1_zkvm::entrypoint!(main);
...
-        assert_eq!(da_block_header.hash(), stf_public_data.slot_hash);
-
-        println!("[guest] verifying proof {index}");
-        let public_values_digest: [u8; 32] = Sha256::digest(&proof_input.public_values).into();
-        sp1_zkvm::lib::verify::verify_sp1_proof(&proof_input.vkey_hash, &public_values_digest);
-        println!("[guest] verified proof {index}");
+        let stf_public_data =
+            deserialize_pub_data::<S, Da>(proof_input.public_values.as_slice(), index);
+
+        {
```
bkolad, line 47: If `jump = 10`, the check guarantees consistency only between 10 da blocks. To guarantee consistency across the entire chain back to genesis, we will need to recursively verify the previous aggregated proof. That will be done in the next PR.

crates/full-node/sov-aggregated-proof/program/src/main.rs
```diff
@@ -4,35 +4,88 @@ sp1_zkvm::entrypoint!(main);
...
+}
+
+fn verify<S: Spec, Da: DaSpec>(proof_inputs: Vec<DeferredProofInput<Da>>, vkey_hash: [u32; 8]) {
+    assert!(
+        !proof_inputs.is_empty(),
+        "Aggregated proof must contain at least one proof input"
+    );
+
+    // `None` means no predecessor to check against (first iteration).
+    let mut expected_prev_hash = None;
```
citizen-stig, line 39: If we don't check that height is increased, does it mean that it is possible to verify transitions that are going backwards?

citizen-stig (approved):

crates/full-node/sov-aggregated-proof/program/src/main.rs
bkolad: For two reasons no:
1. We explicity call da_block_header.prev_hash() so it encodes the infromation that we are moving forward
2. In the next PR we will encode the information that the whole chain has to start from genesis.

PR #2595, opened 2026-03-17
Verify outer aggregated proofs recursively.

bkolad:
# Description

1. Add support for supplying a previous outer aggregated proof to the aggregation circuit, so the circuit can verify it as part of recursive chaining. This change is limited to wiring the necessary witness data into the circuit.
2. Introduce a PreviousOuterProofWitness struct containing the previous outer proof’s public values and verification key hash.

The remaining pieces—adding the relevant constraints, creating aggregated public data, and linking the previous aggregate to the current batches—will be implemented in a separate PR focused solely on constraints.

## Linked Issues
- Related to #2551


crates/full-node/sov-aggregated-proof/program/src/main.rs
```diff
@@ -25,8 +27,16 @@ pub fn main() {
     let witness = sp1_zkvm::io::read::<AggregatedProofWitness<MockDaSpec>>();
     let proof_inputs = witness.proof_inputs;
     let vkey_hash = witness.vkey_hash;
+    let prev_outer_proof_witness = witness.prev_outer_proof_witness;
 
     verify::<S, MockDaSpec>(proof_inputs, vkey_hash);
+
+    if let Some(prev_outer_proof_witness) = prev_outer_proof_witness {
+        verify_sp1_proof(
```
bkolad, line 35: At this stage, we do not yet enforce any relationship between the previous proof’s public values and the individual proofs we just verified. That work will be handled in a separate PR. We want to avoid mixing the mechanical code that simply submits data into the circuit with the constraints PR, since the latter will require much more careful review. Keeping that feature PR as small as possible should make it easier to review.

citizen-stig (approved):

PR #2599, opened 2026-03-17
EVM: Paymaster-aware affordance checks

citizen-stig:
# Description

This PR implements paymaster-aware affordability checks in the sov-ethereum RPC layer for `eth_estimateGas`, `eth_sendRawTransaction`, and `eth_sendTransaction`.

Previously, balance-based upfront cost validation was removed from the EVM module because the RPC layer couldn't determine whether a paymaster would cover the caller's gas (see the removed TODO in [`sov-evm/src/rpc/mod.rs`](https://github.com/Sovereign-Labs/sovereign-sdk/blob/f6038895ee21b253716f9a7d73a7ccd6c32224fb/crates/module-system/module-implementations/sov-evm/src/rpc/mod.rs#L151). This PR resolves that gap by lifting affordability logic into sov-ethereum's handler layer, where full runtime context (including paymaster resolution) is available.

## Core changes

1. `eth_estimateGas` moved from `sov-evm` to `sov-ethereum`
  - The #[rpc_method]-annotated eth_estimate_gas was removed from `sov-evm` and re-registered manually in `sov-ethereum` as `Handlers::eth_estimate_gas`.
  - The underlying gas estimation logic remains in `sov-evm` as a plain `pub fn eth_estimate_gas(...)` (no longer auto-registered via `#[rpc_gen]`).
  - A new validate_estimate_gas_request() method in `sov-evm` handles request-shape and base-fee validation independently of affordability.

2. Paymaster-aware affordability preflight
  - `run_affordability_preflight()` resolves the runtime's `TransactionAuthorizer` to determine the actual gas payer via `try_reserve_gas()`. If a paymaster covers gas, only the transfer value is checked against the sender's balance. Otherwise the full gas_cost + value is checked.
  - For `eth_estimateGas`: when gas is omitted, the EVM estimate is obtained first, then affordability is checked with the estimated gas. When gas is provided, a stale-nonce preflight runs first, then affordability.
  - For raw/local sends: `authenticate_and_preflight_send()` chains authentication → uniqueness (nonce) preflight → affordability preflight, all against a single pinned state snapshot.

3. Request preflight auth
  - New build_request_preflight_auth() function synthesizes `AuthenticatedTransactionData` + `AuthorizationData` from a `TransactionRequest` (no signed transaction needed). Used by `eth_estimateGas` to run the same gas-reservation logic as the send path.
  - Refactored `create_auth_tx_and_hash()` to extract `build_authenticated_tx_data()` — shared by both signed-tx and request-preflight paths.
  - saturating_mul for gas limit replaced with checked_mul to surface overflow errors.

4. Wiring: `EthRpcConfig` now carries sequencer context
  - `EthRpcConfig` gains `sequencer_rollup_address`, `sequencer_da_address`, and `sequencer_type fields` — needed by the authorizer to resolve paymaster context.
  - All 6 rollup blueprints updated to pass these fields through.
  - `FullNodeBlueprint::register_rpc` signature gains `sequencer_da_address` parameter.

5. Stale-nonce preflight
  - validate_request_stale_nonce_preflight() and validate_send_uniqueness_preflight() reject transactions with nonces below the account's current nonce before running affordability checks, ensuring nonce errors take precedence over
  balance errors.

## Tests

- Unit tests in handlers/mod.rs: supports_request_affordability_preflight with/without overrides.
- Integration test in sov-evm: request_preflight.rs — verifies build_request_preflight_auth uses current nonce when omitted and preserves explicit nonce.
- E2E RPC compliance tests:
  - rpc2_002 (previously #[ignore]): un-ignored — affordability consistency between estimate and send now works.
  - rpc2_002b (new): stale nonce must win over affordability rejection on all three paths.
  - rpc2_015 (previously #[ignore]): un-ignored — paymaster-aware estimate/send affordability consistency.
  - Existing fee-history and effective-gas-price tests updated to use estimate_gas_and_check_affordability helper to stay within sender balance.
  
### Why these test changes are correct

These changes do not relax coverage. They remove assertions that were targeting the wrong mechanism and replace them with assertions on the behavior the code actually guarantees today in examples/demo-rollup/tests/evm/evm_rpc_compliance_validation_2.rs.

**RPC2-002**
- What was wrong with the original test: it funded the sender above the real affordability floor and then tried to force an insufficient-funds error by using an unrealistic `HIGH_MAX_FEE_PER_GAS`. That is not what the implementation checks. The real preflight uses actual affordability (gas_limit * rollup_gas_price), not the advertised EIP-1559 ceiling.
- Why it had been passing before: before the current affordability preflight landed, the unrealistic fee ceiling was not being rejected in the same early path, so the test setup was not exposed as invalid. Once estimate/send started sharing the real affordability gate, the test failed for the wrong reason.
- Why this change is done this way: the sender is now funded with a balance strictly below gas_limit * rollup_gas_price, `eth_estimateGas` omits gas, and the tx uses a normal fee cap. That isolates the affordability condition the code really enforces.
- What it achieves now: the test proves that `eth_estimateGas` and `eth_sendRawTransaction` both reject for the same genuine insufficient-funds reason. It will fail again if estimate/send diverge or if affordability regresses back to using the wrong metric.

**RPC2-002c**

- What was wrong with the original test: removing it lost useful wrapper-level coverage. Restoring it literally was also not sufficient, because it assumed a lone future-nonce `eth_sendRawTransaction` should immediately succeed and used a brittle transfer payload. In the current preferred-sequencer RPC path, send is synchronous and waits for queue resolution; a future tx only succeeds if the missing nonce arrives before timeout.
- Why the old idea had seemed valid before: the underlying feature is real and still passes in the sequencer integration test, but that test submits out-of-order nonces in the shape the queue expects. The old wrapper test was asserting the right feature with the wrong RPC timing model.
- Why this change is done this way: the test now uses a stable `SimpleStorage` call, estimates gas once, submits the future-nonce raw tx in the background, then submits the missing nonce tx while the first request is still queued. It also checks final state, not just receipt presence.
- What it achieves now: it verifies the real end-to-end contract at the RPC layer: a future raw tx is accepted into the reorder path and executes successfully once the gap is filled before timeout. It still catches queue, ordering, and execution regressions.

**RPC2-002d**

- What was wrong with the original test: same issue as RPC2-002c, but on eth_sendTransaction. The literal restored version assumed immediate success for a lone future nonce and used a payload shape that was not a stable success case in this harness.
- Why the old idea had seemed valid before: the preferred sequencer does support future nonces, but only as queued transactions that later drain in order. The previous wrapper test shape no longer matched that synchronous RPC behavior.
- Why this change is done this way: it mirrors the raw-send fix exactly, but through eth_sendTransaction: stable contract call, estimated gas, future nonce submitted first and kept in-flight, current nonce submitted second, then both receipts and final storage are asserted.
- What it achieves now: it restores wrapper-level coverage for future-nonce local sends without hiding bugs behind timing assumptions or unrelated gas/payload issues.

**RPC2-015**

- What was wrong with the original test: phase 2 expected paymaster-backed estimate/send to fail whenever gas_limit * HIGH_MAX_FEE_PER_GAS exceeded payer balance. That is the wrong invariant. Paymaster-backed affordability is about actual gas cost and paymaster coverage, not the declared fee ceiling by itself.
- Why it had been passing before: that expectation only matches an older, incorrect interpretation where the advertised ceiling was effectively treated as the spend amount. After the paymaster-aware preflight landed, that became a false-failure test.
- Why this change is done this way: phase 1 uses a reasonable fee cap to prove the baseline zero-balance sender success path. Phase 2 keeps the very high declared fee, but now expects success and checks nonce advancement, because the paymaster can still cover the real gas cost.
- What it achieves now: the test guards the intended paymaster behavior: estimate/send must not falsely reject a paymaster-covered transaction just because the declared maxFeePerGas is large. If the code regresses to ceiling-based false rejection, this test will fail immediately.

## Key design decisions

- State snapshot pinning: each RPC handler pins a single checkpoint snapshot so validation, authentication, and affordability cannot observe different heads.
- Best-effort preflight: if paymaster context resolution fails (e.g. missing config), the preflight is skipped rather than erroring — sequencer admission still enforces the real checks.
- Overrides skip preflight: when state/block overrides are present, account balances are arbitrary, so real-state affordability is skipped.

## Follow up

1. Make tx_hash optional in `AuthenticationError::FatalError` variant, so sentinel hash from this PR can be omitted for simulation endpoints.
2. Cleanup: `EthRpcConfig` now carries three related pieces of sequencer identity, `sequencer_rollup_address`, `sequencer_da_address`, and `sequencer_type`, because the Ethereum RPC wrapper needs the same context as sequencer admission when running estimate/send preflights. A good follow-up would be to group these into a single `SequencerIdentity` struct so the intent is explicit and the plumbing across the blueprint and demo rollups is less repetitive, without changing behavior

----

## Linked Issues
- Related to #2271  

## Testing
All exsting and new tests are passing

## Docs
Docstrings as usual

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2599" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -134,6 +165,456 @@ where
...
+        request: TransactionRequest,
+        block_id: Option<BlockId>,
+        state_overrides: Option<StateOverride>,
+        block_overrides: Option<Box<BlockOverrides>>,
+        ethereum: &Arc<Ethereum<S, Seq>>,
+    ) -> RpcResult<U64> {
+        // Pin one checkpoint snapshot for the full request so validation,
+        // affordability preflight, and estimation cannot observe different heads.
+        let snapshot_state = ethereum.api_state_accessor();
+        Self::estimate_gas_request_with_snapshot(
```
bkolad: We have `estimate_gas_request` and `estimate_gas_request_with_snapshot`. `estimate_gas_request` adds only one line to `estimate_gas_request_with_snapshot` but it makes it more dificult to trace the code as this is one more layer of nesting.  Let's remove `estimate_gas_request` and just call 
` ethereum.api_state_accessor() & estimate_gas_request_with_snapshot`

crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -134,6 +165,456 @@ where
...
+
+        let should_run_affordability_preflight = supports_request_affordability_preflight(
+            &request,
+            state_overrides.as_ref(),
+            block_overrides.as_deref(),
+        );
+
+        if request.gas.is_none() {
+            let mut state = snapshot_state.clone_without_local_writes();
+            let estimated_gas = evm.eth_estimate_gas(
```
bkolad: Inside the evm let's rename `eth_estimate_gas` to `estimate_gas_helper` or something similar so it is clear that this is not the final rpc method.

crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -134,6 +165,456 @@ where
...
+        )
+    }
+
+    fn estimate_gas_request(
+        request: TransactionRequest,
+        block_id: Option<BlockId>,
+        state_overrides: Option<StateOverride>,
+        block_overrides: Option<Box<BlockOverrides>>,
+        ethereum: &Arc<Ethereum<S, Seq>>,
+    ) -> RpcResult<U64> {
```
bkolad: A lot of the new methods seem to follow the same pattern: they take a very similar set of arguments, and they each instantiate `evm`. Instantiating `evm` here is not wrong, but it does add quite a bit of repetition.

I’d suggest moving the new gas estimation logic into a separate file, `estimate_gas.rs`, and introducing a struct that holds the shared context for these methods, such as:

```rust
request: TransactionRequest,
block_id: Option<BlockId>,
state_overrides: Option<StateOverride>,
block_overrides: Option<Box<BlockOverrides>>,
snapshot_state: &'a ApiStateAccessor<S>,
ethereum: &'a Arc<Ethereum<S, Seq>>,
// ...other shared fields
```
Then the current functions could be refactored into methods on that struct. That would reduce the repeated arguments, keep the gas estimation code more self-contained, and make the overall flow easier to follow.

bkolad: Refactoring it into a set of methods on sturct is optuonal but we should move the code to a separate file regardless.

crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -134,6 +165,456 @@ where
...
+                block_id,
+                state_overrides.clone(),
+                block_overrides.clone(),
+                &mut state,
+            )?;
+
+            if should_run_affordability_preflight {
+                let mut request_with_estimated_gas = request;
+                request_with_estimated_gas.gas = Some(estimated_gas.to::<u64>());
+                Self::run_request_affordability_preflight(
```
bkolad: This is a nit, but I had to go back and forth to the implementation of this method because it was not clear to me that the paymaster logic lives here. It would be great to rename it so that this is more obvious from the method name.

Maybe something like `run_request_affordability_preflight_and_paymaster_check`

crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -134,6 +165,456 @@ where
...
+                )?;
+            }
+
+            return Ok(estimated_gas);
+        }
+
+        Self::validate_request_stale_nonce_preflight(&request, block_id, snapshot_state)?;
+
+        if should_run_affordability_preflight {
+            Self::run_request_affordability_preflight(
```
bkolad: Let;s remoe one level of method nesting.
1. Let's remove `run_request_affordability_preflight`
2. Let's call
```
let evm = Evm::<S>::default();
        let mut api_state = snapshot_state.clone_without_local_writes();
        let mut preflight_state = evm
            .preflight_state_for_block_id(block_id, &mut api_state)
            .map_err(ErrorObjectOwned::from)?;
```
in line 234
3. Let's call `request_affordability_preflight` and change it to return RpcResult<()> instead of 
```
AffordabilityPreflight::Affordable
```

crates/full-node/sov-ethereum/src/handlers/mod.rs
```diff
@@ -134,6 +165,456 @@ where
...
+                .map_err(map_authentication_error)?;
+        Ok((authenticated_tx, auth_data))
+    }
+
+    /// Authenticates and runs affordability preflight for a raw/local send.
+    ///
+    /// A single state snapshot is taken up-front so that authentication and the
+    /// subsequent affordability check observe the same head — even if new blocks
+    /// land between the two steps.
+    fn authenticate_and_preflight_send(
```
bkolad, line 242: Looks like `authenticate_and_preflight_send` is just wpraper on `authenticate_and_preflight_send_with_snapshot` that adds single line. But the wraper is called only in one place so it doesn't reduce any repetitions. Let's remove and call:
```
 let snapshot_state = ethereum.api_state_accessor();
Self::authenticate_and_preflight_send_with_snapshot()
```
directly.

crates/full-node/sov-ethereum/src/handlers/mod.rs
citizen-stig: I don't fully get what we try to achieve here? 

Do you want to remove second call to `run_req_affordability_preflight_incl_paymaster` ?

This is different request, without estimated gas explicitly set. 

This breaks some of the tests.

crates/full-node/sov-ethereum/src/handlers/mod.rs
citizen-stig: Done

crates/full-node/sov-ethereum/src/handlers/mod.rs
citizen-stig: Moved to a separate file, will check feasibiliyt of adding a struct at the end

crates/full-node/sov-ethereum/src/handlers/mod.rs
citizen-stig: Renamed

crates/full-node/sov-ethereum/src/handlers/mod.rs
citizen-stig: Done (`authenticate_and_preflight_send` became `authenticate_and_preflight_send_with_snapshot` effectively)

crates/full-node/sov-ethereum/src/handlers/mod.rs
citizen-stig: Digging deeper

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
```diff
@@ -0,0 +1,322 @@
...
+        // These guards are defensive: omitted-gas callers synthesize `gas` from a pinned estimate
+        // before calling this helper, and the caller already validates `from` and a fee field.
+        let Some(from) = request.from else {
+            return Ok(());
+        };
+        if request.gas.is_none() {
+            return Ok(());
+        }
+        if request.max_fee_per_gas.or(request.gas_price).is_none() {
+            return Ok(());
```
bkolad, line 135: This 4 conditions are the same as `should_run_affordability_preflight`.  In if they rae not meet the function just returns early.
1. here we can call `should_run_affordability_preflight`.
2. If we return early here we don't need to call `should_run_affordability_preflight` outside of the method.

bkolad (approved):

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
citizen-stig: Split the validation

PR #2600, opened 2026-03-17
CI: Dedicated cache tag for nextest_all_features for preventing cache overflow

citizen-stig:
# Description

This should prevent consumption of full cache volume.

This is what happening currently:
Screenshot 2026-03-18 at 15 53 57

By adding dedicated cache tag, I hope it will be able to reduce amount of cache being evicted.

Screenshot 2026-03-18 at 15 54 19

**High level goal**: Faster nextest step. It should be roughly equal runtime of the tests + 1-3 minutes for compilation. If cache holds, we can work on tests runtime improvement.

## Dockerfile changes

This docker file should be used for building namespace ci profiles, but build process has been unstable there, so I just keep those changes for the future use.

## Testing
All existing tests are pasing

## Docs
No updates

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2600" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -1,6 +1,7 @@
 ARG NAMESPACE_BASE_IMAGE_REF=""
 # Your image must build FROM NAMESPACE_BASE_IMAGE_REF
 FROM ${NAMESPACE_BASE_IMAGE_REF} AS base
+ARG MOLD_VERSION=2.40.4
 
 # Prevent interactive prompts during package installation
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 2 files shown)


docker/ci/Dockerfile
```diff
@@ -2,95 +2,102 @@ ARG NAMESPACE_BASE_IMAGE_REF=""
...
-    protobuf-compiler \
-    clang \
-    cmake \
-    libclang-dev \
-    lld \
-    zstd \
-    unzip \
-    && sudo rm -rf /var/lib/apt/lists/*
+ARG RUST_STABLE_VERSION=1.93.0
+ARG RUST_NIGHTLY_VERSION=nightly-2026-03-10
```
theodorebugnet, line 6: Do we want to pin this date specifically? Will we remember to keep this updated?
My instinct would be to keep nightly unpinned and if we ever have issues, then we can pin it - I don't think we've ever had breakages due to using nightly in CI yet. But I don't have a strong opinion so if you think it's better this way then no problem.

theodorebugnet (approved):

docker/ci/Dockerfile
citizen-stig: Idea is to cut time from each job that require nightly toolchain and installs it.

We can periodically bump this version and rebuild base image

citizen-stig: and, also this is not yet added to base image

PR #2601, opened 2026-03-17
Implement aggregation algorithm with recursive proof verification

bkolad:
# Description

1. Extracts the aggregation logic from `main.r`s into a dedicated `aggregation.rs` module, with a generic `run_aggregation_program` function for verifying inner proof chains and recursive outer proofs.

2. Adds a `build.rs` step that precomputes the inner verifying key hash at compile time, removing it from the runtime witness.

3. Renames `CodeCommitment` to `OuterCodeCommitmentHash` and `vkey_hash` to `outer_vkey_hash` throughout the codebase for improved clarity.

Note for reviewers: The substantive change is in aggregation.rs. The remaining changes are primarily renaming and script-related improvements, and are not relevant to the security of the outer circuit.                                                                                                                                        

## Linked Issues
- Related to #2551

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2601" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/full-node/sov-aggregated-proof/program/build.rs
+++ b/crates/full-node/sov-aggregated-proof/program/build.rs
@@ -1,0 +1,37 @@
+use std::env;
+use std::fs;
+use std::path::PathBuf;
+
+use sp1_sdk::prelude::HashableKey;
+use sp1_sdk::SP1VerifyingKey;
+
+fn main() {
+    let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR"));
+    let data_dir = manifest_dir
+        .parent()
+        .expect("program crate must live under the sov-aggregated-proof workspace root")
+        .join("data");
+    let inner_vk_path = data_dir.join("inner_vk.bin");
+    let out_dir = PathBuf::from(env::var("OUT_DIR").expect("OUT_DIR"));
+
+    println!("cargo:rerun-if-changed={}", inner_vk_path.display());
+
+    let inner_vk_bytes = fs::read(&inner_vk_path).unwrap_or_else(|error| {
+        panic!(
+            "Failed to read saved inner verifying key fixture at {}: {error}",
+            inner_vk_path.display()
+        )
+    });
+    let inner_vk: SP1VerifyingKey = bincode::deserialize(&inner_vk_bytes).unwrap_or_else(|error| {
+        panic!(
+            "Failed to deserialize saved inner verifying key fixture at {}: {error}",
+            inner_vk_path.display()
+        )
+    });
+    let inner_vk_hash = inner_vk.hash_u32();
+
+    let generated = format!("const INNER_VKEY_HASH: [u32; 8] = {inner_vk_hash:?};\n");
+
+    fs::write(out_dir.join("inner_vk_hash.rs"), generated)
+        .expect("Failed to write generated inner VK hash file");
+}
--- a/crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
+++ b/crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
@@ -1,0 +1,197 @@
+use sha2::{Digest, Sha256};
+use sov_aggregated_proof_shared::{
+    AggPubData, AggregatedProofWitness, DeferredProofInput, StfPubData,
+};
+use sov_modules_api::da::BlockHeaderTrait;
+use sov_modules_api::{DaSpec, OuterCodeCommitmentHash, Spec, Storage};
+use sov_rollup_interface::common::SlotNumber;
+
+struct BoundaryData<Hash, Root> {
+    slot_hash: Hash,
+    state_root: Root,
+    slot_number: SlotNumber,
+}
+
+struct VerifiedProofData<Address, Hash, Root> {
+    initial_boundary: BoundaryData<Hash, Root>,
+    final_boundary: BoundaryData<Hash, Root>,
+    rewarded_addresses: Vec<Address>,
+}
+
+type VerifyResult<S, Da> = VerifiedProofData<
+    <S as Spec>::Address,
+    <Da as DaSpec>::SlotHash,
+    <<S as Spec>::Storage as Storage>::Root,
+>;
+
+pub(crate) fn run_aggregation_program<S: Spec<Da = Da>, Da: DaSpec>(
+    witness: AggregatedProofWitness<Da>,
+    inner_vkey_hash: [u32; 8],
+) {
+    let proof_inputs = witness.proof_inputs;
+    let outer_vkey_hash = witness.outer_vkey_hash;
+    let prev_outer_proof_witness = witness.prev_outer_proof_witness;
+
+    // Verify the previous aggregation proof if one exists. On the first aggregation
+    // after genesis, there is no predecessor, the chain starts here.
+    let previous_public_data = if let Some(prev_outer_proof_witness) = prev_outer_proof_witness {
+        Some(deserialize_and_verify_pub_data::<AggPubData<S, Da>>(
+            &prev_outer_proof_witness.public_values,
+            outer_vkey_hash,
+        ))
+    } else {
+        None
+    };
+
+    let verified_proof_data: VerifyResult<S, Da> =
+        verify_proof_chain::<S, Da>(proof_inputs, inner_vkey_hash, previous_public_data.as_ref());
+
+    let VerifiedProofData {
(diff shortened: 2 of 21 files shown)


crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
```diff
@@ -0,0 +1,202 @@
...
+    rewarded_addresses: Vec<Address>,
+}
+
+type VerifyResult<S, Da> = VerifiedProofData<
+    <S as Spec>::Address,
+    <Da as DaSpec>::SlotHash,
+    <<S as Spec>::Storage as Storage>::Root,
+>;
+
+pub(crate) fn run_aggregation_program<S: Spec<Da = Da>, Da: DaSpec>(
```
bkolad, line 27: This is the most relevant part in thie PR

crates/full-node/sov-aggregated-proof/program/build.rs
```diff
@@ -0,0 +1,37 @@
+use std::env;
+use std::fs;
+use std::path::PathBuf;
+
+use sp1_sdk::prelude::HashableKey;
+use sp1_sdk::SP1VerifyingKey;
+
+fn main() {
```
bkolad, line 8: This `build.rs` will no longer be needed once `program/aggregation.rs` is moved to the rollup interface and `program/main.rs` is moved to the demo rollup (the we will just have access to thesp1::SP1_GUEST_MOCK_ELF).

crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
```diff
@@ -0,0 +1,202 @@
...
+    let aggregated_public_data = AggPubData::<S, Da> {
+        initial_slot_number: initial_boundary.slot_number,
+        final_slot_number: final_boundary.slot_number,
+        genesis_state_root,
+        initial_state_root: initial_boundary.state_root,
+        final_state_root: final_boundary.state_root,
+        initial_slot_hash: initial_boundary.slot_hash,
+        final_slot_hash: final_boundary.slot_hash,
+        outer_vk_hash,
+        rewarded_addresses,
```
bkolad, line 74: devin points out that we only reward the current addresses, but that is the intended behavior.
see https://github.com/Sovereign-Labs/sovereign-sdk/pull/2601/changes#diff-b31e5c0d240ef1563ea09cbecde859e737319b8487289b40c964e0ea071a0867R100

crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
```diff
@@ -0,0 +1,202 @@
...
+        // transition.
+        {
+            if let Some(expected_prev_state_root) = &expected_prev_state_root {
+                assert_eq!(
+                    expected_prev_state_root, &stf_public_data.initial_state_root,
+                    "State root discontinuity at index {index}: previous final_state_root != current initial_state_root"
+                );
+            }
+
+            expected_prev_state_root = Some(stf_public_data.final_state_root.clone());
```
ross-weir, line 149: we assert eq to initial state root and then set to final state root - just confirming that's correct?

crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
```diff
@@ -0,0 +1,202 @@
...
+    witness: AggregatedProofWitness<Da>,
+    inner_vkey_hash: [u32; 8],
+) {
+    let proof_inputs = witness.proof_inputs;
+    let outer_vkey_hash = witness.outer_vkey_hash;
+    let prev_outer_proof_witness = witness.prev_outer_proof_witness;
+
+    // Verify the previous aggregation proof if one exists. On the first aggregation
+    // after genesis, there is no predecessor, the chain starts here.
+    let previous_public_data = if let Some(prev_outer_proof_witness) = prev_outer_proof_witness {
```
citizen-stig, line 37: nit: `.map()` works on Option too

crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
```diff
@@ -0,0 +1,202 @@
...
+    vkey_hash: [u32; 8],
+    previous_agg_proof_public_data: Option<&AggPubData<S, Da>>,
+) -> VerifyResult<S, Da> {
+    assert!(
+        !proof_inputs.is_empty(),
+        "Aggregated proof must contain at least one proof input"
+    );
+
+    let mut expected_prev_slot_hash =
+        previous_agg_proof_public_data.map(|public_data| public_data.final_slot_hash.clone());
```
citizen-stig, line 93: Do I understand correctly, that it is possible to skip some chunks of chain, if malicious prover just does not pass `previous_agg_proof_public_data`

crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
bkolad: Yes, this guarantees that the final state of the previous slot is equal to the previous state of the next one.

crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
citizen-stig: That's what Devin says:


Yes, your concern is valid — within the zkVM circuit itself, there is no protection against omitting the previous aggregation proof. Here's the precise issue:

In run_aggregation_program (), if prev_outer_proof_witness is None, the program simply treats this as the "first aggregation after genesis" — no predecessor check is performed. Then at lines 92-96, expected_prev_slot_hash and expected_prev_state_root are both set to None, meaning the first inner proof in the batch is accepted without any continuity check against prior history.

So a malicious prover could:

    Take inner proofs for slots 100–110
    Set prev_outer_proof_witness = None
    Produce a valid aggregation proof that claims to start a fresh chain at slot 100, with an arbitrary genesis_state_root (which would just be the initial_state_root of slot 100)

The proof would verify successfully inside the zkVM.

However, the security depends on the external verifier. The committed AggPubData () includes initial_slot_number, genesis_state_root, initial_slot_hash, etc. The AggregateProofVerifier at line 93-114 deserializes and returns this public data. So an L1 contract or external verifier must check:

    That genesis_state_root matches the known genesis root
    That initial_slot_hash / initial_state_root continue from the last accepted aggregation's final_slot_hash / final_state_root
    That initial_slot_number follows the last accepted final_slot_number

The circuit alone does not enforce this. The circuit only guarantees internal consistency (the inner proofs chain correctly, and if a previous outer proof is provided it chains correctly). But the decision of whether to accept a "genesis" aggregation vs. requiring a predecessor is pushed to the external verifier.

This is a meaningful gap worth flagging. If the L1 verifier doesn't enforce continuity from the last accepted proof, a malicious prover can indeed skip chunks of the chain. The host-side script () does perform these checks, but that's the honest prover's script — it's not enforced in the circuit.

Recommendation: Consider adding a boolean or enum to the committed AggPubData (e.g., is_genesis: bool or previous_final_slot_hash: Option<SlotHash>) so the external verifier can distinguish a legitimate first-ever aggregation from one that simply omitted its predecessor. Alternatively, the circuit could require that if initial_slot_number > 1 (or whatever the genesis slot is), then a previous outer proof must be provided.

bkolad: If `previous_agg_proof_public_data = None` (and the prover is malicious, so it sets it after genesis), `genesis_state_root` falls back to `initial_boundary.state_root`. As a result, the aggregate proof is constructed with the initial state root of the current collections of inner proofs as its genesis. The prover will still succeed, but the resulting public data will have a genesis state that does not match the rollup’s canonical one, so the proof verification will fail. 


Related code:
```
    // Propagate the genesis state root forward through recursive aggregations.
    // For the very first aggregation, the genesis root is the initial state root
    // of the first inner proof (i.e. the state root at chain genesis).
    let genesis_state_root = previous_public_data
        .as_ref()
        .map(|public_data| public_data.genesis_state_root.clone())
        .unwrap_or_else(|| initial_boundary.state_root.clone());
````        ```

citizen-stig: Ok, I see!

citizen-stig (approved):

crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
bkolad: Will fix in the next PR.

PR #2605, opened 2026-03-18
Bug Fix: Make batch production  non-optional during recovery. Add sequencer proof processing tests

preston-evans98:
# Description

This PR rewrites and expands the sequencer proof processing test suite to include resync and recovery, as well as handling of proofs that arrive while an open batch is in progress. This should fix the flaky test as well as improving coverage. 

The new tests uncovered a pre-existing bug in recovery mode. The sequencer used `produce_batch_if_convenient`, but assumed that the requested batches were always produced. Trigger a failure was easier thanks to proofs (the blob sender is much more likely to be busy) which allowed the new tests to detect the issue. This PR fixes the issue by removing the "is_convenient" checks from batch production during recovery mode only.

## Linked Issues
- #2571


crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
```diff
@@ -214,6 +307,7 @@ where
             // Update the next height to receive
             self.stf_info_receiver
                 .inc_next_height_to_receive_by(num_proofs_to_create as u64);
+            self.proof_progress.finish_aggregate_proof_creation();
```
bkolad, line 315: `begin_aggregate_proof_creation` sets `aggregate_proof_in_flight = true`, and `finish_aggregate_proof_creation` clears it. However, if `create_aggregate_proof_with_retries` or `publish_proof_blob_with_metadata`  fails in this code block, the flag is never reset.

bkolad: I think this is fine as the `ZkProofManager` just exits if this happens

crates/utils/sov-test-utils/src/test_rollup.rs
```diff
@@ -123,6 +123,7 @@ pub struct RollupBuilderConfig<S: Spec> {
 /// background to test node APIs.
 #[derive(Clone)]
 pub struct RollupBuilder<R: FullNodeBlueprint<Native>> {
+    blueprint: Arc<R>,
```
bkolad, line 126: This will make `RollupBuilder` stateful, which means some state can persist across restarts.
```
builder.with_blueprint(..);
let rollup = builder.start()
let builder = rollup.shutdown() / rollup.restart()
rollup_after_restart = builder.star()

rollup_after_restart <- inherited r`eady_proof_count`
```
In this flow, `rollup_after_restart` may inherit state from the previous run. For example, in `ManualProofPostingRtAgnosticBlueprint`, we keep shared state such as `ready_proof_count`, `available_releases`, and `is_open`. That shared state would also be inherited by `rollup_after_restart`.

Would it be possible to add an `Option<R>` in `start_test_rollup`. So we would provide the bluerint only when `TestRollup` is created?

bkolad: Or we could jut add another method `start_test_rollup_with_manual_proof_posting` and call `ManualProofPostingRtAgnosticBlueprint::<TestSpec, RT>::new_with_control()` in it.

crates/full-node/sov-sequencer/src/preferred/mod.rs
```diff
@@ -247,9 +247,7 @@ where
                 }
 
                 self.synchronized_state_updator
-                    .trigger_batch_production_if_convenient_msg(
-                        "recover_and_catch_up:dump_catchup_batches",
-                    )
+                    .trigger_batch_production_msg("recover_and_catch_up:dump_catchup_batches")
```
bkolad, line 252: One of the checks that is skipped now is:
```
if !self.seq_config.automatic_batch_production {
       warn!("Skipping batch production due to settings");
       return;
 }
```

This means that the sequencer will produce batches during recovery, although it is configured to never do it.  This only matters in tests, but it can still be confusing. Let's log an error if we hit recovery and `automatic_batch_production = true`?

bkolad: Another option: we could panic when this happens, and `cfg!(debug_assertions) = true`

preston-evans98: Would something like this address the concern? 
```
 if !self.seq_config.automatic_batch_production {  
            warn!("Triggered batch production despite automatic batch production being disabled. This is likely a bug in test code.");  
        }  

```

PR #2610, opened 2026-03-20
Fix: validate replica visible slot before applying leader `BatchStart`

bkolad:
# Description

When a preferred leader falls behind and enters recovery, it may later resume batch production with a visible slot number that diverges from the replica's local state. Previously the replica would apply this inconsistent `BatchStart` blindly, leading to state corruption.

## Changes

- **Validate visible slot on replica `BatchStart`**: before applying a `BatchStart` from the leader, compute the expected visible slot number locally and compare it against the leader's value. If they diverge, mark the replica as syncing and return `NotReady` instead of applying inconsistent state.
- **Integration test**: add `test_db_elected_leader_recovery_replica_keeps_running` which forces the leader into recovery (by exceeding the deferred slots threshold), verifies the replica stays operational throughout, and confirms both nodes return to ready with successful transaction processing afterward.

## Test plan

- [x] New integration test covers the leader-recovery → replica-continues → cluster-resumes scenario
- [x] Existing replica tests pass with the shared `RecoveryStrategy::TryToSave` default

---

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2610" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/examples/demo-rollup/tests/replica/recovery.rs
+++ b/examples/demo-rollup/tests/replica/recovery.rs
@@ -1,0 +1,96 @@
+use super::*;
+
+/// Test that if the leader enters recovery after falling behind the deferred-slots threshold,
+/// the replica continues to function correctly once recovery completes.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_db_elected_leader_recovery_with_replica() {
+    std::env::set_var("SOV_TEST_CONST_OVERRIDE_DEFERRED_SLOTS_COUNT", "40");
+
+    let Some(setup) = NodeDiscoveryTestSetup::new().await else {
+        return;
+    };
+
+    let key_and_address = read_private_key::<S>("tx_signer_private_key.json");
+
+    let node_1 = setup
+        .start_node("node_1", ConfiguredNodeRole::DbElected)
+        .await;
+    let node_2 = setup
+        .start_node("node_2", ConfiguredNodeRole::DbElected)
+        .await;
+
+    node_1.wait_for_sequencer_ready().await.unwrap();
+    node_2.wait_for_sequencer_ready().await.unwrap();
+
+    let (leader, replica) = establish_leader_and_replica(node_1, node_2).await;
+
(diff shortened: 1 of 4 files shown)


examples/demo-rollup/tests/replica/recovery.rs
```diff
@@ -0,0 +1,96 @@
...
+        &replica,
+        &key_and_address.private_key,
+        token_id,
+        receiver_addr,
+        0,
+    )
+    .await;
+
+    // Pause the sequencer update_state loop to prevent batch production.
+    leader.pause_preferred_batches().await;
```
bkolad, line 42: This pauses batches for both the replica and the sequencer. This is intended in this test, but I will extend `pause_preferred_batches` to allow pausing batches only for the leader.  (In separate PR)

examples/demo-rollup/tests/replica/recovery.rs
```diff
@@ -0,0 +1,96 @@
+use super::*;
+
+/// Test that if the leader enters recovery after falling behind the deferred-slots threshold,
+/// the replica continues to function correctly once recovery completes.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_db_elected_leader_recovery_with_replica() {
```
theodorebugnet, line 6: Is it useful to have it as a demo-rollup test, rather than a TestRollup test for the sequencer? The downside is that during development this won't run on the default workspace, you have to compile the demo-rollup for it. It's also in a different place to the other replica/recovery tests.
But if there's a good reason to have it here then it's not a big deal, so non-blocking

theodorebugnet (approved):

examples/demo-rollup/tests/replica/recovery.rs
bkolad: All the tests related to cluster setup are in the demo rollup, mainly because they require Testcontainers, which is slow to run, so we separated them there.

PR #2613, opened 2026-03-20
Add support for a one-time gas limit update to existing rollups

preston-evans98:
# Description

This PR allows a one-time update to the gas limit of a rollup. This update also disables preferred sequencer fee escrow, which eliminates a potential maintenance headache when gas prices rise. 

The change to eliminate the fee escrow is safe as long as the prover and the sequencer aren't separate entities. 

## Old Flow of Funds

Blob selector escrows funds for sequencer 
- Transfer from sequencer_registry -> bank
- Sequencer balance in sequencer registry is marked down

If the sequencer is penalized
- Transfer penalty amount from bank -> prover immediately 
- After returning to apply_batch, net rewards are decremented by `gas_used.value(gasPrice)`

If we reach `reserve_gas`, 
- Transfer funds from `payer` -> bank (entire tx.max_value)

If we reach refund_remaining_gas
- Transfer from bank -> payer `tx.max_value - gas_used.value(gas_price)`

If we reach reward_prover
- Transfer from bank -> prover `gas_used.value(base_fee)`. Note that this leaves `gas_used.value(priority_fee)` to reward the seuqencer

At the end of apply_batch, the initial bond plus (or minus) net rewards/penalties is transferred from Bank -> sequencer_registry

### Safety Justification
The new setup where the preferred sequencer does not have any funds escrowed is safe (meaning that it can't cause a shortfall) as long as we skip "penalization". (Penalization is the only case when funds are logically transferred from the sequencer's escrow to another party). This prevents a shortfall in the bank. Note that sequencer rewards may still be paid out, since they are populated from the *user* balance, not the sequencer escrow. Note also that prover rewards may be paid out from the user funds as before - only the "unhappy" sequencer penalization path is impacted. 

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -825,6 +825,9 @@ where
     #[cfg(feature = "test-utils")]
     async fn process_force_close_current_batch(&mut self, reason: &'static str) {
         let mut inner = self.get_inner_with_timing(reason).await;
+        if !inner.executor.has_in_progress_batch() {
```
bkolad, line 844: Let’s return a bool or an enum indicating whether the batch was closed, so the caller knows.

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -933,9 +933,14 @@ where
         .base_fee_per_gas(&mut accessor)
         .unwrap_or(S::initial_base_fee_per_gas());
     // Note that we need to escrow gas for each batch and proof. The escrow amount is `max_tx_check_costs` in each case.
-    let needed_gas_escrow = S::max_tx_check_costs()
-        .checked_value(next_gas_price)
-        .expect("Gas price overflow! This is a bug, please report it.");
+    let needed_gas_escrow =
```
bkolad: This path is used for both preferred and standard batches:
`block_executor.rs:L971`

In `validation::validate_preferred_blob`, we override `funds_needed`, but we do not do the same in `validation::validate_blob`. As a result, `funds_needed` gets overridden for standard blobs in the sequencer, but not in the node.

Later, in the STF, we only skip this check for preferred blobs:
`registered.rs:L740`

```
if sequencer_bond.amount() < max_tx_check_value {
    if !(scratchpad.rollup_height_to_access() > <S as GasSpec>::change_gas_limit_after_height()
        && matches!(sequencer_bond, SequencerBondForTx::Preferred(Amount::ZERO)))
```

This looks that standard blobs may follow different code paths in the sequencer and the node.

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -933,9 +933,14 @@ where
...
+        if target_rollup_height > <S as GasSpec>::change_gas_limit_after_height() {
+            Amount::ZERO
+        } else {
+            S::max_tx_check_costs()
+                .checked_value(next_gas_price)
+                .expect("Gas price overflow! This is a bug, please report it.")
+        };
```
bkolad, line 947: Another issue appears in the same code path.

In this file, we call `kernel.increment_rollup_height(&mut accessor, next_visible_slot_number)` on line 923, so the height check happens after the height has already been incremented (in the sequencer).

In the node, the order is different in `apply_slot_with_control_flow`:
- `select_and_validate_blobs` on line 492 (which runs the height check)
- ` runtime.chain_state().increment_rollup_height` on line 520

So we can have an off-by-one height difference check between the node and the sequencer.

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
preston-evans98: Very good catch. The funds actually get reserved by the blob storage module, but I forgot that we source the amount here. Fixed.

crates/full-node/sov-ledger-apis/tests/integration/snapshots/integration__get_batch_include_children.snap
preston-evans98, line 1: Note for reviewers: the snapshot tests here changed because they use progenitor. The `data` field was always included in the response from the server, but previously the OpenAPI spec didn't account for it. This caused progenitor to silently drop the data. Since these tests use progenitor to fetch the data and then call `serde_json::to_value` on the output, the spec change causes a change in the observed results even though the server response is untouched.

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
preston-evans98: Very good catch. Confirmed and fixed

bkolad (approved):

PR #2615, opened 2026-03-23
Fix proof handle leak in Network prover service

ross-weir:
# Description

Previously if the DA verifier failed its `verify_relevant_tx_list` call then we would leak a proof network submission because the network submission happened first.

This re-orders the operations so we first verify the tx list and then submit the proof request if the tx list is valid.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
+++ b/crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
@@ -118,6 +118,18 @@ where
             prover_address: self.prover_address.clone(),
         };
 
+        // Verify DA BEFORE submitting to the network to avoid leaking a proof handle
+        verifier
+            .da_verifier
+            .verify_relevant_tx_list(
+                &data.stf_witness.da_block_header,
+                &data.stf_witness.relevant_blobs,
+                &data.stf_witness.relevant_proofs,
+            )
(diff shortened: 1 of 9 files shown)


crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
```diff
@@ -118,6 +118,18 @@ where
             prover_address: self.prover_address.clone(),
         };
 
+        // Verify DA BEFORE submitting to the network to avoid leaking a proof handle
+        verifier
+            .da_verifier
+            .verify_relevant_tx_list(
+                &data.stf_witness.da_block_header,
+                &data.stf_witness.relevant_blobs,
+                &data.stf_witness.relevant_proofs,
```
bkolad: Why could we not keep the old API, where verify_relevant_tx_list simply takes  `data.stf_witness.relevant_proofs`, and clone them at the call site when needed?                                                       
                                                                                                                                                                                                                                                              
They are cloned inside verify_relevant_tx_list now, but it feels a bit cleaner to leave that decision to the caller. That way, other code using verify_relevant_tx_list could pass owned data, such as when it no longer needs it afterward, and avoid an     
unnecessary clone.

bkolad (approved):

PR #2616, opened 2026-03-23
add timeout for polling outer proof

ross-weir:
# Description

Adds a timeout when polling for outer proof result

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
+++ b/crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
@@ -56,6 +56,7 @@ pub(crate) struct NetworkProver<
     outer_vm: OuterVm::Network,
     tracker: tokio::sync::RwLock<ProofStatusMap<Address, StateRoot, Da::Spec, InnerVm>>,
     code_commitment: OuterCodeCommitmentHash,
+    outer_proof_timeout: std::time::Duration,
     phantom: PhantomData<Witness>,
 }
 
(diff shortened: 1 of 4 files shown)


crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
```diff
@@ -265,21 +268,30 @@ where
...
+                    Ok(Some(proof_bytes)) => {
+                        break Ok(SerializedAggregatedProof {
+                            raw_aggregated_proof: proof_bytes,
+                        });
+                    }
+                    Ok(None) => {
+                        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+                    }
+                    Err(e) => {
+                        break Err(anyhow::anyhow!("Outer network proving failed: {}", e));
```
bkolad, line 283: `create_aggregated_proof` is called by `create_aggregate_proof_with_retries`. As written, each retry appears to submit a brand-new request, which means the proving process starts over from scratch every time. Shouldn’t we poll the original request instead?

Or is the intended behavior to exit the prover on retry? If so, we should handle it similarly to how we already handle `error_message.contains("Elf parse error")` in `create_aggregate_proof_with_retries`.

ross-weir: For now, the latter. Added handling and some tests to ensure the error message is correct.

I think the retry handling here could be improved when we start real testing and understand the error variants/network responses better

bkolad (approved):

PR #2622, opened 2026-03-23
Expose the SDK specs as `rustdocs` 

nuke-web3:
# Description

This PR renders the spec via rustdocs (WIP, as there is frustrating workarounds needed to render on github and IDEs and rustdocs well).

Motivation: the present `make docs` fails with a lib name conflict, that is resolved in ffd8422e28b92b869909649c75df31d705d523c9 - if the rustdocs stuff is unwanted, feel free to cherry pick that commit!

Inspiration: Substrate uses a sophisticated rustdocs setup to render some docs, although it is somewhat of a pain... See https://github.com/paritytech/polkadot-sdk/issues/6191

- To render markdown pages in rustdocs, a dummy crate is introduced in the `rollup-interface/spec/`, but it has no actual code, and is only compiled for `cargo doc`
- Updated `make docs` to open a main README (the spec or some other README) directly alongside a sidebar that includes all SDK crates (I think all the relevant ones, would appreciate a check on that!)

  - I have not noted in changelog, I can if we think it's needed!

## Testing

Run `make docs` to view the result, and navigate pages _not_ with the (presently broken) inline links, but using the `Modules` section at the bottom of each page (overview -> interfaces -> da ...)

## Docs

> Rendered docs are available at `https://sovlabs-ci-rustdoc-artifacts.us-east-1.amazonaws.com/<BRANCH_NAME>/index.html`

^^ from the issue template, so I think reviewers can check if things work there, but the entrypoint/index for the docs overall is likely not the new spec docs dummy crate, that would need to be worked out to fix it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12826,6 +12826,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "sov-sdk-specs"
+version = "0.3.0"
+
+[[package]]
 name = "sov-sequencer"
 version = "0.3.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ resolver = "2"
 exclude = ["crates/module-system/module-implementations/extern/hyperlane-solana-register/solana"]
 members = [
     "crates/rollup-interface",
+    "crates/rollup-interface/specs",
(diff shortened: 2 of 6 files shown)


nuke-web3:

PR #2626, opened 2026-03-24
EVM: Remove apply_margin and sync gas used value from estimate to receipt

citizen-stig:
# Description

## Changes summary

This PR adds a more precise gas estimation based on rollup-stf-blueprunt execution hellpers.

For state and block overrides it adds "fallback" estimation mechanism that tries to mimic `rollup-stf-blueprint` 

## Decisions made

To reduce complexity of this PR, following decision were made:

* `estimateGas` endpoint will be backwards incompatible with querying on blocks below `EVM_RECEIPT_ACTUAL_FEE_HEIGHT`
* `estimateGas` endpoint won't be entirely consistent with `sendRawTx` for non-preferred sequencers, and future nonces there will be rejected.

Some of it might be addressed in follow up PR with proptests: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2620
But we will stay pragmatic.

This has been highlighted in following agents observations:

> - High: the patch removes the runtime-parity fallback for every eligible request, which will make eth_estimateGas fail for preferred-sequencer future-nonce flows that sendRawTx can still accept. The wrapper now hard-returns any runtime-parity failure at crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs:114 and crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs:190. The parity path runs STF uniqueness checks in normal sequencer context, and nonce uniqueness outside warm-up requires an exact current nonce at crates/module-system/module-implementations/sov-uniqueness/src/capabilities.rs:25 and crates/module-system/module-implementations/sov-uniqueness/src/nonces.rs:14. Preferred sendRawTx still queues future nonces through the nonce buffer at crates/full-node/sov-sequencer/src/preferred/mod.rs:430 and crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs:891, and you already have compliance coverage for that flow at examples/demo-rollup/tests/evm/evm_rpc_compliance_validation_2.rs:46 and examples/demo-rollup/tests/evm/evm_rpc_compliance_validation_2.rs:159. This is a real behavior regression, not just a code simplification.

> Runtime parity path returns gas estimate instead of revert error for non-preferred sequencers
>
> The estimate_gas_via_runtime_parity path matches TxEffect::Successful to return a gas estimate (). However, an EVM-level revert produces TxEffect::Successful at the STF level whenever on_revert in does not bail — specifically when the node is a non-preferred sequencer (context.sequencer_is_preferred() == false) or when preferred_sequencer_publish_reverted_txs is enabled. In these cases, the runtime parity path silently returns a gas estimate for a failing transaction instead of a revert error. The fallback path (eth_estimate_gas_helper at ) correctly returns RpcInvalidTransactionError::Revert. Since the runtime parity path is the default for all requests with from set and no overrides (should_use_runtime_parity_estimate at ), non-preferred sequencer RPC nodes will return gas values for reverting calls, violating Ethereum RPC semantics and the AGENTS.md rule: "eth_estimateGas errors on revert with revert payload; no success value on revert".

---

## Linked Issues
- Related to #2271 

## Testing
New test is unignored, 2 more tests added to validate case better.

## Docs
No updates

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2626" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
```diff
@@ -85,54 +89,153 @@ where
...
+            prepared.authenticated_tx,
+            prepared.auth_data,
+            prepared.runtime_call,
+        );
+
+        let mut runtime = Seq::Rt::default();
+        let (result, mut tx_scratchpad, _) = process_tx_and_reward_prover(
+            &mut runtime,
+            prepared.pre_exec_working_set,
+            <S::Gas>::MAX,
```
citizen-stig, line 176: Why is it MAX?

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
```diff
@@ -85,54 +89,153 @@ where
...
+            )
+    }
+
     pub(crate) fn run_request_affordability_preflight(
         request: &TransactionRequest,
         block_id: Option<BlockId>,
         snapshot_state: &ApiStateAccessor<S>,
         ethereum: &Arc<Ethereum<S, Seq>>,
-    ) -> RpcResult<()> {
+    ) -> RpcResult<AffordabilityPreflight> {
```
citizen-stig: Why this?

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
```diff
@@ -179,14 +270,6 @@ where
         state: &mut ApiStateAccessor<S>,
         ethereum: &Arc<Ethereum<S, Seq>>,
     ) -> RpcResult<AffordabilityPreflight> {
-        let evm = Evm::<S>::default();
```
citizen-stig, line 182: Why is this removed?

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
```diff
@@ -117,6 +136,92 @@ where
...
+            metrics,
+            ethereum.sequencer_type,
+        );
+
+        match result {
+            Ok(apply_tx_result) => match apply_tx_result.receipt.receipt {
+                TxEffect::Successful(_) => evm
+                    .read_runtime_parity_estimate_from_pending_tail(&mut tx_scratchpad)
+                    .map_err(|err| ErrorObjectOwned::owned(-32603, err, None::<()>)),
+                other => Err(Self::tx_effect_to_rpc_error(other)),
```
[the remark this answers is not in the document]
citizen-stig: Ackbnowledge

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
```diff
@@ -94,14 +98,29 @@ where
...
 
-        let mut state = snapshot_state.clone_without_local_writes();
-        let estimated_gas = evm.eth_estimate_gas_helper(
-            request.clone(),
-            block_id,
-            state_overrides,
-            block_overrides,
-            &mut state,
-        )?;
+        let estimate_with_fallback = || {
```
bkolad: Wys is it a clojure? Can we just run  `evm.eth_estimate_gas_helper(..)` in the else branch?

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
```diff
@@ -117,6 +136,92 @@ where
...
+            ExecutionContext::Sequencer,
+            &NoOpControlFlow,
+            prepared.operating_mode,
+            metrics,
+            ethereum.sequencer_type,
+        );
+
+        match result {
+            Ok(apply_tx_result) => match apply_tx_result.receipt.receipt {
+                TxEffect::Successful(_) => evm
```
bkolad: If `preferred_sequencer_publish_reverted_txs = true`, this will return success (evm runtime behaviour) even for reverted transactions, but I think `estimate_gas` should return an error if the tx is reverted. 

So we should set `preferred_sequencer_publish_reverted_txs=false` before the tx execution.

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
```diff
@@ -117,6 +136,92 @@ where
...
+                    .read_runtime_parity_estimate_from_pending_tail(&mut tx_scratchpad)
+                    .map_err(|err| ErrorObjectOwned::owned(-32603, err, None::<()>)),
+                other => Err(Self::tx_effect_to_rpc_error(other)),
+            },
+            Err((error, _)) => Err(ErrorObjectOwned::owned(
+                alloy_rpc_types::error::EthRpcErrorCode::TransactionRejected.code(),
```
bkolad: I think this is why a test was updated to assert 32003 instead of 32603. Is it possible to return a finer-grained error here?

examples/demo-rollup/tests/evm/evm_contract_creation_allowlist.rs
```diff
@@ -22,6 +22,6 @@ async fn denied() -> anyhow::Result<()> {
     let client = alloy_client_with_signer(rollup.http_addr, SECONDARY_SENDER_PRIV_KEY);
 
     let err = SimpleStorage::deploy(client).await.unwrap_err();
-    assert_eq!(err.to_string(), "server returned an error response: error code -32603: Contract creation is only allowed from allowed addresses. 0x3FE0233e6cf3c9753fcB7449987EC49C88aDDE71 is not on the list");
+    assert_eq!(err.to_string(), "server returned an error response: error code -32003: Contract creation is only allowed from allowed addresses. 0x3FE0233e6cf3c9753fcB7449987EC49C88aDDE71 is not on the list");
```
bkolad: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2626/changes#r2994192317

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
bkolad: wallets and SDKs use error.code and error.data to classify failures, so this might be important. We are also losing the revert reason

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
citizen-stig: Good point! It is a leftover from more fallback branches

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
citizen-stig: Hm, that's valid.

But where this should be set? It looks like `EVM_EXECUTION_CONFIG` is OnceLock...

I will investigate

crates/module-system/module-implementations/sov-evm/src/rpc/handlers.rs
```diff
@@ -603,93 +614,360 @@ where
...
+            gas_info.gas_value,
+            &mut metered_state,
+        )
+        .unwrap_infallible();
+
+        Ok(U64::from(pending_tx.receipt.gas_used))
+    }
+}
+
+impl<S: Spec> Evm<S>
```
bkolad: Let's move all the gas-estimate related logic to a separate file.

crates/module-system/module-implementations/sov-evm/src/rpc/handlers.rs
```diff
@@ -603,93 +614,360 @@ where
...
+        if request.from.is_none() {
+            request.from = Some(Address::ZERO);
+        }
+
+        if request.gas.is_none() {
+            request.gas = Some(block_gas_limit.min(tx_gas_limit.unwrap_or(block_gas_limit)));
+        }
+
+        if request.max_fee_per_gas.is_none() && request.gas_price.is_none() {
+            request.gas_price = Some(default_gas_price);
```
bkolad: if `request.max_fee_per_gas.is_none() && request.gas_price.is_none() = true` shouldn't we skip the affordability check and only return the consumed gas from eth_estimateGas?

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
citizen-stig: Simplest approach is to check in the receipt, which has been done

crates/module-system/module-implementations/sov-evm/src/rpc/handlers.rs
citizen-stig: Done!

crates/module-system/module-implementations/sov-evm/src/rpc/handlers.rs
citizen-stig: This is normalization of the request before sumilation. affordance checks are done before.

crates/full-node/sov-ethereum/src/handlers/estimate_gas.rs
citizen-stig: I've double checked and -32003(Transaction Rejected) is better than -32603(InternalServerErrror.

Claude did a deeper investigation:

> The pre-check `check_for_evm_revert` already returns code `3` with raw EVM revert bytes for actual EVM reverts (Solidity revert/require). The STF-level rejections (allowlist, gas
reservation) that reach `tx_effect_to_rpc_error` are NOT EVM reverts and have no raw bytes. Should we use `-32003` (TransactionRejected) for those STF rejections?


I've also removed all usages of plain number with named constants.

examples/demo-rollup/tests/evm/evm_contract_creation_allowlist.rs
citizen-stig: Replied there.

citizen-stig: This is another endpoint, there's a way to improve errors and make them more granular,

```
❯ Can we inspect `verify_contract_creation_allowlist` and return granular error from there instead of anyhow? Something from reth or revm family?

⏺ All 7 tests pass. Now to address your question about verify_contract_creation_allowlist:

  Currently it returns anyhow::Error with bail!(). You're asking if we can return a proper typed error from the reth/revm family instead.


⏺ This is a good improvement but it's a separate concern from the gas estimation cleanup we've been doing. The function is used in both call.rs:266 (STF execution path) and rpc/mod.rs:1018 (RPC simulation path), so changing its return
  type requires updating both callers.

  That said, this would be scope creep beyond the original 3 changes. Should I note it as a follow-up, or do you want me to tackle it now?

```

bkolad (approved):

PR #2627, opened 2026-03-24
Add per-node sequencer pause and leader-only recovery test.

bkolad:
# Description

1. Extend SOV_TEST_PAUSE_SEQUENCER_UPDATE_STATE to accept a specific node_id value (in addition to "1" for all nodes), enabling per-node pause/resume of preferred batch production in tests                                                                 
2. Add pause_preferred_batches_for_node / resume_preferred_batches_for_node helpers to TestRollup
3. Add test_db_elected_leader_recovery_with_replica_when_only_leader_is_paused which verifies that pausing only the elected leader (while the replica continues) correctly triggers recovery and restores cluster functionality    

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/mod.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/mod.rs
@@ -627,6 +627,17 @@ pub(crate) enum PreferredSeqOperation<S: Spec, Rt: Runtime<S>> {
     ),
 }
 
+/// Returns `true` when the test pause flag is set and applies to this node.
+///
+/// * `"1"` pauses all nodes (backward-compatible).
+/// * A specific `node_id` pauses only the node whose postgres config matches.
+fn should_skip_update_state(postgres_config: Option<&PostgresConfig>) -> bool {
+    let Ok(flag_value) = std::env::var("SOV_TEST_PAUSE_SEQUENCER_UPDATE_STATE") else {
+        return false;
+    };
+    flag_value == "1" || postgres_config.is_some_and(|c| c.node_id == flag_value)
+}
+
 #[tracing::instrument(skip_all, level = "debug")]
 async fn update_state_task_inner<S, Rt, Da>(
     seq: PreferredSequencer<S, Rt, Da>,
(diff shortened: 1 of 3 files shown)


crates/full-node/sov-sequencer/src/preferred/mod.rs
```diff
@@ -627,6 +627,17 @@ pub(crate) enum PreferredSeqOperation<S: Spec, Rt: Runtime<S>> {
     ),
 }
 
+/// Returns `true` when the test pause flag is set and applies to this node.
+///
+/// * `"1"` pauses all nodes (backward-compatible).
+/// * A specific `node_id` pauses only the node whose postgres config matches.
+fn should_skip_update_state(postgres_config: Option<&PostgresConfig>) -> bool {
+    let Ok(flag_value) = std::env::var("SOV_TEST_PAUSE_SEQUENCER_UPDATE_STATE") else {
```
theodorebugnet, line 635: I think if we ever want to pause more than one node, this would need to be rewritten again. How about using per-node env variables, something like `"SOV_TEST_PAUSE_SEQUENCER_UPDATE_STATE_GLOBAL" == 1` and `"SOV_TEST_PAUSE_SEQUENCER_UPDATE_STATE_NODE_{node_id}" == 1`?

Not blocking for this PR, especially if you don't think we'll be adding tests that pause multiple nodes any time soon.

examples/demo-rollup/tests/replica/recovery.rs
```diff
@@ -74,6 +71,77 @@ async fn test_db_elected_leader_recovery_with_replica() {
...
+        token_id,
+        receiver_addr,
+        0,
+    )
+    .await;
+
+    // Pause only the elected leader's sequencer update_state loop.
+    leader.pause_preferred_batches_for_node().await;
+
+    for _ in 0..30 {
```
theodorebugnet, line 116: Idea for the future: maybe we could have a `produce_blocks_to_trigger_recovery()` method that reads `std::env::var("SOV_TEST_CONST_OVERRIDE_DEFERRED_SLOTS_COUNT")` and calculates the number of blocks automatically. Otherwise we're manually calculating the threshold every time in each test

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/mod.rs
bkolad: Yeah I don't think we will have more than 2 nodes anytime soon.

examples/demo-rollup/tests/replica/recovery.rs
bkolad: I think this won't work because in this tests we are using automatic block production, so the number won't be precise. But I think I can improve the loop in the next PR.

PR #2629, opened 2026-03-24
Replica: Reject PG batches that outrun the executor rebase window.                                                                                                                                                                                         

bkolad:
# Description

Before this fix, a replica could accept an unbounded number of BatchStart messages from the leader while its update_state (node replay) was paused or slow.  But after enough PG batches, the next batch depends on kernel state newer than that rebase anchor (the last state the replica received from the node).

This PR adds a safety guard `ensure_replica_batch_start_within_rebase_window` that prevents a replica sequencer from accepting too many batches (via PG) without its executor rebasing to a fresh node checkpoint. It also adds a test for replica recovery when only the replica is paused, and bumps `STATE_ROOT_DELAY_BLOCKS` from 3 to 5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2026-03-30
+- #2629 Sequencer: Replica rejects batch starts that outrun the executor rebase window. `STATE_ROOT_DELAY_BLOCKS` was increased from 3 to 5 in `constants.toml`.
+
 # 2026-03-24
 - #2626 EVM: Fixes estimateGas value to match what will end up in the receipt of actually executed transaction
 - #2634 Test only changes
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # 2026-03-30
+<<<<<<< blaze/replica_test_5
```
preston-evans98: nit: Looks like we have a checked in merge conflict

constants.toml
```diff
@@ -63,7 +63,7 @@ DEFERRED_SLOTS_COUNT = 120
 MAX_VISIBLE_HEIGHT_INCREASE_PER_SLOT = 10
 # How many *rollup blocks* we wait before making a state root visible to the rollup.
 # This value should always be set to at least 1
-STATE_ROOT_DELAY_BLOCKS = 3
+STATE_ROOT_DELAY_BLOCKS = 5
```
preston-evans98, line 66: (Non-blocking, note): The rollup_starter repo already has this constant set to 5. Looks like a good change for that reason as well.

preston-evans98 (approved):

CHANGELOG.md
bkolad: fixed

PR #2630, opened 2026-03-25
AggregatedProof: Make inner proof verification independent of SP1 by implementing ZkVerifier for SP1.

bkolad:
# Description

This PR makes the `run_aggregation_program` method independent of the verifier type by implementing `ZkVerifier` for `SP1Verifier`. This PR doesn't change the circuit's logic; it is just code reorganization. 

  - Implement `SP1Verifier::verify` for target_os = "zkvm" using `sp1_zkvm::lib::verify::verify_sp1_proof`, replacing the previous todo!() stub                                                                                                                   
  - Rename `OuterCodeCommitmentHash(Vec<u8>)` to `CodeCommitmentHash([u32; 8])` — the type is now used for both inner and outer proof verification.
  - Refactor aggregation program to use `V: ZkVerifier` trait for proof verification instead of calling `sp1_zkvm` directly, making verification backend-agnostic                                                                                                 
  - Remove `CodeCommitment::encode/decode` methods.
  - Remove deserialize_and_verify_pub_data and outer_vk_hash_from_vkey_hash helper functions that are no longer needed 

## Linked Issues
- Related to  #2551

### DIFF
--- a/crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
+++ b/crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
@@ -3,7 +3,8 @@ use sov_aggregated_proof_shared::{
     AggPubData, AggregatedProofWitness, DeferredProofInput, StfPubData,
 };
 use sov_modules_api::da::BlockHeaderTrait;
-use sov_modules_api::{DaSpec, OuterCodeCommitmentHash, Spec, Storage};
+use sov_modules_api::ZkVerifier;
+use sov_modules_api::{CodeCommitmentHash, DaSpec, Spec, Storage};
 use sov_rollup_interface::common::SlotNumber;
 
 struct BoundaryData<Hash, Root> {
@@ -24,9 +25,13 @@ type VerifyResult<S, Da> = VerifiedProofData<
     <<S as Spec>::Storage as Storage>::Root,
 >;
 
-pub(crate) fn run_aggregation_program<S: Spec<Da = Da>, Da: DaSpec>(
+pub(crate) fn run_aggregation_program<
+    S: Spec<Da = Da>,
+    Da: DaSpec,
+    V: ZkVerifier<CodeCommitment = CodeCommitmentHash>,
+>(
     witness: AggregatedProofWitness<Da>,
-    inner_vkey_hash: [u32; 8],
+    inner_vkey_hash: CodeCommitmentHash,
 ) {
     let proof_inputs = witness.proof_inputs;
     let outer_vkey_hash = witness.outer_vkey_hash;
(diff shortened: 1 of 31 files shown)


crates/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -118,17 +118,6 @@ pub trait ZkvmHost: Clone + Send + Sync + 'static {
...
-    type DecodeError: Debug;
-
-    /// Encodes the commitment into a byte sequence. Any kind of serializer may be used,
-    /// as long as this method is inverted by the [`CodeCommitment::decode`] method.
-    fn encode(&self) -> Vec<u8>;
-
-    /// Decodes the commitment from a byte sequence.
-    ///
-    /// This method must be the inverse of the [`CodeCommitment::encode`] method.
-    fn decode(data: &[u8]) -> Result<Self, Self::DecodeError>;
```
bkolad, line 131: These two methods were never used, so I removed them to simplify the `CodeCommitment` trait.

crates/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -143,10 +132,10 @@ pub trait ZkVerifier: Default + Clone + Send + Sync + 'static {
     /// The error type which is returned when a proof fails to verify
     type Error: Debug;
 
-    /// Interpret a sequence of a bytes as a proof and attempt to verify it against the code commitment.
-    /// If the proof is valid, return a public outputs of the proof.
+    /// Interpret a sequence of a bytes as a receipt and attempt to verify it against the code commitment.
+    /// If the receipt is valid, return a public outputs of the proof.
     fn verify<T: DeserializeOwned>(
-        serialized_proof: &[u8],
+        serialized_receipt: &[u8],
```
bkolad, line 132: When verifying an aggregated proof, we pass public values rather than the proof itself, so I made the naming more generic.

crates/full-node/sov-aggregated-proof/program/src/aggregation.rs
```diff
@@ -24,27 +25,30 @@ type VerifyResult<S, Da> = VerifiedProofData<
...
-            outer_vkey_hash,
-        ))
-    } else {
-        None
-    };
-
-    let verified_proof_data: VerifyResult<S, Da> =
-        verify_proof_chain::<S, Da>(proof_inputs, inner_vkey_hash, previous_public_data.as_ref());
+    let previous_public_data = prev_outer_proof_witness.map(|prev_outer_proof_witness| {
+        V::verify::<AggPubData<S, Da>>(&prev_outer_proof_witness.public_values, &outer_vkey_hash)
```
bkolad, line 43: This is the important part. We no longer use `deserialize_and_verify_pub_data`, which depended on `SP1`. Now the verification part is generic and will work for any ZkVm.

crates/rollup-interface/src/state_machine/zk/mod.rs
ross-weir: Do we need this trait at all anymore?

Could have the bounds on `ZkVerifier::CodeCommitment` directly and remove the need to impl an empty `CodeCommitment` trait on the types

ross-weir (approved):

crates/rollup-interface/src/state_machine/zk/mod.rs
bkolad: Fixed

PR #2631, opened 2026-03-25
demo-rollup: same STF setup as in starter

citizen-stig:
# Description

* Extract demo-stf-declaration from demo-stf, to unify approach we have in starter.
  * Prevent partial autogenerated.rs reads
  * Reduce build time by selecting spec at compile time in build.rs 
* Remove arbitrary feature from demo-rollup dependencies.

hyperlane and revenue share modules will be added in follow PR

----

## Testing

## Docs
No updates

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2631" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


examples/demo-rollup/build.rs
```diff
@@ -1,8 +1,3 @@
-use demo_stf::runtime::Runtime;
```
citizen-stig, line 1: I believe whole file can be deleted

examples/demo-rollup/stf/src/test_utils.rs
```diff
@@ -27,42 +22,3 @@ where
         &config.sequencer_registry
     }
 }
-
-impl<S: Spec> GenesisConfig<S>
```
citizen-stig, line 31: Hm, why not keep it here....

examples/demo-rollup/stf/build.rs
```diff
@@ -1,3 +1,17 @@
...
+use sov_mock_da::MockDaSpec;
+use sov_mock_zkvm::MockZkvm;
+use sov_modules_api::execution_mode::Native;
+
+type S = sov_modules_api::configurable_spec::ConfigurableSpec<
+    MockDaSpec,
+    MockZkvm,
+    MockZkvm,
+    MultiAddressEvmSolana,
+    Native,
```
citizen-stig: Hm, why native....

examples/demo-rollup/stf/build.rs
```diff
@@ -1,3 +1,22 @@
...
+use sov_mock_da::MockDaSpec;
+use sov_mock_zkvm::MockZkvm;
+
+#[cfg(feature = "native")]
+type ExecMode = sov_modules_api::execution_mode::Native;
+
+#[cfg(not(feature = "native"))]
+type ExecMode = sov_modules_api::execution_mode::Zk;
+
+type S = sov_modules_api::configurable_spec::ConfigurableSpec<
```
theodorebugnet, line 12: Technically I think this isn't correct, as the types depend on the Spec used in the actual rollup, e.g. if a callmessage were to depend on type from DaSpec then the generated schema would be wrong. But getting this right doesn't seem possible because we don't feature-gate the different adapters (unlike the starter). Do you know if there's a reason demo-rollup is configured with CLI args rather than with cargo features?

It seems that this is adding `mock_da` and `mock_zkvm` as dependencies of the prover guests, which is also an odd side-effect.

Anyway this is how it used to be so non blocking for this PR, just wondering if we can improve this as well.

theodorebugnet (approved):
Nice, good to have parity in how we set this up

examples/demo-rollup/stf/build.rs
citizen-stig: > doesn't seem possible because we don't feature-gate the different adapters (unlike the starter). 

Yes, exactly.

> Do you know if there's a reason demo-rollup is configured with CLI args rather than with cargo features?

I think it has purely historical: convinience and cargo hack speed. 

I think it can be improved in follow up PRs

PR #2633, opened 2026-03-25
feat: window-based nonces

vmmon:
# Description

This implements a hybrid approach between Nonces and Generations to solve the performance problems we observed with both approaches.

The differences to Nonces are:
- numbers don't have to be consecutive
- instead a certain window of older TXs are still accepted to have multiple TXs in-flight

The difference to Generations are:
- nonces need to be unique to avoid tracking the full TX hash 
- this reduces the storage size by 256x

Known issues:
- it still has the same counter saturating problem as the generations
- it is a breaking change as a new Uniqueness variant is added

*Summary of the changes...*

## Linked Issues
- Related to #2564

## Testing
- added unit-tests and tried the code in our rollup

## Docs


crates/module-system/module-implementations/sov-uniqueness/src/lib.rs
```diff
@@ -34,6 +35,10 @@ pub struct Uniqueness<S: Spec> {
     #[state]
     pub(crate) nonces: StateMap<CredentialId, u64>,
 
+    /// Mapping from a credential id to a window of seen nonces.
+    #[state]
+    pub(crate) window: StateMap<CredentialId, (u64, Vec<u8>)>,
```
preston-evans98, line 40: Note to reviewers: This is non-breaking since it's added at the end of the module declaration

crates/module-system/module-implementations/sov-uniqueness/src/window.rs
```diff
@@ -0,0 +1,268 @@
...
+// #[derive(Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, Default)]
+// pub struct Window {
+//     highest_seen_nonce: u64,
+//     /// A bitmap representing PAST_TRANSACTION_WINDOW nonces below the highest seen nonce.
+//     /// Each entry contains a 1 if that nonce has been seen, 0 otherwise.
+//     ///
+//     /// For example, if the highest seen nonce is 10 and PAST_TRANSACTION_WINDOW is 5, then index `4` in the bitmap represents nonce `9`
+//     /// and index `0` represents nonce `5`.
+//     bits: BitMap,
+// }
```
theodorebugnet:
```suggestion

```

crates/module-system/module-implementations/sov-uniqueness/src/window.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        anyhow::ensure!(
+	    nonce >= start,
+	    "Tx outdated for credential id: {credential_id}, expected at least: {start}, but found: {nonce}");
+
+        // The offset into the bits array that represents the nonce. This is (nonce - start - 1). For example,
+        // if nonce is 11 and start is 10, then offset is 0.
+        //
+        // If the nonce is very large the outcome could be larger than usize::MAX; that's fine, `get_bit` will return false in that case.
+        // Safety: Nonce > start, so nonce - start is positive, and that quantity minus 1 is non-negative.
```
theodorebugnet:
```suggestion
```

crates/module-system/module-implementations/sov-uniqueness/src/window.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    pub(crate) fn mark_window_tx_attempted(
+        &mut self,
+        credential_id: &CredentialId,
+        nonce: u64,
+        state: &mut impl StateAccessor,
+    ) -> anyhow::Result<()> {
+        let mut window = self.window.get(credential_id, state)?.unwrap_or_default();
+
+        // this assertion ensures that `check_window_uniqueness()` was executed beforehand
```
theodorebugnet: Nit: I don't think it does. If we want a defensive assert here, we could make add_nonce fallible (for example: have `set_bit` throw an error if `self.0[byte_index] & (1 << bit_index) != 0` before setting it - there's no legitimate circumstance where we would need to double-set the same bit).

But we already assume the capability is used correctly (e.g. if it wasn't, `nonce`s could get double-incremented and there's no protection against that), so the simplest fix would be to just drop the comment

constants.testing.toml
```diff
@@ -46,6 +46,8 @@ UNREGISTERED_BLOBS_PER_SLOT = 5
 # How many past transaction generations into the past to keep per-account buckets for
 # Higher numbers give more flexibility to avoid collisions, lower numbers reduce the maximum amount of storage used
 PAST_TRANSACTION_GENERATIONS = 5_000
+# How many past transactions to keep in the windowed uniqueness.
+PAST_TRANSACTION_WINDOW = { const = 1_000 }
```
theodorebugnet: Bikeshedding: I feel like `PAST_TRANSACTIONS_WINDOW` (plural) would make more sense.
Alternatively something completely different like `PAST_NONCES_WINDOW` because it's not actually measuring anything about past transaction(s) but the oldest available nonce when using windowing.

constants.testing.toml
```diff
@@ -46,6 +46,8 @@ UNREGISTERED_BLOBS_PER_SLOT = 5
 # How many past transaction generations into the past to keep per-account buckets for
 # Higher numbers give more flexibility to avoid collisions, lower numbers reduce the maximum amount of storage used
 PAST_TRANSACTION_GENERATIONS = 5_000
+# How many past transactions to keep in the windowed uniqueness.
```
theodorebugnet: Might be worth adding a note that it's the _maximum_ number, and in reality up to 8 fewer might be allowed depending on pruning (since we prune the bitfield a byte at a time)?

I doubt the precise window size will be critical very often, but better to document than not IMO.

crates/module-system/sov-modules-api/src/runtime/capabilities/authorization.rs
```diff
@@ -76,6 +76,11 @@ pub enum UniquenessData {
     /// Transactions older than this buffer are invalid, transactions falling within it or with a
     /// higher generation are valid but must have a unique hash within their generation
     Generation(u64),
+    /// Window-based uniqueness: each transaction of an account must have a unique but not
+    /// necessarily consecutive nonce. Older unique nonces upto `PAST_TRANSACTION_WINDOW` are still
+    /// accepted. Note that `PAST_TRANSACTION_WINDOW` will be rounded down to the next multiple of
+    /// eight to simplify the code.
```
theodorebugnet:
```suggestion
    /// accepted.
```
As far as I can tell that's not the case anymore

crates/module-system/module-implementations/sov-uniqueness/src/window.rs
```diff
@@ -0,0 +1,268 @@
...
+            bits: BitMap(VecDeque::from(bits)),
+        }
+    }
+}
+
+// Cases:
+// nonce = 8, start = 0, PAST_TRANSACTION_WINDOW = 7; unaligned = r. Round up to 8
+// nonce = 8, start = 0, PAST_TRANSACTION_WINDOW = 8; unaligned = 0. Stay the same
+impl Window {
+    /// Adds a nonce to the set of seen nonces, adjusting the window if necessary.
```
theodorebugnet, line 47:
```suggestion
    /// Adds a nonce to the set of seen nonces, adjusting the window if necessary.
    ///
    /// Safety: will panic if `nonce` < `self.start_nonce`.
```

It looks like this function assumes that as an invariant, so worth documenting IMO

crates/module-system/module-implementations/sov-uniqueness/src/window.rs
```diff
@@ -0,0 +1,268 @@
...
+    fn add_nonce(&mut self, nonce: u64) {
+        let old_start_nonce = self.start_nonce;
+
+        // The current window is start_nonce..(start_nonce + PAST_TRANSACTION_WINDOW);
+        // That means the max offset we can have without adjusting the window is PAST_TRANSACTION_WINDOW - 1.
+        let new_start_nonce =
+            if nonce >= self.start_nonce.saturating_add(PAST_TRANSACTION_WINDOW - 1) {
```
theodorebugnet:
```suggestion
            if nonce >= self.start_nonce.saturating_add(PAST_TRANSACTION_WINDOW) {
```
As the comment says, the max offset we can have without adjusting the window is `PAST_TRANSACTION_WINDOW - 1`, so we don't need to branch until `offset >= PAST_TRANSACTION_WINDOW`.

I believe right now the calculations would result in a no-op when this branch is one value too early (as it would "round up" to `old_start_nonce` exactly), but it's semantically wrong.

theodorebugnet (approved):
LGTM with some doc adjustments and a couple of nits, plus a naming question

crates/module-system/sov-modules-api/src/runtime/capabilities/authorization.rs
vmmon: That is correct.

crates/module-system/module-implementations/sov-uniqueness/src/window.rs
```diff
@@ -0,0 +1,268 @@
...
+                // In this branch, we're adjusting the window.
+                let unaligned_start_nonce = nonce - (PAST_TRANSACTION_WINDOW - 1);
+                // Safety: PAST_TRANSACTION_WINDOW is greater than 7, so unaligned_start_nonce + 7 <= u64::MAX
```
vmmon: There is an overflow possible here if the `start_nonce` is zero.

preston-evans98: I'm not seeing the overflow here. If start_nonce is `0`, then `nonce >= (PAST_TRANSACTION_WINDOW - 1)`, so subtracting `PAST_TRANSACTION_WINDOW - 1` should be safe. What am I missing?

crates/module-system/module-implementations/sov-uniqueness/src/window.rs
preston-evans98: Removed.

constants.testing.toml
preston-evans98: Good idea. Done.

crates/module-system/module-implementations/sov-uniqueness/src/window.rs
preston-evans98: Good catch! Adjusted!

constants.testing.toml
preston-evans98: I like it. Switched to the plural

crates/module-system/module-implementations/sov-uniqueness/src/window.rs
vmmon: I have read a `saturating_sub` here. So sorry my fault.

PR #2645, opened 2026-03-26
Move aggregated proof types and circuit logic into `sov-rollup-interface`

bkolad:
# Description

This PR does not change the circuit logic. It just moves the circuit to `sov-rollup-interface`.

 - Convert `aggregated_proof.rs` into a module folder with circuit and common submodules in `sov-rollup-interface`                                                                                                                                               
  - Move shared witness types `(DeferredProofInput, AggregatedProofWitness, PreviousOuterProofWitness)` from `sov-aggregated-proof-shared` into `aggregated_proof::common`
  - Move aggregation circuit logic from `program/src/aggregation.rs` into `aggregated_proof::circuit`, re-parameterized with direct type params (Address, Da, Root) instead of Spec/Storage to avoid circular dependencies                                        
  - Remove the sov-aggregated-proof-shared crate entirely 

## Linked Issues
- Related to #2551

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2645" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/rollup-interface/src/state_machine/zk/aggregated_proof/circuit.rs
```diff
@@ -19,36 +24,44 @@ struct VerifiedProofData<Address, Hash, Root> {
...
+/// proofs and an optional previous outer proof, checks DA and state-root
+/// continuity, then commits [`AggregatedProofPublicData`] as the public output.
+pub fn run_aggregation_program<Address, Da, Root, V, G>(
+    inner_vkey_hash: CodeCommitmentHash,
+    guest: G,
+) where
+    Address: Clone + Serialize + DeserializeOwned,
     Da: DaSpec,
+    Root: Clone + Debug + PartialEq + Serialize + DeserializeOwned,
     V: ZkVerifier<CodeCommitment = CodeCommitmentHash>,
```
citizen-stig, line 39: I think it is pretty valid comment from Devin:

```
crates/rollup-interface/src/state_machine/zk/aggregated_proof/circuit.rs:R33-42

**ZkvmGuest verifier type is not constrained to match the explicit V type parameter**

In `run_aggregation_program`, the function takes both `V: ZkVerifier<CodeCommitment = CodeCommitmentHash>` and `G: ZkvmGuest` as independent type parameters. There is no constraint that `G::Verifier == V`. In principle, a caller could pass a guest whose associated verifier differs from the one used for proof verification. The current caller in `program/src/main.rs:22-28` correctly uses `SP1Verifier` and `SP1Guest` (where `SP1Guest::Verifier = SP1Verifier`), so this is not a runtime issue today. However, since the function is now `pub` in a core crate, a future caller could accidentally mismatch these. Consider adding a bound like `G: ZkvmGuest<Verifier = V>` to make the constraint explicit.
```

crates/rollup-interface/src/state_machine/zk/aggregated_proof/common.rs
```diff
@@ -0,0 +1,35 @@
+use serde::{Deserialize, Serialize};
+
+use super::CodeCommitmentHash;
+use crate::da::DaSpec;
+
+/// A single deferred proof input containing its public values and associated DA block header.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct DeferredProofInput<Da: DaSpec> {
```
citizen-stig, line 12: Same good comment about serde bound:

```
crates/rollup-interface/src/state_machine/zk/aggregated_proof/common.rs:R7-13

**DeferredProofInput lacks explicit serde bounds, relying on default derive behavior**

The `DeferredProofInput<Da: DaSpec>` struct at `common.rs:7-13` uses `#[derive(Serialize, Deserialize)]` without explicit `#[serde(bound(...))]`. Serde's default derive behavior adds bounds on the type parameter itself (`Da: Serialize`), not on the associated type (`Da::BlockHeader: Serialize`). This is the same as the old code in `shared/src/lib.rs` and works for `MockDaSpec` (a unit struct), but could be fragile for DA specs that don't implement `Serialize` directly. The parent struct `AggregatedProofWitness` correctly uses explicit serde bounds — consider doing the same for `DeferredProofInput` for consistency and robustness.
```

crates/rollup-interface/src/state_machine/zk/aggregated_proof/circuit.rs
bkolad: fixed

crates/rollup-interface/src/state_machine/zk/aggregated_proof/common.rs
bkolad: fixed

citizen-stig (approved):

Issue #2652, opened 2026-03-27
State root mismatch during sequencer recovery


bkolad:
## Summary

When a node completes recovery, it can panic with the following error:
```
thread 'tokio-runtime-worker' (51082957) panicked at /Users/blaze/programming/sovereign-workspaces/sdk_2/sovereign-sdk/crates/full-node/sov-sequencer/src/preferred/state_root_compute.rs:82:9:
User state root has changed at rollup_height=1 and slot_number=10. This is a bug.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Steps to reproduce:
1. git checkout `blaze/recovery_debug`
2. `cargo test test_recovery_in_loop -- --nocapture`

**Root cause**
Recovery does not continue from the old speculative executor state. Instead, the sequencer creates a fresh executor and resets its transaction and event caches, so all data associated with the old executor should be cleared at that point. However, it also runs a side task, `StateRootTask`, which maintains a separate cache that does not appear to be cleared during recovery. As a result, the sequencer may reuse stale cached entries from `StateRootTask`, which can lead to the panic above.

PR #2655, opened 2026-03-30
 Fix recovery panic caused by stale StateRootTask cache entries

bkolad:
# Description

See the relevant issue: #2652

  This PR fixes a bug where the preferred sequencer could reuse stale `StateRootTask` cache entries after the speculative executor was replaced.

  When recovery creates a fresh executor, the old speculative state is intentionally discarded. However, `StateRootTask` was keeping a separate cache keyed only by `rollup_height`, so the new executor could still hit cache entries produced by the old executor. That could trigger the panic:
  `User state root has changed at rollup_height=...`

  ## What this fixes

  - Prevents stale state-root cache entries from surviving an executor replacement
  - Fixes the recovery panic caused by cross-executor cache reuse
  - Preserves cache entries built by the replacement executor during replay/catchup

  ## How

  Each state-root request now carries the executor `executor_id`, and `StateRootTask` caches roots by `(executor_id, rollup_height)` instead of just rollup_height.

  That means:
  - old executor entries stay isolated to the old generation
  - the new executor cannot reuse stale roots from the discarded executor
  - replay still keeps the replacement executor’s own cache after swap

## Linked Issues
- Related to #2652

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/state_root_compute.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/state_root_compute.rs
@@ -12,6 +12,7 @@ use sov_state::{NativeStorage, ProvableNamespace, SlotKey, SlotValue, StateAcces
 use tokio::sync::{mpsc, oneshot};
 use tokio::task::JoinHandle;
 use tracing::{debug, error, info, span, trace, Level};
+use uuid::Uuid;
 
 /// The memory limit for old write sets that we keep around. These old sets are useful because they tell us what keys have been changed.
 /// Which lets us output a much handier error message when the state root computation changes.
@@ -21,6 +22,7 @@ const NUM_STATE_ROOT_COMPUTE_REQUESTS: usize = 50;
 
 type Hasher<S> = <<S as Spec>::CryptoSpec as CryptoSpec>::Hasher;
 pub(crate) struct StateRootComputeRequest<S: Spec> {
+    pub executor_id: Uuid,
     pub raw_state_changes: Arc<RawStateChanges>,
     pub uncommitted_changes: SequencerStateChanges<Hasher<S>>,
     pub storage: S::Storage,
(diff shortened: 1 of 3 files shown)


crates/full-node/sov-sequencer/src/preferred/state_root_compute.rs
```diff
@@ -323,7 +327,7 @@ impl<S: Spec> StateRootTask<S> {
     }
 
     fn prune_cache(
-        cached_results: &mut BTreeMap<RollupHeight, StateRootCacheEntry<S>>,
+        cached_results: &mut BTreeMap<(Uuid, RollupHeight), StateRootCacheEntry<S>>,
```
preston-evans98: This will prune entries for older Uuid versions first, which I think might cause problems. IIRC we sometimes have two concurrent executors - the "main" one that's being used to accept new transactions, plus a replacement executor that is currently re-executing old batches to catch up to the main one. 

IIRC, the replacement executor also sends items to the state root compute task. Since it has a newer Uuid and Uuid::v7 starts with a timestamp, this will cause entries from the "main" executor to get pruned even if they're for a newer rollup height. 

I think it's safer to sort by RollupHeight first, then Uuid. That way, older heights get pruned first and the risk of pruning itmes that are needed by the live executor is lower

bkolad: excellent point I will fix that

preston-evans98: Alternatively, we could move the `self.id = other.id` exeuctor id change from `replace_state` (which is shared between the normal `resync` code path and `recovery`) to `force_overwrite_state` which is only used during recovery. Since the state root is only allowed to change during recovery, I think this should give us the guarantee we want without changing the semantics of resync at all

preston-evans98 (approved):

crates/full-node/sov-sequencer/src/preferred/state_root_compute.rs
bkolad: fixed

PR #2658, opened 2026-03-30
Solana-offchain authenticator: add multisig support

theodorebugnet:
# Description

Adds multisig signing suport to the `sov-solana-offchain-auth` authenticator, for `Simple` (i.e. non-Ledger) transactions. (Ledger support has separate considerations, so it's been deprioritised for now: #2659)
The serialized message uses an `0x80` discriminator byte for deserializing. This mirrors the `0xff` byte used for ledger messages, but from the other side of the extended ASCII range (the first extended byte rather than the last). It also shares the property that it's not a valid UTF-8 starting byte.

The multisig commits to the version, and the credential_id encoded as the rollup's default address. I considered a few alternatives here:
* Encoding as `Base58Address`: this would make a difference for rollups using a MultiAddress enum; users signing with Solana wallets would see a Solana-formatted multisig address. I decided against this because the multisig address is entirely synthetic, and in theory the same multisig could be signed from any wallet; having a single canonical format in all authenticators (i.e. using the rollup's default `Spec::Address::From<CredentialId>`) ensures the multisig credential is consistent when presented to users.
* Resolving the actual account address from `sov_accounts`: not only would this complicate the implementation (and the client) significantly, but it's also useful to commit to the actual multisig credential rather than the account it might map to. E.g. if several different multisigs with different thresholds or signing sets are mapped to the same account, it's important for a user to be able to sign a partial transaction for one specific multisig set and not have it be reusable for another multisig set even for the same account. But this does need to be clearly documented for users, that the `"multisig_address"` is the "raw" address of the specific multisig credential and may not necessarily match the "account address".

Typescript support coming in a follow-up PR.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2658" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/module-system/sov-solana-offchain-auth/src/authentication.rs
+++ b/crates/module-system/sov-solana-offchain-auth/src/authentication.rs
@@ -83,11 +83,11 @@ pub struct SolanaOffchainUnsignedTransactionV1<R: TransactionCallable, S: Spec> 
     /// from malicious chains (if the chain name matches some other chain the use but didn't expect
     /// to be signing for right now).
     pub chain_name: SafeString,
-    /// The credential_id derived from the multisig parameters (hash of min_signers + sorted
-    /// pubkeys). Included in the signed message so that signers commit to the multisig
-    /// configuration and prevent credential_id malleability from reusing signed bytes in a
-    /// different multisig envelope.
-    pub credential_id: CredentialId,
+    /// The credential_id (i.e. multisig address) derived from the multisig parameters (hash of
+    /// min_signers + sorted pubkeys). Included in the signed message so that signers commit to
+    /// the multisig configuration and prevent credential malleability from reusing signed bytes
+    /// in a different multisig envelope.
+    pub multisig_address: CredentialId,
     /// Message format version. Must be `1` for this struct.
     pub version: u8,
 }
@@ -594,7 +594,7 @@ where
                 ));
             }
             verify_multisig_commitment::<S>(
-                tx.credential_id,
+                tx.multisig_address,
                 signatures,
                 unused_pub_keys,
                 *min_signers,
--- a/crates/module-system/sov-solana-offchain-auth/src/authentication.rs
+++ b/crates/module-system/sov-solana-offchain-auth/src/authentication.rs
@@ -89,13 +89,28 @@ pub struct SolanaOffchainUnsignedTransactionV1<R: TransactionCallable, S: Spec> 
     /// from malicious chains (if the chain name matches some other chain the use but didn't expect
     /// to be signing for right now).
     pub chain_name: SafeString,
-    /// The multisig address derived from the multisig parameters (hash of min_signers + sorted
+    /// The multisig credential derived from the multisig parameters (hash of min_signers + sorted
     /// pubkeys), formatted in the rollup's native address format. Included in the signed message
     /// so that signers commit to the multisig configuration and prevent credential malleability
     /// from reusing signed bytes in a different multisig envelope.
-    pub multisig_address: S::Address,
+    /// This is the "multisig address" except if the credential is mapped to another address in
+    /// `sov-accounts`.
+    pub multisig_id: S::Address,
     /// Message format version. Must be `1` for this struct.
+    #[serde(deserialize_with = "deserialize_version_1")]
     pub version: u8,
+}
+
+fn deserialize_version_1<'de, D: serde::Deserializer<'de>>(
+    deserializer: D,
+) -> Result<u8, D::Error> {
+    let v = <u8 as serde::Deserialize>::deserialize(deserializer)?;
+    if v != 1 {
+        return Err(serde::de::Error::custom(format!(
+            "expected message version 1, got {v}"
+        )));
+    }
+    Ok(v)
 }
 
 impl<R, S> SolanaOffchainUnsignedTransactionV1<R, S>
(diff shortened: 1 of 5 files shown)


crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -202,52 +442,85 @@ fn unpack_solana_message<S: Spec>(raw_tx: &[u8]) -> Result<UnpackedSolanaMessage
...
+    let raw_message: SolanaOffchainSimpleMessage<S> =
+        borsh::from_slice(raw_tx).map_err(|e| FatalError::DeserializationFailed(e.to_string()))?;
+
+    Ok(UnpackedSolanaMessage::V0 {
+        pub_key: raw_message.pubkey,
+        signature: raw_message.signature,
+        chain_hash: raw_message.chain_hash,
+        signed_bytes: raw_message.signed_message,
+        json_start: 0,
+    })
```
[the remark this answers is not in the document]
theodorebugnet: That's true. Added `deny_unknown_fields`

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -60,7 +68,52 @@ where
...
+    pub details: TxDetails<S>,
+    /// The chain name, so that users can verify the destination chain and avoid replay attacks
+    /// from malicious chains (if the chain name matches some other chain the use but didn't expect
+    /// to be signing for right now).
+    pub chain_name: SafeString,
+    /// The multisig address derived from the multisig parameters (hash of min_signers + sorted
+    /// pubkeys), formatted in the rollup's native address format. Included in the signed message
+    /// so that signers commit to the multisig configuration and prevent credential malleability
+    /// from reusing signed bytes in a different multisig envelope.
+    pub multisig_address: S::Address,
```
preston-evans98: nit: As discussed at standup
```suggestion
    pub multisig_id: S::Address,
```

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -60,7 +68,52 @@ where
...
+    /// Message format version. Must be `1` for this struct.
+    pub version: u8,
```
preston-evans98, line 101: (Question): Where does this version field come from?

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -173,27 +289,158 @@ fn verify_solana_signature<S: Spec>(
...
+            meter
+                .charge_linear_gas(S::gas_to_charge_for_credential(), num_keys)
+                .map_err(|e| AuthenticationError::OutOfGas(e.to_string()))?;
+
+            let all_keys: Vec<_> = signatures
+                .iter()
+                .map(|s| s.pub_key.clone())
+                .chain(unused_pub_keys.iter().cloned())
+                .collect();
+            let multisig = Multisig::new(*min_signers, all_keys);
```
preston-evans98, line 412: note for other reviewers - the multisig struct sorts the keys internally so the credential id is stable

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -202,52 +449,85 @@ fn unpack_solana_message<S: Spec>(raw_tx: &[u8]) -> Result<UnpackedSolanaMessage
...
 
-        // Calculate actual message length (excluding preamble)
-        let actual_message_length = envelope.signed_message_with_preamble.len() - PREAMBLE_LEN;
+    let preamble: RawSolanaOffchainMessagePreamble =
+        borsh::from_slice(&envelope.signed_message_with_preamble[0..PREAMBLE_LEN])
+            .map_err(|e| FatalError::DeserializationFailed(e.to_string()))?;
 
-        // Validate the preamble
-        preamble.validate(actual_message_length)?;
+    let actual_message_length = envelope.signed_message_with_preamble.len() - PREAMBLE_LEN;
```
preston-evans98:
```suggestion
    // Safety: we just checked that `envelope.signed_message_with_preamble.len()` >= `PREAMBLE_LEN`, so this can't underflow
    let actual_message_length = envelope.signed_message_with_preamble.len() - PREAMBLE_LEN;
```

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
```diff
@@ -282,39 +573,61 @@ where
...
+        UnpackedSolanaMessage::V1 {
+            signatures,
+            unused_pub_keys,
+            min_signers,
+            ..
+        } => {
+            let tx = SolanaOffchainUnsignedTransactionV1::<D, S>::unmetered_deserialize(json_slice)
+                .map_err(deser_err)?;
+            if tx.version != 1 {
+                return Err(AuthenticationError::FatalError(
```
preston-evans98: non-blocking: should we consider adding these checks into `unmetered_deserialize`? In general, maintaining the valid that invalid data is unrepresentable make the code easier to reason about

preston-evans98 (commented):
Changes LGTM overall - only open questions are the tx version field and the `0x80` discriminant byte (should we consider `\xffsolanaoffchainmultisig`, for example?)

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
theodorebugnet: That means V1 transaction (i.e. multisig).

crates/module-system/sov-solana-offchain-auth/src/authentication.rs
theodorebugnet: That's a good idea, at least for the version check. I'll probably leave the multisig commitment check outside, since it pertains more to signature verification than the data being invalid, and passing the envelope multisig data into the deserialize method would be unusual.

preston-evans98 (approved):

PR #2677, opened 2026-04-01
Fix minor proving workflow bugs

preston-evans98:
# Description
This small PR fixes two bugs in the sequencer. It ports only the functional changes from https://github.com/Sovereign-Labs/sovereign-sdk/pull/2605. The tests will be ported later, but require redesign due to recent changes to the sequencer.

The first bug was that recovery mode used `trigger_batch_production_if_convenient`. This meant that if the blob sender got busy (or for a variety of other reasons), batches would not actually be produced and catch up would not occur. 

Second, the new proof processing logic had a bug where we pruned proofs from our list to process based on the *target* batch version not the current batch version. (This is hard to explain but easy to see when you look at the code - I've flagged it in a github comment below.)

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/db/mod.rs
@@ -295,9 +295,9 @@ impl BlobsCache {
     /// Fetch all proofs that need to be played at the start of the batch with the given sequence number.
     pub fn proofs_for_replay(
         &self,
-        batch_sequence_number: SequenceNumber,
+        target_batch_sequence_number: SequenceNumber,
     ) -> Vec<PreferredProofToReplay> {
-        self.sanity_check_batch_sequence_number_is_in_range(batch_sequence_number);
+        self.sanity_check_batch_sequence_number_is_in_range(target_batch_sequence_number);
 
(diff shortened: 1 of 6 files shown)


crates/full-node/sov-sequencer/src/preferred/db/mod.rs
```diff
@@ -306,11 +306,11 @@ impl BlobsCache {
                 ReadBlob::Batch(batch) => {
                     // Since we're looking for all proofs in between two batches, we either need to return (if the batch has the sequence number we're looking for)
                     // or we need to reset our output (since the proofs we've already seen would have been handled during processing of the batch we just hit).
-                    if batch.sequence_number == batch_sequence_number {
+                    if batch.sequence_number == target_batch_sequence_number {
                         break;
                     }
                     output.retain(|p: &PreferredProofToReplay| {
-                        p.sequence_number > batch_sequence_number
+                        p.sequence_number > batch.sequence_number
```
preston-evans98, line 313: This was the bug: we should prune proofs which are older than the batch we just encountered - the target batch has nothing to do with it.

theodorebugnet (approved):

PR #2679, opened 2026-04-01
More defensive StorableMockDaLayer block producing

citizen-stig:
# Description

The test `seq_many_invalid_txs` intermittently fails with:
```
     (code: 14) unable to open database file
```

The intermittent `SQLITE_CANTOPEN (code 14)` error was caused by excessive SQLite contention in `seq_many_invalid_txs`. 

These fixes applied:

1. Reduced connection pool from 50 to 5 The primary fix. SQLite is single-writer, and the `StorableMockDaLayer` is already behind a tokio::RwLock. 50 connections created ~150 file descriptors for one database (.sqlite + .sqlite-wal + .sqlite-shm per connection), causing `SQLITE_CANTOPEN` under heavy contention. Tests use concurrent_sync_tasks=1, so max concurrent readers is ~3. Pool of 5 = 3 readers + 1 writer + 1 buffer.
2. crates/adapters/mock-da/src/storable/entity/mod.rs:33-40 — Added PRAGMA busy_timeout = 5000 Defensive safety net. Tells SQLite to retry for up to 5 seconds instead of failing immediately on lock contention.

3. Wrap produce_block_with_timestamp in a transaction. This is the key fix. Previously, each of the 5 DB operations inside produce_block_with_timestamp independently checked out a pool connection, ran a query, and returned it. Under system-wide FD pressure, any checkout could fail.
  Now conn.begin() acquires one pool connection at the start and all operations run on it:
  ```
   begin()          ← checks out 1 connection from pool
     SELECT prev block hash
     SELECT blobs
     INSERT block header
     UPSERT finalized height
     SELECT finalized header
  commit()         ← returns connection to pool
  ```

  This eliminates per-operation pool contention entirely. The connection is reserved for the duration of the block production — no other operation can steal it. As a bonus, if any operation fails mid-way, the transaction rolls back automatically, preventing partial state.

4. retry_db helper for idempotent operations. A retry helper with exponential backoff (50ms base, 5 attempts) for transient SQLITE_CANTOPEN and database is locked errors. Only used for idempotent operations:
  - Conditional deletes in rewind_to_height (DELETE WHERE height > X — safe to re-execute) Not used for inserts (they can commit-then-error, causing UNIQUE violations on retry).
5. anyhow::Context on all DB operations. Every DB call now has a .context("operation name") so failures identify exactly which operation failed (e.g., insert block_header, get prev block hash).
6. finalized_height::update_value accepts impl ConnectionTrait. Changed from &DatabaseConnection to &impl ConnectionTrait so it works inside a transaction.

---

## Testing
Existing tests should pass

## Docs
No updates

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2679" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/adapters/mock-da/src/storable/layer.rs
+++ b/crates/adapters/mock-da/src/storable/layer.rs
@@ -561,23 +561,23 @@ impl StorableMockDaLayer {
         }
 
         let conn = &self.conn;
-        retry_db(|| async {
-            Blobs::delete_many()
-                .filter(blobs::Column::BlockHeight.gt(height))
-                .exec(conn)
-                .await?;
-            Ok(())
-        })
-        .await?;
+        let txn = retry_db(|| async { Ok(conn.begin().await?) })
+            .await
+            .context("begin transaction for rewind")?;
 
-        retry_db(|| async {
-            BlockHeaders::delete_many()
-                .filter(block_headers::Column::Height.gt(height))
-                .exec(conn)
-                .await?;
-            Ok(())
-        })
-        .await?;
+        Blobs::delete_many()
+            .filter(blobs::Column::BlockHeight.gt(height))
+            .exec(&txn)
+            .await
+            .context("delete blobs above rewind height")?;
+
+        BlockHeaders::delete_many()
+            .filter(block_headers::Column::Height.gt(height))
+            .exec(&txn)
+            .await
+            .context("delete block_headers above rewind height")?;
+
+        txn.commit().await.context("commit rewind transaction")?;
 
         let past_next_height = self.next_height;
         self.next_height = height + 1;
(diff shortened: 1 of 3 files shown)


bkolad (approved):

crates/adapters/mock-da/src/storable/layer.rs
```diff
@@ -121,12 +159,17 @@ impl StorableMockDaLayer {
...
 
+        let conn = &self.conn;
         let prev_block_hash = if self.next_height > 1 {
-            let block = BlockHeaders::find()
-                .filter(block_headers::Column::Height.eq(self.next_height - 1))
-                .one(&self.conn)
-                .await?
-                .expect("Previous block is missing from the database");
+            let prev_height = self.next_height - 1;
+            let block = retry_db(|| async {
```
citizen-stig: Concurrency issue here for some reason...

PR #2682, opened 2026-04-02
Upgrade axum from 0.7 to 0.8

citizen-stig:
# Description

* axum: 0.7.9 → 0.8.8
* axum-server: 0.7 → 0.8
* utoipa: 4.2 → 5.4
* utoipa-swagger-ui: 7.1.1-rc.0 → 9.0
* tokio-tungstenite: 0.28.0 → 0.29.0

Core changes

1. Route syntax — 13 routes across 5 files: :param → {param}
2. async_trait removal — 20 occurrences across 14 files
3. WebSocket — Message::Text/Binary/Ping now use Utf8Bytes/Bytes; close() → send(Close(None))
4. tcp_nodelay — ListenerExt::tap_io() replaces .tcp_nodelay(true)
5. utoipa 5 — ToSchema trait redesign (no lifetime), SchemaType moved, ContentBuilder::schema() takes Option, OpenAPI 3.0→3.1
6. Progenitor compat — downconvert_to_3_0() function in build.rs strips 3.1-specific constructs
7. axum-server 0.8 — Handle now generic: Handle<std::net::SocketAddr>
8. Custom extractors — Added OptionalFromRequestParts for Query<T>, S: Send + Sync bounds
9. Router — nest("/", ...) → merge(...) in generated code

---

## Testing
No updates, existing test are passing

## Docs
No updates

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2682" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/module-system/sov-modules-api/src/rest/__private/openapi-templates/runtime-base.yaml
```diff
@@ -113,8 +112,7 @@ components:
           title: "Boolean"
         - type: array
           title: "Array"
-          items:
-            x-stainless-any: true
```
citizen-stig, line 117: I think this might be a blocker

examples/demo-rollup/tests/resync/mod.rs
```diff
@@ -88,15 +87,8 @@ async fn check_value(client: &demo_stf_json_client::Client, expected: u64) {
...
-        RuntimeAnyJsonValue::Object(inner) => {
-            let state_value = inner.get("value").unwrap();
-            println!("State value: {state_value:?}");
-            let heavy_vec = state_value.as_array().expect("HeavyVec is not an array");
-            assert_eq!(heavy_vec.len(), expected as usize);
-        }
-        _ => panic!("Getting SyntheticLoad state value returned unexpected JSON shape."),
-    }
+    println!("State value: {:?}", response.value);
+    assert_eq!(response.value.len(), expected as usize);
```
theodorebugnet, line 91: Is `response.value.len()` definitely equivalent to the checks we had before? I'm wondering if we could get false-positives if, for example, `response.value` is a String rather than an Array (e.g. some error string), and happens to have the same `len()` as what we expected.

theodorebugnet (approved):
Approved with one question

examples/demo-rollup/tests/resync/mod.rs
citizen-stig: That's a good point, let me double check. I would prefer to be conservative in this test.

citizen-stig: Double checked. WIth new axum, it returns already typed:

```rust
 pub struct SyntheticLoadHeavyStateGetStateValueResponse {
        pub value: Vec<i64>,
    }
```

so this assert is equivalent of what was on line 96.

PR #2688, opened 2026-04-02
Versioned unsigned transaction with multisig credential

theodorebugnet:
# Description

This is the initial implementation of the credential_id malleability fix, by splitting `UnsignedTransaction` into an enum with a V0 (matching the old UnsignedTransaction struct) and a V1, now containing also the multisig address. This ensures that signed bytes commit to a) the version (as the enum discriminant), preventing cross-version reuse/replay of signed messages; and b) for multisigs, to the exact credential_id, preventing replace of the signed bytes with a different min_signers or pubkey set.

The key changes to review are in `crates/module-system/sov-modules-api/src/transaction/*`. The rest of the changes mostly amount to switching to explicit use of `UnsignedTransactionV0` for all the tests and helpers which explicitly only deal with V0 transactions. (It's more convenient to use V0 rather than the outer enum, because functions like `sign()` are not well-defined for the versioned enum.)

As discussed this uses `theodore/multisig-upgrade` as the base branch to avoid propagating breaking changes to `dev` until we're ready.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/sov-modules-api/src/transaction/types/v1.rs
```diff
@@ -102,15 +135,22 @@ impl<Call: BorshSerialize, S: Spec, C: CryptoSpecExt> Version1<Call, S, C> {
         self.add_signature(signature, key.pub_key())
     }
 
-    /// Serializes only the `UnsignedTransaction` part of the transaction and appens the chain_hash
+    /// Serializes the versioned unsigned transaction and appends the chain_hash, producing the
+    /// bytes to be signed. The output is identical to
+    /// `borsh(UnsignedTransaction::V1(UnsignedTransactionV1{...})) || chain_hash`.
     pub fn serialize_for_signing(&self, chain_hash: &[u8; 32]) -> Vec<u8> {
-        let mut out = Vec::with_capacity(64); // Preallocate a little capacity to avoid excessive reallocations
+        let credential_address = self.credential_address();
```
theodorebugnet, line 142: I will try replacing this with `Transaction::serialize_with_chain_hash()` in a follow-up PR, because the generics for `Version0` and `Version1` types need to be changed for that to work. Right now this functoin duplicates the V1 serialization logic.

crates/module-system/sov-modules-api/src/transaction/types/v1.rs
```diff
@@ -102,15 +135,22 @@ impl<Call: BorshSerialize, S: Spec, C: CryptoSpecExt> Version1<Call, S, C> {
...
+        let mut out = Vec::with_capacity(128);
+        // V1 enum discriminant
+        out.push(1u8);
```
bkolad, line 145: Manually serializing the enum is very error-prone. Let’s create an issue and make sure this is addressed in the follow-up.

crates/module-system/sov-modules-api/src/transaction/unsigned.rs
```diff
@@ -10,33 +10,172 @@ use crate::{
...
+        serialize = "R::Call: BorshSerialize",
+        deserialize = "R::Call: BorshDeserialize",
+    ))]
+    pub runtime_call: R::Call,
+    /// The uniqueness identifier
+    pub uniqueness: UniquenessData,
+    /// Data related to fees and gas handling.
+    pub details: TxDetails<S>,
+    /// The multisig credential address, derived as
+    /// `hash(min_signers || sorted(pub_keys))` in the rollup's native address format.
```
bkolad: Where do we ensurte that the pub_keys are sorted (we don't sort in  `pub fn credential_address`)?

crates/module-system/sov-modules-api/src/transaction/unsigned.rs
```diff
@@ -10,33 +10,172 @@ use crate::{
...
 
-/// An unsent transaction with the required data to be submitted to the DA layer
+/// V0 unsigned transaction (single-sig). This is the consumer-facing builder type used
+/// to construct transactions before signing.
 #[derive(
     derive_more::Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize, UniversalWallet,
 )]
 #[serde(bound = "R::Call: serde::Serialize + serde::de::DeserializeOwned")]
-pub struct UnsignedTransaction<R: TransactionCallable, S: Spec> {
+pub struct UnsignedTransactionV0<R: TransactionCallable, S: Spec> {
```
bkolad: For Transaction, the `V0 and V1` variants live in `types/v0.rs` and `types/v1.rs`, but here they’re defined in the same file. Let's reorganize this so it follows the same code pattern. `unsigned.rs` could be a Rust module with separate v0 and v1 files.

bkolad (approved):

crates/module-system/sov-modules-api/src/transaction/unsigned.rs
```diff
@@ -10,33 +10,172 @@ use crate::{
...
+/// in signed bytes: `V0` = `0x00`, `V1` = `0x01`.
+///
+/// This is the type serialized (with the chain hash appended) to produce the bytes that
+/// are signed and verified. The enum discriminant ensures V0 and V1 signed bytes are
+/// distinct, preventing cross-version replay attacks.
+#[derive(
+    derive_more::Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize, UniversalWallet,
+)]
+#[serde(bound = "R::Call: serde::Serialize + serde::de::DeserializeOwned")]
+pub enum UnsignedTransaction<R: TransactionCallable, S: Spec> {
```
bkolad: The following crates still use the old `UnsignedTransaction` (struct) which I think will break clients.
1. crates/web3/src/schema.rs (it has its own copy of `UnsignedTransaction` that is outdated)
2. python/py_sovereign_web3/rust/src/lib.rs
3. typescript/packages/web3/src/rollup/standard-rollup.ts

bkolad: We should also check why none of our unit tests caught it.

crates/module-system/sov-modules-api/src/transaction/types/v1.rs
```diff
@@ -88,6 +90,37 @@ pub struct Version1<Call, S: Spec, C: CryptoSpecExt = <S as Spec>::CryptoSpec> {
     pub details: TxDetails<S>,
 }
 
+impl<Call, S: Spec, C: CryptoSpecExt> Version1<Call, S, C> {
+    /// Computes the multisig credential address from the transaction's multisig parameters.
+    /// This is the single source of truth for the derivation used in both signing and verification.
+    pub fn credential_address(&self) -> S::Address {
```
bkolad, line 96: The TypeScript helper still uses the old pattern. `/typescript/packages/multisig/src/index.ts#L123`, `fromTransactions()` only accepts V0 transactions and copies their `{ pub_key, signature }`. Then  `typescript/packages/multisig/src/index.ts#L261` wraps those same signatures into a V1 transaction. That means the helper is taking signatures
  that were made for a V0 message and claiming they authorize a V1 multisig message. After this upgrade, those are different bytes, so verification fails.

  Concretely, the broken flow is:

  1. Each signer signs an individual V0 tx.
  2. fromTransactions() extracts those V0 signatures.
  3. asTransaction() builds a V1 multisig envelope.
  4. The node verifies those signatures against V1 bytes, not V0 bytes.
  5. Result: bad signature, transaction rejected.

 The current integration test is actually built around the broken pattern: multisig.integration-test.ts#L70

crates/module-system/sov-modules-api/src/transaction/types/v1.rs
theodorebugnet: That's blocked by the generics in `Version0<>` and `Version1<>` being different, I am going to address this in an immediate follow-up if doing so doesn't break anything else.

crates/module-system/sov-modules-api/src/transaction/unsigned.rs
theodorebugnet: We do internally in `Multisig::credential_id()`

crates/module-system/sov-modules-api/src/transaction/types/v1.rs
theodorebugnet: I'll be updating typescript in follow-ups (since this is on a feature branch the breakage is not an issue for now)

crates/module-system/sov-modules-api/src/transaction/unsigned.rs
theodorebugnet: Good catch, I'll update the SDKs in a follow-up. I'll look into why the web3 integration tests didn't catch this

theodorebugnet: Also the python and rust web3 tests never run: #2733

PR #2692, opened 2026-04-03
  Simplify zkVM proof APIs by removing execute mode and `Proof` wrappers

bkolad:
# Description

1. Remove the `Proof<F, T>` enum from `rollup-interface` — adapters now serialize the proof/public data type directly instead of wrapping it                                                                                                                 
2. Remove the `with_proof: bool` parameter from `ZkvmHost::run()` — the method now always generates a proof
3. Remove `RollupProverConfig::Execute` variant, collapsing three prover modes (Skip/Execute/Prove) into two (Skip/Prove)                                                                                                                                    
4. Update all adapters (mock-zkvm, risc0, sp1), prover service, tests, and docs .

`RollupProverConfig::Execute` did not produce a real proof, so it was unclear how its output should be handled during proof aggregation. In addition, this mode could be used in production, which should not have been allowed. This PR removes the mode. It is safe to use `RollupProverConfig::Prove` instead, since the SP1 prover is sufficiently fast.`

## Linked Issues
- Related to #2551


crates/adapters/mock-zkvm/src/lib.rs
```diff
@@ -107,19 +103,14 @@ impl sov_rollup_interface::zk::ZkVerifier for MockZkVerifier {
...
-                pub_data: input,
-            }) => {
-                if is_valid {
-                    Ok(bincode::deserialize(&input)?)
-                } else {
-                    anyhow::bail!("Proof is not valid")
-                }
-            }
-            Proof::Full(_) => unimplemented!("MockZkVerifier doesn't support full zk proofs"),
+        let Inner {
```
preston-evans98: (non-blocking, nit): Maybe we should rename `Inner` to `MockProof` after this change?

preston-evans98 (approved):

crates/adapters/mock-zkvm/src/lib.rs
bkolad: fixed

PR #2693, opened 2026-04-03
Add RESTful API for state map iteration

preston-evans98:
# Description

This PR adds support for iterating State maps via API. Each call is internally consistent (i.e. maintains a snapshot of state), but the underlying data may change as callers paginate.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
```diff
@@ -307,6 +307,188 @@ impl<S: Spec> PinnedCacheAccessor<S> for ApiStateAccessor<S> {
...
+
+        // Next, iterate over the writes on the checkpoint. These are also unordered, so we iterate over all of them.
+        for ((key, entry_namespace), value) in self.checkpoint_and_read_txn.read_txn.iter() {
+            if *entry_namespace == namespace
+                && matches_prefix(key)
+                && (cursor_key.is_none() || cursor_key.as_ref().is_some_and(|cursor| key > cursor))
+            {
+                match value {
+                    Some(value) => {
+                        // Only consider the key if it wasn't delted by a local write
```
bkolad, line 427: Typo `delted`

crates/module-system/sov-modules-api/src/rest/__private/state.rs
```diff
@@ -252,6 +302,117 @@ where
...
+        let pagination = query.pagination;
+        let prefix = Prefix::new(
+            state.module_discriminant,
+            state.state_item_info.item_discriminant,
+        );
+        let state_map = NamespacedStateMap::<N, K, V, Codec>::with_codec(prefix, Codec::default());
+        let slot_prefix = SlotKey::singleton(&prefix);
+        let accessor = state
+            .api_state
+            .build_api_state_accessor(None)
```
bkolad, line 334: This endpoint paginates across separate HTTP requests. Because each request reads from the current head rather than a fixed snapshot, the underlying map can change between page 1 and page 2. That means the second page may be evaluated against a different state, leading to inconsistencies. Can we pin pagination to a specific rollup height? But then we `query.historical_height_requested` is explicitly unimplemented.

preston-evans98: Yeah, unfortunately we can't pin to a specific height because the archival table doesn't have a sorted index. I think we'll have to just document that it's pinned to the live table

bkolad (approved):

crates/module-system/sov-modules-api/src/state/accessors/http_api.rs
preston-evans98: Will fix in a follow up (to not block CI)

PR #2695, opened 2026-04-06
Add new RPC endpoint for fetching storage proofs

preston-evans98:
# Description

This PR completely reworks the `get_with_proof` API on `NativeStorage` to fix bugs in the NOMT implementation and accommodate NOMT's constraint that only the latest merkle tree can be used. The core changes are that...
- get_with_proof no longer accepts a slotnumber param
- get_with_proof now accounts for races between the flat store and NOMT during commit. Previously, it would return incorrect data (where the merkle proof and the value didn't match). 
- get_with_proof returns the slot number that its state root came from. The new RPC method uses this to fetch the correct EVM block number from accessory state.

In addition to the changes to `get_with_proof`, this PR actually implements the new RPC for fetching merkle proofs `ext_getStorageProof`. A POC verifier script will be included in a separate PR to rollup-starter. 

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/module-implementations/integration-tests/tests/concurrent_prover_storages.rs
```diff
@@ -482,29 +481,12 @@ fn assert_values<S: NativeStorage>(
     let witness_stub = S::Witness::default();
     let get_value = |version: Option<SlotNumber>| -> Option<SlotValue> {
         match namespace {
-            ValueNamespace::StateKernel => {
-                let just_value = storage
-                    .get_historical::<Kernel>(key, version, &witness_stub)
-                    .unwrap();
-                let with_proof = storage
```
preston-evans98, line 489: We can no longer `get_with_proof` at an arbitrary height. @citizen-stig, do you think we need additional test changes to compensate for the loss of coverage here?

crates/module-system/module-implementations/sov-evm/src/rpc/handlers.rs
```diff
@@ -135,6 +146,43 @@ where
...
+            .expect("NativeStorage broke its API contract; returned empty accessory values when non-empty accessory values were provided") // Error 2: Empty accessory values
+            .as_ref()
+            .expect("evm.block_numbers returned None at the latest height. This is a bug, block numbers must always be set."); // Error 3: Slot key isn't set
+
+        let block_number = *self
+            .block_numbers
+            .decode_unwrap(block_number_slot_value)
+            .end();
+        let block_number = block_number.saturating_add(config_value!("STATE_ROOT_DELAY_BLOCKS")); // Add state root delay blocks to get the state root for the block number
+        println!("proof: {}", serde_json::to_string(&proof).unwrap());
```
preston-evans98: TODO: remove

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -711,16 +720,64 @@ where
...
+    // Aside: Why do we need to fetch the root hash twice?
+    // If we only fetch once, there's a subtle race condition. With a single read, one or the other of these interleavings is possible: (Read_rocksdb_value-> (other thread)commit_rocksdb -> Read_rocksdb_root) or (Read_rocksdb_root-> (other thread)commit_rocksdb -> Read_rocksdb_value)
+    // In either case, the value we read from RocksDB will be inconsistent with the root hash we fetched from RocksDB. By reading the root hash before and after the value and cross-checking, we eliminate this possibility.
     fn get_with_proof<N: ProvableCompileTimeNamespace>(
         &self,
-        key: SlotKey,
-        slot_number: Option<SlotNumber>,
-    ) -> anyhow::Result<StorageProof<Self::Proof>> {
+        proven_key: SlotKey,
+        accessory_keys: Option<Vec<SlotKey>>,
```
citizen-stig: What is the purpose of adding this accessory key retrieval in this function?

it looks like it can be extracted into a separate function and simplify function and return type.

I understand that those keys need to be a particular version, but we can return version with storage proof and then read accessory on that version separately. Also this reduces disk accesses in concurrency-sensitive area.

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -711,16 +720,64 @@ where
...
+    // Fetch the requested key with proof. Atomically retrieve the values of any provided accessory keys at the same storage version.
+    //
+    // This method is complicated because NOMT only maintains the live version of the merkle tree
+    //
+    // We want to ensure that the merkle proof we fetch (from NOMT) and the value (from RocksDB) are consistent. But RocksDB and NOMT commit at slightly different times.
+    // So, what we do is...
+    // 1. Fetch the root hash from RocksDB at the latest version.
+    // 2. Read the value from RocksDB
+    // 3. Fetch the root hash from RocksDB again.
+    // 4. Open a NOMT session (which locks NOMT and prevents any commits)
```
citizen-stig: What if NOMT committed between step 3 and 4? This means we will get different root hash from nomt, and return an error. 
If we open nomt session before reading from rocks db, then there's a less likely that commit happens during rocksdb reads.

crates/utils/sov-rpc-eth-types/src/eth_api_error.rs
```diff
@@ -83,6 +83,9 @@ pub enum EthApiError {
     /// Invalid block count
     #[error("invalid block count: {0}")]
     InvalidBlockCount(u64),
+    /// Thrown when a storage proof is not found
+    #[error("storage proof not found")]
+    StorageProofNotFound,
```
citizen-stig: nit: it might be helpful to add index of input as param and ideally underlying error from the storage.

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -711,16 +720,64 @@ where
...
+    // so they fall back to archival state if the underlying DB has changed. But NOMT doesn't have an archival tree to make merkle proofs against, so this behavior would break consistency.
+    // The "unbound" methods bypass this behavior and always show the newest available value (either a value from the in-memory deltas in the storage, or - if it's newer - the value from the underlying database).
+    //
+    // Aside: Why do we need to fetch the root hash twice?
+    // If we only fetch once, there's a subtle race condition. With a single read, one or the other of these interleavings is possible: (Read_rocksdb_value-> (other thread)commit_rocksdb -> Read_rocksdb_root) or (Read_rocksdb_root-> (other thread)commit_rocksdb -> Read_rocksdb_value)
+    // In either case, the value we read from RocksDB will be inconsistent with the root hash we fetched from RocksDB. By reading the root hash before and after the value and cross-checking, we eliminate this possibility.
     fn get_with_proof<N: ProvableCompileTimeNamespace>(
         &self,
-        key: SlotKey,
-        slot_number: Option<SlotNumber>,
```
bkolad, line 717: For optimistic rollups, we use the `AttesterIncentives` module.

`AttesterIncentives::process_attestation` calls `AttesterIncentives::check_bonding_proof`:
``` comment in the code:
/// The bonding proof is now a proof that an attester was bonded during the last `finality_period` range.
```
So in the optimistic rollup case, we require proofs for historical slots. It looks like this was already have been broken for NOMT, since I see the following:

```
#[ignore = "NOMT does not support archival proof generation - proofs are always generated against the current state"]
fn test_archival_proof_gen()
```

This is outside the scope of this PR, but I wanted to flag it here because the proposed change would further solidify that behavior.

crates/module-system/module-implementations/sov-evm/src/rpc/handlers.rs
preston-evans98: Done

crates/module-system/sov-state/src/nomt/prover_storage.rs
preston-evans98: Good question. I initially thought this wasn't possible because of the different underlying DBs, but I found a way to make it work with only a new trait method (`get_accessory_unbounded`) and a tweak to the existing API to return the slotnumber. Done!

crates/utils/sov-rpc-eth-types/src/eth_api_error.rs
preston-evans98: I've reworked the error handling. Now, only internal server errors can occur and those should be logged appropriately

crates/module-system/sov-state/src/nomt/prover_storage.rs
citizen-stig: We can discuss what to do with attester module. 

I think there's a way, attester does not need proof for arbitrary state keys, only for bonded attesters. So probably it can have end slot hooks or something like this to store proof separately, or emit them as events

P.s. getting state root for historical height should be possible, it is in rocksdb

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -100,20 +100,26 @@ where
                 },
         } = group;
 
+        // ======================= DANGER ZONE ======================
+        // The commit order here is relied on by NomtProverStorage::get_with_proof
+        // If you change the order, you'll need to update merkle proof generation.
```
citizen-stig, line 105: Does it mean, that we won't be able to concurrent commits? 

If I remember correctly, this is one of the biggest bottlenecks in the whole data flow

crates/full-node/sov-db/src/state_db_nomt.rs
```diff
@@ -202,6 +204,10 @@ where
             let snapshots = self.all_snapshots.read().expect("Snapshots lock poisoned");
             for overlay_ref in &self.relevant_snapshot_refs {
                 let Some(state_overlay) = snapshots.get(overlay_ref) else {
+                    // Note: This assumption does not hold when the underlying chain can fork. The snapshot might have been fresh but discarded
+                    // because a different fork was committed. In this case, the session we create here will *not* match the `HistoricalStateReader`,
+                    // since that uses RocksDB changesets whose references are held in the storage itself.  In other words,
+                    // forking can break the consistency of `Storage` snapshots on abandoned forks.
```
citizen-stig:
```suggestion
                    // forking can break the consistency of `Storage` snapshots **on abandoned forks**.
```

crates/module-system/module-implementations/integration-tests/tests/concurrent_prover_storages.rs
citizen-stig: If I understand correctly, we still can get with proof for the latest storage, right?

So I would check `value_with_proof` for cases when `version` is None. it can be done outside of the closure.

If we have other test that verifies that value from `get_with_proof` matches normal get I would we fine.

This test is nice, because it has several competing storages

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -344,6 +373,102 @@ where
...
+    // RocksDB committing slightly out of phase. All other failures are returned immediately.
+    //
+    // NOMT only maintains the live version of the merkle tree
+    //
+    // We want to ensure that the merkle proof we fetch (from NOMT) and the value (from RocksDB) are consistent. But RocksDB and NOMT commit at slightly different times.
+    // So, what we do is...
+    // 1. Fetch the root hash from RocksDB at the latest version.
+    // 2. Read the value from RocksDB
+    // 3. Fetch the root hash from RocksDB again.
+    // 4. Open a NOMT session (which locks NOMT and prevents any commits)
```
citizen-stig, line 403: Since now RPC call now directly opens NOMT session, which wasn't the case before, is it possible that it will open up DoS attack on the whole node if many client request proof?

examples/demo-rollup/out.log
bkolad, line 1: Looks like this file was committed by accident.

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -344,6 +373,102 @@ where
...
+        let multi_proof = MultiProof::from_path_proofs(vec![path_proof]);
+
+        Ok((
+            StorageProof {
+                key: proven_key,
+                value,
+                proof: NomtMultiProof(multi_proof),
+                namespace,
+            },
+            committed_slot_number,
```
bkolad, line 469: I think we should take `committed_slot_number` in line `435` instead of `424` in order to be consistent with `post_fetch_state_root`.  If `PROVABLE_NAMESPACE=USER` and we change only kernel values the state root check in 451 will pass, but the `POST_committed_slot_number` will be different than the one from line 424

bkolad (approved):

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
preston-evans98: > Does it mean, that we won't be able to concurrent commits?

No, I don't think it means that we won't be able to - we'll just need to assess`get_with_proof` and ensure that we keep the code paths in sync

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -344,6 +373,102 @@ where
...
+                .map_err(GetWithProofError::from)?,
+        };
+
+        if pre_fetch_state_root != post_fetch_state_root_namespace
+            || post_fetch_state_root_namespace != session.prev_root().as_ref()
+        {
+            return Err(GetWithProofError::StateRootMismatch);
+        }
+
+        let key_path: KeyPath = S::Hasher::digest(proven_key.as_ref()).into();
```
citizen-stig: nit: Probably this is micro optimization, but I would prefer deriving key path before session is started, so lock is held a litte lest

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -344,6 +373,102 @@ where
...
+        let path_proof = session.prove(key_path).map_err(GetWithProofError::from)?;
+        let multi_proof = MultiProof::from_path_proofs(vec![path_proof]);
```
citizen-stig, line 460: Another small nit to reduce time lock is held
```suggestion
        drop(session);
        let multi_proof = MultiProof::from_path_proofs(vec![path_proof]);
```

preston-evans98: Good idea

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -713,34 +838,43 @@ where
...
+                    let elapsed = start.elapsed();
+                    if elapsed >= RETRY_DEADLINE {
+                        anyhow::bail!(
+                            "State root mismatch between pre-fetch and post-fetch after {attempts} attempts over {:?}",
+                            elapsed
+                        );
+                    }
+
+                    let remaining = RETRY_DEADLINE.saturating_sub(elapsed);
+                    std::thread::sleep(Duration::from_millis(backoff_ms).min(remaining));
```
citizen-stig, line 875: I know this is bigger change, but a lot of that will be called from async context (RPC handlers) and this kind of sleeps might starve tokio runtime.

Would it be too big of a change to make high level function async and then call async sleep and `tokio::task::block_in_place` for blocking ops?

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -344,6 +373,102 @@ where
...
+    // wrapper retries only `StateRootMismatch`, which is the transient case caused by NOMT and
+    // RocksDB committing slightly out of phase. All other failures are returned immediately.
+    //
+    // NOMT only maintains the live version of the merkle tree
+    //
+    // We want to ensure that the merkle proof we fetch (from NOMT) and the value (from RocksDB) are consistent. But RocksDB and NOMT commit at slightly different times.
+    // So, what we do is...
+    // 1. Fetch the root hash from RocksDB at the latest version.
+    // 2. Read the value from RocksDB
+    // 3. Fetch the root hash from RocksDB again.
```
citizen-stig, line 402: Copy this one here: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2695#discussion_r3043390850

crates/module-system/sov-state/src/nomt/prover_storage.rs
preston-evans98: Hmm, that's a good question. Session creation can also block (probably for much longer than the sleep). Let me think about this

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -845,3 +990,564 @@ impl sov_metrics::Metric for NomtProverComputeStateResult {
...
+        (
+            crate::StorageProof<<TestStorage as Storage>::Proof>,
+            sov_rollup_interface::common::SlotNumber,
+            <TestStorage as Storage>::Root,
+        ),
+        GetWithProofError,
+    >;
+
+    fn test_db_config(path: std::path::PathBuf) -> RollupDbConfig {
+        RollupDbConfig {
```
citizen-stig: Does `RollupDbConfig::default_in_path()` does not work here? It is available via `test-utils` feature in sov-db

preston-evans98: Oop, good catch. I let chatgpt add this helper...

crates/module-system/sov-state/src/nomt/prover_storage.rs
citizen-stig: alternatively, we can call "block_in_place" in the RPC handler and call it a day =)

crates/module-system/sov-state/src/nomt/prover_storage.rs
preston-evans98: The tradeoff here is that we have to keep the session open quite a bit longer if we open it before reading from RocksDB. I think that's the better tradeoff for now given the perf concerns - WDYT?

crates/module-system/sov-state/src/nomt/prover_storage.rs
preston-evans98: jsonrpsee has a nice method for this. I can mark the whole `ext_getWithProof` as blocking, which uses `tokio::task::spawn_blocking`. I've done that in the latest revision

crates/module-system/sov-state/src/nomt/prover_storage.rs
preston-evans98: Replied there. Resolving this one to avoid duplicate conversation :)

crates/full-node/sov-db/src/state_db_nomt.rs
preston-evans98: Done

examples/demo-rollup/out.log
preston-evans98: Removed, thanks!

crates/module-system/sov-state/src/nomt/prover_storage.rs
preston-evans98: I've switched over!

crates/module-system/sov-state/src/nomt/prover_storage.rs
preston-evans98: Done

crates/module-system/sov-state/src/nomt/prover_storage.rs
preston-evans98: Chatted offline. We think the risk is relatively low since merkle generation is fast and it only holds a read lock

crates/module-system/module-implementations/integration-tests/tests/concurrent_prover_storages.rs
preston-evans98: After analyzing, there are some cases where this comparison holds and some where it doesn't. `get_with_proof` will diverge whenever the storage is stale (because NOMT only has the latest merkle tree). I've added a new assertion helper to account for this, and to handle `version == None`. 

Assertions have been added back wherever possible

PR #2715, opened 2026-04-09
  Proof aggregation: commit inner vkey hash as public output.

bkolad:
# Description

Previously, the aggregation guest binary embedded a build-time-generated inner_vk_hash, which complicated the build process. This PR removes that generated dependency by having the host supply inner_vk_hash in the aggregation witness, and the guest commit it into the proof’s public outputs.

1. The aggregation circuit no longer hardcodes the inner verification key hash at build time from `inner_vk.bin`. Instead, the host provides the value, and it is committed as part of `AggregatedProofPublicData`. The sov-prover-incentives module now verifies this against the inner code commitment registered on-chain and slashes with `IncorrectInnerVkeyHash` if they do not match. The circuit also checks that this hash stays the same across recursive aggregation steps.
2. Added `CodeCommitmentTrait::to_hash()`, giving every adapter commitment type—Risc0, SP1, and Mock—a consistent way to convert into a canonical CodeCommitmentHash.
3. Removed the `outer_vk_hash parameter from Prover, NetworkProver, and all blueprint factory interfaces. As part of that cleanup, program/build.rs was deleted, along with its sp1-sdk build dependency.

## Linked Issues
- Related to #2551


crates/adapters/sp1/src/host.rs
```diff
@@ -90,10 +90,14 @@ impl SP1AggregationHost {
         }
 
         let aggregation_vk_hash = self.aggregation_vk.hash_u32();
+        let inner_vk: sp1_sdk::SP1VerifyingKey = bincode::deserialize(&self.inner_method_id.0)
+            .map_err(|e| anyhow::anyhow!("Failed to deserialize inner SP1VerifyingKey: {e}"))?;
```
bkolad, line 94: We deserialize this each time we create an aggregated proof. As an optimization, we cache the SP1VerifyingKey in SP1AggregationHost, but this is not currently a bottleneck because deserialization is fast and aggregated proofs are created infrequently.

crates/module-system/module-implementations/sov-prover-incentives/src/capabilities.rs
```diff
@@ -144,6 +144,28 @@ impl<S: Spec> ProverIncentives<S> {
...
+        // from host advice, so without this check a prover could verify inner
+        // proofs against an arbitrary inner VK. The outer VK proof alone no
+        // longer transitively pins the inner circuit.
+        let inner_code_commitment = self
+            .chain_state
+            .inner_code_commitment(state)
+            .map_err(Into::<anyhow::Error>::into)?
+            .expect("The inner code commitment should be set at genesis");
+
+        if inner_code_commitment.to_hash()? != public_outputs.inner_vkey_hash {
```
bkolad, line 157: Here, we recompute to_hash each time we verify an aggregated proof. This could be optimized by storing the hash in the module, but doing so would be a breaking change because it would alter the rollup’s state. We can revisit this once the proving workflow is finalized.

crates/adapters/sp1/src/host.rs
bkolad: Added to #2551

crates/module-system/module-implementations/sov-prover-incentives/src/capabilities.rs
bkolad: Added to #2551

citizen-stig (approved):

PR #2716, opened 2026-04-10
   Move aggregated-proof artifacts into demo-rollup prover directories

bkolad:
# Description

This PR relocates the SP1 `aggregated-proof` workspace from `crates/full-node` to `demo-rollup/provers/sp1/guest-aggregation-mock`, placing the demo-rollup aggregation circuit next to the other circuits.

  Changes:
1. Move the aggregation host workspace to examples/demo-rollup/tests/prover/sov-aggregated-proof
2. Move the SP1 guest program to examples/demo-rollup/provers/sp1/guest-aggregation-mock
3. Remove the old crates/full-node/sov-aggregated-proof workspace
4. update top-level cleanup paths for the relocated workspace

## Linked Issues
- Related to #2551


Makefile
```diff
@@ -45,7 +46,12 @@ check-provers:   ## cargo check in non attached crates
 	@set -e; for dir in $(PROVER_DIRS); do \
 		echo "$$(date) Running cargo fmt + check in $$dir"; \
 		cargo fmt --all --check --quiet --manifest-path "$$dir/Cargo.toml"; \
-		cargo check --all-targets --all-features --manifest-path "$$dir/Cargo.toml"; \
+		case "$$dir" in \
```
bkolad, line 49: Before, we were only running a standard cargo check on the guest. That was incorrect, because we actually need to check it with target_os = "zkvm". So the previous command was not validating the right configuration.

This change fixes that for SP1. For RISC Zero, we skip the equivalent command because it downloads very large Docker images. Skipping it is not a regression, since the previous version was not doing the correct check anyway.

ross-weir (approved):

PR #2720, opened 2026-04-13
CI clean up

citizen-stig:
# Description

* remove unnecessary libclang installation #2717 , because it is already installed in the base image
* use separate base image for nexttest-all-features to improve hash matches
* base images have updated risc0 version

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -254,7 +254,7 @@ jobs:
   nextest_all_features:
     name: nextest_all_features
     # Privileged feature is needed for proper `io_uring` functionality needed by NOMT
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -110,8 +110,6 @@ jobs:
         with:
           cache: rust
           path: ./crates/fuzz/target
-      - name: Install system dependencies
-        run: sudo apt-get update && sudo apt-get install -y clang libclang-dev
```
bkolad, line 114: Do we need to add this in `make install-dev-tools`?

bkolad (approved):

.github/workflows/rust.yml
citizen-stig: Good question, usually we don't, because this is distro specifc (MacOs vs. Ubuntu vs. CentOS, etc)

I think people already know what headers and toolchains are needed for basic development

PR #2721, opened 2026-04-13
Celestia: Add support for gRPC fallback endpoints

citizen-stig:
# Description

Allow specify several gRPC endpoints for celestia adapter, so blobs can continue being submitted.

## Linked Issues
- Related to #1630  

## Testing
Tests for configuration parsing has been added. Core functionality is outside of SDK code and assumed to be working. 

## Docs
No updates

### DIFF
--- a/crates/adapters/celestia/src/config.rs
+++ b/crates/adapters/celestia/src/config.rs
@@ -219,6 +219,9 @@ impl CelestiaConfig {
             builder = builder.rpc_auth_token(rpc_auth_token);
         }
         // Submission section.
+        if self.grpc_url.is_none() && !self.grpc_fallback_endpoints.is_empty() {
+            anyhow::bail!("`grpc_fallback_endpoints` requires `grpc_url` to be set");
+        }
         if let Some(grpc_url) = &self.grpc_url {
             let mut endpoint = celestia_client::Endpoint::new(grpc_url.clone());
             if let Some(grpc_auth_token) = &self.grpc_auth_token {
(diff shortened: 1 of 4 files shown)


crates/adapters/celestia/src/config.rs
```diff
@@ -192,12 +219,25 @@ impl CelestiaConfig {
             builder = builder.rpc_auth_token(rpc_auth_token);
         }
         // Submission section.
+        if self.grpc_url.is_none() && !self.grpc_fallback_endpoints.is_empty() {
```
bkolad, line 222: Would it make sense to
1 remove grpc_url 
2 rename self.grpc_fallback_endpoints to self.grpc_endpoints
3 set the current value as the forst entry in self.grpc_endpoints 
?

citizen-stig: That would be config breaking for existing clients.

citizen-stig: Oh, you mean in rust code.

So always set it via `grpc_endpoints` and just have a vector of a different size. That can be done.

bkolad (approved):

crates/adapters/celestia/src/config.rs
citizen-stig: Actually changing this changes semantics. Original code requires explicit single gRPC endpoint and 1 or more fallback endpoints. But now it is possible to skip original grpc_url and use only fallback. Which is probably fine, but a little bit confusing. And only upside, is that rust is code slightly more tidy. So I am leaning towards keeping original approach. What do you think?

PR #2723, opened 2026-04-13
Add CI compile time monitoring jobs

citizen-stig:
# Description

Adds 3 new jobs that monitor what take the most time compiling in CI in jobs like check and test.
This is intended to gather information about our cache efficiency as well as main bottlenecks in CI.
Can be gathered and analyzed by LLM agent

## Testing
No impact

## Docs
No updates


ross-weir (approved):
Cool, `--timings` flag seems neat

citizen-stig:
> Cool, `--timings` flag seems neat

Yep, wanted to add something basic so we can start harvesting cargo specific data. If something else comes up as useful, we can just add it to this job.

And overall idea: you can just tell Claude to use `gh` and analyze jobs in last 5 days and it should get enough signal

PR #2736, opened 2026-04-14
test_mock_proof_public_data_matches_witnesses

bkolad:
What Was Broken
  The failing test was showing that native witness generation and guest-side replay produced different public data, specifically a different final state root. The important part is that the STF logic was not the problem. The bug was in how Sovereign eplayed NOMT state updates inside the guest.

  Before the fix, Sovereign effectively treated NOMT like:

  - one compact multiproof for reads
  - one flat sorted list of writes for updates

  That is too lossy. NOMT update verification is path-scoped, not just “all writes sorted by key”.

  1. Add a Regression Test
  The new test in examples/demo-rollup/tests/prover/sp1_cpu_prover.rs:79 runs the mock SP1 guest for each witness and checks that the guest’s public outputs match the witness:

  - initial_state_root
  - final_state_root
  - slot_hash

  The helper in examples/demo-rollup/tests/prover/sp1_cpu_prover.rs:160 deserializes public_values from the mock proof so the test compares exactly what the guest computed.

  That test is what exposed the NOMT replay bug.

  2. Preserve the Full NOMT Witness on the Prover Side
  The key prover-side change is in crates/module-system/sov-state/src/nomt/prover_storage.rs:537.

  Before this change, the prover took NOMT’s full witness and collapsed it into a MultiProof, discarding:

  - original witnessed paths
  - NOMT read/write operation metadata
  - path-to-operation structure

  Now it stores the full nomt::Witness in the Sovereign witness:

  - path_proofs
  - operations.reads
  - operations.writes

  That is necessary because the verifier needs more than a flat multiproof if it wants to replay updates correctly.

  3. Rewrite Guest-Side Replay Around the Full Witness
  The main fix is in crates/module-system/sov-state/src/nomt/zk_storage.rs:37.

  The flow now is:

  1. Build canonical expected read/write maps from Sovereign state accesses.
     Code: crates/module-system/sov-state/src/nomt/zk_storage.rs:46
  2. Deserialize the full NomtWitness.
     Code: crates/module-system/sov-state/src/nomt/zk_storage.rs:69
  3. Reconstruct a MultiProof from the witnessed paths and use it only for reads.
     Code: crates/module-system/sov-state/src/nomt/zk_storage.rs:70
  4. Verify all reads against that multiproof.
     Code: crates/module-system/sov-state/src/nomt/zk_storage.rs:83
  5. Validate the witnessed write set against expected writes:

  - no unexpected writes
  - no missing writes
  - no duplicates
  - witnessed value hash must exactly match the expected authenticated write hash
    Code: crates/module-system/sov-state/src/nomt/zk_storage.rs:108

  6. Verify each path proof independently into a VerifiedPathProof.
     Code: crates/module-system/sov-state/src/nomt/zk_storage.rs:133
  7. Group each write onto the deepest matching path prefix using:

  - the verified path depth
  - a canonical truncated prefix key
  - a map keyed by (depth, truncated_prefix)
    Code: crates/module-system/sov-state/src/nomt/zk_storage.rs:151
    Support function: crates/module-system/sov-state/src/nomt/zk_storage.rs:194

  8. Build PathUpdates and call NOMT’s path-based verify_update.
     Code: crates/module-system/sov-state/src/nomt/zk_storage.rs:179

  Why This Fix Works
  The old approach was wrong because it replayed updates as a flat batch. Reads can be checked that way, but writes cannot. NOMT updates need each write attached to a path that actually scopes that key.

  The final version fixes that by:

  - keeping the full NOMT witness
  - using the multiproof only for read verification
  - validating the exact write set
  - regrouping writes by actual path scope
  - using path-based verify_update

  That is why the final_state_root mismatch disappeared.

  Why There Was a Second Failure on Larger Data
  After the first fix, larger data still hit OpOutOfScope. That showed another issue: you could not safely rely on NOMT’s serialized write-to-path bookkeeping alone for guest replay.

  The current code avoids that by deriving write grouping from actual path prefixes instead of trusting the witness path_index. The likely upstream cause is NOMT witness assembly under larger workloads; that part is an inference. The proven part is that
  regrouping by prefix scope fixes the failure.

  Net Effect
  The PR changes the protocol from:

  - “store compact multiproof, replay flat writes”

  to:

  - “store full NOMT witness, verify reads globally, verify updates path-by-path”

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/sov-state/src/nomt/zk_storage.rs
+++ b/crates/module-system/sov-state/src/nomt/zk_storage.rs
@@ -105,44 +105,64 @@ impl<S: MerkleProofSpec> NomtVerifierStorage<S> {
             }
         }
 
-        let mut writes_by_path = vec![Vec::<(KeyPath, Option<ValueHash>)>::new(); nomt_witness.path_proofs.len()];
-
+        let mut witnessed_writes = BTreeMap::new();
         for write in nomt_witness.operations.writes {
             let Some(expected_value) = expected_writes.remove(&write.key) else {
                 anyhow::bail!("Unexpected or duplicate NOMT write witness entry");
             };
-            let Some(path_writes) = writes_by_path.get_mut(write.path_index) else {
-                anyhow::bail!("NOMT write witness path index out of bounds");
+            if write.value != expected_value {
+                anyhow::bail!(
+                    "Mismatched NOMT write witness value for key {:?}: witness={:?}, expected={:?}",
+                    write.key,
+                    write.value,
+                    expected_value
+                );
+            }
+            if witnessed_writes
+                .insert(write.key, expected_value)
+                .is_some()
+            {
+                anyhow::bail!("Duplicate NOMT write witness entry");
             };
-            path_writes.push((write.key, expected_value));
         }
 
         if !expected_writes.is_empty() {
             anyhow::bail!("Missing NOMT write witness entries");
         }
 
-        let mut path_entries = nomt_witness
+        let mut verified_paths = nomt_witness
             .path_proofs
             .into_iter()
-            .enumerate()
-            .map(|(index, path)| {
-                (
-                    path,
-                    std::mem::take(&mut writes_by_path[index]),
-                )
+            .map(|path| {
+                let verified = path
+                    .inner
+                    .verify::<BinaryHasher<S::Hasher>>(path.path.path(), prev_root)
+                    .map_err(|e| anyhow::anyhow!("Failed to verify path proof: {:?}", e))?;
+                Ok((verified, Vec::<(KeyPath, Option<ValueHash>)>::new()))
             })
-            .collect::<Vec<_>>();
-        path_entries.sort_by(|a, b| a.0.path.path().cmp(b.0.path.path()));
+            .collect::<anyhow::Result<Vec<_>>>()?;
+        verified_paths.sort_by(|a, b| a.0.path().cmp(b.0.path()));
+
+        for (key, value) in witnessed_writes {
+            let matching_index = verified_paths
+                .iter()
+                .enumerate()
+                .filter_map(|(index, (verified, _))| {
+                    verified
+                        .confirm_nonexistence(&key)
+                        .ok()
+                        .map(|_| (index, verified.path().len()))
+                })
+                .max_by_key(|(_, path_len)| *path_len)
+                .map(|(index, _)| index)
+                .ok_or_else(|| anyhow::anyhow!("No NOMT path proof covers write key {:?}", key))?;
+
+            verified_paths[matching_index].1.push((key, value));
+        }
 
         let mut updates = Vec::new();
-        for (path, mut writes) in path_entries {
-            let verified = path
-                .inner
-                .verify::<BinaryHasher<S::Hasher>>(path.path.path(), prev_root)
-                .map_err(|e| anyhow::anyhow!("Failed to verify path proof: {:?}", e))?;
-
+        for (verified, writes) in verified_paths {
             if !writes.is_empty() {
-                writes.sort_by(|a, b| a.0.cmp(&b.0));
                 updates.push(PathUpdate {
                     inner: verified,
                     ops: writes,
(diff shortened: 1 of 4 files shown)


crates/module-system/sov-state/src/nomt/zk_storage.rs
```diff
@@ -41,69 +43,173 @@ impl<S: MerkleProofSpec> NomtVerifierStorage<S> {
...
+                // Authenticated write is hash of a combination of size and original value hash.
+                S::Hasher::digest(slot_value.combine_val_hash_and_size::<S::Hasher>()).into()
+            });
+            if expected_writes.insert(key_hash, value_hash).is_some() {
+                anyhow::bail!("Duplicate key write in state accesses: {:?}", key);
+            }
+        }
 
-        let multi_proof: MultiProof = array_witness.get_hint();
+        let nomt_witness: NomtWitness = array_witness.get_hint();
```
citizen-stig, line 69: This duplicates the information we send to ZK. 

NomtWitness contains `operations`, which has matching data to `state_accesses`

```rust
pub struct NomtWitness {
    pub path_proofs: Vec<WitnessedPath>,
    pub operations: WitnessedOperations,
}

pub struct WitnessedOperations {
    pub reads: Vec<WitnessedRead>,
    pub writes: Vec<WitnessedWrite>,
}

pub struct WitnessedPath {
    pub inner: PathProof,
    pub path: TriePosition,
}

pub struct WitnessedRead {
    pub key: KeyPath,
    pub value: Option<ValueHash>,
    pub path_index: usize,
}
```

crates/module-system/sov-state/src/nomt/zk_storage.rs
```diff
@@ -41,69 +43,173 @@ impl<S: MerkleProofSpec> NomtVerifierStorage<S> {
...
+                    path.path.depth() as usize,
+                    truncate_key_path(path.path.raw_path(), path.path.depth() as usize),
+                    Vec::<(KeyPath, Option<ValueHash>)>::new(),
+                ))
             })
-            .collect::<Vec<(KeyPath, Option<ValueHash>)>>();
+            .collect::<anyhow::Result<Vec<_>>>()?;
+        verified_paths.sort_by(|a, b| a.0.path().cmp(b.0.path()));
+
+        let mut path_index_by_prefix = BTreeMap::new();
```
citizen-stig, line 151: What is this done? This seems like this should be part of nomt

crates/module-system/sov-state/src/nomt/zk_storage.rs
```diff
@@ -41,69 +43,173 @@ impl<S: MerkleProofSpec> NomtVerifierStorage<S> {
...
+            let matching_index = path_depths_desc
+                .iter()
+                .find_map(|depth| {
+                    path_index_by_prefix
+                        .get(&(*depth, truncate_key_path(key, *depth)))
+                        .copied()
+                })
+                .ok_or_else(|| anyhow::anyhow!("No NOMT path proof covers write key {:?}", key))?;
+
+            verified_paths[matching_index].3.push((key, value));
```
citizen-stig, line 176: What is `verified_paths` exactly in the end? What does it represent?

crates/module-system/sov-state/src/nomt/zk_storage.rs
```diff
@@ -41,69 +43,173 @@ impl<S: MerkleProofSpec> NomtVerifierStorage<S> {
...
+            {
+                anyhow::bail!("Duplicate NOMT write witness entry");
+            };
+        }
+
+        if !expected_writes.is_empty() {
+            anyhow::bail!("Missing NOMT write witness entries");
+        }
+
+        let mut verified_paths = nomt_witness
```
citizen-stig, line 133:
```suggestion
        // Vec<(VerifiedPathProof, usize, KeyPath, Vec<(KeyPath, Option<ValueHash>)>)>
        let mut verified_paths = nomt_witness
```

citizen-stig: This is pretty complicated variable.

PR #2737, opened 2026-04-14
Simplify Transaction versioned struct generics

theodorebugnet:
# Description

`Transaction` and `UnsignedTransaction` were generic in `R: TransactionCallable`, but `Version0` and `Version1` were generic on `R::Call`. This made it awkward to convert between `VersionX` and `UnsignedTransactionVX` as it required re-introducing the R generic.

This PR changes both VersionX structs to be generic on the same TransactionCallable, allowing convenient cross-conversion. This simplifies serialization of V1 transactions in particular, and some general ergonomics as well.

To simplify the signing serialization this also moves `serialize_with_chain_hash` to be a method on `UnsignedTransaction` itself, removing the need to branch on enum variants in the signing helpers.

Fixes some feedback from #2688 

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/sov-modules-api/src/transaction/types/v0.rs
+++ b/crates/module-system/sov-modules-api/src/transaction/types/v0.rs
@@ -47,7 +47,7 @@ pub struct Version0<R: TransactionCallable, S: Spec, C: CryptoSpecExt = <S as Sp
 
 impl<R: TransactionCallable, S: Spec, C: CryptoSpecExt> Version0<R, S, C> {
     /// Extracts the versioned unsigned transaction data from this signed envelope.
-    pub fn to_unsigned(&self) -> UnsignedTransaction<R, S> {
+    pub fn as_unsigned(&self) -> UnsignedTransaction<R, S> {
         UnsignedTransaction::V0(UnsignedTransactionV0::new_with_details(
             self.runtime_call.clone(),
             self.uniqueness,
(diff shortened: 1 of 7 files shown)


crates/module-system/sov-modules-api/src/transaction/types/v0.rs
```diff
@@ -33,16 +36,25 @@ pub struct Version0<Call, S: Spec, C: CryptoSpecExt = <S as Spec>::CryptoSpec> {
...
     /// Uniqueness identifier of this transaction. see [`UniquenessData`] for more details.
     pub uniqueness: UniquenessData,
     /// The transaction metadata. Contains gas parameters and the chain ID.
     pub details: TxDetails<S>,
 }
 
-impl<Call, S: Spec, C: CryptoSpecExt> Version0<Call, S, C> {
+impl<R: TransactionCallable, S: Spec, C: CryptoSpecExt> Version0<R, S, C> {
+    /// Extracts the versioned unsigned transaction data from this signed envelope.
+    pub fn to_unsigned(&self) -> UnsignedTransaction<R, S> {
```
citizen-stig: nit: can we consume it here and avoid clones? If no, maybe let's call this method `as_unsigned` to follow rust convention

citizen-stig (approved):

crates/module-system/sov-modules-api/src/transaction/types/v0.rs
theodorebugnet: Great spot

theodorebugnet: Consuming is somewhat annoying in places and in most authenticators we need to clone anyway, so I went ahead and renamed.

PR #2742, opened 2026-04-15
Switch universal-wallet macro to be standalone

theodorebugnet:
# Description

Opinionated fix: changes the default macro exported from `sov-universal-wallet` to expand crate paths to `::sov_universal_wallet` rather than `::sov_rollup_interface::sov_universal_wallet`. This makes the default macro usable standalone and decouples `sov_universal_wallet` from the SDK.

Most user-facing rollup code will be using the re-export from `sov_modules_api::macros::UniversalWallet`, which is unaffected. What _is_ affected is what was previously a re-export from `sov-rollup-interface`. This PR makes the choice to simply drop the re-export: most internal crates (i.e. rollup-interface itself, sov-modules-api internally, and adapter crates) now explicitly depend upon `sov-universal-wallet` rather importing it from rollup-interface.

An alternative would be to have _three_ separate macro definitions: a standalone one, the sov_modules_api one and a separate re-export for rollup-interface. The cleanest way to do that would probably be to have sov-rollup-interface define its own macro. This would reduce the diff significantly here, but IMO it's not worth it for internal crates: the UniversalWallet macro should conceptually work standalone, so internal crates should be fine importing it standalone.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 - #2744 **Manual intervention might be needed**: Adds `serde(deny_unknown_fields)`, which can fail rollup at startup if genesis config is not tidy.
   The change also affects call message de-serialization in sov-paymaster, for all call messages that use `PaymasterPolicyInitializer`
 
+# 2026-04-15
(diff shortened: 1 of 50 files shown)


citizen-stig (approved):
Zen of Python, line 5:

> Flat is better than nested.

PR #2743, opened 2026-04-15
Cleaning up dependencies with cargo-machette

theodorebugnet:
# Description

Our makefile's `install-dev-tools` section installs cargo-udeps, but it seems broken because it requires a nightly compiler and it breaks sp1 crates when I tried it.

This PR replaces it with `cargo machete`, which is must faster and works fine, except it has a lot of false positives. Codex worked for some hours in the background adding `[package.metadata.cargo-machete]` sections to Cargo.toml files where necessary to get everything to compile again.

Net changes:
* -150 in Cargo.lock, a couple of dependency versions removed from the workspace
* Net minus in sp1 and risc0 crate Cargo.locks
* 10 completely unused dependencies removed from Cargo.toml
* A number of `[package.metadata.cargo-machete]` noise added to many Cargo.toml files
* With those metadata sections, `cargo machete` can run in CI for the future (once we update the Namespace base image)

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -46,9 +46,6 @@ runs:
     - uses: taiki-e/install-action@nextest
       env:
         GITHUB_TOKEN: ${{ inputs.github_token }}
-    - uses: taiki-e/install-action@machete
-      env:
-        GITHUB_TOKEN: ${{ inputs.github_token }}
     - uses: taiki-e/install-action@cargo-llvm-cov
       env:
         GITHUB_TOKEN: ${{ inputs.github_token }}
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -46,6 +46,9 @@ runs:
     - uses: taiki-e/install-action@nextest
       env:
         GITHUB_TOKEN: ${{ inputs.github_token }}
+    - uses: taiki-e/install-action@machete
+      env:
+        GITHUB_TOKEN: ${{ inputs.github_token }}
     - uses: taiki-e/install-action@cargo-llvm-cov
       env:
         GITHUB_TOKEN: ${{ inputs.github_token }}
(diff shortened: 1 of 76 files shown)


.github/actions/setup/action.yml
```diff
@@ -46,6 +46,9 @@ runs:
     - uses: taiki-e/install-action@nextest
       env:
         GITHUB_TOKEN: ${{ inputs.github_token }}
+    - uses: taiki-e/install-action@machete
```
citizen-stig: I would prefer avoid installation on every CI job and add it to base image.

citizen-stig: Actually, do we run it in CI at all? Do we need this installation step?

crates/module-system/module-implementations/sov-revenue-share/Cargo.toml
```diff
@@ -36,7 +39,7 @@ serde_json = { workspace = true }
 sov-mock-da = { workspace = true }
 sov-mock-zkvm = { workspace = true }
 sha2 = { workspace = true }
-sov-address = { workspace = true, features = ["evm"] }
+sov-address = { workspace = true, features = ["native", "evm"] }
```
citizen-stig, line 42: this is concerning. Why it does it needs native? It might trigger cascading native unification

citizen-stig: Oh, that's dev. missed collaps from diff

citizen-stig (approved):
Good work. Happy to merge when CI changes are rolled back

.github/actions/setup/action.yml
theodorebugnet: I don't think we run it actually. Yeah, adding it to the base image will be the better course of action.

PR #2745, opened 2026-04-16
Migrate `demo-rollup/Readme.md` to SP1

bkolad:
# Description

Migrates the `MockDemoRollup` example from `Risc0` to `SP1` as the default inner zkVM. Introduces a Skip mode for `SP1Host` so tests that never actually prove don't need a real `ELF`. 

## Linked Issues
- Related to #2551


crates/adapters/sp1/src/host.rs
```diff
@@ -137,13 +137,27 @@ impl SP1AggregationHost {
...
 pub struct SP1Host {
+    inner: SP1HostInner,
+}
+
+#[derive(Clone)]
+enum SP1HostInner {
+    /// Non-proving host; any call that needs a proving/verifying key errors.
+    /// Used by tests that run with `RollupProverConfig::Skip`, which drops
+    /// the host arguments before reaching `create_prover_service`.
+    Skip,
```
bkolad, line 148: This is needed so `SP1Host` can work with Mock proof aggregation. I will explore a way to remove it in the next PR.

crates/module-system/module-implementations/sov-prover-incentives/src/capabilities.rs
```diff
@@ -144,28 +145,30 @@ impl<S: Spec> ProverIncentives<S> {
...
-            );
-            self.slash_prover(prover_address, state)?;
-            return Err(ProcessProofError::ProverSlashedNoRevert(format!(
-                "Invalid output {}",
-                SlashingReason::IncorrectInnerVkeyHash
-            )));
-        }
-
+        // TODO #2551: We don’t handle real inner_code_commitment yet. Re-enable it at the end of #2551.
+        /*
```
bkolad, line 148: This check has to be disabled for now because every change to the on-chain logic will require a new code_commitment. We will need a separate upgrade process, similar to the one we use for chain hashes. This will be handled once the proving workflow is ready.

examples/demo-rollup/configs/mock_rollup_config.toml
```diff
@@ -10,7 +10,7 @@ finalization_blocks = 0 # Defines how many blocks progress to finalization. Set
 # failure_behavior = "none" # Configures failure injection for testing. Options: "none", or structured variants for testing
 
 [da.block_producing.periodic]
-block_time_ms = 1_000
+block_time_ms = 1000
```
citizen-stig, line 13: seems like unnencessary change

examples/demo-rollup/src/zk.rs
```diff
@@ -31,3 +41,13 @@ fn should_skip_guest_build() -> bool {
         Ok("0") | Ok("false") | Ok(_) | Err(_) => false,
     }
 }
+
+fn should_skip_sp1_guest_build() -> bool {
```
citizen-stig, line 45: can we use `sov-zkvm-utils` here, similar how it is attempted here:

https://github.com/Sovereign-Labs/sovereign-sdk/pull/2725/changes#diff-b0519fefbc26273e74e27f02706f74d00588b64d7c03aa2645ee044a5d79d44dR132

```rust
  if sov_zkvm_utils::should_skip_guest_build("sp1")
```

citizen-stig (approved):

ross-weir (approved):

examples/demo-rollup/configs/mock_rollup_config.toml
bkolad: Will fix in the next pr

examples/demo-rollup/src/zk.rs
bkolad: will fix in the next PR

examples/demo-rollup/configs/mock_rollup_config.toml
citizen-stig: Point was to not have a git log entry for this file pointing to this PR, so if someone investigates some change, it won't have to check this PR.

PR #2757, opened 2026-04-19
Update `SP1Verifier` to use `verify_compressed_with_public_values`, which uses the verification key hash instead of the key itself.

bkolad:
# Description
In the `Chain-State` module, we store verification key hashes rather than the keys themselves. For this reason, we updated the Sp1Verifier to use `verify_compressed_with_public_values`, which works with hashes.

The remaining changes are related to cleaning up `CodeCommitmentHash` vs. `SP1MethodId` usage.

  - Change `SP1MethodId` from a bincode-serialized `SP1VerifyingKey` (`Vec<u8>`) to the canonical `[u32; 8]` vkey hash, eliminating repeated deserialization roundtrips in the host, verifier, and network adapters.                                          
  - Rework `SP1Verifier::verify` (native target) to call `sp1_verifier::compressed::SP1CompressedVerifier` directly against the vkey hash. Adds `sp1-verifier`, `sp1-primitives`, and `slop-algebra` as host-side deps.                                       
  - Make `CodeCommitmentTrait::to_hash` infallible and add a `from_hash` constructor; drop the `CodeCommitment = CodeCommitmentHash` bound on the aggregated-proof circuit so adapters can keep native commitment types end-to-end.                           
  - Thread `SP1VerifyingKey` (instead of a serialized `SP1MethodId`) through `SP1AggregationHost::new` and expose `SP1Host::verifying_key()` / `method_id()` helpers. Update `mock-zkvm` and `risc0` adapters, plus the aggregated-proof test fixtures, to the
   new trait shape

## Linked Issues
- Related to #2551

### DIFF
--- a/crates/adapters/mock-zkvm/src/lib.rs
+++ b/crates/adapters/mock-zkvm/src/lib.rs
@@ -72,13 +72,18 @@ impl Zkvm for MockZkvm {
 pub struct MockCodeCommitment(pub [u8; 8]);
 
 impl sov_rollup_interface::zk::CodeCommitmentTrait for MockCodeCommitment {
-    fn to_hash(
-        &self,
-    ) -> anyhow::Result<sov_rollup_interface::zk::aggregated_proof::CodeCommitmentHash> {
+    fn to_hash(&self) -> sov_rollup_interface::zk::aggregated_proof::CodeCommitmentHash {
         // Pad the 8-byte mock commitment to 32 bytes to match the canonical hash layout.
         let mut bytes = vec![0u8; 32];
         bytes[..8].copy_from_slice(&self.0);
-        Ok(sov_rollup_interface::zk::aggregated_proof::CodeCommitmentHash(bytes))
+        sov_rollup_interface::zk::aggregated_proof::CodeCommitmentHash(bytes)
+    }
+
+    fn from_hash(hash: sov_rollup_interface::zk::aggregated_proof::CodeCommitmentHash) -> Self {
+        let mut bytes = [0u8; 8];
+        let len = hash.0.len().min(8);
+        bytes[..len].copy_from_slice(&hash.0[..len]);
+        Self(bytes)
     }
 }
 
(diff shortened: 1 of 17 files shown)


crates/adapters/mock-zkvm/src/lib.rs
```diff
@@ -72,13 +72,18 @@ impl Zkvm for MockZkvm {
...
         bytes[..8].copy_from_slice(&self.0);
-        Ok(sov_rollup_interface::zk::aggregated_proof::CodeCommitmentHash(bytes))
+        sov_rollup_interface::zk::aggregated_proof::CodeCommitmentHash(bytes)
+    }
+
+    fn from_hash(hash: sov_rollup_interface::zk::aggregated_proof::CodeCommitmentHash) -> Self {
+        let mut bytes = [0u8; 8];
+        let len = hash.0.len().min(8);
+        bytes[..len].copy_from_slice(&hash.0[..len]);
+        Self(bytes)
```
bkolad, line 86: We convert u32 to u8 here but the problem exist only for `MockCodeCommitment` which we initialize to 0 anyway

crates/adapters/sp1/src/lib.rs
```diff
@@ -93,21 +88,40 @@ impl ZkVerifier for SP1Verifier {
...
+            let sp1_sdk::SP1Proof::Compressed(ref compressed) = proof.proof else {
+                anyhow::bail!("SP1Verifier only supports compressed proofs");
+            };
+
+            let vkey_hash: [sp1_primitives::SP1Field; 8] = code_commitment
+                .0
+                .map(sp1_primitives::SP1Field::from_canonical_u32);
+
+            sp1_verifier::compressed::SP1CompressedVerifier::new()
+                .verify_compressed_with_public_values(
```
bkolad, line 108: This is the most important change in tis PR, now we use API that accepts vkey_hash instead of the full vkey.

crates/adapters/sp1/src/host.rs
```diff
@@ -34,32 +35,30 @@ pub struct SP1AggregationHost {
 struct Inner {
     host: SP1Host,
     aggregation_vk: sp1_sdk::SP1VerifyingKey,
```
ross-weir, line 37: unrelated side-note: should we name this `outer_vk` to follow existing inner/outer naming conventions?

bkolad: I will do it in a folloup

ross-weir (approved):

PR #2758, opened 2026-04-20
Typescript multisig upgrade

theodorebugnet:
# Description

Updates the Typescript SDK according to the new multisig signing format for the upgrade.

* `UnsignedTransaction` is now `UnsignedTransactionV0`, with its three basic fields. The new `UnsignedTransction` type is the enum used for serializing for signing, as per the new rust implementation. (See below for some more thoughts on this.)
* Remove `payload` from Multisig package, it's now only used to manage and aggregate signatures (similar to the rust `Multisig` type)
* `web3` now depends on `multisig` (and the `web3` dependency in `multisig` is removed, which was only used for the integration test, which is moved into the `integration-test` package)

Some thoughts on the types: V0 and V1 both share the same basic fields that define a user-created transaction, and which are also included in both Transaction variants. It may or may not be worthwhile to define a special `UnsignedTransactionPayload` type or similar, with the semantics being that the "payload" is what users construct, then an `UnsignedTransaction` is the actual signed bytes (in the V0 case that's the payload + chain hash, in V1 that's payload + multisig credential + chain hash). Or keep `UnsignedTransaction` as the old type and introduce `SigningPayload`.
However, this change should probably be coordinated across both Rust and TS; right now I've kept the TS following my current Rust implementation. If we do rename, the functionality will mostly not change, so I think I'd rather go ahead with this PR and then consider the type ergonomics as a follow-up.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/typescript/packages/multisig/src/index.ts
+++ b/typescript/packages/multisig/src/index.ts
@@ -1,7 +1,7 @@
 import { sha256 } from "@noble/hashes/sha2";
 import type { SignatureAndPubKey } from "@sovereign-sdk/types";
 import type { HexString } from "@sovereign-sdk/utils";
-import { hexToBytes } from "@sovereign-sdk/utils";
+import { hexToBytes, normalizeHexString } from "@sovereign-sdk/utils";
 import * as borsh from "borsh";
 
 const MAX_SIGNERS = 21;
@@ -41,10 +41,17 @@ export class Multisig {
   private unusedPubKeys: Set<HexString>;
 
   constructor({ signatures = [], unusedPubKeys, minSigners }: MultisigParams) {
-    assertValidSignerSet(signatures, unusedPubKeys, minSigners);
+    const normalizedSignatures = signatures.map(normalizeSignatureAndPubKey);
+    const normalizedUnusedPubKeys = unusedPubKeys.map(normalizeHexString);
 
-    this.signatures = [...signatures];
-    this.unusedPubKeys = new Set(unusedPubKeys);
+    assertValidSignerSet(
+      normalizedSignatures,
+      normalizedUnusedPubKeys,
+      minSigners,
+    );
+
+    this.signatures = normalizedSignatures;
+    this.unusedPubKeys = new Set(normalizedUnusedPubKeys);
     this.minSigners = minSigners;
   }
 
@@ -62,10 +69,14 @@ export class Multisig {
     signatureOrPair: HexString | SignatureAndPubKey,
     pubKey?: HexString,
   ): void {
-    const pair =
+    const pair = normalizeSignatureAndPubKey(
       typeof signatureOrPair === "string"
-        ? { signature: signatureOrPair, pub_key: assertDefined(pubKey) }
-        : signatureOrPair;
+        ? {
+            signature: normalizeHexString(signatureOrPair),
+            pub_key: normalizeHexString(assertDefined(pubKey)),
+          }
+        : signatureOrPair,
+    );
 
     if (!this.unusedPubKeys.delete(pair.pub_key)) {
       throw new InvalidMultisigParameterError(
@@ -136,9 +147,9 @@ function assertValidSignerSet(
     );
   }
 
-  if (allPubKeys.length < 1 || allPubKeys.length > MAX_SIGNERS) {
+  if (allPubKeys.length < 2 || allPubKeys.length > MAX_SIGNERS) {
     throw new InvalidMultisigParameterError(
-      `expected 1-${MAX_SIGNERS} total signers, got ${allPubKeys.length}`,
+      `expected 2-${MAX_SIGNERS} total signers, got ${allPubKeys.length}`,
     );
   }
 
@@ -158,6 +169,15 @@ function assertValidSignerSet(
     seenPubKeys.add(pubKey);
     hexToBytes(pubKey);
   }
+}
+
+function normalizeSignatureAndPubKey(
+  pair: SignatureAndPubKey,
+): SignatureAndPubKey {
+  return {
+    signature: normalizeHexString(pair.signature),
+    pub_key: normalizeHexString(pair.pub_key),
+  };
 }
 
 function assertDefined<T>(value: T | undefined): T {
--- a/typescript/packages/multisig/src/index.ts
+++ b/typescript/packages/multisig/src/index.ts
@@ -1,5 +1,9 @@
 import { sha256 } from "@noble/hashes/sha2";
-import type { SignatureAndPubKey } from "@sovereign-sdk/types";
+import type {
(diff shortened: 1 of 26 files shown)


typescript/packages/multisig/src/index.ts
```diff
@@ -1,35 +1,16 @@
...
-  TransactionV0,
-  UnsignedTransaction,
-} from "@sovereign-sdk/types";
+import type { SignatureAndPubKey } from "@sovereign-sdk/types";
 import type { HexString } from "@sovereign-sdk/utils";
-import { bytesToHex, hexToBytes } from "@sovereign-sdk/utils";
+import { hexToBytes } from "@sovereign-sdk/utils";
 import * as borsh from "borsh";
 
+const MAX_SIGNERS = 21;
```
ross-weir, line 11: this would be a useful export or a static property on `Multisig`

typescript/packages/multisig/src/index.ts
```diff
@@ -40,210 +21,93 @@ export class InvalidMultisigParameterError extends MultisigError {
...
-   * @throws {InvalidTransactionVersionError} If the transaction is not V0 variant
-   * @throws {TransactionMismatchError} If the transaction doesn't match the unsigned transaction
-   * @throws {InvalidMultisigParameterError} If the public key is not a member or has already signed
-   */
-  addSignedTransaction(tx: Transaction<unknown>): void {
-    assertTxVariant(tx);
-    assertTxMatchesUnsignedTx(tx, this.unsignedTx);
+  addSignature(signature: HexString, pubKey: HexString): void;
+  addSignature(pair: SignatureAndPubKey): void;
+  addSignature(
```
ross-weir, line 72: Lets keep it simple and only have a single method signature, there's not much benefit to these overloads & it adds cognitive overhead

typescript/packages/multisig/src/index.ts
```diff
@@ -253,74 +117,55 @@ export class MultisigTransaction {
...
+    if (seenPubKeys.has(pubKey)) {
+      throw new InvalidMultisigParameterError(
+        `Duplicate multisig public key: ${pubKey}`,
+      );
+    }
 
-  if (JSON.stringify(txAsUnsigned) !== JSON.stringify(unsignedTx)) {
-    throw new TransactionMismatchError();
+    seenPubKeys.add(pubKey);
+    hexToBytes(pubKey);
```
ross-weir, line 191: If this is just being used for the side-effect of validating the pubkey is valid hex would be worth a comment. Or alternatively having a simple wrapper that throws a multisig exception so library consumers can still `catch MultisigError` and catch all potential errors raised by this lib

typescript/packages/multisig/src/index.ts
```diff
@@ -253,74 +117,55 @@ export class MultisigTransaction {
...
   }
 }
 
-function assertIsNonceBasedTx(unsignedTx: UnsignedTransaction<unknown>): void {
-  if (!("nonce" in unsignedTx.uniqueness)) {
+function assertDefined<T>(value: T | undefined): T {
```
ross-weir, line 204: If this is intended to be a general assertion lets remove `pubKey` from the error message, or if its specifically for pubkeys maybe name it `assertPubKeyDefined`

typescript/packages/web3/src/rollup/standard-rollup.ts
```diff
@@ -22,10 +28,11 @@ export type Dedup = {
 
 export type StandardRollupContext = {
   defaultTxDetails: TxDetails;
+  credentialIdToAddress?: CredentialIdToAddress;
 };
 
 export type StandardRollupSpec<RuntimeCall> = {
-  UnsignedTransaction: UnsignedTransaction<RuntimeCall>;
+  UnsignedTransaction: UnsignedTransactionV0<RuntimeCall>;
```
ross-weir, line 33: Should this be `UnsignedTransaction`?

typescript/packages/web3/src/rollup/standard-rollup.ts
```diff
@@ -94,6 +101,110 @@ export class StandardRollup<RuntimeCall> extends Rollup<
...
+  ): Promise<SignatureAndPubKey> {
+    const signingUnsignedTx = await this.multisigUnsignedTxForSigning(
+      unsignedTx,
+      multisig,
+    );
+
+    return this.signVersionedUnsignedTx(signingUnsignedTx, signer);
+  }
+
+  async signMultisigTransaction(
```
ross-weir: It would be really nice if we didnt need to add these `*multisig*` method variants. It seems like a lot of the "transformation" methods could be defined separately - LMK thoughts or we can discuss when PR ready for real review

typescript/packages/web3/src/rollup/solana-signable-rollup.ts
```diff
@@ -709,36 +715,66 @@ export class SolanaSignableRollup<RuntimeCall> {
...
+    multisig: Multisig,
+    params: SolanaMultisigSubmitParams,
+    options?: SovereignClient.RequestOptions,
+  ): Promise<SovereignClient.Sequencer.TxCreateResponse>;
   async submitMultisigTransaction(
     multisigTx: TransactionV1<RuntimeCall>,
     params: SolanaMultisigSubmitParams,
+    options?: SovereignClient.RequestOptions,
+  ): Promise<SovereignClient.Sequencer.TxCreateResponse>;
+  async submitMultisigTransaction(
```
ross-weir: The overloads here are complex, lets see if we can simplify (or just remove the overloads preferably)

typescript/packages/web3/src/rollup/standard-rollup.ts
theodorebugnet: Nope - the standard rollup lets users feed in an `UnsignedTransactionV0` and internally wraps it in the `V0: { ... }` envelope for serialization by overriding the new `unsignedTxForSigning()` helper.

The type name is not super ergonomic but see my note on typing in the PR description - this works, but I'll see if I can introduce a separation between "the bytes serialized for signing" and "the payload the user assembles" (which are currently both represented by `UnsignedTransaction`, leading to awkward naming like this). This also affects the Rust side in the same way, so if I write a fix it'd be applied to both the rust and TS types.

typescript/packages/multisig/src/index.ts
theodorebugnet: Yep, this was redundant after I cleaned up the validation function above :+1:

typescript/packages/web3/src/rollup/standard-rollup.ts
theodorebugnet: Improved as per discussion on slack

ross-weir (approved):

PR #2763, opened 2026-04-20
Adds metrics to ZK proving related tasks

ross-weir:
# Description

- `sov_rollup_zk_stf_info_channel`: channel depth and capacity of the mpsc channel between state manager and ZK prover — detects backpressure that would block the node
- `sov_rollup_zk_proof_manager`: proving lag (slots between latest received and first unproven), current aggregation batch size, and current slot number
- `sov_rollup_zk_aggregated_proof`: wall-clock duration of aggregated proof generation
- `sov_rollup_zk_network_proof_submit`: duration of the add_hint_and_submit call to SP1 network (from manual observation this can take around 7 seconds)

## Linked Issues
- Related to #2514

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
+++ b/crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
@@ -141,10 +141,9 @@ where
             .await
             .map_err(ProverServiceError::Other)?;
         sov_metrics::track_metrics(|tracker| {
-            tracker.submit_inline(
-                "sov_rollup_zk_network_proof_submit",
-                format!("duration_ms={}i", submit_start.elapsed().as_millis()),
-            );
+            tracker.submit(crate::processes::metrics::ZkNetworkProverMetrics {
+                submit_duration_ms: submit_start.elapsed().as_millis(),
+            });
         });
 
         let StateTransitionWitnessWithAddress {
--- a/crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
+++ b/crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
@@ -143,7 +143,7 @@ where
         sov_metrics::track_metrics(|tracker| {
             tracker.submit_inline(
                 "sov_rollup_zk_network_proof_submit",
-                format!("duration_ms={}", submit_start.elapsed().as_millis()),
+                format!("duration_ms={}i", submit_start.elapsed().as_millis()),
             );
         });
 
--- a/crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
+++ b/crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
@@ -134,11 +134,18 @@ where
                 ProverServiceError::Other(anyhow::anyhow!("DA verification failed: {:?}", e))
             })?;
 
+        let submit_start = std::time::Instant::now();
         let handle = self
             .inner_vm
             .add_hint_and_submit(&data)
             .await
             .map_err(ProverServiceError::Other)?;
+        sov_metrics::track_metrics(|tracker| {
+            tracker.submit_inline(
+                "sov_rollup_zk_network_proof_submit",
+                format!("duration_ms={}", submit_start.elapsed().as_millis()),
+            );
+        });
 
         let StateTransitionWitnessWithAddress {
             stf_witness:
--- a/crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
+++ b/crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
@@ -215,13 +215,9 @@ where
             let aggregation_duration = proving_start.elapsed();
 
             sov_metrics::track_metrics(|tracker| {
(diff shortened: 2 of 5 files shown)


crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
```diff
@@ -134,11 +134,18 @@ where
...
 
+        let submit_start = std::time::Instant::now();
         let handle = self
             .inner_vm
             .add_hint_and_submit(&data)
             .await
             .map_err(ProverServiceError::Other)?;
+        sov_metrics::track_metrics(|tracker| {
+            tracker.submit_inline(
+                "sov_rollup_zk_network_proof_submit",
```
citizen-stig: Could we please avoid inline metrics? I am working on removal of them, because it is not type safe and easy to mess up, and harder for LLM to find

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
```diff
@@ -203,13 +204,25 @@ where
...
                 .create_aggregate_proof_with_retries(
                     metadata,
                     prover_service,
                     &self.genesis_state_root,
                 )
                 .await?;
+            let aggregation_duration = proving_start.elapsed();
+
+            sov_metrics::track_metrics(|tracker| {
+                tracker.submit_inline(
```
citizen-stig: Same about inline metrics

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
```diff
@@ -224,6 +237,17 @@ where
             self.stf_info_receiver
                 .inc_next_height_to_receive_by(num_proofs_to_create as u64);
         }
+
+        sov_metrics::track_metrics(|tracker| {
+            tracker.submit(super::metrics::ZkProofManagerMetrics {
+                proving_lag: received_slot_number
+                    .get()
+                    .saturating_sub(first_height_unproven.get()),
```
citizen-stig: Would saturation hide some bugs? What if we make this value signed, so we can observe "negative" values if they happen?

crates/full-node/sov-stf-runner/src/processes/prover_service/network/prover.rs
ross-weir: done

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
ross-weir: Added logging that checks for this issue explicitly

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
```diff
@@ -168,6 +168,15 @@ where
         >,
     ) -> anyhow::Result<()> {
         let first_height_unproven = self.stf_info_receiver.next_height_to_receive();
+        let received_slot_number = stf_info.slot_number;
+
+        if received_slot_number.get() < first_height_unproven.get() {
+            tracing::warn!(
```
citizen-stig: Should this be error? Or even assert?

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
citizen-stig: Then we safe to use checked sub unwrap, right?

citizen-stig (approved):

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
citizen-stig: This not solves problem originally described here: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2763/changes#r3110066902

Operator sees on the graph, all good, plain zero. 

Operator somehow needs to know to check for a particular warning. If they have bunch of other warnings, from tower, this remains hidden.

ross-weir: Warnings should be much more visible in observability then a signed integer being negative. Why would an operator assume a negative signed int is a issue?

If an invariant around this breaks then there's a big bug upstream, it shouldn't be on metric emission code to try to catch it

I can change this to assert or something if you feel strongly about it

citizen-stig: > Warnings should be much more visible in observability

They probably should, but currently they are not, unfortunately. 
For instance if tower logging is not configured properly warnings/errors would be flooded.

And when there's a issue in the rollup, there's usually bunch of other warnings, which makes it hard for customers to identify issue themself and they ask us to take a look.

> I can change this to assert or something if you feel strongly about it

I believe assert would be appropriate here. I don't want a situation in 3 months, where something like this is hidden and we have a postmortem discussing "Who could've thought". 

If there's "impossible" condition in code and it cannot be expressed via type system, then assert should be used.

ross-weir: done, changed to assert

PR #2766, opened 2026-04-20
Sequencer DA address rotation

citizen-stig:
# Description

**Breaking change (code, wire, consensus).** Adds `CallMessage::UpdateDaAddress { old_da_address, new_da_address }`, letting a registered sequencer move to a new DA address without unstaking.

### Why
Recovery from a compromised DA signing key: rotation is authorized by the *rollup* key (still safe), so the operator keeps stake / balance / preferred status instead of withdrawing and re-registering. The operator must restart the node
with the new DA keys once the tx lands — the registry can't enforce that from state.

### Semantics
- Authorized by `context.sender()` = the rollup address registered at `old_da_address` (blocks hijacking another sequencer's DA).
- Rejected when: `new == old`, `old` not registered, sender doesn't own `old`, `new` already registered, `new` is retired, or a rotation is already pending this block.
- Preserves the `known_sequencers` entry (rollup address, balance, balance_state) and moves the `preferred_sequencer` pointer if it pointed at `old`.
- Dispatch depends on who produces the current batch:
  - preferred sequencer rotating its own active DA → deferred to `end_rollup_block_hook` (one per block; event emitted at scheduling time);
  - non-preferred active producer → rejected (`CannotUnregisterDuringOwnBatch`);
  - anyone else → applied immediately.

### Retired addresses — the load-bearing bit
On rotation `A→B`, `A` is recorded in `retired_da_addresses`. This blocks reusing a retired address, and redirects escrow refunds: `add_to_stake` walks the retired chain (`A→B→C`) before crediting.

The redirect is **not** defensive padding. Deferred blobs (`sov-blob-storage`) capture the sender DA at deferral and settle slots later via `return_escrowed_funds_to_sequencer(sender) → add_to_stake(sender).expect(...)`. If the
sequencer rotated in between (the compromised-key flow routes the rotation through a third party), `add_to_stake(old)` finds nothing, `bail!`s, and the `.expect()` **panics the STF**. The redirect keeps that refund reachable.

⚠️ **Asymmetry to confirm in review:** refunds (`add_to_stake`) redirect, penalties (`remove_part_of_the_stake`) don't. Believed safe because penalties apply at batch admission, before any rotation takes effect — but worth confirming no
flow penalizes a rotated-away DA.

### New state / types (`lib.rs`)
- `retired_da_addresses: KernelStateMap<DaAddress, RetiredDaAddress>`
- `pending_da_address_update: StateValue<PendingDaAddressUpdate>`
- `Event::DaAddressUpdated`, `CustomError::NewDaAddressSameAsOld`

### Tests
16 new integration tests: auth/validation rejections, state preservation (incl. `PendingWithdrawal.ready_at`), preferred-pointer move, the deferred preferred-sequencer path (defer / one-per-block / new-address reservation), and chained
rotations with refund redirect. Full crate suite green (37/37).

### Notes for reviewers
- The penalty asymmetry above is the one spot not covered by tests.
- `CHANGELOG.md` still has a `#PR_NUMBER` placeholder to fill.
- `Cargo.toml` drops the `default-members` list — looks unrelated to this feature; confirm intent.

----

## Testing
New tests are added

## Docs
Docstrings updated


crates/module-system/module-implementations/sov-sequencer-registry/src/call.rs
```diff
@@ -270,6 +298,169 @@ impl<S: Spec> SequencerRegistry<S> {
...
+    /// - If `old_da_address` is not registered.
+    /// - If the caller's rollup key does not own the entry at `old_da_address`.
+    /// - If another DA address update is already pending for this rollup block
+    ///   (by any sequencer — at most one rotation may be pending per block).
+    /// - If the caller is the active non-preferred batch producer for this slot
+    ///   (`CannotUnregisterDuringOwnBatch`). Preferred sequencer self-rotation
+    ///   is deferred until the end of the rollup block.
+    /// - If `new_da_address` is already registered.
+    /// - If `new_da_address` was retired by a previous rotation.
+    pub(crate) fn update_da_address<ST: TxState<S>>(
```
citizen-stig, line 311: simplify this function, logic is convoluted

PR #2767, opened 2026-04-21
Add type safety to the ZK proving workflow

bkolad:
# Description
This PR introduces strong types in palce of `Vec<u8>` for proving workflow, and splits  `ZkVerifier::verify` into `verify_with_proof` (host, takes a `SerializedZkProof`) and `verify_with_pub_values` (guest-only, `#[cfg(target_os = "zkvm")]`, takes a `SerializedPubValues`). Previously both call sites passed ambiguous `&[u8]`.

Changes:
  - Introduce `SerializedZkProof` as the shared wrapper for zk proof bytes
  - Update zkVM host/verifier interfaces and call sites to use typed proof wrappers
  - Split verifier handling between native proof verification and zkVM public-value verification
  - Adapt SP1 aggregation proof plumbing to pass typed inner and aggregated proofs

## Linked Issues
- Related to #2551


crates/module-system/module-implementations/sov-attester-incentives/src/capabilities.rs
```diff
@@ -346,10 +349,11 @@ where
...
 
-        let public_outputs_opt = <<S::InnerZkvm as Zkvm>::Verifier as ZkVerifier>::verify::<
-            StateTransitionPublicData<S::Address, S::Da, <S::Storage as Storage>::Root>,
-        >(proof, &code_commitment)
-        .map_err(|e| anyhow::format_err!("{:?}", e));
+        let public_outputs_opt =
+            <<S::InnerZkvm as Zkvm>::Verifier as ZkVerifier>::verify_with_proof::<
+                StateTransitionPublicData<S::Address, S::Da, <S::Storage as Storage>::Root>,
+            >(proof, &code_commitment)
+            .map_err(|e| anyhow::format_err!("{:?}", e));
```
citizen-stig: nit: Hm, I think we can just do anyhow::anyhow!(e) or something like that to let anyhow to handle type conversion rather then interpolating debug value

crates/module-system/module-implementations/sov-prover-incentives/src/capabilities.rs
```diff
@@ -144,6 +145,12 @@ impl<S: Spec> ProverIncentives<S> {
             }
         };
 
+        tracing::info!(
```
citizen-stig: are we sure we want this to be info and not debug?

citizen-stig (approved):

crates/module-system/module-implementations/sov-attester-incentives/src/capabilities.rs
bkolad: fixed

crates/module-system/module-implementations/sov-prover-incentives/src/capabilities.rs
bkolad: fixed

PR #2771, opened 2026-04-21
Accounts Refactor PR 2: Adding `address_override` param to tx

citizen-stig:
# Description

Adds a signed `address_override`: `Option<S::Address>` field to V0 and V1 (multisig) transactions, threaded into `AuthorizationData.address_override`. 
Resolve_context branches on it: Some(X) verifies (X, credential_id) ∈ account_owners via PR #2770 's is_authorized;
None keeps today's resolve_sender_address flow. 
**CHAIN HASH BREAKING for V1.**

First commit: 7d2ab42284a58299a2f5eb05b69e91d1855315ad

PR #2770 froze writes to the legacy credential_id → address map; 
InsertCredentialId now only marks (sender, credential) authorized, without rerouting that credential's default. 
That means a rotated multisig (new keys → new credential_id.into()) has no way to keep acting as the original address.
`address_override` is the signed channel that lets a signer declare which authorized address to execute as.

## Design tradeoffs

- V0 and V1 wire break, no grace period. `CHAIN_HASH_OVERRIDES` covers signature-level grace only (post-deserialize). Adding a borsh field breaks deserialization of pre-upgrade V0/V1 bytes regardless of chain hash, so a grace period isn't meaningful.
- Unauthorized Some(X) is Skipped, not Reverted. We haven't resolved a sender yet, so there's no account to charge — sequencer pays inclusion cost, user is unaffected. Matches the existing pre-execution-error receipt category (same path as bad signatures). Error string is preserved in the receipt.
- Read-only check. The Some(X) path never writes to account_owners. Auto-registration only happens on the None path.
- Explicit builder parameters. `to_multisig_tx` and `sign_utx_v1_in_place` take the target as an argument rather than defaulting to None with callers assigning the public field afterwards — avoids a half-baked API in an SDK recommended pattern.

## Audit invariants

1. V1's credential_address field (multisig malleability commitment) is untouched.
2. address_overred is inside the V1 signed bytes — post-sign tampering invalidates the signature (test: test_v1_target_tamper_breaks_signature).
3. `Some(_)` path does exactly one read of account_owners; never writes.
4. Failed `Some(_) → TxEffect::Skipped`, sequencer penalized, user never charged.

## Notable side-effects

- `test_multisig_signature_verification` in the EIP-712 integration test was already silently broken by PR #2770 (the multisig credential no longer routed to admin after `InsertCredentialId`, leaving gas on an unfunded default). Fixed here by signing with address_override = Some(admin.address()) — doubles as an EIP-712-meets-target-address smoke test.

---

## Linked Issues
- Related to #2772

## Testing
New tests are added

## Docs
Updated doc-strings and comments


crates/full-node/sov-rollup-apis/src/endpoints/simulate.rs
```diff
@@ -304,23 +304,34 @@ impl<S: Spec, R: Runtime<S>> SovereignSimulate<S, R> {
...
+    ) -> Result<AuthorizationData<S>, SimulateError> {
+        let credential_id = CredentialId::from_str(&params.sender).map_err(|_| {
+            SimulateError::InvalidInput("failed to parse sender credential id".to_owned())
+        })?;
+        let address = params
+            .target_address
+            .as_deref()
+            .map(S::Address::from_str)
+            .transpose()
+            .map_err(|_| {
```
citizen-stig, line 316: why original error is swallowed?

crates/module-system/sov-capabilities/src/lib.rs
```diff
@@ -340,11 +362,7 @@ impl<S: Spec, T> TransactionAuthorizer<S> for StandardProvenRollupCapabilities<'
...
     ) -> anyhow::Result<Context<S>> {
-        let sender = self.accounts.resolve_sender_address(
-            &auth_data.default_address,
-            &auth_data.credential_id,
-            state,
-        )?;
+        let sender = self.resolve_sender(auth_data, state)?;
         // The tx sender & sequencer are the same entity
         Ok(Context::new(
             sender,
```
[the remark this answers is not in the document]
citizen-stig: The broadening is intentional. `auth_data.address` is a signed user declaration carried inside V1 bytes — enforcing `is_authorized(X, credential_id)` uniformly across `resolve_context` and `resolve_unregistered_context` keeps the V1 `target_address` invariant in one place. A V1 self-registering sequencer with `target_address = Some(X)` should still be required to prove the multisig credential is authorized for `X`; the alternative (skipping the check during registration only) would make the signed `target_address` field weaker in registration paths than in normal execution, which is a foot-gun. V0 sets `address = None` and is byte-identical to pre-PR behavior.

Added a doc comment above `resolve_unregistered_context` capturing this invariant in-code.

crates/module-system/module-implementations/integration-tests/tests/stf_blueprint/operator/auth_eip712.rs
theodorebugnet: Now that `target_address` is added, this workaround can be removed and the test can continue exercising the normal semantics of an authorized multisig, making the test simpler

crates/module-system/module-implementations/sov-accounts/tests/integration/main.rs
```diff
@@ -619,3 +620,247 @@ fn test_disable_custom_account_mappings() {
...
+    TransactionType::<RT, S>::PreSigned(RawTx {
+        data: borsh::to_vec(&tx).unwrap(),
+    })
+}
+
+/// V1 tx with `target_address = None` routes through `resolve_sender_address`
+/// and executes as the multisig's default address. Evidence: the `InsertCredentialId`
+/// call writes `(multisig_default, inner_credential)` to `account_owners`.
+#[test]
+fn test_v1_target_none_uses_default_resolver() {
```
theodorebugnet: IMO would be good to also have a test for `target_addess: None` while there _is_ a valid mapping that oculd be used. I.e. add an `alice_address` mapping in genesis, and verify that submitting with `target_address: None` does not edit alice_address.

crates/module-system/sov-solana-offchain-auth/src/authentication/payload.rs
```diff
@@ -78,6 +78,13 @@ pub struct SolanaOffchainUnsignedTransactionV1<R: TransactionCallable, S: Spec>
     /// This is the "multisig address" except if the credential is mapped to another address in
     /// `sov-accounts`.
     pub multisig_id: S::Address,
+    /// Signer-declared target address; routing semantics in
+    /// [`sov_modules_api::capabilities::AuthorizationData::address`].
+    ///
+    /// Omitted from the serialized JSON when `None`, so pre-change signed messages (which have no
+    /// `target_address` field) remain byte-identical and verify against the same signature.
```
theodorebugnet: Omitting from the JSON when None is probably good UX, but compatibility with pre-change messages is not relevant since this branch is going to be a hard fork anyway. (Also not very useful in a doc comment, IMO.)

theodorebugnet (commented):
LGTM, but is there a reason we avoid adding it to V0 transactions? Note that the `theodore/multisig-upgrade` branch is going to be a hard fork and require a new rollup binary regardless, with a coordinated upgrade height using the rollup manager. The previous changes on this branch already break the V0 format (`UnsignedTransaction::V0 {...}` is not byte-compatible with the old `UnsignedTransaction`, by design).

Adding it to both cases would IMO simplify things conceptually: this way, the differences between V0 and V1 are the signing protocol, while the actual transaction features remain the same.

crates/module-system/sov-solana-offchain-auth/src/authentication/payload.rs
citizen-stig: Updated!

crates/module-system/module-implementations/sov-accounts/tests/integration/main.rs
citizen-stig: I think new `test test_v1_address_override_none_ignores_existing_mapping` Should to the job

crates/module-system/sov-capabilities/src/lib.rs
```diff
@@ -334,7 +349,10 @@ impl<S: Spec, T> TransactionAuthorizer<S> for StandardProvenRollupCapabilities<'
         _state: &mut impl StateAccessor,
         execution_context: ExecutionContext,
     ) -> anyhow::Result<Context<S>> {
-        // The tx sender & sequencer are the same entity
+        // `address_override` is intentionally ignored: this path runs for sequencer
+        // self-registration before `account_owners` has any entries to authorize against,
```
citizen-stig: > self-registration before `account_owners` has any entries to authorize against,

Is this true? What if the new sequencer wants to be recovered against existing address controlled by multisig?

crates/module-system/sov-modules-api/tests/integration/transaction.rs
```diff
@@ -75,7 +76,8 @@ fn test_schema_and_native_serialization_consistency() {
                     "max_fee": 10000,
                     "gas_limit": [500, 500],
                     "chain_id": 1337
-                }
+                },
+                "address_override": null
```
citizen-stig, line 80: Do we still need this if we allow it to be unspecified?

crates/module-system/module-implementations/integration-tests/tests/stf_blueprint/operator/auth_eip712.rs
citizen-stig: Discussed offline + test got some refactor

crates/module-system/sov-capabilities/src/lib.rs
citizen-stig: Addressed

crates/module-system/sov-modules-api/tests/integration/transaction.rs
citizen-stig: explicit null is required by the universal-wallet json_to_borsh path.
Optionally mention adding serde(default) propagation to universal-wallet as a follow-up.

crates/module-system/sov-capabilities/src/lib.rs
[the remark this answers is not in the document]
citizen-stig: Revisit this, as V0 also has ~~`target_address`~~ `address_override`. Added comment as follow up in parent issue

crates/module-system/sov-capabilities/src/lib.rs
```diff
@@ -331,14 +346,32 @@ impl<S: Spec, T> TransactionAuthorizer<S> for StandardProvenRollupCapabilities<'
...
         sequencer: &<<S as Spec>::Da as DaSpec>::Address,
-        _state: &mut impl StateAccessor,
+        state: &mut impl StateAccessor,
         execution_context: ExecutionContext,
     ) -> anyhow::Result<Context<S>> {
-        // The tx sender & sequencer are the same entity
+        // The tx sender & sequencer are the same entity on this path. If the signer
+        // declared an `address_override`, the credential must be explicitly authorized
+        // for it (same invariant as `resolve_context`); otherwise we fall back to the
+        // credential's canonical address.
```
theodorebugnet: Minor nit: is there anything actually special about the sequencer address here? I found this comment a bit confusing because it seems to imply there's special handling, but it looks like the logic is identical to `resolve_context()`. If so I'd either remove this comment, or move it to `resolve_context()` so there's no red herring about the unregistered context being special.

We could also even extract the common address resolution logic here into a helper but it's simple enough that I don't think it matters too much

crates/module-system/sov-modules-api/src/runtime/capabilities/authorization.rs
```diff
@@ -102,4 +102,14 @@ pub struct AuthorizationData<S: Spec> {
...
     pub default_address: S::Address,
+
+    /// Signer-declared override of the default execution address.
+    ///
+    /// `None` => resolve to the credential's default address.
+    /// `Some(X)` => requires an explicit `(X, credential_id)` entry in `account_owners`;
+    /// the transaction is skipped otherwise.
+    ///
+    /// Set by authenticators that carry a signed `address_override` (sov-tx V0/V1,
+    /// Solana off-chain auth). Other authenticators leave it `None`.
```
theodorebugnet: I'd remove this - authenticator implementation choices are out of scope for the `TransactionAuthorizer` trait docs IMO, also as written I think this is wrong since sov-eip712-auth supports the field too

crates/module-system/sov-modules-api/src/transaction/types/v0.rs
```diff
@@ -41,16 +41,21 @@ pub struct Version0<R: TransactionCallable, S: Spec, C: CryptoSpecExt = <S as Sp
     pub uniqueness: UniquenessData,
     /// The transaction metadata. Contains gas parameters and the chain ID.
     pub details: TxDetails<S>,
+    /// Signer-declared address override. This field is part of the signed bytes.
```
theodorebugnet:
```suggestion
    /// Signer-declared address override.
```
We don't specify this for any of the other fields that are part of the signed bytes, so I don't see a reason to call it out specifically here. It's also implied by `UnsignedTransaction` containing a corresponding field

crates/module-system/sov-modules-api/src/transaction/types/v1.rs
```diff
@@ -87,6 +87,10 @@ pub struct Version1<R: TransactionCallable, S: Spec, C: CryptoSpecExt = <S as Sp
     pub uniqueness: UniquenessData,
     /// The transaction metadata. Contains gas parameters and the chain ID.
     pub details: TxDetails<S>,
+    /// Signer-declared address override. This field is part of the signed bytes.
```
theodorebugnet:
```suggestion
    /// Signer-declared address override.
```
Same as V0

crates/module-system/sov-modules-api/src/transaction/unsigned/v0.rs
```diff
@@ -27,6 +29,10 @@ pub struct UnsignedTransactionV0<R: TransactionCallable, S: Spec> {
     pub uniqueness: UniquenessData,
     /// Data related to fees and gas handling.
     pub details: TxDetails<S>,
+    /// Signer-declared address override. This field is part of the signed bytes.
```
theodorebugnet:
```suggestion
    /// Signer-declared address override.
```
Same here, all fields in `UnsignedTransaction` are part of the signed bytes

crates/module-system/sov-modules-api/src/transaction/unsigned/v1.rs
```diff
@@ -31,6 +31,10 @@ pub struct UnsignedTransactionV1<R: TransactionCallable, S: Spec> {
     /// and prevent credential malleability from reusing signed bytes in a different
     /// multisig envelope.
     pub credential_address: S::Address,
+    /// Signer-declared address override. This field is part of the signed bytes.
```
theodorebugnet:
```suggestion
    /// Signer-declared address override.
```
Same as V0

crates/module-system/sov-modules-api/src/transaction/unsigned/v0.rs
```diff
@@ -104,19 +114,22 @@ impl<R: TransactionCallable, S: Spec> UnsignedTransactionV0<R, S> {
...
+            address_override: self.address_override,
+        })
     }
 
     /// Creates a new `V1` transaction from this unsigned transaction.
+    /// See [`crate::capabilities::AuthorizationData::address_override`] for routing semantics.
     pub fn to_multisig_tx(
         self,
         multisig: Multisig<<S::CryptoSpec as CryptoSpec>::PublicKey>,
+        address_override: Option<S::Address>,
```
theodorebugnet: Now that we have `self.address_override`, I think this arg is redundant, right?

crates/module-system/sov-solana-offchain-auth/src/authentication/mod.rs
```diff
@@ -122,9 +122,12 @@ fn verify_signatures<S: Spec>(
 }
 
 /// Builds authorization data for either single-sig or multisig transactions.
+/// `address_override` is forwarded from the signed payload — see
+/// [`sov_modules_api::capabilities::AuthorizationData::address_override`] for routing semantics.
```
theodorebugnet: This is to preference, but I feel like this is redundant, e.g. `uniqueness` is also forwarded from the signed payload but is not mentioned

crates/module-system/sov-solana-offchain-auth/src/authentication/payload.rs
```diff
@@ -33,6 +33,10 @@ pub struct SolanaOffchainUnsignedTransactionV0<R: TransactionCallable, S: Spec>
     /// from malicious chains (if the chain name matches some other chain the use but didn't expect
     /// to be signing for right now).
     pub chain_name: SafeString,
+    /// Signer-declared address override. This field is part of the signed bytes.
```
theodorebugnet:
```suggestion
    /// Signer-declared address override.
```
True for all UnsignedTransaction fields

crates/module-system/sov-solana-offchain-auth/src/authentication/payload.rs
```diff
@@ -78,6 +83,11 @@ pub struct SolanaOffchainUnsignedTransactionV1<R: TransactionCallable, S: Spec>
     /// This is the "multisig address" except if the credential is mapped to another address in
     /// `sov-accounts`.
     pub multisig_id: S::Address,
+    /// Signer-declared address override.
+    /// This field is part of the signed bytes.
```
theodorebugnet:
```suggestion
```
Same as V0

theodorebugnet (approved):
LGTM, almost all comments are doc nits except the redundant argument to `to_multisig_tx()` in unsigned/v0.rs which I think needs fixing

crates/module-system/sov-modules-api/src/transaction/unsigned/v0.rs
citizen-stig: That's good point! Slipped through cracks while adding support of this field to v0

citizen-stig:
> LGTM, almost all comments are doc nits except the redundant argument to `to_multisig_tx()` in unsigned/v0.rs which I think needs fixing

Thanks @theodorebugnet , keeping the bar. Many things slipped through while repurposing PR to support of the target_address for V0 tx too.

theodorebugnet (approved):

PR #2774, opened 2026-04-22
Add fulfillment duration metric for network sp1

ross-weir:
# Description

Emit a metric for a single proof request fulfillment time

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/adapters/sp1/src/metrics.rs
+++ b/crates/adapters/sp1/src/metrics.rs
@@ -1,4 +1,9 @@
 //! Defines utilities for collecting runtime metrics from inside a SP1 VM
+use std::io::Write;
+
+use sov_metrics::Metric;
+
+#[cfg(feature = "bench")]
 use sp1_sdk::HookEnv;
 
 /// A custom callback for extracting metrics from the SP1 zkvm.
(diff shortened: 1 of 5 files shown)


crates/adapters/sp1/src/network.rs
```diff
@@ -11,6 +11,39 @@ use sp1_sdk::prover::{ProveRequest, Prover};
...
 
+#[cfg(feature = "metrics")]
+mod metrics {
+    use std::io::Write;
+
+    use sov_metrics::Metric;
+
+    /// Metrics emitted when the SP1 proving network fulfills a proof request.
+    #[derive(Debug)]
+    pub(super) struct SP1ProofFulfillmentMetrics {
```
bkolad: Let's move it to sp1/src/mod so it can be reused for other povers than Network

crates/adapters/sp1/src/network.rs
```diff
@@ -38,6 +71,27 @@ impl SP1Network {
     }
 }
 
+#[cfg(feature = "metrics")]
```
bkolad: In general, we don’t add extra feature flags for metrics in other crates. Could this just be gated behind native instead so  dn't increase the amount of features?

crates/adapters/sp1/src/network.rs
ross-weir: It seems useful to be able to distinguish metrics from different provers rather than diluting cpu and network metrics, also what would we include for request id in the cpu case?

crates/adapters/sp1/src/network.rs
ross-weir: I can but we will always incur an extra REST API call which would be wasted if we get to a point that we don't care about this metric any more (i.e we use reservation network)

IMO `metrics` is a feature flag that should be separate from `native` because we might not always want the overhead of capturing metrics. I can re-use `native` though since it's the current convention

crates/adapters/sp1/src/network.rs
bkolad: We could use Uuid. We can add a dedicated field to distinguish them.

The main benefit is that it avoids duplicating nearly identical code for the CPU prover. It would also help keep the metrics uniform, so we can reuse the same Grafana dashboards across different kinds of provers.

crates/adapters/sp1/src/network.rs
citizen-stig: > In general, we don’t add extra feature flags for metrics in other crates.

I second that. our cargo hack jobs already float between 20 and 30 minutes.

And amount of times we had to deal with weird feature unification is over budget =)

bkolad: Also the metircs are submitted in a separte thread so it won't slowdown the proof processing code.

crates/adapters/sp1/src/network.rs
ross-weir: Added a prover_type field

crates/adapters/sp1/src/network.rs
ross-weir: I think it's a bad idea not to have a way to disable metrics but will move to `native` since that's already what we're already doing everywhere

ross-weir: > Also the metircs are submitted in a separte thread so it won't slowdown the proof processing code.

It adds a `get_proof_request` API call for each proof that isn't done on the submission thread. Also, speaking generally it's not always about submission, sometimes producing the metrics you want to capture can be non-negligible and it all adds up across the system

crates/adapters/sp1/src/metrics.rs
```diff
@@ -1,13 +1,68 @@
...
+#[cfg(feature = "bench")]
 use sp1_sdk::HookEnv;
 
 /// A custom callback for extracting metrics from the SP1 zkvm.
 ///
 /// When the "bench" feature is enabled, this callback is registered as a syscall
 /// in the SP1 VM and invoked whenever a function annotated with the `cycle_tracker`
 /// macro is invoked.
+#[cfg(feature = "bench")]
 pub fn metrics_hook(_env: HookEnv, buf: &[u8]) -> Vec<Vec<u8>> {
```
ross-weir, line 15: I'm going to remove this in a follow-up, it's dead code

bkolad (approved):

PR #2785, opened 2026-04-25
Restore outer-host aggregation state on restart 

bkolad:
# Description

This PR makes the outer-VM aggregated proof pipeline survive a node restart. Before this change, after a restart, the prover would forget the previous aggregated proof, and either start a fresh aggregation that wasn't contiguous with what was already on disk, or fail recursive verification.                                                                                                                                                                                                                       

Core idea:        
                                                                                                                                                                                                           On startup, load the latest aggregated proof from the ledger DB, verify it, extract its public data, and seed two things with it:                                                                                                                           
  1. The outer VM host — so the next aggregation is recursively chained to the previous one.
  2. The STF-info channel — so the runner replays slots starting at previous_final_slot + 1.

Changes: 

  - On startup, load the latest aggregated proof from the ledger DB and feed it into the outer ZKVM host (mock + SP1) so that aggregation-continuity assertions hold across a restart.                                                                        
  - Use that proof's `final_slot_number` to anchor the STF-info channel at `final_slot + 1`.                                                                                                                                                            
  - Port `external_mock_rollup.rs` from Risc0 to `MockZkvm` so the new restart test runs without real proving.                                                                                                                                                

## Linked Issues
- Related to #2551

### DIFF
--- a/crates/adapters/mock-zkvm/src/host.rs
+++ b/crates/adapters/mock-zkvm/src/host.rs
@@ -1,7 +1,11 @@
+use std::sync::{Arc, Mutex};
+
 use crate::notifier::NotificationManager;
 use crate::{MockCodeCommitment, MockProof, MockZkGuest};
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
+use sov_rollup_interface::common::SlotNumber;
 use sov_rollup_interface::da::DaSpec;
+use sov_rollup_interface::zk::aggregated_proof::AggregatedProofPublicData;
 use sov_rollup_interface::zk::aggregated_proof::{
     BlockProof, OuterZkvmHost, SerializedAggregatedProof,
 };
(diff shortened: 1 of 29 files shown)


crates/adapters/mock-zkvm/src/host.rs
```diff
@@ -12,6 +17,48 @@ use sov_rollup_interface::zk::SerializedZkProof;
 pub struct MockZkvmHost {
     notification_manager: NotificationManager,
     wait_for_proof: bool,
+    /// Anchor extracted from the most recently produced aggregated proof.
+    /// Shared across clones so that continuity assertions in
+    /// [`OuterZkvmHost::run_proof_aggregation`] hold across the whole prover
+    /// service.
+    previous_anchor: Arc<Mutex<Option<PreviousAggregatedProofAnchor>>>,
```
bkolad, line 24: This is added in MockZkvmHost to test that consecutive aggregated proofs are continuous.

ross-weir (approved):

PR #2787, opened 2026-04-27
Add timelock capability to runtime, and with placeholder module

theodorebugnet:
# Description

Implements the timelock capability. Adds a barebones `sov-timelock` crate that implements no-op timelocks for now - this will be filled in in a follow-up PR.

I've left it out of `StandardCapabilities` and instead it's used through an accessor method (which defaults to a no-op implementation). This means rollups won't have to implement timelocks if they don't want to, in particular this means we can avoid changing the chain_hash for existing rollups that don't need it or forcing them to add a new module into their Runtime. But it does have limited value past this one-time upgrade, so if we want to make it mandatory we can add it back to the StandardCapabilities at the cost of a chain_hash change when it's merged.

Main files to review are the capability trait (`crates/module-system/sov-modules-api/src/runtime/capabilities/timelock.rs`) and the modifications to attempt_tx (in `crates/module-system/sov-modules-stf-blueprint/src/sequencer_mode/common.rs`)

- [ ] ~~I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.~~ Will be added after the module is implemented. Leaving it undocumented for now.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/module-implementations/sov-timelock/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+        Ok(())
+    }
+
+    fn call(
+        &mut self,
+        _msg: Self::CallMessage,
+        _context: &Context<S>,
+        _state: &mut impl TxState<S>,
+    ) -> Result<(), Self::Error> {
+        unreachable!()
```
preston-evans98: Let's not panic here even though this should be unreachable. Just `Ok(())` is probably fine

crates/module-system/sov-modules-api/src/runtime/capabilities/timelock.rs
```diff
@@ -0,0 +1,133 @@
...
+/// Identifier for a timelock proposal.
+///
+/// This is the hash of an encoded runtime call message, not the hash of a raw transaction.
+pub type ProposalId = HexHash;
+
+/// Timelock policy returned by a runtime for call messages that must be delayed.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub struct TimelockPolicy {
+    /// Number of seconds that must pass before the proposal may be unlocked.
+    pub unlock_seconds_from_now: NonZeroU64,
```
preston-evans98: (Non-blocking) I'm curious about the choice of `NonZeroU64` here. It means that users can't easily unify code paths for timelock and standard txs by just setting the lock time to 0. Is that intentional?

crates/module-system/sov-modules-api/src/runtime/capabilities/timelock.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Number of seconds after unlock during which the proposal may be executed.
+    pub expire_seconds_after_unlock: NonZeroU64,
```
preston-evans98: Same question about `NonZeroU64` here

preston-evans98 (approved):

crates/module-system/sov-modules-api/src/runtime/capabilities/timelock.rs
theodorebugnet: I think it could go either way; IMO having `Some(TimelockPolicy)` should imply that there's a timelock going on, and returning `Some(TimelockPolicy { unlock_seconds: 0 })` would be semantic abuse: just return `None` if you intend for it to not be timelocked.

On the other hand it's probably not a huge deal so we can relax this, I don't have a strong opinon. What do you think?

crates/module-system/sov-modules-api/src/runtime/capabilities/timelock.rs
theodorebugnet: Hmm. If we set it to 0 here then a proposal will only be executable on the exact same second that it gets unlocked. On the other hand setting it to 1 does not make it much easier to execute. I feel like using `NonZero` makes it harder to hit the footgun of just leaving the expiry time on "default" and having proposals nearly impossible to execute, but also doesn't provide very useful validation.

I'm implementing the `sov-timelock` module right now and I've started thinking that maybe requiring an explicit expiry time is actually an unnecessary burden on users and developers; perhaps it'd be better to define a `DEFAULT_MAX_EXPIRY_TIME` as a sane value, e.g. 24h, and use that by default. Then make `TimelockPolicy::expire_seconds_after_unlock` an `Option<u64>`, and then:
 * `None` would use `min(unlock_seconds_from_now, DEFAULT_MAX_EXPIRY_TIME)` which should give a sane value for the majority of usecases, scaling down for very short unlocks while having a decent default for longer ones;
 * Users would still have the freedom to override it with `Some(u64)` if they really want to (which would also allow 0 if the user really wants to explicitly override it that way).

Thus removing some of the congitive burden of specifying an expiry time for every single timelock. What do you think?

theodorebugnet: Looking at how other chains do it, it seems like the most common approaches are either a default expiry time of several days, or just leaving it valid forever.
So I went ahead and just added a simple default, since it seems there's not much precedent for very short expiry times, and users can override if needed.

PR #2788, opened 2026-04-27
  Pipeline proof aggregation.

bkolad:
# Description

 This PR pipelines the aggregated-proof workflow so the prover can keep ingesting STF infos while the previous aggregation/publication is in flight.

Before this change, the flow was more serialized than it needed to be.
Assume aggregated_proof_block_jump = 5:

  1. For each proof window, we would submit up to 5 batch-proof jobs.
  2. Once those 5 batch proofs were finished, we would start aggregate-proof generation.
  3. Only after aggregate-proof generation completed would we start submitting the next 5 batch proofs.

  The key observation is that steps 2 and 3 do not actually depend on each other. While one aggregate proof is being generated and published, we can already start proving the next batch window. This PR fixes exactly that: ZkProofManager now overlaps
  batch-proof production with aggregate-proof generation, so both stages can run in parallel.

  This PR also changes the worker-count logic to num_threads = 2 * aggregated_proof_block_jump + 1.

  Previously, num_threads defaulted to num_cpus, which on the dev machine was 97. That was too aggressive for the network prover: the remote service could not keep up, and we were frequently hitting invalid nonce errors due to excessive parallel
  submissions.

  With aggregated_proof_block_jump = 5, the new setting gives num_threads = 11, so we submit at most 11 proving jobs in parallel. That has been a much more reasonable limit so far, though this value is still a tuning parameter and can be adjusted further
  if needed.

## Linked Issues
- Related to #2551

### DIFF
--- a/crates/full-node/full-node-configs/src/runner.rs
+++ b/crates/full-node/full-node-configs/src/runner.rs
@@ -139,6 +139,15 @@ fn default_eager_proof_submission() -> bool {
     true
 }
 
+impl<Address> ProofManagerConfig<Address> {
+    /// Number of prover threads required to fully pipeline aggregation:
+    /// `2 * aggregated_proof_block_jump + 1` covers inner proofs for the
+    /// current and next batch plus one outer-aggregation worker.
+    pub fn prover_thread_count(&self) -> usize {
+        2 * self.aggregated_proof_block_jump.get() + 1
+    }
+}
+
 /// Rollup Configuration
 #[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
 #[schemars(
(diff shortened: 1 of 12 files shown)


crates/full-node/full-node-configs/src/runner.rs
```diff
@@ -139,6 +139,15 @@ fn default_eager_proof_submission() -> bool {
     true
 }
 
+impl<Address> ProofManagerConfig<Address> {
+    /// Number of prover threads required to fully pipeline aggregation:
+    /// `2 * aggregated_proof_block_jump + 1` covers inner proofs for the
+    /// current and next batch plus one outer-aggregation worker.
+    pub fn prover_thread_count(&self) -> usize {
+        2 * self.aggregated_proof_block_jump.get() + 1
```
ross-weir, line 147: If we decide to use a big jump we could run into this again:

> Previously, num_threads defaulted to num_cpus, which on the dev machine was 97. That was too aggressive for the network prover: the remote service could not keep up, and we were frequently hitting invalid nonce errors due to excessive parallel
submissions.

Is it worth protecting against?

ross-weir (approved):

crates/full-node/full-node-configs/src/runner.rs
bkolad: I think let's leave it for now and if we hit it then we implement a smarte heuristics in prover_thread_count

PR #2792, opened 2026-04-28
Bug fix: switch from systemtime to std::time::instant for execution time

preston-evans98:
# Description

This PR changes how we track the execution time of transactions in the sequencer. Previously, we relied on two calls to `SystemTime::now`. This was incorrect, since systemtime is not monotonic; NTP or manual clock changes can change its results. This PR replaces those calls with std::time::Instant, storing the result in a RefCell rather than relying on an atomic number of nanoseconds since the epoch.

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/async_batch.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/async_batch.rs
@@ -1,6 +1,4 @@
-use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
-use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::preferred::cache_warm_up_executor::FullyBakedTxWithMaybeChangeSet;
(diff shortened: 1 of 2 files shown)


bkolad (approved):

crates/full-node/sov-sequencer/src/preferred/async_batch.rs
```diff
@@ -172,7 +170,7 @@ pub struct AsyncBatchResponder<S: Spec> {
     /// The timestamp of the start of the latest tx in microseconds since the UNIX epoch
```
theodorebugnet: Nit: the comment is outdated

theodorebugnet (approved):

crates/full-node/sov-sequencer/src/preferred/async_batch.rs
preston-evans98: Good catch. I renamed the field as well

PR #2794, opened 2026-04-29
  Add optional sleep throttle for mock provers

bkolad:
# Description

  - Adds `maybe_mock_sleep` helper to both `MockZkvmHost` and SP1's mock backend that sleeps for a duration read from an env var before proving/aggregating.
  - `MOCK_PROVE_SLEEP_MS` throttles inner proof generation; `MOCK_AGGREGATION_SLEEP_MS` throttles outer aggregation.                                                                                                                                          
  - No-op when the env var is unset/unparseable, or (for SP1) when not running under `EnvProver::Mock`. Real provers are unaffected.  

## Linked Issues
- Related to # (issue)

### DIFF
--- a/crates/adapters/mock-zkvm/src/host.rs
+++ b/crates/adapters/mock-zkvm/src/host.rs
@@ -131,6 +131,19 @@ impl MockZkvmHost {
             pub_data,
         })?)
     }
+
+    /// Sleeps for the duration (in milliseconds) read from `env_var`. Used in
+    /// tests/soak runs to throttle the otherwise-instant mock prover. No-op
+    /// if the env var is unset or not parseable.
+    fn maybe_mock_sleep(env_var: &str) {
+        let Some(ms) = std::env::var(env_var)
+            .ok()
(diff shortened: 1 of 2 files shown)


crates/adapters/mock-zkvm/src/host.rs
```diff
@@ -151,6 +164,7 @@ impl sov_rollup_interface::zk::ZkvmHost for MockZkvmHost {
         item: &T,
         _agg_proofs: Vec<SerializedAggregatedProof>,
     ) -> anyhow::Result<SerializedZkProof> {
+        Self::maybe_mock_sleep("MOCK_PROVE_SLEEP_MS");
```
ross-weir: We usually prefix these env vars with `SOV_*` prefix

ross-weir (approved):

crates/adapters/mock-zkvm/src/host.rs
bkolad: Adding it

PR #2796, opened 2026-04-30
  Make prover thread count and aggregated-proof buffer configurable.

bkolad:
# Description

This PR introduces the folloing changes:

1. Add `prover_thread_count_override: Option<NonZero<usize>>` to `ProofManagerConfig`. When `None` (default), `prover_thread_count()` keeps its previous value of `2 * aggregated_proof_block_jump + 1`; when `Some(n)`, the override is used directly.
2. Add `max_number_of_aggregated_proofs_in_memory: NonZero<usize>` to `ProofManagerConfig` and thread it through `start_zk_workflow_in_background` → `ZkProofManager`. This becomes the capacity of the intake → aggregator mpsc (previously hardcoded to 1),
   so operators can let more completed aggregation windows queue in memory before back-pressure stalls intake.                                                                                                                                                
3. Update the runner's config validation: `max_number_of_transitions_in_db & max_number_of_transitions_in_memory`                                                                                                                                                                          
4. Set `max_number_of_aggregated_proofs_in_memory = 5` in the example/demo TOMLs and test rollup builders, and bump `max_number_of_transitions_in_db`/`_in_memory` accordingly so the new validation passes.

## Linked Issues
- Related to #2551

### DIFF
--- a/crates/full-node/full-node-configs/src/runner.rs
+++ b/crates/full-node/full-node-configs/src/runner.rs
@@ -134,7 +134,8 @@ pub struct ProofManagerConfig<Address> {
     #[serde(default = "default_eager_proof_submission")]
     pub eager_proof_submission: bool,
     /// Override for the number of prover threads. When `None`, defaults to
-    /// `2 * aggregated_proof_block_jump + 1`.
+    /// `2 * aggregated_proof_block_jump + 1`, which covers inner proofs for
+    /// the current and next batch plus one outer-aggregation worker.
     #[serde(default)]
     pub prover_thread_count_override: Option<NonZero<usize>>,
     /// Maximum number of completed aggregation windows that can be buffered in
--- a/crates/full-node/full-node-configs/src/runner.rs
+++ b/crates/full-node/full-node-configs/src/runner.rs
@@ -133,10 +133,15 @@ pub struct ProofManagerConfig<Address> {
     /// When false, submission is deferred until the batch is full, then submitted concurrently.
     #[serde(default = "default_eager_proof_submission")]
     pub eager_proof_submission: bool,
-    /// Number of prover threads. If unset, defaults to
-    /// `2 * aggregated_proof_block_jump + 1`, which fully pipelines aggregation.
+    /// Override for the number of prover threads. When `None`, defaults to
+    /// `2 * aggregated_proof_block_jump + 1`.
     #[serde(default)]
-    pub prover_thread_count: Option<NonZero<usize>>,
+    pub prover_thread_count_override: Option<NonZero<usize>>,
+    /// Maximum number of completed aggregation windows that can be buffered in
+    /// memory between the intake task and the aggregator task. When the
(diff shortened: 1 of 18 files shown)


crates/full-node/full-node-configs/src/runner.rs
```diff
@@ -133,18 +133,27 @@ pub struct ProofManagerConfig<Address> {
     /// When false, submission is deferred until the batch is full, then submitted concurrently.
     #[serde(default = "default_eager_proof_submission")]
     pub eager_proof_submission: bool,
+    /// Override for the number of prover threads. When `None`, defaults to
+    /// `2 * aggregated_proof_block_jump + 1`.
+    #[serde(default)]
+    pub prover_thread_count_override: Option<NonZero<usize>>,
```
bkolad, line 140: Now we can control the thread pool size in the parallel prover.

crates/full-node/full-node-configs/src/runner.rs
```diff
@@ -133,18 +133,27 @@ pub struct ProofManagerConfig<Address> {
...
+    /// Maximum number of completed aggregation windows that can be buffered in
+    /// memory between the intake task and the aggregator task. When the
+    /// aggregator falls behind by this many windows, intake stalls and
+    /// back-pressure propagates upstream.
+    pub max_number_of_aggregated_proofs_in_memory: NonZero<usize>,
```
bkolad, line 145: Now `max_number_of_aggregated_proofs_in_memory` is configurable. Before this PR, it was hardcoded to 1.
This means that if proving the outer proof is much slower than `jump * inner_proof_time`, we can buffer outer proof metadata in memory instead of leaving the thread pool underutilized.

I set the value to 5 in the configs. If the value is too small, even a slight delay in proof aggregation can propagate through the system. The buffer helps mitigate that. If the value is too large, we may continue proving inner proofs even when outer proofs are much slower, which can cause a large amount of data to accumulate in memory. That work would also be lost after a rollup restart.

crates/full-node/full-node-configs/src/runner.rs
```diff
@@ -133,18 +133,27 @@ pub struct ProofManagerConfig<Address> {
     /// When false, submission is deferred until the batch is full, then submitted concurrently.
     #[serde(default = "default_eager_proof_submission")]
     pub eager_proof_submission: bool,
+    /// Override for the number of prover threads. When `None`, defaults to
+    /// `2 * aggregated_proof_block_jump + 1`.
```
theodorebugnet: Very minor nit: I think the previous implementation had useful clarification in the comments, so maybe we can keep it in the option docs:
```suggestion
    /// Override for the number of prover threads. When `None`, defaults to
    /// `2 * aggregated_proof_block_jump + 1`, which covers inner proofs for
    /// the current and next batch plus one outer-aggregation worker.
```

Up to preference though

theodorebugnet (approved):

PR #2798, opened 2026-04-30
`sov-timelock` module

theodorebugnet:
# Description

Implements `sov-timelock` for timelock support.

Main things to review:
1. The `TimelockCapablity` introduced in a previous PR has been simplified, since a single method makes the timelock state machine more ergonomic to use. Changes in `sov-modules-api/runtime/capabilities/timelock.rs` and `stf-blueprint/src/sequencer_mode/common.rs`.
2. The `sov-timelock` module itself, main logic being in lib.rs and call.rs.

`sov-timelock` stores timelocked proposals keyed by `(Address, ProposalId)`. The ProposalId is a domain-separated hash so we can support both Runtime::Decodable callmessages, and separately arbitrary module-owned data; the latter is used in `sov-timelock` to implement a timelock on the `ModifyCancellationPolicy` callmessage.

We also track the number of timelocks per user in a separate state item, to enforce the `MAX_PENDING_PROPOSALS_PER_ADDRESS` upper bound.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/module-implementations/integration-tests/tests/stf_blueprint/optimistic/timelock.rs
+++ b/crates/module-system/module-implementations/integration-tests/tests/stf_blueprint/optimistic/timelock.rs
@@ -1,14 +1,20 @@
 use std::num::NonZeroU64;
 
-use sov_modules_api::capabilities::TimelockPolicy;
+use sov_modules_api::capabilities::{
+    calculate_timelock_proposal_id, TimelockPolicy, TimelockProposalHashData,
+    DEFAULT_EXPIRE_SECONDS_AFTER_UNLOCK,
+};
+use sov_modules_api::da::Time;
 use sov_modules_api::prelude::UnwrapInfallible;
+use sov_modules_api::EncodeCall;
 use sov_test_modules::hooks_count::TxHooksCount;
 use sov_test_utils::runtime::genesis::optimistic::HighLevelOptimisticGenesisConfig;
 use sov_test_utils::runtime::{TestRunner, ValueSetter};
 use sov_test_utils::{
     generate_optimistic_runtime_with_kernel, AsUser, TestUser, TransactionTestCase,
 };
-use sov_value_setter::{CallMessage, ValueSetterConfig};
+use sov_timelock::{CancellationPolicy, Timelock, TimelockKey};
+use sov_value_setter::{CallMessage as ValueSetterCallMessage, ValueSetterConfig};
 
 use crate::stf_blueprint::S;
 
@@ -22,7 +28,7 @@ generate_optimistic_runtime_with_kernel!(
     ],
     timelock_policy_wrapper: |call: &TimelockRuntimeCall<S>| {
         match call {
-            TimelockRuntimeCall::ValueSetter(CallMessage::SetValue { value: 7, .. }) => {
+            TimelockRuntimeCall::ValueSetter(ValueSetterCallMessage::SetValue { value: 7, .. }) => {
                 Some(TimelockPolicy {
                     unlock_seconds_from_proposal: NonZeroU64::new(60).unwrap(),
                     expire_seconds_after_unlock_override: Some(60),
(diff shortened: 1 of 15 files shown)


crates/module-system/module-implementations/sov-timelock/src/lib.rs
```diff
@@ -3,23 +3,192 @@
...
+    }
+}
+
+impl<S: Spec> Display for ProposalKey<S> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}{PROPOSAL_KEY_SEPARATOR}{}", self.0, self.1)
+    }
+}
+
+impl<S> FromStr for ProposalKey<S>
```
bkolad, line 72: Is this implementation used anywhere?

crates/module-system/module-implementations/integration-tests/tests/stf_blueprint/optimistic/timelock.rs
```diff
@@ -2,27 +2,27 @@ use std::num::NonZeroU64;
...
+    NoopTimelockRuntime <=
     kernel_type: sov_test_utils::runtime::BasicKernel<'a, S>,
     modules: [
-        value_setter: ValueSetter<S>,
-        tx_hooks_count: TxHooksCount<S>,
-        timelock: sov_timelock::Timelock<S>
+        value_setter: ValueSetter<S>
     ],
-    timelock_policy_wrapper: |call: &TimelockRuntimeCall<S>| {
+    timelock_policy_wrapper: |call: &NoopTimelockRuntimeCall<S>| {
```
bkolad, line 21: Why is it `NoopTimelockRuntimeCall`?

crates/module-system/sov-modules-stf-blueprint/src/sequencer_mode/common.rs
```diff
@@ -139,23 +140,18 @@ fn attempt_tx<S: Spec, RT: Runtime<S>, I: StateProvider<S>>(
...
-            {
-                let mut timelock = runtime.timelock();
-                timelock.try_unlock_proposal(ctx.sender(), &proposal_id, state)?;
-            }
-            runtime.dispatch_call(message, state, ctx)?;
-        } else {
+        let outcome = {
             let mut timelock = runtime.timelock();
-            timelock.register_proposal(ctx.sender(), proposal_id, policy, state)?;
+            timelock.register_or_try_unlock_proposal(
```
bkolad, line 145: `pre_dispatch_tx_hook and post_dispatch_tx_hook` are triggered events if a tx is only registered but not executed. Is that correct? Shouldn't we trigger the hooks only when the tx gets executed?

crates/module-system/module-implementations/sov-timelock/src/call.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    pub(super) fn clean_expired_proposals(
+        &mut self,
+        proposal_keys: SafeVec<ProposalKey<S>, MAX_PROPOSAL_KEYS_PER_CLEANUP>,
+        state: &mut impl TxState<S>,
+    ) -> Result<(), Error<S>> {
+        let current_time = self.current_time_secs(state)?;
+        for key in proposal_keys {
+            let mut attempt_state = state.to_revertable();
```
bkolad: Why do we need a reversible state here? If we rearrange the code to return an error and use `?`, the state will be reverted automatically. Also, in `clean_expired_proposal`, `ProposalNotFound & ProposalNotExpired` don't modify the state.

crates/module-system/module-implementations/sov-timelock/src/lib.rs
theodorebugnet: It's a necessary bound for StateMap keys

crates/module-system/sov-modules-stf-blueprint/src/sequencer_mode/common.rs
theodorebugnet: IMO yes; I built this with the logic that registering a timelock is a pseudo-dispatch to `sov-timelock`. You get back a receipt, with events (the event is "timelock registered"), so to the user it looks like a normal transaction; we just replace the user's CallMessage with a call to sov-timelock.

Happy to discuss if you disagree or think there's any reason not to do this though

crates/module-system/module-implementations/sov-timelock/src/call.rs
theodorebugnet: Reversible state lets us return errors internally, so this function can skip over non-existing proposals and do a best-effort cleanup.

It's true that right now the error paths in `clean_expired_proposal()` don't modify state, so reverting state is not really necessary, we can just ignore the errors. Do you think it would be clearer if there was no reverting?

crates/module-system/module-implementations/integration-tests/tests/stf_blueprint/optimistic/timelock.rs
theodorebugnet: That's what the macro generates for `NoopTimelockRuntime`. And the runtime is called that because it's a runtime configured to have the noop timelock (the impl on `()`)

bkolad: Yeah, my question was about the name,

crates/module-system/sov-modules-stf-blueprint/src/sequencer_mode/common.rs
bkolad: Ok cool

crates/module-system/module-implementations/sov-timelock/src/call.rs
bkolad: Looking at the flow and the function:

```
CallMessage::CleanExpiredProposals =>{
self.clean_expired_proposals(proposal_keys, state)
                    .map_err(ModuleError::from)?;
}
```
Already reverts on any error, and in:
```

    pub(super) fn clean_expired_proposals(
        &mut self,
        proposal_keys: SafeVec<ProposalKey<S>, MAX_PROPOSAL_KEYS_PER_CLEANUP>,
        state: &mut impl TxState<S>,
    ) -> Result<(), Error<S>> {
        let current_time = self.current_time_secs(state)?;
        for key in proposal_keys {
            let mut attempt_state = state.to_revertable(); 
            match self.clean_expired_proposal(&key, current_time, &mut attempt_state) {
                Ok(()) => {
                    attempt_state.commit();
                }
                Err(Error::Timelock(TimelockError::ProposalNotFound))
                | Err(Error::ProposalNotExpired { .. }) => {
                    attempt_state.revert();
                }
                Err(error) => {
                    attempt_state.revert();
                    return Err(error);
                }
            }
        }

        Ok(())
    }
 ```
There is this `attempt_state`. We do manual reverts. 
In my opinion Err(Error::Timelock(TimelockError::ProposalNotFound))| Err(Error::ProposalNotExpired { .. })` should also return an error instead of `()`. This way, we won't need attempt_state, and the semantics will be "Revert on any error." And it will be handled automaically

bkolad (approved):

crates/module-system/module-implementations/sov-timelock/src/call.rs
theodorebugnet: As discussed on Slack, simplified the cleanup CallMessage to fail-fast, which is easier to reason about and likely less unexpected for users.

PR #2806, opened 2026-05-04
Add proving cycles/pgu metrics

ross-weir:
# Description

*Summary of the changes...*

## Linked Issues
- Related to #2551

## Testing

## Docs

### DIFF
--- a/crates/adapters/sp1/src/host.rs
+++ b/crates/adapters/sp1/src/host.rs
@@ -3,9 +3,10 @@
 use std::sync::{Arc, Mutex};
 
 use crate::guest::SP1Guest;
-use crate::metrics::submit_network_proving_metric;
+use crate::metrics::submit_proving_metric;
 use crate::SP1MethodId;
 use serde::Serialize;
+use sov_metrics::ZkCircuit;
 use sov_rollup_interface::da::DaSpec;
 use sov_rollup_interface::reexports::anyhow;
 use sov_rollup_interface::zk::aggregated_proof::common::{
@@ -56,7 +57,7 @@ impl SP1AggregationHost {
         inner_vk: sp1_sdk::SP1VerifyingKey,
         previous_aggregated_proof: Option<SerializedAggregatedProof>,
     ) -> anyhow::Result<Self> {
-        let prover = SP1Prover::new(elf)?;
+        let prover = SP1Prover::new_outer(elf)?;
         let outer_vk = prover.verifying_key().clone();
 
         let prev_agg_proof = previous_aggregated_proof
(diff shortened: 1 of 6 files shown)


crates/adapters/sp1/src/host.rs
```diff
@@ -212,6 +214,16 @@ impl SP1Prover {
     }
 
     fn run(&self, stdin: SP1Stdin) -> anyhow::Result<sp1_sdk::SP1ProofWithPublicValues> {
+        // For network proving we go through the explicit two-step API
+        // (`request()` + `wait_proof()`) so we can capture the request_id and
+        // afterwards fetch the canonical cycles + PGUs the network recorded.
+        if let EnvProver::Network(network) = &self.prover {
```
ross-weir, line 232: We can't use the normal `run` method because it doesn't return the request id which is needed to check the proof gas usage. So we need to specifically use the NetworkProver and retrieve the details from the network

crates/adapters/sp1/src/metrics.rs
```diff
@@ -0,0 +1,67 @@
...
+/// fulfilled request without them, we still want the (empty) datapoint so the
+/// dashboard surfaces the gap.
+#[derive(Debug)]
+pub(crate) struct Sp1NetworkProvingMetric {
+    /// Identifier for the proven program — `program_name` from the network when
+    /// set, otherwise the hex-encoded `vk_hash`. Different ELFs (inner state
+    /// transition vs. outer aggregation, or upgraded versions of either)
+    /// produce different values, so this tag is what distinguishes circuits in
+    /// the dashboard.
+    pub program: String,
```
bkolad: We have `ZkCircuit::Inner/Outer` enum that is used in other metrics. Let's use this enum for indentifying the program.

crates/adapters/sp1/src/metrics.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub cycles: Option<u64>,
+    pub gas_used: Option<u64>,
+}
+
+impl Metric for Sp1NetworkProvingMetric {
```
bkolad: Let’s drop “Network” from the name. There may be a way to get the same metrics for the mock/CPU proofs, so we can reuse the metric. We can add the other mock/CPU metrics another time; for now, this is only about the metric name.

bkolad (approved):

crates/adapters/sp1/src/metrics.rs
ross-weir: The reason I went with this was because it would allow the program type to be derived whereas `ZkCircuit` required threading the value around `SP1Prover`

Moved to using the enum though, passing the enum around seems fine

crates/adapters/sp1/src/metrics.rs
ross-weir: done

PR #2808, opened 2026-05-04
Add `max_concurrent_proof_blobs` cap with rollup shutdown on saturation. 

### LINKED ISSUE #2800: Aggregated ZK proofs posted to DA during node sync are never processed onchain
### Summary

  When the prover is persistently slower than block production, back-pressure forces the node into a syncing phase. During syncing, aggregated proofs are still produced and posted to DA, but the sequencer never picks them up for onchain verification. The proofs sit on DA indefinitely.                                                                                                                                                                                                                               

### Background

  The runner has an internal buffer that holds pending STF infos (witnesses needed for proving). This buffer acts as a back-pressure mechanism between block production and proving:                                                                          

  - A temporarily slow prover is fine — the buffer absorbs the spike and drains once proving catches up.                                                                                                                                                      
  - A persistently slow prover keeps the buffer full. Back-pressure then throttles DA-block ingestion to give the prover room to keep producing proofs. That throttling manifests as the node entering the syncing phase (synced_da_height < target_da_height for an extended period).

 While the node is syncing under back-pressure:                                                                                                                                                                                                              
  
  1. Proof production continues. Aggregated proofs keep being created and posted to DA (the log shows Sending aggregated proof to DA continuing every minute throughout the stall).                                                                           
  2. Preferred-batch production stops. No new preferred batches are produced while syncing.
                                                                                                                                                                                                                                                              
  The sequencer's proof-intake path only retrieves a ZK proof from DA once a preferred batch with a sequence number greater than the pending proof's is available. With no preferred batches being produced, that gating condition is never satisfied, so the proofs that landed on DA during sync are never pulled in for onchain processing.

bkolad:
# Description
This PR introduces the following changes.                     
1. Add `max_concurrent_proof_blobs` config field that caps the number of proof blobs in flight on the DA layer, mirroring the existing `max_concurrent_batch_blobs`.
2. Introduce `BlobSenderStatus { in_flight, max_concurrent }` with `is_busy()` / `Display`, and expose it via `ProofSender::proof_blob_sender_status` and `ProofBlobSender::proof_blob_sender_status` on both standard and preferred sequencers.             
3. Make the ZK aggregator check proof-sender status before each aggregation. When saturated, it logs and triggers the primary shutdown signal — proofs aren't being confirmed on time, so continuing only deepens the backlog.  

## Linked Issues
- Fixes #2800


crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
```diff
@@ -274,6 +279,19 @@ where
                     FutureOrShutdownOutput::Output(Some(item)) => item,
                 };
 
+            let status = self.proof_sender.proof_blob_sender_status().await?;
+            if status.is_busy() {
```
bkolad, line 283: This is the major change in this PR

crates/adapters/mock-zkvm/src/host.rs
```diff
@@ -186,6 +186,14 @@ impl MockZkvmHost {
         };
         std::thread::sleep(std::time::Duration::from_millis(ms));
     }
+
+    /// Polls `env_var` every 10ms and blocks while it is set to
+    /// `"STOP_PROVING"`. Returns immediately once the var is unset.
+    fn wait_while_stop_proving(env_var: &str) {
+        while std::env::var(env_var).as_deref() == Ok("STOP_PROVING") {
```
bkolad, line 193: Now we can block the prover from a test, which allows us to test various edge conditions. This is currently implemented only for mock-zkVM.

ross-weir, line 193: nit: `SOV_*` prefix

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -542,17 +547,18 @@ where
         )
     }
 
-    fn blob_sender_busy(&self) -> Option<usize> {
-        // Only batch blobs gate batch production. Proof blobs in flight must
-        // not block new batch creation, otherwise a saturated proof buffer
-        // would prevent the very thing that drains queued proofs (a new batch
-        // start consumes them via `proofs_for_replay`).
-        let num_current_in_flight_batches = self.nb_of_concurrent_batch_blob_submissions();
+    fn batch_blob_sender_busy(&self) -> BlobSenderStatus {
```
ross-weir, line 550: nit: maybe `batch_blob_sender_status` is a better name now

examples/demo-rollup/tests/bank/zk_rollup/max_concurrent_proof_blobs.rs
```diff
@@ -0,0 +1,56 @@
...
+        .subscribe_finalized_slots()
+        .await
+        .context("Failed to subscribe to aggregated proof")?;
+
+    for _ in 0..20 {
+        let _ = finalized_slots.next().await.unwrap()?;
+    }
+    resume_proving();
+    test_rollup
+        .wait_for_rollup_to_shutdown(Duration::from_secs(10))
```
ross-weir, line 52: unrelated thought but I wonder if we should add a way to declare the reason for a shutdown. We assume it's because of the proofs falling behind in this case but in reality it could be due to anything

ross-weir (approved):

crates/adapters/mock-zkvm/src/host.rs
bkolad: I think we are good here. The env name has the prefix.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
bkolad: Will fix in a followup

examples/demo-rollup/tests/bank/zk_rollup/max_concurrent_proof_blobs.rs
bkolad: +1 but this should be done outside of this PR

PR #2810, opened 2026-05-05
chore: upgrade schemars 1.x

citizen-stig:
# Description

This PR makes four code changes and two test snapshot updates, with no runtime or API changes.

- Public-key schemas are now implemented manually for `crates/module-system/sov-address/src/evm/public_key.rs:81`, `crates/adapters/risc0/src/crypto.rs:177`, and `crates/adapters/sp1/src/crypto.rs:169`.
 Each schema now explicitly describes the human-readable JSON form as a fixed-length hex string.
- The exported `bech32` schema macro in `crates/module-system/sov-modules-api/src/common/address.rs:81` and the `impl_hash32_type` macro now build the schema via `$crate::prelude::schemars::json_schema!` and qualify all `JsonSchema`/`SchemaGenerator`/`Schema` paths through `$crate::prelude::schemars::*`. Macro expansion no longer assumes downstream crates have `schemars` in scope.
- Two trybuild stderr fixtures were refreshed because compiler diagnostics changed after the real macro bug was fixed.
- The one place a downstream consumer could break is the Amount schema flip (integer → string), but those consumers were already broken at runtime, so this is a correction.

**Bonus:** less changes for 2024 edition upgrade: https://github.com/Sovereign-Labs/sovereign-sdk-wip/pull/3202

## Motivation

The `schemars` upgrade changed how derived schemas are produced. For these public-key wrapper types, derived `JsonSchema` no longer matched what serde actually emits in human-readable mode. That broke the schema-validity tests.
The manual impls make the schema match the real wire format instead of relying on derive behavior that changed under the upgrade.

The macro fix addresses a separate regression: the generated code compiled inside this repo, but failed in trybuild and downstream-style contexts because `schemars` paths inside the macro were unqualified. Routing them through `$crate::prelude::schemars` makes the macros hygienic.

## Trade-offs

- The manual schemas duplicate a small amount of serialization knowledge, but they are explicit and stable across schemars versions.
- The public-key schemas validate hex shape and length, not full cryptographic validity. That is intentional: it matches the JSON representation contract without widening scope.
  - The macro fix keeps the existing implementation style rather than refactoring it to a more abstract schemars builder API. That keeps the diff small, but still leaves the schema construction expressed through `schemars::json_schema!{}` internally.

The net effect is narrow: schema metadata is corrected, macro hygiene is fixed, and the failing tests pass.

---

## Testing
All existing test are passing

## Docs
No updates


crates/adapters/sp1/src/crypto.rs
```diff
@@ -9,6 +9,10 @@ use sov_rollup_interface::crypto::{PublicKeyHex, SigVerificationError};
 use sov_rollup_interface::reexports::schemars::{self, JsonSchema};
 use sov_universal_wallet::schema::OverrideSchema;
 
+/// Length of an ed25519 public key in bytes. `ed25519_consensus` does not expose
```
theodorebugnet, line 12: Out of curiosity, what was wrong with `ed25519_consensus::VerificationKey::LENGTH` that we were using? Is it not `const`?

theodorebugnet (approved):

crates/adapters/sp1/src/crypto.rs
citizen-stig: Good point. Previously it was just 32 number everywhere. I will update that.

theodorebugnet: I think having a named constant is fine, I'm just curious about creating a new one rather than importing `LENGTH`. If we can't use the imported one then IMO it's fine to leave the new constant (up to you though)

citizen-stig: Oh, I've misread the original request. I've focused on 

The derived jsonschema has changed in schemars and now does not match exactly json as it was before. So why it is implemented manually now.

PR #2813, opened 2026-05-05
Multisig solana v0 improvement

theodorebugnet:
# Description

Adds a `"version": 0` field to the V0 solana-offchain JSON. This brings it in line with the normal UnsignedTransaction format which includes an enum discriminator byte for both formats, and makes it less ambiguous and more secure for the future (each version is explicitly marked, no assumptions are needed when the "version" field is missing).

Background: `"version: 1"` was added as part of adding V1 support to the solana authenticator, but V0 was left untouched because the V1 support addition was non-breaking. Now V0 is being brough inline with V1 and with the standard authenticator on this point as part of the multisig hard fork upgrade.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/sov-solana-offchain-auth/tests/integration/main.rs
+++ b/crates/module-system/sov-solana-offchain-auth/tests/integration/main.rs
@@ -249,6 +249,7 @@ fn create_transfer_tx_json_with_address_override(
         details: unsigned_tx.details,
         chain_name: config_value!("CHAIN_NAME").to_string().try_into().unwrap(),
         address_override,
+        version: 0,
     };
 
     serde_json::to_string(&solana_unsigned_tx).unwrap()
@@ -354,7 +355,7 @@ async fn test_submit_ledger_signed_transaction() {
     // updated.)
     assert_eq!(
         transfer_json_tx,
-        r#"{"runtime_call":{"bank":{"transfer":{"to":"4zdwHNaEa5npHtRtaZ3RL1m6rptuQZ6RBLHG6cAyVHjL","coins":{"amount":"5000","token_id":"token_1nyl0e0yweragfsatygt24zmd8jrr2vqtvdfptzjhxkguz2xxx3vs0y07u7"}}}},"uniqueness":{"generation":0},"details":{"max_priority_fee_bips":0,"max_fee":"100000000000","gas_limit":[1000000000,1000000000],"chain_id":4321},"chain_name":"TestChain"}"#,
+        r#"{"runtime_call":{"bank":{"transfer":{"to":"4zdwHNaEa5npHtRtaZ3RL1m6rptuQZ6RBLHG6cAyVHjL","coins":{"amount":"5000","token_id":"token_1nyl0e0yweragfsatygt24zmd8jrr2vqtvdfptzjhxkguz2xxx3vs0y07u7"}}}},"uniqueness":{"generation":0},"details":{"max_priority_fee_bips":0,"max_fee":"100000000000","gas_limit":[1000000000,1000000000],"chain_id":4321},"chain_name":"TestChain","version":0}"#,
         "JSON changed - re-sign on Ledger and update the hardcoded signature"
     );
     let encoded_tx = transfer_json_tx.as_bytes().to_vec();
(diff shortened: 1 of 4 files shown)


crates/module-system/sov-solana-offchain-auth/tests/integration/main.rs
```diff
@@ -371,6 +372,17 @@ async fn test_submit_ledger_signed_transaction() {
 
     let raw_tx_bytes = borsh::to_vec(&message).unwrap();
 
+    {
+        let request = AcceptTx {
+            body: sov_sequencer::rest_api::Base64Blob {
+                blob: raw_tx_bytes.clone(),
+            },
+        };
+        println!(
```
citizen-stig, line 401: why do we need println?

crates/module-system/sov-solana-offchain-auth/tests/integration/main.rs
```diff
@@ -405,6 +417,7 @@ async fn test_submit_raw_signed_message_transaction() {
     let tx_str = create_transfer_tx_json(Amount(10_000), RECIPIENT_ADDRESS);
     let encoded_tx = tx_str.as_bytes().to_vec();
     let signer = admin.private_key();
+    println!("SINGLE SIG SIGNER KEY: {}", hex::encode(signer));
```
citizen-stig, line 440: Same about println

crates/module-system/sov-solana-offchain-auth/tests/integration/main.rs
```diff
@@ -416,6 +429,17 @@ async fn test_submit_raw_signed_message_transaction() {
     };
     let raw_tx_bytes = borsh::to_vec(&message).unwrap();
 
+    {
+        let request = AcceptTx {
+            body: sov_sequencer::rest_api::Base64Blob {
+                blob: raw_tx_bytes.clone(),
+            },
+        };
+        println!(
```
citizen-stig, line 458: same about println

crates/module-system/sov-solana-offchain-auth/tests/integration/main.rs
theodorebugnet: These are used to copy the artifacts to the typescript tests, for the byte-compatibility tests. It doesn't normally get printed when the test passes, and these aren't flaky so they will almost never show up in CI, but this way when a change breaks the existing representation you can just manually run them with `--no-capture` and paste the new values into the typescript test.

citizen-stig: Ok, I see. 

So if test ts test fails, you need to run this test with `--nocapture` to get new data?

I had similiar issue with Ethereum rollup. More robust solution is to write this data to a json file. So rust test just always writes to a file. File is commited and updated accordingly. 

Typescript then reads the same file to use value "as expected". So if rust is changed, file will be changed and ts tests will start failing automatically.

typescript/packages/web3/src/rollup/solana-signable-rollup.ts
```diff
@@ -40,14 +41,13 @@ export type SolanaOffchainUnsignedTransaction<RuntimeCall> = Omit<
...
   RuntimeCall,
   MultisigId = unknown,
-> = Omit<SolanaOffchainUnsignedTransaction<RuntimeCall>, "address_override"> & {
+> = Omit<SolanaOffchainUnsignedTransaction<RuntimeCall>, "version"> & {
   multisig_id: MultisigId;
   /**
    * Signer-declared address override.
    * See `AuthorizationData::address_override` (Rust) for routing semantics.
    */
-  address_override?: string;
```
citizen-stig, line 49: Just to clarify, why we don't support address override here? Also comment needs to be updated

theodorebugnet: The address_override comes from the V0 type. The old diff did `Omit<V0, "address_override"> & address_override` which doesn't seem necessary.

Thanks for catching comment! Updating

crates/module-system/sov-solana-offchain-auth/tests/integration/main.rs
theodorebugnet: That's a good idea. I'd like to do this as a follow-up, I've made an issue: #2929 
Since changes to transaction encoding are breaking changes, we probably won't need to do this until the next hard fork. So it'll be a very good improvement but probably lower priority for now, unless you think it's worth eliminating these `println!`s in the short term

citizen-stig (approved):

PR #2816, opened 2026-05-06
Fix universal wallet case conversion difference

ross-weir:
# Description

Fixes a mismatch in case conversion between `convert_case` and `serde`.

We're now using `serde`s "private" crate that performs case conversion, the main trade-off being that it's not a stable semantically versioned crate so might change, I think this is fine given the API surface we need is small and it will ensure we always match `serde`s behaviour

CHAIN HASH BREAKING CHANGE

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


citizen-stig (approved):

theodorebugnet (approved):
Nice!

PR #2830, opened 2026-05-07
Add zk-proving integration test module and consolidate external mock-DA helper.                                                                                                                                                                        

bkolad:
# Description

This PR introduces the following changes:
1. Moves `max_concurrent_proof` into its own test module. In the next PRs, we will add more tests in the same area, so it makes sense to give them a dedicated module.
2. Switches `max_concurrent_proof` to use the external mock DA and refactors the shared external mock DA functionality. Future tests will cover proof generation behavior after restart, resync from genesis, and similar scenarios. For that, we need an external DA that can progress independently of the rollup state

- [x] ~I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. If my PR removes a feature or changes its behavior, I provide help for users on how to migrate to the new behavior.~
- [x] I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. (Are all new dependencies necessary? Is any module dependency leaked into the full-node (hint: it shouldn't)?)

## Linked Issues
- Related to #2551

PR #2832, opened 2026-05-07
Store the latest successfully verified aggregated proof in `sov-prover-incentives`.

bkolad:
# Description

Depends on #2830

This PR updates aggregated proof handling so we propagate an `origin_state_root` through recursive aggregation instead of passing a separate `genesis_state_root` through the prover pipeline.

  It also stores the latest successfully verified aggregated proof in `sov-prover-incentives`, which lets tests and callers inspect verified aggregation outputs via module state.

  ## Changes
  - rename aggregated proof public data from `genesis_state_root` to `origin_state_root`
  - derive and propagate the origin root inside aggregation instead of threading it through zk manager / prover service APIs
  - persist the latest successfully verified aggregated proof in `sov-prover-incentives`
  - validate slashing expectations against `origin_state_root`
  - add a helper for querying `latest-proof-succesfully-verified` from the demo rollup
  - strengthen zk proving coverage by asserting that a verified aggregated proof is produced after proving resumes

## Linked Issues
- Related to #2551


crates/adapters/mock-zkvm/src/host.rs
```diff
@@ -265,16 +265,26 @@ impl OuterZkvmHost for MockZkvmHost {
...
+                "Aggregated proof continuity violated: origin_state_root differs from previous aggregation",
             );
             assert_eq!(
                 public_data.initial_state_root, prev_final_state_root,
                 "Aggregated proof continuity violated: new initial_state_root does not match previous final_state_root",
             );
+
+            public_data
         } else {
+            let origin_state_root = block_proofs_data[0].st.initial_state_root.clone();
```
bkolad, line 278: This now mirrors how the real aggregation circuit derives `origin_state_root`: carry it forward from the previous aggregation when one exists, otherwise take the initial state root of the first inner proof (the chain's genesis).

crates/rollup-interface/src/state_machine/zk/aggregated_proof/mod.rs
```diff
@@ -87,14 +86,16 @@ impl core::fmt::Display for CodeCommitmentHash {
...
     /// Initial rollup slot.
     pub initial_slot_number: SlotNumber,
     /// Final rollup slot.
     pub final_slot_number: SlotNumber,
-    /// The genesis state root of the aggregated proof.
-    pub genesis_state_root: Root,
+    /// The origin state root of the aggregated proof: the initial state root of the
+    /// first inner proof in this aggregation's chain. When the chain extends unbroken
+    /// from the rollup's genesis, this equals the rollup's genesis state root.
+    pub origin_state_root: Root,
```
bkolad, line 98: Proof aggregation currently does not allow gaps: each aggregated proof must cover a continuous chain of state transitions starting from genesis. A follow-up PR will relax this so aggregated proofs may have gaps between them, while each individual aggregated proof still covers a continuous range internally. The field has now been renamed in preparation for that change.

crates/module-system/module-implementations/sov-prover-incentives/src/lib.rs
```diff
@@ -47,6 +48,12 @@ pub struct ProverIncentives<S: Spec> {
     #[state]
     pub last_claimed_reward: StateValue<SlotNumber>,
 
+    /// Public data from the most recently verified aggregated proof.
+    #[state]
+    #[allow(clippy::type_complexity)]
+    pub latest_proof_succesfully_verified:
```
bkolad, line 54: This is the moos importnat part in this PR

ross-weir (approved):

PR #2838, opened 2026-05-09
Add Sequencer::pending_tx_count() for operator metrics

### LINKED ISSUE #2827: Expose mempool depth (e.g., Sequencer::pending_tx_count) for operator metrics
## Context

Wiring Prometheus observability for Ligate Chain (built on SDK pin `f89964c66bcfb6a31712e43278378b54c33d3779`). Mempool depth (pending tx count) is a baseline operator metric every chain wants on its dashboard.

Today: `StdSequencer.inner.mempool` is private, and `SequencerCreationReceipt` only hands the chain `Arc<dyn ProofBlobSender>`. Downstream chains can't read pending tx count without monkey-patching the SDK.

## Proposed shape

A `Sequencer` trait method:

```rust
pub trait Sequencer<S: Spec> {
    fn pending_tx_count(&self) -> usize { 0 }
    // existing methods
}
```

Default returns 0 (no observability) so non-`StdSequencer` impls aren't forced to track mempool depth. `StdSequencer` overrides to expose its actual queue length.

## Out of scope

- Per-account or per-fee-bucket mempool stats (deferred)
- Mempool gossip metrics (separate concern)

## Willingness to PR

Happy to write the PR. Roughly 50 LOC plus a test that pins the count under synthetic load.

If the shape is wrong (e.g., you'd prefer it on the receipt instead of a trait method), let me know before I file the PR.

## Related

- Companion issue (filed concurrently): DA failure categorization in `BlobExecutionStatus`
- Downstream tracking: https://github.com/ligate-io/ligate-chain/issues/164 (Phase 2 metrics for Ligate Chain)

proofmancer:
# Description

Adds `Sequencer::pending_tx_count(&self) -> usize` as a default-`0` trait method, overrides it on `StdSequencer` to return the actual mempool length. Wiring up Prometheus dashboards for our chain and `mempool_depth` was one of the baseline gauges we wanted; chains using a different `Sequencer` impl don't have to track anything.

Originally proposed sync in #2827. Switched to `async fn` because `StdSequencer.inner` lives behind a `tokio::sync::Mutex`. Sync would need either `try_lock` (racy under load) or a parallel `AtomicUsize` (more invasive). Open to either if you'd prefer.

- [x] I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. *(Non-breaking; will add a CHANGELOG entry once the PR number is assigned.)*
- [x] I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. *(No `Cargo.toml` changes.)*

## Linked Issues
- Fixes #2827

## Testing
One new test in `standard_sequencer.rs`: `pending_tx_count_tracks_mempool_through_accept_and_drain`. Counts 0 at start, 1 after one `accept_tx`, 2 after two, back to 0 after `produce_and_submit_batch`. Passes locally.

Verified with `cargo fmt --check`, `cargo check -p sov-sequencer`, `cargo clippy -p sov-sequencer --all-targets --all-features -- -A clippy::too_many_arguments -D warnings`, and the new test under `cargo test -p sov-sequencer --test integration`.

One thing I ran into while running `make lint`, flagging in case it's news: zepter 1.88's auto-fix tried to add bare `"serde"` to the `serde` feature in `crates/universal-wallet/schema/Cargo.toml`. That creates a `cyclic feature dependency: feature serde depends on itself` because `serde` is both a feature name and an optional dep in that crate. The right syntax there is `"dep:serde"`. I left this alone in the PR since it's outside scope, but worth knowing about. Happy to file a separate PR for it if you want.

## Docs
Two-line doc comment on the trait method matching the existing `Sequencer` method doc style. No other public-API doc changes.


proofmancer:

preston-evans98 (approved):

PR #2839, opened 2026-05-09
Add typed BlobSubmissionError variant to BlobSubmissionStatus

### LINKED ISSUE #2828: Surface BlobSubmissionError taxonomy in BlobExecutionStatus (or via on_failed_blob hook)
## Context

Wiring Prometheus observability for Ligate Chain. `BlobSenderHooks` already gives us submission latency cleanly, which is great. The next step is a `{reason}` failures counter so dashboards can distinguish "namespace too large" from "DA endpoint unreachable" from "TIA balance too low" etc.

Today: the SDK's internal classification isn't exposed. From a hooks consumer all we observe is "something retried", with no error type, no reason label, no stable shape we can put on a Prometheus counter.

### Option A: typed error in BlobExecutionStatus (preferred)

```rust
pub enum BlobExecutionStatus {
    Pending,
    Submitted,
    Confirmed,
    Failed { error: BlobSubmissionError },
}

pub enum BlobSubmissionError {
    NamespaceTooLarge,
    EndpointUnreachable,
    InsufficientFunds,
    InternalRpcError,
    Other(String),
}
```

Lets observability code match on the variant for a stable label.

### Option B: new BlobSenderHooks method

```rust
pub trait BlobSenderHooks {
    fn on_blob_submitted(&self, ...) {}
    fn on_blob_confirmed(&self, ...) {}
    fn on_failed_blob(&self, error: &BlobSubmissionError) {}
}
```

Same `BlobSubmissionError` enum, different exposure surface.

I'd lean A because the status field already exists; adding a typed error is a natural extension. B requires new hooks plumbing.

## Willingness to PR

Happy to write either. Roughly 50 LOC plus tests. A pointer on which shape you'd accept saves a wasted PR.

## Related

- Companion issue (filed concurrently): mempool depth exposure
- Downstream tracking: https://github.com/ligate-io/ligate-chain/issues/164 (Phase 2 metrics for Ligate Chain)

proofmancer:
# Description

Adds a typed `Failed` variant to `BlobSubmissionStatus` paired with a new `BlobSubmissionError` enum, so observers subscribed to the blob-sender broadcast can label retry / shutdown counts by cause. Before this change, internal failures were either silently retried or led to a logged-then-shutdown with no observable signal partners could put on a Prometheus counter.

```rust
pub enum BlobSubmissionError {
    Submission(String),
    PublishTimeout { retries_attempted: u32 },
    Reorg,
    FinalityCheck(String),
    MaxRetriesExhausted { retries_attempted: u32 },
}

pub enum BlobSubmissionStatus<Da: DaSpec> {
    MustSubmit,
    Published { ... },
    Processed { ... },
    Finalized { ... },
    Failed { error: BlobSubmissionError, will_retry: bool },  // NEW
}
```

5 internal failure sites in `BlobSender::manage_blob_submission_inside_task` now emit `Failed` through the existing `send_notification` channel before the existing retry-or-shutdown action runs. No new transport or hook trait, just typed plumbing through what the broadcast already carries.

Picked Option A (typed enum on `BlobSubmissionStatus`) over Option B (`on_failed_blob` hook) per the reply on #2828 ("Happy to enumerate a status, PR welcome").

- [x] I have updated `CHANGELOG.md` with a new entry if my PR makes any breaking changes or fixes a bug. *(Adds a new enum variant; existing match sites in this crate were updated. Will add a CHANGELOG entry once the PR number is assigned.)*
- [x] I have carefully reviewed all my `Cargo.toml` changes before opening the PRs. *(No `Cargo.toml` changes.)*

## Linked Issues
- Fixes #2828

## Testing

One new integration test in `tests/blob_sender.rs`: `blob_sender_emits_failed_status_before_max_retries_shutdown`. Drives a blob to `MAX_NB_OF_BLOB_SUBMISSION_RETRIES`, asserts a `Failed { MaxRetriesExhausted, will_retry: false }` notification fires on the broadcast before the shutdown signal. Passes locally.

Verified with `cargo fmt --check`, `cargo check --workspace` (excluding prover crates that need risc0/sp1 toolchains I don't have locally), `cargo clippy -p sov-blob-sender --all-targets --all-features -- -A clippy::too_many_arguments -D warnings`, and the new test under `cargo test -p sov-blob-sender --test blob_sender`.

Two pre-existing issues to flag (neither caused by this PR):

1. The blob_sender test suite has a pre-existing `tracing_subscriber` race. Two tests (`blob_sender_shutdown_task` and `blob_sender_exit_if_blob_not_processed`) both call `subscriber.init()` which sets a global; they collide when run in parallel. I confirmed this fails the same way without my new test added (5 passed, 2 failed either way). Probably a `set_default()` switch in a separate PR.

2. Same zepter 1.88 cyclic-feature fix in `crates/universal-wallet/schema/Cargo.toml` that we encountered on #2838. Reverted in this PR for the same reason (zepter adds bare `"serde"` to the `serde` feature, but `serde` is also an optional dep in that crate, which creates a self-cycle). The right fix is `"dep:serde"`.

## Docs

Doc comments on `BlobSubmissionError`, the `Failed` variant, and the failure-site code paths match the surrounding terse style. No public-API docs change beyond the new enum and the new variant on the existing one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2026-05-09
+- #2839 (Minor breaking change, code) Adds a typed `Failed { error: BlobSubmissionError, will_retry: bool }` variant to `BlobSubmissionStatus`, plus a new `BlobSubmissionError` enum (`Submission`, `PublishTimeout`, `Reorg`, `FinalityCheck`, `MaxRetriesExhausted`). Internal blob-sender failure sites now emit `Failed` on the broadcast channel before retrying or shutting down, so observers can label retry / shutdown counts by cause. Downstream consumers that match on `BlobSubmissionStatus` exhaustively need to add a `Failed` arm.
+
 # 2026-05-05
 - #2808 **Breaking config change**: Adds a required `sequencer.max_concurrent_proof_blobs` field to rollup TOML configs, capping the number of proof blobs in flight on the DA layer. When the cap is reached, the ZK aggregator triggers a rollup shutdown.
 # 2026-04-20
--- a/crates/full-node/sov-blob-sender/src/lib.rs
+++ b/crates/full-node/sov-blob-sender/src/lib.rs
@@ -645,6 +645,14 @@ impl<Da: DaService, FM: FinalizationManager> TaskState<Da, FM> {
                                 ?blob_status,
                                 "BlobSender: unable to send blob. Shutting down."
                             );
+                            self.send_notification(BlobExecutionStatus {
+                                blob_submission_status: BlobSubmissionStatus::Failed {
+                                    error: BlobSubmissionError::Submission(format!("{err:?}")),
+                                    will_retry: false,
+                                },
+                                blob_selector_status: None,
+                            })
+                            .await;
                             let _ = self.shutdown_sender.send(());
                             return;
                         }
@@ -686,6 +694,16 @@ impl<Da: DaService, FM: FinalizationManager> TaskState<Da, FM> {
                                     %blob_hash,
                                     "Shutting down the rollup. Blob processing wasn't completed on time."
                                 );
+                                self.send_notification(BlobExecutionStatus {
+                                    blob_submission_status: BlobSubmissionStatus::Failed {
+                                        error: BlobSubmissionError::MaxRetriesExhausted {
+                                            retries_attempted: nb_of_retries_attempted as u32,
+                                        },
+                                        will_retry: false,
+                                    },
+                                    blob_selector_status: None,
+                                })
+                                .await;
                                 let _ = self.shutdown_sender.send(());
                                 return;
                             }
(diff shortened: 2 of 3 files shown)


bkolad:
@sstefdev Thanks for the PR. I have just noticed it and the corresponding issue — sorry I didn't comment at an earlier stage. Here is my general feedback:
  
  If the main goal is to have more metrics, we already have the `sov_metrics` crate. It uses `InfluxDB / Telegraf` , and the simplest path would be to just add another `BlobSubmissionFailed` metric there. You'd also get it alongside all the other rollup metrics that way.

  If you specifically need the notification rather than a metric, then I'd suggest the following (InfluxDB is not an option) :

  1. Keep BlobSubmissionStatus::Failed as you added it.
  2. Instead of:
  
  ```
  self.send_notification(BlobExecutionStatus {
      blob_submission_status: BlobSubmissionStatus::Failed {
          error: BlobSubmissionError::FinalityCheck(format!("{e:?}")),
          will_retry: false,
      },
      blob_selector_status: None,
  })
  .await;
  // If we can't check the finality status, we shut down.
  let _ = self.shutdown_sender.send(());
  ```

  we should (notice that shutdown signal is removed):

```
  blob_status = BlobExecutionStatus {
      blob_submission_status: BlobSubmissionStatus::Failed {
          error: BlobSubmissionError::FinalityCheck(format!("{e:?}")),
          will_retry: false,
      },
      blob_selector_status: None,
  };
  break;
  ````

  And then handle everything centrally in the outer match:

  ```
  BlobSubmissionStatus::Failed { will_retry, .. } => {
      self.send_notification(blob_status.clone()).await;
      if will_retry {
          blob_status = BlobExecutionStatus {
              blob_submission_status: BlobSubmissionStatus::MustSubmit,
              blob_selector_status: None,
          };
      } else {
          let _ = self.shutdown_sender.send(());
          return;
      }
  }
```

  This way `BlobSubmissionStatus::Failed` is a normal case in the state machine, and all the notification + shutdown logic lives in one place. 


 Let me know if `sov_metrics` works for your use case, or if you actually need the notification-based approach. If it's the second one, lets implement the above solution and I'll do a more detailed pass on the rest of the PR.

  On the `tracing_subscriber` issue — we run our tests with `cargo nextest` instead of cargo test, which executes each test in its own process and sidesteps the global-subscriber race entirely.

PR #2841, opened 2026-05-11
Remove `sea-orm` from MockDA and upgrade `strum`

citizen-stig:
# Description

MockDA has been stabilized long time ago and with modern LLMs it is easier to maintain 2 set of queries for postgresql and sqlite than big dependency to `sea-orm`

## Testing

## Docs
No updates


crates/adapters/mock-da/src/storable/db/sqlite.rs
```diff
@@ -0,0 +1,348 @@
...
+    .await?;
+    sqlx::query("CREATE INDEX IF NOT EXISTS idx_blobs_block_height ON blobs (block_height)")
+        .execute(pool)
+        .await?;
+    Ok(())
+}
+
+pub(super) async fn query_last_saved_block<'e, E>(executor: E) -> anyhow::Result<MockBlockHeader>
+where
+    E: Executor<'e, Database = Sqlite>,
```
bkolad, line 43: would it be possible to be generic over the Database and share the same sql for sqlite and postgres?

bkolad (approved):

crates/adapters/mock-da/src/storable/db/sqlite.rs
citizen-stig: I've explored that, unfortunately no. Claude tried to explore that.
Params interpolation relies on types, so the only approach would be to use `Any` type which will reduce overall robustness.

Another way would be macros, but this is also more complicated than I think is worth it for this task.

PR #2844, opened 2026-05-11
Collapsing test jobs into 1

citizen-stig:
# Description

Since default-features only tests have been removed in https://github.com/Sovereign-Labs/sovereign-sdk/pull/2847 We are going to cut CI resource usage by always running all features tests

Code coverage that will be missing from default-feautres, based on static analysis from Codex:

 - crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs:167
  Default-features covers sov-db without migration-script: asserts expected_latest.is_none() and calls flat_state.commit(...).
  --all-features covers the migration-script branch instead, which may call commit_at_latest_checked(...).
- crates/full-node/sov-db/src/schema/tables.rs:109
  Default-features covers sov-db table structs as pub(crate).
  --all-features covers migration-script, where those structs are pub.
- crates/adapters/mock-da/src/storable/db/mod.rs:132
  Default-features covers sov-mock-da without postgres: postgres URLs error with “requires the postgres feature.”
  --all-features compiles postgres URL support instead.
- crates/adapters/mock-da/src/storable/layer.rs:1291
  Default-features has the postgres test ignored when postgres is off.
  --all-features makes that test runnable.
- crates/module-system/sov-modules-api/src/state/traits.rs:328
  Default-features covers expensive-observability off: maybe_trace_span! is a no-op wrapper.
  --all-features creates trace spans.
- crates/module-system/sov-modules-api/src/state/traits.rs:755
  Default-features covers delete-state metric behavior returning None.
  --all-features records size metrics under expensive observability.
- crates/module-system/sov-modules-api/src/module/gas_spec.rs:19
  Default-features constructs gas constants without .with_name(...).
  --all-features constructs named gas constants.
- crates/module-system/sov-modules-macros/src/manifest.rs:186
  Default-features generated gas config uses unnamed gas values.
  --all-features generated gas config uses .with_name(...).

**Which I believe is acceptable trade-off**

---

## Testing
No updates

## Docs
No updates


theodorebugnet (approved):
LGTM. A couple of the default-features tests look potentially useful, though, in particular `migration-scripts` and `expensive-observability` verifying no-ops when they're disabled. I wonder if we can extract a tiny test suite with just these tests and run it with default-features?

Downside would be we would still need to compile most of the SDK even just for a few tests, so I'm not sure of the tradeoff here.

citizen-stig:
> migration-scripts

I think this feature is going away in sov-migration script

citizen-stig:
> LGTM. A couple of the default-features tests look potentially useful, though, in particular `migration-scripts` and `expensive-observability` verifying no-ops when they're disabled. I wonder if we can extract a tiny test suite with just these tests and run it with default-features?
> 
> Downside would be we would still need to compile most of the SDK even just for a few tests, so I'm not sure of the tradeoff here.

`expensive-observability` is a fair comment. But we pay 15-25 minutes of compiling and running full CI suite on each commit in each PR just to double check that =)

Issue #2846, opened 2026-05-11
[TRACKING] ZK gas calibration


ross-weir:
## Description

ZK gas values are currently dummy values in `constants.toml`. We want to determine real values based on SP1 gas usage.

## Checklist

- [x] Create a microbench harness that can be used to measure all the different gas-consuming primitives (hashing, signature verification, (de)serialization, etc) #2849
- [x] Implement microbenches for primitives to determine costs
  - [x] Hashing
    - [x] sha256 #2849 
    - [ ] `keccak`
  - [x] Signature verficiation
    - [x] `ed25519` #2857 
  - [x] Serialization
    - [x] Borsh #2890  (TBD do we need another constant to capture vec allocations?)
- [ ] Investigate if any current consants can be removed or if there's any missing
- [ ] Once we’re confident the new method produces satisfactory numbers, we can remove the old machinery

## Open Questions

- [ ] `process_tx_pre_exec_checks_gas_per_tx_byte` is unused - remove?

PR #2849, opened 2026-05-11
SP1 microbenches for ZK gas calibration 

ross-weir:
# Description

Adds the microbench harness for running guests containing ZK primitives we're interested in (hashing, sig verification, etc) as well as an initial bench for hashing sha256.

We run the guest once per input size with `n` iterations. We want many different sized inputs so we can produce (size, prover cost) data points which we can fit with OLS. This gives us the `bias` and `per_byte` values in the cost equation `cost(size) = bias + size * per_byte`. It also produces `R` which allows us to determine the quality of the fit, which ensures a linear model is the right shape for this data. I think linear should fit all our primitives accept maybe state reads/writes.

We need to use a `bias` value otherwise we would greatly undercharge for small input operations (like small hashing) which could lead to ddosing.

We're using prover gas instead of cycle count because it's the most accurate way to measure prover costs: https://docs.succinct.xyz/docs/sp1/optimizing-programs/prover-gas

Output:

```
=== raw measurements ===
 bytes   iters  prover gas (total)    gas/iter    total cycles   region cycles     region/iter
     0    1000             1240334     1240.33          834094          830297          830.30
     1    1000             1238600     1238.60          831467          827297          827.30
    32    1000             1240021     1240.02          834684          830297          830.30
    64    1000             2189071     2189.07         1340908         1336297         1336.30
   128    1000             2920584     2920.58         1560356         1555297         1555.30
   256    1000             4383619     4383.62         1999252         1993297         1993.30
   512    1000             7309673     7309.67         2877044         2869297         2869.30
  1024    1000            13161777    13161.78         4632628         4621297         4621.30
  4096    1000            48274483    48274.48        15166132        15133297        15133.30
 16384    1000           188725252   188725.25        57300148        57181297        57181.30
 65536    1000           750783261   750783.26       225836212       225373297       225373.30

=== linear fit (prover gas per call) ===
Model: gas_per_call = bias + per_byte * input_size
  bias         = 1339.75 prover gas / call
  per_byte     = 11.4358 prover gas / byte
  R²           = 0.999999
  max residual = 465.67 prover gas

=== suggested constants.toml values (raw SP1 prover gas, 1:1) ===
  GAS_TO_CHARGE_HASH_UPDATE[1]          ≈ 1340
  GAS_TO_CHARGE_PER_BYTE_HASH_UPDATE[1] ≈ 11
```

## Linked Issues
- Related to #2846

## Testing

## Docs


bkolad (approved):
great

Issue #2772, opened 2026-04-22
`sov-accounts` refactoring and extending


citizen-stig:
## Description

Today `sov-accounts` is a one-way `credential_id → address` map (`StateMap<CredentialId, Account<S>>` where `Account<S> = { addr: S::Address }`).

This limits three things:

1. **Multisig key rotation.** The default multisig address is `credential_id.into()` where `credential_id = hash(min_signers || sorted_pubkeys)`. Rotating any key changes the credential_id and therefore the signing identity. Users can work around this today because `sov-accounts` allows registering a multisig credential against an arbitrary *existing* address via `InsertCredentialId`, but there is **no way to remove the old multisig credential** — the old keys remain authorized forever.
2. **Multi-address credentials.** The legacy map forces a 1:1 credential→address binding (`exit_if_credential_exists`). One signing key cannot spend from two logical accounts.
3. **Unknown / orphan addresses.** There is no way to create an address with no associated private key (DAO treasuries, contract-like addresses).

The current one-shot registration model is looser than the invariants we want going forward; move to an explicit many-to-many authorization relation and tighten the first-registration semantics. 

## Multi-PR plan

All PRs are based on top of [`theodore/multisig-upgrade`](https://github.com/Sovereign-Labs/sovereign-sdk/tree/theodore/multisig-upgrade) branch, so we have both breaking changes in one set.

- [ ] PR 1 — Data model + full legacy freeze: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2770. Foundational change. Introduce the new `account_owners` map and stop writing to the legacy `accounts` map from all code paths (genesis, auto-register, `InsertCredentialId`). Legacy map stays readable as a fallback for pre-upgrade on-chain state only. Accepts a user-visible behavior change for `InsertCredentialId` and `get_account`.\
- [ ] PR 2 - Extend V1 with `target_address` + `AuthorizationData.address`: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2771 Add the signing channel. V0 untouched. No new call messages yet. **Chain hash changes.**
- [ ] PR 3 — Credential rotation call messages: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2773 : Core user-facing feature: add / remove credentials on an owned address. TBD: Swap credential too
[...]

PR #2850, opened 2026-05-11
Accounts Refactor PR 4: Adding address registration without private key

citizen-stig:
# Description

Adds the final missing piece to accounts-refactor workstream (tracking #2772) by: adding way to create addresses that have no naturally-associated private key (DAO treasuries, governance vaults, escrow recipients) and authorize a credential for them in a single atomic call. Today such an address can hold balance in sov-bank but can never be signed for, since every existing authorization path requires either the credential's canonical-address fallback or a self-mutation gated by message.address == context.sender().

Fist commit: 311b52157aa165eb8aae5ac2b37b7897f861a31d

##   What's new in sov-accounts (PR 4)

- `CallMessage::CreateSyntheticAddress { salt: [u8; 32] }` — synthetic (a.k.a. counterfactual; cf. ERC-4337, CREATE2) address creation:
    - Address derived as `sha256(domain || visible_slot_hash || sender_addr || sender_credential || salt) → CredentialId → S::Address`.
    - Auto-authorizes the caller's current credential (`account_owners[(new_addr, caller_cred)] = true`).
    - Idempotent within a slot: re-submitting the same (caller, salt) is a no-op — no duplicate error, no event re-emit, no flip of an explicit false if the caller previously revoked.
    - Supports V0 single-signer + V1 multisig; EVM/Solana credentials bail (documented limitation).
- `Event<S>` enum with `SyntheticAddressCreated { address, creator, credential };` `Module::Event` changes from `()` to `Event<S>`.
  - `sov-chain-state` becomes a `#[module]` dependency of sov-accounts (used only to read the visible slot hash). All current runtimes already wire chain_state, so no runtime wiring change is needed.

##  Breaking changes
- `sov_accounts::CallMessage` wire format shifts (new variant)
- `sov_accounts::Module::Event`: `() → Event<S>`.

## Review focus (where the logic actually lives)

- crates/module-system/module-implementations/sov-accounts/src/call.rs — create_synthetic_address handler + caller_credential_id helper.
- crates/module-system/module-implementations/sov-accounts/src/lib.rs — Event definition + chain_state module field + account_owners invariant note.
- crates/module-system/module-implementations/sov-accounts/tests/integration/main.rs — property coverage.

**Mechanical / regenerated (skim, don't review line-by-line):**

- examples/demo-rollup/demo-rollup-schema.json, .bin — universal-wallet schema regen.
- typescript/packages/__fixtures__/demo-rollup-schema.json — mirror of above.
- examples/demo-rollup/README.md, README_CELESTIA.md — chain-hash + tx-hash refresh.
- examples/demo-rollup/tests/wallet/mod.rs — detect_schema_has_breaking_change snapshot (still #[ignore]'d, see follow-up).

---

## Linked Issues
- Related to #2772

## Testing
New tests added, existing tests are passing

## Docs
Updated doc-strings


preston-evans98:
This looks reasonable to me. We probably want to add a way to do this at genesis with a known address though! I think Zeta would like their admin address to be synthetic. 

The only other thing to consider is whether we should add a way to authorize a *different* credential in the callmessage instead of relying on the signer of the current message. With the current setup, you have to fund the address corresponding to your current credential. In other words, the flow is probably something like:
- Generate a fresh multisig
- Send gas tokens to multisig address
- Use multisig to send `CreateSyntheticAddress`
- (Optional) Drain remaining gas tokens from multisig address 

If we also let you specify the registered credential ID, the flow would be
- Generate a fresh multisig
- Use any existing funded address to send `CreateSyntheticAddress(multisig_credential_id)`

crates/module-system/module-implementations/sov-accounts/src/call.rs
```diff
@@ -57,6 +66,27 @@ pub enum CallMessage<S: Spec> {
...
         new_credential: CredentialId,
     },
+
+    /// Creates a new *synthetic* address — an address whose authorization
+    /// lives purely in `account_owners` and which has no
+    /// naturally-corresponding private key — and auto-authorizes the
+    /// caller's current credential for it. This is the same construct other
+    /// ecosystems call a *counterfactual* address (cf. ERC-4337, CREATE2):
+    /// the address is deterministically derivable from public inputs and
+    /// can receive funds before any controller exists for it on-chain.
```
theodorebugnet: I think including the visible slot hash makes this impractical and very risky. Getting the sequencer to include a transaction on _exactly_ the slot you want is often possible but never guaranteed (from my experience trying to do just that in the acceptance tests). And obviously if you miss it, you only had one chance, so now any pre-sent funds are lost.
I'm not sure if this matters for our usecase. But if it did, then removing the visible slot as an input would actually make it usable. If not, I would remove this part of the documentation.

crates/module-system/module-implementations/sov-accounts/README.md
```diff
@@ -25,6 +25,23 @@ addresses and records which credentials may act for which addresses.
    There is no orphan guard on remove — revoking the last credential leaves
    the address unspendable via `account_owners`.
 
+6. It is possible to create a new *synthetic* address — an address whose
+   authorization lives purely in `account_owners` and which has no
+   naturally-corresponding private key — with
+   `CallMessage::CreateSyntheticAddress { salt }`. This is the same construct
+   other ecosystems call a *counterfactual* address (cf. ERC-4337, CREATE2):
+   the address is deterministically derivable from public inputs and can
+   receive funds before any controller exists for it on-chain. The new
```
theodorebugnet: Same comment as above w.r.t. using it to receive funds before it exists

examples/demo-rollup/tests/wallet/mod.rs
```diff
@@ -151,6 +151,9 @@ fn test_display_signed_tx() {
 #[ignore = "Ignored for rapid schema iteration, re-enable when CI timebomb goes off"]
```
theodorebugnet: We should probably unignore this... (in another PR)

theodorebugnet (approved):
LGTM

crates/module-system/module-implementations/sov-accounts/tests/integration/main.rs
```diff
@@ -620,6 +641,106 @@ fn submit_v1(tx: Version1<RT, S>) -> TransactionType<RT, S> {
     })
 }
 
+/// Mirrors `CallMessage::CreateSyntheticAddress` derivation for test setup.
+const SYNTHETIC_ADDRESS_DOMAIN: &[u8] = b"sov_accounts::synthetic_address::v1";
+
+/// Destructures the single expected `SyntheticAddressCreated` event from a
+/// receipt's event list.
+fn synthetic_address_created_event(
```
theodorebugnet, line 650: By the way, I've recently learned about the `#[track_caller]` function attribute: it makes panics and asserts print the location of the caller in the panic message. IMO this would be a good idea in test helpers like these that wrap a panic, in order to point at the actual location in the test where the panic failed, rather than requiring digging through the backtrace (if enabled) to see the callsite of the helper.

crates/module-system/module-implementations/sov-accounts/src/call.rs
citizen-stig: This is critical part of requirement.
We need to ensure that if original sender private key is compromised, it would be impossible to recreate same synthetic address. This requirement is more important that ability to prefund this account. First you create synthetic address, get it finalized and then use.

citizen-stig: Probably I need to update comment

citizen-stig: Updated documentation

crates/module-system/module-implementations/sov-accounts/README.md
citizen-stig: updated documentation

examples/demo-rollup/tests/wallet/mod.rs
citizen-stig: Unignored

crates/module-system/module-implementations/sov-accounts/tests/integration/main.rs
citizen-stig: Added!

theodorebugnet (approved):

PR #2853, opened 2026-05-12
Accounts Refactor PR 5: Cleanup, Events and other minor follow ups 

citizen-stig:
# Description
 ## Changes

- **Events**: `CredentialInserted`, `CredentialAdded`, `CredentialRemoved`,`CredentialRotated` emitted by their respective `CallMessage` handlers
- **Genesis invariant**: `init_module` rejects `AccountConfig` entries whose address parses as a valid public key. On 32-byte specs (Solana-style`Base58Address`) this enforces synthetic-only registration; on shorter specs (28-byte default, 20-byte EVM) the size check makes it a no-op.
- **REST `AuthorizationResponse`**: gains `admit_as_override` and `admit_as_default`, which mirror the on-chain admit-path predicates. The existing `authorized` field is `#[deprecated]` — it uses the canonical-fallback view and disagrees with the chain for authenticators (notably EVM) whose `default_address` is not `canonical(credential_id)`.
- **Regression test**: pins that `is_explicitly_authorized` returns `false` for `Some(false)` (an earlier `.is_some()` would have allowed it).
- **Docs**: `README` rewritten with a three-predicate table and a worked EVM divergence example; `AuthorizationData::address_override` documents the `Some(default_address) ≠ None` footgun.
- **Typo**: `oprating_mode` → `operating_mode` in `GasEnforcer` and `StandardProvenRollupCapabilities`.

## Breaking changes

- **Consensus (32-byte specs only)**: genesis configs containing on-curve addresses are rejected. 28-byte and EVM unaffected.
- **REST**: additive only — `authorized` is deprecated but still returned.
- **API**: `GasEnforcer` named parameter renamed (typo fix).

## Testing

Unit tests for genesis (incl. a `base58_spec` submodule exercising the synthetic check on Solana-style addresses); event assertions added to the existing per-call success-path tests; new tests pin admit-path / canonical divergence and full revoke semantics; the Solana off-chain auth integration test was updated to use a synthetic off-curve target so it complies with the new genesis invariant.

First commit ae7b4e5d4b040358831e00622753e27673eb8a1d

----

## Issues

Relevant to #2772

## Testing
New tests are added

## Docs
No updates


crates/module-system/module-implementations/sov-accounts/src/genesis.rs
```diff
@@ -1,11 +1,28 @@
...
+///
+/// For 32-byte address specs (e.g., Solana-style `Base58Address` with ed25519) this
+/// rejects addresses whose bytes are on the curve — addresses that correspond to a
+/// real keypair and therefore have a natural off-chain owner.
+///
+/// For address schemes shorter than the public-key size (default 28-byte `Address`,
+/// 20-byte `EthereumAddress`), `TryFrom` errors on size and this check is a no-op:
+/// those schemes lose information through truncation/hashing and cannot be checked
+/// from the address bytes alone. Closing that gap requires a different address scheme.
+fn is_synthetic_address<S: Spec>(address: &S::Address) -> bool {
```
theodorebugnet, line 21: Given that this is effectively a no-op, I think it would be really nice if this returned `Option<bool>`. That way the semantics of "not synthetic address" (`false`) and "cannot know" (`None`) would be clearly communicated. The callsites can still use `is_none_or(false)` but it would make the semantics explicit.

However this would complicate the check beyond just `.is_err()` so leaving it up to preference

crates/module-system/module-implementations/sov-accounts/src/lib.rs
```diff
@@ -61,6 +61,40 @@ pub enum Event<S: Spec> {
...
+        credential: CredentialId,
+    },
+    /// Emitted by [`CallMessage::AddCredentialToAddress`] when a credential
+    /// is authorized for an explicit address (which must be the caller).
+    CredentialAdded {
+        /// The address whose credential set was extended.
+        address: S::Address,
+        /// The newly authorized credential.
+        credential: CredentialId,
+    },
```
theodorebugnet: These two events are basically identical.
I see that one of them is used when inserting the credential for the caller, and another when adding a credential to a third-party authorized account. But I wonder, is it actually useful to have two events for this?

If it is, I would probably rename them so it's clear what the difference is. "CredentialInserted" and "CredentialAdded" mean almost the same thing so it's confusing.

Another option is maybe having an `authorizer_address: Option<S::Address>` which would be None for the "Inserted" case. Or even non-optional and just have `authorizer_address == address` then.

crates/module-system/module-implementations/sov-accounts/src/lib.rs
```diff
@@ -61,6 +61,40 @@ pub enum Event<S: Spec> {
...
+    /// Emitted by [`CallMessage::RemoveCredentialFromAddress`] when a
+    /// credential is revoked from an address.
+    CredentialRemoved {
```
theodorebugnet, line 79: By the way, thinking about the above made me think, maybe having `authorizer_address` would also be userful here (and in Rotated too, perhaps), for auditing purposes - in case it's different from the target `address`

crates/module-system/module-implementations/sov-accounts/tests/integration/main.rs
```diff
@@ -2730,3 +2853,114 @@ fn test_create_synthetic_address_rejected_when_custom_account_mappings_disabled(
...
+        let canonical_view = accounts
+            .is_authorized_for(&non_canonical_address, &credential, state)
+            .unwrap();
+
+        assert!(
+            !explicit,
+            "no explicit account_owners entry exists: is_explicitly_authorized must be false"
+        );
+        assert!(
+            default_path,
```
theodorebugnet, line 2922: I'm a bit confused here, why does creating a second account then result in `is_default_address_authorized(account_2.address(), account_1.credential_id()) == true`? I must have missed something in the test setup

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2026-05-13
+- #2853 Accounts: `AuthorizationResponse` (REST `GET /authorizations/{address}/{credential_id}`) gains `admit_as_override` and `admit_as_default` fields that mirror the on-chain admit-path predicates directly. The existing `authorized` field is deprecated — it uses the canonical-fallback semantic, which can disagree with the chain for authenticators whose `default_address` is not the credential's canonical address (notably EVM, where `default_address` is a `MultiAddress::Vm` variant while `canonical(credential_id)` is `MultiAddress::Standard`). REST consumers should read `admit_as_default` for the equivalent of the chain's admit-path. Additive change; existing callers continue to work.
```
theodorebugnet: We're gonna need to reorganise the changelog when merging into `dev`, I'd rather keep it all centralised in the `# MULTISIG UPGRADE` heading than add independent items (but we can always check the diff when doing the final merge so not a big deal)

theodorebugnet (approved):

CHANGELOG.md
citizen-stig: Addressed in 8c931301bd54f63fee5e7fbc8bdab07a13cbf75a

crates/module-system/module-implementations/sov-accounts/src/lib.rs
citizen-stig: Addressed in 8c931301bd54f63fee5e7fbc8bdab07a13cbf75a

crates/module-system/module-implementations/sov-accounts/src/lib.rs
citizen-stig: Good point!
Addressed in 8c931301bd54f63fee5e7fbc8bdab07a13cbf75a

crates/module-system/module-implementations/sov-accounts/src/genesis.rs
citizen-stig: Hm, but for code to answer this:

> "cannot know" (None) 

It needs to know how to know. Basically now it should compare length of address as bytes, but it won't be robust for the future.

I personally prefer not to invest to much into it, as for production genesis should be reviewed carefully and trusted. This code adds simple guard-rail, nothing more.

theodorebugnet: Yeah makes sense. I think it'd be useful to clearly document that it's best effort then, e.g. in the doc comment.

citizen-stig: Added!

PR #2866, opened 2026-05-14
mock-zkvm: enforce real aggregation constraints in the mock host

bkolad:
# Description

 Before this PR, `MockZkvmHost::run_proof_aggregation` was a hand-rolled reimplementation of `run_aggregation_program`: same constraints (slot/DA/state continuity), but maintained in two places. Now the mock executes the shared circuit verbatim — only the `ZkVerifier` and `ZkvmGuest` impls differ from SP1.

Changes:

1. Lifted `#[cfg(target_os = "zkvm")]` from `aggregated_proof::circuit` and `ZkVerifier::verify_with_pub_values` so the shared circuit is callable on native.

2.  The shared circuit previously assumed every aggregation chain roots at rollup genesis. Two checks encoded that assumption, both fired only when the aggregation has no predecessor outer proof:
   - Slot assertion. `verify_proof_chain` asserted `first_inner_slot == SlotNumber::ONE`.
  - Origin default. run_aggregation_program set the new aggregation's o`rigin_slot_number = SlotNumber::GENESIS`.

This PR relaxes both: the slot assertion is dropped when no predecessor exists, and o`rigin_slot_number` defaults to `first_inner_slot.prev()` instead. 
The motivation is  `start_fresh_outer_proof_on_resync`: when a node resumes with that flag, the prover service intentionally drops the on-disk anchor and begins a new chain at the current slot, which the old assertion forbade

## Linked Issues
- Related to #2551

### DIFF
--- a/crates/adapters/mock-zkvm/src/guest.rs
+++ b/crates/adapters/mock-zkvm/src/guest.rs
@@ -1,17 +1,52 @@
+use std::sync::Mutex;
+
 use serde::Serialize;
 
 use crate::MockZkVerifier;
+
 /// A mock implementing the Guest.
 #[derive(Default)]
-pub struct MockZkGuest {}
+pub struct MockZkGuest {
+    hint: Mutex<Option<Vec<u8>>>,
+    committed: Mutex<Option<Vec<u8>>>,
+}
+
+impl MockZkGuest {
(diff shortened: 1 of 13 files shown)


theodorebugnet (approved):
LGTM, though I don't have the full context of the proof flow

PR #2871, opened 2026-05-15
fix(sequencer): structured JSON errors for skipped transactions

0xtristan:
## Summary

When a transaction is skipped (e.g. due to a bad generation number or nonce), the sequencer was serializing the `TransactionReceipt` using `format!("{:?}", receipt)` — a Rust debug string stuffed into a JSON field. This made it impossible for SDK clients to deserialize the error in a structured way; they had to regex-match an opaque string.

`TransactionReceipt`, `SkippedTxContents`, and `TxProcessingError` all already derived `serde::Serialize`, so the fix was straightforward.

### Commit 1 — serialize receipt as JSON

Replaces `format!("{:?}", receipt)` with `serde_json::to_value(&receipt)` in the `UnsuccessfulTransaction` arm of `block_executor.rs`. The `Reverted` case already handled this correctly; the `Skipped`/`Ignored` cases did not.

### Commit 2 — `CheckUniquenessError` typed enum

`TxProcessingError::CheckUniquenessFailed(String)` gave clients no machine-readable way to identify the failure type or extract fields. This replaces the `String` with a proper enum:

```rust
pub enum CheckUniquenessError {
    BadGeneration { latest_generation: u64, provided_generation: u64 },
    DuplicateGeneration { generation: u64 },
    GenerationCapacityExceeded { current_generation: u64, next_valid_generation: u64 },
    BadNonce { expected_nonce: u64, provided_nonce: u64 },
    NonceTooLow { minimum_nonce: u64, provided_nonce: u64 },
    Internal(String), // state-access failures
}
```

The enum uses `#[serde(tag = "kind", rename_all = "snake_case")]` so the JSON output is fully typed:
```json
{"kind": "bad_generation", "latest_generation": 1778774165791000, "provided_generation": 1778806984782}
```

The change threads through: `generations.rs` / `nonces.rs` → `Uniqueness::check_uniqueness` → `TransactionAuthorizer` trait → `StandardProvenRollupCapabilities` → `registered.rs` / `unregistered.rs`. All test assertions are updated to match on variant fields rather than substrings.

## Test plan

- [ ] Existing uniqueness integration tests pass (`sov-uniqueness`)
- [ ] `nonce_buffer_task` tests pass
- [ ] `tx_revert_tests` pass
- [ ] Verify JSON error response shape from a live sequencer for a bad-generation tx

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/block_executor.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/block_executor.rs
@@ -95,9 +95,9 @@ impl<S: Spec> RollupBlockExecutorError<S> {
                     sov_rollup_interface::stf::TxEffect::Reverted(reverted) => {
                         reverted.reason.error_detail().unwrap_or(json_obj!({}))
                     }
-                    _ => json_obj!({
+                    _ => serde_json::to_value(&receipt).unwrap_or_else(|_| json_obj!({
                         "error": format!("{:?}", receipt),
-                    }),
+                    })),
                 };
                 ErrorObject {
                     status: StatusCode::BAD_REQUEST,
--- a/crates/full-node/sov-sequencer/src/preferred/block_executor.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/block_executor.rs
@@ -95,9 +95,11 @@ impl<S: Spec> RollupBlockExecutorError<S> {
                     sov_rollup_interface::stf::TxEffect::Reverted(reverted) => {
                         reverted.reason.error_detail().unwrap_or(json_obj!({}))
                     }
-                    _ => serde_json::to_value(&receipt).unwrap_or_else(|_| json_obj!({
-                        "error": format!("{:?}", receipt),
-                    })),
+                    _ => serde_json::to_value(&receipt).unwrap_or_else(|_| {
+                        json_obj!({
+                            "error": format!("{:?}", receipt),
+                        })
+                    }),
                 };
                 ErrorObject {
                     status: StatusCode::BAD_REQUEST,
--- a/crates/full-node/sov-sequencer/src/preferred/block_executor.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/block_executor.rs
@@ -13,9 +13,9 @@ use sov_modules_api::capabilities::{
 };
 use sov_modules_api::macros::config_value;
 use sov_modules_api::{
-    call_message_repr, Amount, BlobDataWithId, ChangeSet, DaSpec, ExecutionContext, FullyBakedTx,
-    Gas, GasSpec, HexString, KernelStateAccessor, NoOpControlFlow, RejectReason, Runtime,
-    RuntimeEventProcessor, RuntimeEventResponse, SelectedBlob, Spec, StateCheckpoint,
+    call_message_repr, Amount, BlobDataWithId, ChangeSet, DaSpec, ErrorDetail, ExecutionContext,
+    FullyBakedTx, Gas, GasSpec, HexString, KernelStateAccessor, NoOpControlFlow, RejectReason,
+    Runtime, RuntimeEventProcessor, RuntimeEventResponse, SelectedBlob, Spec, StateCheckpoint,
     TransactionReceipt, TxChangeSet, TxHash, VersionReader, VisibleSlotNumber,
 };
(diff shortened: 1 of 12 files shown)


crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -95,9 +95,10 @@ impl<S: Spec> RollupBlockExecutorError<S> {
                     sov_rollup_interface::stf::TxEffect::Reverted(reverted) => {
                         reverted.reason.error_detail().unwrap_or(json_obj!({}))
                     }
-                    _ => json_obj!({
-                        "error": format!("{:?}", receipt),
-                    }),
+                    _ => match serde_json::to_value(&receipt) {
```
ross-weir: Lets impl `ErrorDetail` on `TxProcessingError` & `CheckUniquenessError`, then here we do something similar to reverted:

```
  sov_rollup_interface::stf::TxEffect::Skipped(skipped) => {
      skipped.error.error_detail().unwrap_or(json_obj!({}))
  }
```

In the impl on `TxProcessingError` for the `CheckUniquenessFailed` variant we delegate to `inner.error_detail()` for all other variants fall back to `json_obj!({"error": format!("{:?}", receipt)})`

Then we have everything in place to gradually migrate the other variants on `TxProcessingError` to structured errors

crates/full-node/sov-sequencer/src/preferred/nonce_buffer_task.rs
```diff
@@ -982,19 +982,21 @@ fn err_invalid_nonce<S: Spec, Rt: Runtime<S>>(
...
     );
     let receipt = TransactionReceipt {
         tx_hash,
         body_to_save: None,
         events: Vec::new(),
         receipt: sov_rollup_interface::stf::TxEffect::Skipped(SkippedTxContents {
             gas_used: <S::Gas as Gas>::zero(),
-            error: TxProcessingError::CheckUniquenessFailed(error_msg),
+            error: TxProcessingError::CheckUniquenessFailed(
+                sov_modules_api::CheckUniquenessError::BadNonce {
```
ross-weir, line 1003: Would it be possible to include a `reason` field on the error here so we can supply `InvalidNonceReason`? That enum provides a lot of useful context for exactly why the nonce is bad

crates/module-system/sov-modules-api/src/tx_receipt.rs
```diff
@@ -74,6 +74,59 @@ impl<S: Spec> PartialEq for SkippedTxContents<S> {
 }
 impl<S: Spec> Eq for SkippedTxContents<S> {}
 
+/// Structured error returned when a transaction's uniqueness check fails.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize, thiserror::Error)]
+#[serde(rename_all = "snake_case", tag = "kind")]
+pub enum CheckUniquenessError {
```
ross-weir, line 99: Would it be possible to define this closer to the uniqueness code? This file contains more higher level tx receipt types

Also we should impl `ErrorDetail` for it, which I think could jsut be:

```
impl ErrorDetail for CheckUniquenessError {
    fn error_detail(&self) -> Result<ErrorContext, Box<dyn std::error::Error + Send + Sync>> {
        Ok(err_detail!(self))
    }
}
```

crates/module-system/module-implementations/sov-uniqueness/src/nonces.rs
```diff
@@ -24,13 +30,19 @@ impl<S: Spec> Uniqueness<S> {
...
-
-        anyhow::ensure!(
-            nonce <= transaction_nonce,
-            "Tx bad nonce for credential id: {credential_id}, expected at least: {nonce}, but found: {transaction_nonce}",
-        );
+    ) -> Result<(), CheckUniquenessError> {
+        let nonce = self
+            .nonces
+            .get(credential_id, state)
+            .map_err(|e| CheckUniquenessError::Internal(e.to_string()))?
```
ross-weir: We have a `CoreModuleError` type that we can use for state accessor errors, so this becomes:

```
.map_err(CoreModuleError::state_read)?
```

And we add a `CoreModuleError` variant to `CheckUniquenessError` - see bank `Error` impl for example of how it's used

PR #2872, opened 2026-05-15
 SP1: support start_fresh_outer_proof_on_resync

bkolad:
# Description

  - Implements `start_fresh_outer_proof_on_resync` for `MockSp1DemoRollup`, removing the panic stub in `create_prover_service` (todo #2551). When the flag is on we drop the persisted outer proof and start a fresh aggregation chain.
  - Extracts a shared `get_persisted_outer_proof::<Vm, …>` helper in `examples/demo-rollup/src/lib.rs`,
  - Adds `build_sp1_inner_host` in `mock_sp1_rollup.rs` that runs SP1's setup once per ELF and returns the inner `SP1Host` plus both code commitments. `

## Linked Issues
- Related to #2551

### DIFF
--- a/examples/demo-rollup/src/mock_sp1_rollup.rs
+++ b/examples/demo-rollup/src/mock_sp1_rollup.rs
@@ -11,7 +11,7 @@ use sov_mock_da::storable::StorableMockDaService;
 use sov_mock_da::MockDaSpec;
 use sov_modules_api::configurable_spec::ConfigurableSpec;
 use sov_modules_api::execution_mode::{Native, WitnessGeneration};
-use sov_modules_api::{CryptoSpec, NodeEndpoints, Spec};
+use sov_modules_api::{CodeCommitmentTrait, CryptoSpec, NodeEndpoints, Spec};
 use sov_modules_rollup_blueprint::pluggable_traits::PluggableSpec;
 use sov_modules_rollup_blueprint::proof_sender::SovApiProofSender;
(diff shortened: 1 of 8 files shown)


examples/demo-rollup/src/mock_sp1_rollup.rs
```diff
@@ -188,19 +208,6 @@ impl FullNodeBlueprint<Native> for MockSp1DemoRollup<Native> {
...
 
-        // Validate the persisted proof and extract the `final_slot_number` so
-        // the runner can rewind the STF-info stream to `final_slot + 1`.
-        let latest_proof_final_slot = previous_aggregated_proof.as_ref().map(|proof| {
-            let public_data: AggregatedProofPublicData<
-                <Self::Spec as Spec>::Address,
-                MockDaSpec,
-                <<Self::Spec as Spec>::Storage as Storage>::Root,
-            > = AggregateProofVerifier::<SP1Verifier>::new(outer_vm.code_commitment())
-                .verify(proof)
```
theodorebugnet, line 199: We no longer verify the last persisted proof - is that intentional?
Makes sense when starting fresh, but as far as I can see this was removed globally, even when `start_fresh_outer_proof_on_resync == false`.
I suppose the local DB should be trusted but just wanna make sure it's intended

theodorebugnet (approved):

examples/demo-rollup/src/mock_sp1_rollup.rs
bkolad: Yeh if the proof is invalid it will fail anyway and a later stage, but we trust the local db.

PR #2876, opened 2026-05-15
Split UnsignedTransction from TransactionSigningPayload types

theodorebugnet:
# Description

Refactor core transaction signing types to separate the consumer-facing unsigned payload from the canonical signing payload.

 * What used to be called `UnsignedTransaction{, V0, V1}` is now called `TransactionSigningPayload{, V0, V1}`. This is the type that is serialized for signing.
 * Serialization logic is centralised to this Payload, in particular the `chain_hash` is now a field on the Payload rather than manually apended.
 * A new type, called `UnsignedTransaction`, is added to encapsulate the actual user-defined parts of the transaction.

This removes the old ambiguity where “unsigned transaction” sometimes meant “user payload” and sometimes meant “exact signed preimage”.

### Reviewer Guide

The core logic is in `crates/module-system/sov-modules-api/src/transaction/`, specifically `unsigned.rs` and `signing_payload/*`. The serialization logic also changed accordingly in `crates/module-system/sov-modules-api/src/runtime/capabilities/authentication.rs`.

The majority of the rest of the touched files are switching the usage from `UnsignedTransactionV0` to `UnsignedTransaction` and the signatures of the serialization/signing method on these.

In `sov-solana-offchain-auth`, the rename is mostly mechanical. The outer `Message` types have also been renamed to `Envelope`, because the solana spec defines "message" as the actual (signed) offchain message, which in this context is the Payload.

### Typescript

The Schema root changed from `UnsignedTransaction` to `TransactionSigningPayload`, since the UnsignedTransaction is never actually serialised.

In Typescript, adds `TypeBuilder.transactionSigningPayload` and adjust the signing flow accordingly: the user provides fields for the UnsignedTransaction, the `typeBuilder` builds a `transactionSigningPayload` which is then serialized.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2876" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


crates/module-system/sov-modules-api/src/transaction/signing_payload/v0.rs
```diff
@@ -0,0 +1,59 @@
...
+            details: self.details.clone(),
+            address_override: self.address_override,
+            chain_hash: self.chain_hash,
+        }
+    }
+}
+
+impl<R: TransactionCallable, S: Spec> PartialEq for TransactionSigningPayloadV0<R, S> {
+    fn eq(&self, other: &Self) -> bool {
+        self.runtime_call == other.runtime_call
```
citizen-stig: nit: I believe reordering of elements might give some perf gains. 

I am not sure if this is a real case, but runtime_call is a more complicated structure and will do more memory reads than, for instance chain_hash.

And in practice chain_hash difference is the most stark one, followed by uniqueness, probably, which are simple fields.

crates/module-system/sov-modules-api/src/transaction/signing_payload/v1.rs
```diff
@@ -8,53 +8,60 @@ use crate::{
...
     /// multisig envelope.
-    pub credential_address: S::Address,
+    pub(crate) credential_address: S::Address,
     /// Signer-declared address override.
     /// See [`crate::capabilities::AuthorizationData::address_override`] for routing semantics.
     #[serde(default)]
-    pub address_override: Option<S::Address>,
+    pub(crate) address_override: Option<S::Address>,
+    /// The chain hash binding this signature to the rollup schema and metadata.
+    #[sov_wallet(hidden)]
```
citizen-stig, line 39: Why chain_hash is hidden from wallet?

crates/module-system/sov-modules-api/src/transaction/signing_payload/v1.rs
```diff
@@ -8,53 +8,60 @@ use crate::{
...
-impl<R: TransactionCallable, S: Spec> PartialEq for UnsignedTransactionV1<R, S> {
+
+impl<R: TransactionCallable, S: Spec> PartialEq for TransactionSigningPayloadV1<R, S> {
     fn eq(&self, other: &Self) -> bool {
         self.runtime_call == other.runtime_call
             && self.uniqueness == other.uniqueness
             && self.details == other.details
             && self.credential_address == other.credential_address
             && self.address_override == other.address_override
+            && self.chain_hash == other.chain_hash
```
citizen-stig: same note about order of conditions

crates/module-system/sov-modules-api/src/transaction/types/v0.rs
```diff
@@ -48,14 +47,20 @@ pub struct Version0<R: TransactionCallable, S: Spec, C: CryptoSpecExt = <S as Sp
...
-    /// Extracts the versioned unsigned transaction data from this signed envelope.
-    pub fn as_unsigned(&self) -> UnsignedTransaction<R, S> {
-        UnsignedTransaction::V0(UnsignedTransactionV0 {
-            runtime_call: self.runtime_call.clone(),
-            uniqueness: self.uniqueness,
-            details: self.details.clone(),
-            address_override: self.address_override,
-        })
+    /// Extracts the unsigned transaction payload from this signed envelope.
+    pub fn to_unsigned_transaction(&self) -> UnsignedTransaction<R, S> {
```
citizen-stig, line 51: Can we preserve `as_` prefix, as we don't consume self?

citizen-stig: Or another option would be to consume, but this will probably blow up the size of the PR

crates/module-system/sov-modules-api/src/transaction/types/v1.rs
```diff
@@ -108,23 +107,29 @@ impl<R: TransactionCallable, S: Spec, C: CryptoSpecExt> Version1<R, S, C> {
...
+        UnsignedTransaction::new_with_details(
+            self.runtime_call.clone(),
+            self.uniqueness,
+            self.details.clone(),
+            self.address_override,
+        )
+    }
+
+    /// Serializes the V1 transaction signing payload for this signed envelope.
+    pub fn to_signing_bytes(&self, chain_hash: &[u8; 32]) -> Vec<u8> {
```
citizen-stig, line 121: I think we can reduce size of the PR by preserving old name: `serialize_for_signing`

crates/module-system/sov-modules-api/src/transaction/mod.rs
```diff
@@ -218,15 +222,31 @@ impl<R: TransactionCallable, S: Spec, C: CryptoSpecExt> Transaction<R, S, C> {
...
+        uniqueness: UniquenessData,
+        details: TxDetails<S>,
+    ) -> Self {
+        Self::V0(Version0 {
+            signature,
+            pub_key,
+            runtime_call,
+            uniqueness,
+            details,
+            address_override: None,
```
citizen-stig: is it ok to skip address_override from input params?

citizen-stig (approved):
Overall looks good, nothing blocking.

crates/module-system/sov-modules-api/src/transaction/signing_payload/v1.rs
theodorebugnet: It's an opaque hash value to the user, and its integrity is also validated by the signature so the user never needs to manually care about it.
The trust boundary here is that the user must trust the wallet to use the same schema for display as it uses for hashing. But just showing the hash doesn't help the user verify, because a malicious wallet can show any doctored display while still including the correct hash into the payload.

crates/module-system/sov-modules-api/src/transaction/types/v1.rs
theodorebugnet: I think I'd rather use the better name in the long run than reduce diff right now. Unless you think `serialize_for_signing()` is the better name to keep? Then we can keep it, I just changed it because it's idiosyncratic while `to_signing_bytes()` is more similar to typical rust naming convention for conversion/serialization functions

crates/module-system/sov-modules-api/tests/integration/transaction.rs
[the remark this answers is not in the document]
theodorebugnet: Good call, oversight from rebasing on dev

crates/module-system/sov-modules-api/src/transaction/types/v1.rs
citizen-stig: The thing with rust convention is that `to_` usually implies consuming, and `as_` usually takes reference. `serialize_for_signing` is explicit. If we want to rename, I would prefere `as_signing_bytes`. 

Not a blocker by the way, matter of taste

crates/module-system/sov-eip712-auth/src/lib.rs
```diff
@@ -270,29 +270,29 @@ fn get_eip712_hash<
...
             raw_tx_hash,
         )
     })?;
 
-    // Use the schema provider to get the schema and calculate the EIP712 signing hash
-    let schema = SP::get_schema();
-    let transaction_type_index = schema.rollup_expected_index(sov_modules_api::sov_universal_wallet::schema::RollupRoots::UnsignedTransaction)
+    // Convert the transaction to the canonical signing bytes that are transformed into
+    // EIP-712 typed data.
+    let signing_payload_bytes = tx.to_signing_bytes(&chain_hash);
```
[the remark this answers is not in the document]
theodorebugnet: The EIP712 authenticator actually already didn't support CHAIN_HASH_OVERRIDES. It's just more visible now because the chain_hash has to be included in the payload as well, but note that the SigningPayload is not what's actually signed here, the EIP712 hash is.
I made an issue: #2900

crates/module-system/sov-modules-api/src/transaction/types/v0.rs
theodorebugnet: As far as I understand, `as_` is for returning a reference, `to_` is for returning a copied/cloned value, and `into_` is for returning an owned value after consuming `self`.
So `&self -> UnsignedTransaction` is fine as `to_`, since we don't consume (so it's not `into_`) but we do internally clone so it's not just a view/ref/slice (so not `as_`). I think that's how the convention works, let me know if you disagree

crates/module-system/sov-modules-api/src/transaction/types/v1.rs
theodorebugnet: See my response on the other naming comment - when you consume I think it's `into_`, and when you don't consume _but_ clone internally then `to_` is fine. Here we do clone internally and basically construct a new object, a Vec of copied bytes (not a slice view of the original), so `to_` fits better than `as_`.

Again if you still disagree or have a different understanding of the convention let me know and I'll be happy to change

PR #2880, opened 2026-05-18
Claude command for updating readme tests

citizen-stig:
# Description

Useful when chain hash changes

## Testing
No updates

## Docs
No updates

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2880" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.claude/commands/update-readme-tests.md
+++ b/.claude/commands/update-readme-tests.md
@@ -1,84 +1,103 @@
 # Update demo-rollup README tests
 
-Refresh the `bashtestmd:compare-output` blocks in `examples/demo-rollup/README.md` and `examples/demo-rollup/README_CELESTIA.md` after rebasing/changes that alter on-chain values (chain_hash, tx_hash, supply, etc.).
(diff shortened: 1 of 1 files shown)


theodorebugnet (approved):
This can work but what I usually do is look at the failing job in CI and check the bashtestmd log. The expected outputs are always in the error messages or logs. I feel like a few tool calls to `gh` to fetch the job logs and grep for the correct things output would be a lot faster than having claude derive the chain hash locally and then compile and run the rollup for the token tx hash

citizen-stig:
> This can work but what I usually do is look at the failing job in CI and check the bashtestmd log. The expected outputs are always in the error messages or logs.

Yep, I also has been doing that. But it is slower: ci check job time + demo rollup compile. Fail on chain has. Repeat ci check job time + demo rollup compile, then update tx has.

So I doubt it is faster. Also if it is local, less pressure on CI. 

I've been using claude to get values locally and it was pretty fast, because my machine already has everything needed for pre-compilation.

PR #2884, opened 2026-05-18
Remove `jmt` and `ProverStorage` 

citizen-stig:
# Description

NOMT storage has been production ready and used for a while. 

JMT storage hasn't been used in production and has lower throughput.

In order to simplify mission critical `sov-state` crate and make it easier to understand by humans and analyze by LLMs this PR removes JMT based storage and jmt crate as a dependency

Minor issue with empty user changes has been detected in this PR and fix is included.

Key changes:

* Delete ProverStorage (JMT), ZkStorage (JMT), StateDb, NativeStorageManager, associated schemas, benchmarks, and tests.
* Merge DefaultNomtSpec into DefaultSpec — there is no longer a separate NOMT variant.
* Refactor NomtProverStorage to deduplicate per-namespace read logic into shared helpers.
* Fix NomtVerifierStorage to short-circuit when a namespace has no accesses (avoids trying to verify an empty witness).
* Add write_kernel_marker / push_kernel_marker helpers — NOMT requires both user and kernel namespaces to advance together at every commit, so all tests that only touch user state now also write a dummy kernel value.
* Migrate all tests from `SimpleJmtStorageManager` → `SimpleStorageManager`, `TestJmtSpec` → `TestSpec`, `ZkStorage` → `NomtVerifierStorage`.

---

## Linked Issues
- Related to https://github.com/Sovereign-Labs/sovereign-sdk-wip/issues/2634 

## Testing

## Docs
Updates.


crates/full-node/sov-db/src/namespaces.rs
```diff
@@ -4,15 +4,17 @@ use rockbound::schema::ColumnFamilyName;
 
 pub use crate::schema::namespace::Namespace;
 
+// The `_jmt_values` suffix on the STATE_VALUES_TABLE_NAME constants is a
```
theodorebugnet, line 7: I wonder if we can schedule a full-resync upgrade? This seems like the kind of thing that would be really nice to clean up, so we don't have confusing naming forever into the future.

Definitely not necessary as part of the PR but maybe worth filing an issue and keeping in mind for next time we have a big upgrade. Forcing a resync is not a huge deal, we already did it earlier this year when we changed the hyperlane event counters for example. Maybe something to coordinate with customers. (On our side it's not a very big deal for relay deployments, for example.)

theodorebugnet (approved):
Looks good to me!

As a tiny nit: some comment changes in the diff switch from "JMT state" to "authenticated state". I'm not sure if the state is really authenticated - maybe "merkleized state" would be more accurate? (Or "proven state".)

Not a blocker if it's annoying to change or if you disagree.

crates/full-node/sov-db/src/namespaces.rs
citizen-stig: https://github.com/Sovereign-Labs/sovereign-sdk/issues/2903

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 # 2026-05-18
 - #2881 **Breaking config change**: Moves `max_concurrent_proof_blobs` from `[sequencer]` to `[proof_manager]` in rollup TOML configs. The `[proof_manager]` section is now optional for `operator` rollups and remains required for `zk`/`optimistic` rollups (enforced at startup).
 - #2879 Updates NOMT crate version. **Breaking for existing ZK-rollups**. Proof type has changed.
+- #2884 Removes JMT crate and JMT-base storage.
```
citizen-stig:
```suggestion
- #2884 Removes JMT crate and JMT-base storage. Please update rollup to use `NomtStorageManager` instead of `StorageManager`
```

PR #2886, opened 2026-05-18
Re-enable pruning

citizen-stig:
# Description

This PR re-enables NOMT historical-state pruning by moving user/kernel pruning to `rockbound::VersionedDB::collect_pruning_batch` (PR https://github.com/Sovereign-Labs/rockbound/pull/51), keeping accessory pruning through the existing pruner, and committing both delete batches duringfinalization. It replaces flat pruning config fields with `PrunerConfig::{Off, Periodic, OnceAtStartup}`, adds a startup-only prune hook to HierarchicalStorageManager, and optionally compacts pruned CFs after startup pruning.

Main tradeoffs: defaulting to Off protects fresh nodes from accidental history deletion but makes config migration more important; `OnceAtStartup` avoids live read/write contention but can make startup expensive; range tombstones reduce pruning-index write volume but need compaction to reclaim space.

---

## Testing
New tests have been added

## Docs
No updates


crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -361,114 +361,35 @@ where
...
-                    tracker.submit(PrunerMetric {
-                        db: "versioned_dbs",
-                        keys_inspected,
-                        keys_to_prune,
-                        time: pruning_time,
-                    });
-                });
+                let keep = versions_to_keep as u64;
+                let user_output = user_db.collect_pruning_batch(keep, Some(max_batch_size))?;
+                let remaining = max_batch_size.saturating_sub(user_output.keys_to_prune);
```
citizen-stig: why it is saturating and not checked?

crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
```diff
@@ -169,13 +187,34 @@ where
...
-        let pruner_max_batch_size = config.get_pruner_max_batch_size();
-        assert!(
-            pruner_versions_to_keep >= 1,
-            "Pruner versions to keep should be at least 1, got {pruner_versions_to_keep}",
-        );
+        // Derive the periodic-pruning fields (used by `finalize`) and the optional startup-prune
+        // marker from the pruning policy. `Off` and `OnceAtStartup` both leave
+        // `pruner_block_interval = None`, so `finalize` never spawns a periodic pruner for them;
+        // `OnceAtStartup` instead carries `startup_prune`, consumed once by `prune_once_at_startup`.
+        let (pruner_block_interval, pruner_versions_to_keep, pruner_max_batch_size, startup_prune) =
```
citizen-stig: This section is too convoluted.

crates/module-system/sov-modules-rollup-blueprint/src/native_only/mod.rs
```diff
@@ -441,6 +441,10 @@ pub trait FullNodeBlueprint<M: ExecutionMode>: RollupBlueprint<M> {
         let mut storage_manager =
             startup_step!(self.create_storage_manager(&rollup_config, witness_generation));
 
+        // Run a one-time pruning pass before any state access, if the storage manager is
+        // configured for it (`PrunerConfig::OnceAtStartup`). No-op otherwise.
+        storage_manager.prune_once_at_startup()?;
```
citizen-stig: Why it is here and not in storage manager constructor?

crates/full-node/sov-db/src/storage_manager/nomt_based/pruning.rs
```diff
@@ -0,0 +1,259 @@
...
+use sov_rollup_interface::reexports::digest;
+
+use super::groups::{DbGroup, PrunerJob};
+use crate::config::{PrunerConfig, RollupDbConfig};
+
+/// Pruning schedule resolved from [`PrunerConfig`] at construction time: `NonZero` values are
+/// unwrapped to plain `usize` and `max_batch_size` is resolved to its concrete default. Each
+/// variant carries exactly the parameters that mode needs, so there are no unused/dummy fields.
+#[derive(Debug, Clone, Copy)]
+enum PrunerSchedule {
```
bkolad: This type is very similar to `PrunerConfig`; why do we need a new type instead of using the serde machinery to handle default values?

crates/full-node/sov-db/src/storage_manager/nomt_based/pruning.rs
```diff
@@ -0,0 +1,259 @@
...
+        K: Eq + std::hash::Hash + Clone + std::fmt::Debug,
+    {
+        let is_pruner_ready = self
+            .in_flight
+            .as_ref()
+            .map(|p| p.is_finished())
+            .unwrap_or(false);
+        let mut pruning_commit_time = None;
+
+        if is_pruner_ready {
```
bkolad: Instead of the is_pruner_ready flag and unwrap, let's use `if let Some(..) = ` pattern

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
```diff
@@ -981,9 +1007,22 @@ fn apply_storage_defaults_and_overrides(config: &mut RollupDbConfig, notes: &mut
         config.kernel_preallocate_ht = Some(false);
         notes.push("storage.kernel_preallocate_ht missing; defaulted to false".to_string());
     }
-    if config.pruner_versions_to_keep.is_none() {
-        config.pruner_versions_to_keep = Some(20);
-        notes.push("storage.pruner_versions_to_keep missing; defaulted to 20".to_string());
+    // Only default-enable pruning when the operator left the section out entirely. An explicit
+    // `[storage.pruner]` (including `pruner = "off"`, e.g. an archival node) is respected so the
+    // migration never silently starts deleting history the operator chose to retain.
+    if !pruner_present {
```
bkolad: This pruner override has no effect:
  1. The migration is an offline tool; it never runs the node's finalize loop, and Periodic pruning only fires from finalize. So nothing is pruned while it runs.
   2. The config is never written back to disk, so the override is discarded on exit — after restart the node re-reads the original TOML.

Lets don't deal with this setting in the migration tool.

crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
```diff
@@ -567,44 +589,9 @@ where
...
-            if should_run_pruner {
-                let pruner = self
-                    .db_group
-                    .start_pruner(self.pruner_versions_to_keep, self.pruner_max_batch_size);
-                self.pruner = Some(pruner);
-            }
-        }
+        let pruning_commit_time = self
+            .pruning
+            .on_finalize(&mut self.db_group, block_header.height())?;
```
bkolad: Any pruning error makes finalize return Err, which propagates through the runner and shuts the node down. If that's intended, let's ensure this produces a very clear error message so operators know pruning — not finalization — failed.

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -173,20 +173,49 @@ where
             && version_to_use == self.latest_version()
     }
 
+    /// Resolves the result of an *unbound* (live-tip) read.
+    ///
+    /// Unbound reads bypass the snapshot pin (`get_latest_borrowed_unbound`), so they always
+    /// observe the latest committed version and can therefore *never* surface
+    /// [`HistoricalValueError::PrunedVersion`] — only genuine I/O faults. We encode that invariant
+    /// with a `debug_assert!`, and in production we log-and-return `None` rather than panicking, so
+    /// a transient I/O hiccup cannot crash block processing.
```
bkolad: This doesn’t seem correct. None is a valid storage result meaning “the key is absent”, but here we also return None for an I/O/read error. That makes callers act on potentially false absence. In particular, this can produce incorrect latest-state reads or invalid storage-proof responses instead of surfacing the underlying DB error. Let's keep this path fallible, retry and then return an error, or preserve the previous fail-fast behavior.

crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -173,20 +173,49 @@ where
...
+            Err(error) => {
+                debug_assert!(
+                    !matches!(
+                        error.downcast_ref::<HistoricalValueError>(),
+                        Some(HistoricalValueError::PrunedVersion { .. })
+                    ),
+                    "unbound reads should never observe PrunedVersion (source={source})"
+                );
+                tracing::error!(?error, %source, "unbound read failed; returning None");
+                None
```
bkolad: Here is the problem

crates/full-node/sov-db/src/storage_manager/nomt_based/pruning.rs
citizen-stig: Done! 

The only downside is that we `take` the job, so in new code, it takes it all the time, and if it is not finished, it puts it back.

examples/demo-rollup/src/migrations/mockda_to_celestia.rs
citizen-stig: Addressed, and removed config migration from this PR

crates/full-node/sov-db/src/storage_manager/nomt_based/mod.rs
citizen-stig: Added context

crates/module-system/sov-state/src/nomt/prover_storage.rs
citizen-stig: Good find, addressed

crates/module-system/sov-state/src/nomt/prover_storage.rs
citizen-stig: Same, addressed.

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
```diff
@@ -273,11 +283,18 @@ where
 
     // Flush pruning schema batches to disk.
     pub(crate) fn commit_pruning(&mut self, group: PruneGroup) -> anyhow::Result<()> {
+        #[cfg(test)]
+        if FAIL_NEXT_PRUNING_COMMIT.swap(false, Ordering::SeqCst) {
```
bkolad: In line 42 we have test method that sets it to true, let;s make the code above a separate method as ell

bkolad (approved):

crates/full-node/sov-db/src/storage_manager/nomt_based/groups.rs
citizen-stig: Addressed

crates/full-node/sov-db/src/storage_manager/nomt_based/pruning.rs
citizen-stig: Addressed!

PR #2890, opened 2026-05-19
Add borsh metered reader

ross-weir:
# Description

Adds a metered reader that is used for borsh deserialization.

Now we provide generic borsh deserialization costing:
- An upfront constant
- A per-byte cost
- A per-read cost

This can apply to all borsh structs so we no longer require individual constants for each type.

Here's an AI summary/example of why the per-byte and per-read constants:

>   The two costs:
>   - BORSH_PER_READ_BIAS (~29 gas): fixed cost charged once per read call. Covers function call, metering arithmetic, error handling. Same whether you read 1 byte or 1000.
>   - BORSH_PER_BYTE_READ (~0.27 gas/byte after our calibration): linear cost per byte actually read. Covers the memcpy work itself.
> 
>   Why both are needed — concrete example:
> 
>   Workload A — Vec<u8> of 1000 bytes: 2 reads (4-byte length prefix + 1000-byte body)
>   charge = 2 × 29 + 0.27 × 1004 ≈ 329 gas
> 
>   Workload B — struct of 1000 u8 fields: 1000 reads of 1 byte each, same 1000 total bytes
>   charge = 1000 × 29 + 0.27 × 1000 ≈ 29,270 gas
> 
>   Same byte total, ~90× different gas. That reflects reality — 1000 read calls really is much more expensive than 2 reads, because each call has fixed overhead the prover pays.
> 
>   If we only had BORSH_PER_BYTE_READ: both workloads would charge ~270 gas. Workload B would be massively undercharged → an attacker could craft struct-of-bytes payloads that burn prover time for almost no fee.
> 
>   If we only had BORSH_PER_READ_BIAS: workload A would charge ~60 gas (no byte component). Bulk-read decodes massively undercharged → an attacker submits 10MB blobs for the cost of two reads.
> 
>   The two-parameter model captures both axes. A Vec<u8> decode pays mostly per-byte; a deeply-nested struct decode pays mostly per-read. Neither is treated as the other.

Benchmarks for borsh deserialization will be in next PR

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


constants.toml
```diff
@@ -137,26 +137,18 @@ PROCESS_TX_PRE_EXEC_GAS = [1, 1]
 PROCESS_TX_PRE_EXEC_GAS_PER_TX_BYTE = [1, 1]
 # --- End Gas parameters to specify how to charge gas for signature verification ---
 
-# The cost of deserializing a message using Borsh
-DEFAULT_GAS_TO_CHARGE_PER_BYTE_BORSH_DESERIALIZATION = [1, 1]
```
bkolad, line 144: Let's also remove the constants form `fn reset_constants()` in `integration-tests/tests/stf_blueprint/mod.rs:118`

crates/module-system/sov-modules-api/src/gas/metered_reader.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        // Charge after the inner read so partial reads / EOF / inner failures
+        // don't over-charge for bytes that were not delivered.
+        let n = self.inner.read(buf)?;
+        self.charge(n)?;
+        Ok(n)
+    }
+
+    fn read_exact(&mut self, buf: &mut [u8]) -> io::Result<()> {
+        // Override required: `<&[u8] as Read>::read_exact` bypasses `Read::read`, skipping metering.
```
bkolad: Why the comment `skipping metering`?  In line 72 there is a call to `charge`

crates/module-system/module-implementations/sov-test-modules/src/access_pattern/mod.rs
```diff
@@ -21,41 +20,11 @@ pub const MAX_VEC_LEN_BENCH: usize = 100_000;
 /// Max length of a string for a bench pattern call message
 pub const MAX_STR_LEN_BENCH: usize = 1_024;
 
-/// A newtype struct that deserializes into a string and charges gas.
+/// A newtype struct that deserializes into a string. Metered decode is provided by
+/// the blanket impl on `BorshDeserialize` in `sov-modules-api`.
 #[derive(BorshSerialize, BorshDeserialize, Debug, Clone)]
 pub struct MeteredBorshDeserializeString(pub String);
```
bkolad: Would it be possible to remove the special casing for String deserialisation?

crates/module-system/sov-modules-api/src/gas/metered_reader.rs
```diff
@@ -0,0 +1,74 @@
...
+        self.inner.read_exact(buf)?;
+        self.charge(buf.len())
```
bkolad: Don't we need something like:
```
  fn read_exact(&mut self, mut buf: &mut [u8]) -> io::Result<()> {
      let mut consumed = 0usize;

      while !buf.is_empty() {
          match self.inner.read(buf) {
              Ok(0) => {
                  self.charge(consumed)?;
                  return Err(io::Error::from(io::ErrorKind::UnexpectedEof));
              }
              Ok(n) => {
                  consumed += n;
                  let tmp = std::mem::take(&mut buf);
                  let (_, rest) = tmp.split_at_mut(n);
                  buf = rest;
              }
              Err(err) if err.kind() == io::ErrorKind::Interrupted => continue,
              Err(err) => {
                  self.charge(consumed)?;
                  return Err(err);
              }
          }
      }

      self.charge(consumed)
  }
```
 `Read::read_exact` may perform multiple inner `read` calls. If some of those
   reads succeed and a later one hits EOF, `read_exact` returns `UnexpectedEof`
   after partial progress. In the original code, we delegated to  `self.inner.read_exact(buf)?` and only charged afterward, so any bytes consumed before the error were left unmetered.

constants.toml
ross-weir: done

crates/module-system/sov-modules-api/src/gas/metered_reader.rs
ross-weir: removed

crates/module-system/module-implementations/sov-test-modules/src/access_pattern/mod.rs
ross-weir: removed the `MeteredBorshDeserializeString` newtype

crates/module-system/sov-modules-api/src/gas/metered_reader.rs
ross-weir: fixed

bkolad (approved):

crates/module-system/sov-modules-api/src/gas/metered_reader.rs
```diff
@@ -0,0 +1,92 @@
...
+        let mut consumed = 0usize;
+        while !buf.is_empty() {
+            match self.inner.read(buf) {
+                Ok(0) => {
+                    let _ = self.charge(consumed);
+                    return Err(io::Error::from(io::ErrorKind::UnexpectedEof));
+                }
+                Ok(n) => {
+                    consumed += n;
+                    let tmp = std::mem::take(&mut buf);
```
citizen-stig: Has this been written by claude? I think modern compiler version support direct reslicing. I've checked it locally, and it compiles
```suggestion
                    buf = &mut buf[n..];
```

crates/module-system/sov-modules-api/src/gas/metered_reader.rs
```diff
@@ -0,0 +1,92 @@
...
+                    let (_, rest) = tmp.split_at_mut(n);
+                    buf = rest;
+                }
+                Err(err) if err.kind() == io::ErrorKind::Interrupted => continue,
```
citizen-stig, line 81: nit: I think it is good practice to retry on this one, but I think it would be nice to put some hard limit on number of such retries.

crates/module-system/sov-modules-api/src/gas/metered_reader.rs
ross-weir: yeah it was, nice improvement

crates/module-system/sov-modules-api/src/gas/metered_reader.rs
ross-weir: Going to keep it as is to keep it simple, we can limit if it becomes a problem - the default impl in std also doesn't impose limits

citizen-stig: Ok, I live it up to you. In my experience it was very hard to pull out of my head where "silently stuck rollup" is coming from, but I think it is skill issue.

PR #2893, opened 2026-05-19
Custom stateful EVM precompiles

theodorebugnet:
# Description

Exposes the ability to provide custom precompiles to `sov-evm`, through revm's `PrecompileProvider` trait.

* An `EvmPrecompile` trait allows custom precompile implementers.
* A small declarative macro can generate an `EvmPrecompileSet` impl from a list of `EvmPrecompile`s.
* The `sov-evm::Evm` module is now generic on `EvmPrecompileSet` (defaulting to an empty set implementation)
* Internally it then constructs a `SovPrecompileProvider<P: EvmPrecompileSet>`, exposing every `EvmPrecompile` as a precompile to `revm`

Two precompiles are already implemented to serve as an example and be used in tests: `BankBalancePrecompile` allows querying `sov-bank` for arbitrary modules, and `SequencingTimestampPrecompile` inspects the transaction context to get the `HDTimestamp` provided by the SequencingData (if available).

Using `const fn` it's possible to statically verify that the precompiles in an `EvmPrecompileSet` don't have conflicting addresses, and don't conflict with (known current) Ethereum standard precompiles.

### Limitations
1. Activation-height config is not implemented yet - all configured precompiles are enabled at all times. This will be added in a follow-up PR.
2. State accesses are intended to be read-only because as far as I can tell, compatibility with EVM's call-frame revert semantics would require explicit support. This is sufficient for current usecases. We can add closer integration if read-write precompiles are necessary in the future (it looks potentially a bit thorny though, we'd need to provide a custom `revm::Journal` to hook into its call-frame lifecycle properly, which is doable provided it doesn't cause any extra snags).

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/module-implementations/sov-evm/src/evm/executor.rs
```diff
@@ -46,32 +47,47 @@ pub(crate) fn get_cfg_env(
 }
 
 /// Execute an Ethereum transaction and commit it to the database.
-pub fn transact_commit<DB: Database<Error = E> + TryDatabaseCommit<Error = E>, E: DBErrorMarker>(
+#[cfg(feature = "native")]
```
citizen-stig, line 50: is it intentionally become native only?

crates/module-system/module-implementations/sov-evm/src/precompiles/bank_balance.rs
```diff
@@ -0,0 +1,108 @@
...
+use sov_modules_api::{Spec, TxState};
+
+use super::{
+    Address, EvmPrecompile, EvmPrecompileEnv, EvmPrecompileSet, PrecompileError, PrecompileOutput,
+    PrecompileResult,
+};
+
+/// The gas-token bank balance precompile address.
+pub const BANK_BALANCE_PRECOMPILE_ADDRESS: Address = Address::new([
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x01, 0x00, 0x00,
```
citizen-stig, line 13: Do we ensure that our precompile addresse do not conflict with Ethereum L1 standard precompiles?

crates/module-system/module-implementations/sov-evm/src/precompiles/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        Self {
+            eth,
+            custom,
+            sov_context,
+        }
+    }
+}
+
+impl<'a, S, P, CTX> PrecompileProvider<CTX> for SovPrecompileProvider<'a, S, P>
```
citizen-stig, line 103: nit: I know it is not consistent already, but our eng guidelines have a suggestion how to capitalize acronyms like `Ctx`

crates/module-system/module-implementations/sov-evm/src/precompiles/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+                gas,
+                output: output.bytes,
+            }
+        }
+        Err(PrecompileError::OutOfGas) => InterpreterResult {
+            result: InstructionResult::PrecompileOOG,
+            gas: Gas::new(gas_limit),
+            output: Bytes::new(),
+        },
+        Err(_) => InterpreterResult {
```
citizen-stig, line 190: Can we not swallow error and passed it inside PrecompileError? 
I know Ethereum has long-standing tradition of not giving descriptive errors, but I think we have opportunity here to make life a little bit easier here.

crates/module-system/module-implementations/sov-evm/src/precompiles/traits.rs
```diff
@@ -0,0 +1,235 @@
...
+        env: &mut EvmPrecompileEnv<'_, S, ST>,
+    ) -> PrecompileResult;
+}
+
+/// A precompile set with no custom precompiles.
+#[derive(Debug, Clone, Default)]
+pub struct NoCustomPrecompiles<S>(PhantomData<S>);
+
+impl<S: Spec> EvmPrecompileSet<S> for NoCustomPrecompiles<S> {
+    const ADDRESSES: &'static [Address] = &[];
```
citizen-stig, line 47: Should we add support of default L1 precompiles, such as `ecrecover 0x01`, sha256 0x02

crates/module-system/module-implementations/sov-evm/src/precompiles/bank_balance.rs
citizen-stig: Found: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2893/changes#diff-20741b66a14a5a6bf18a5d0e4d5f4929f910f4b088f38d9a123d1efa2390b401R116

crates/module-system/module-implementations/sov-evm/src/precompiles/bank_balance.rs
```diff
@@ -0,0 +1,108 @@
...
+        }
+        _ => {
+            return Err(PrecompileError::InvalidInput(
+                "expected 20-byte address or 20-byte address plus 32-byte token id",
+            ));
+        }
+    };
+
+    let address = S::Address::from_vm_address(
+        EthereumAddress::try_from(address_bytes)
```
citizen-stig: Why not use `EthereumAddress::new` then if it is infallible?

crates/module-system/module-implementations/sov-evm/src/precompiles/bank_balance.rs
```diff
@@ -0,0 +1,108 @@
...
+    );
+
+    let balance = bank
+        .get_balance_of(&address, token_id, state)
+        .map_err(|e| PrecompileError::State(e.to_string()))?
+        .unwrap_or_default();
+
+    Ok(PrecompileOutput {
+        gas_used: BANK_BALANCE_GAS,
+        bytes: Bytes::copy_from_slice(&U256::from(balance.0).to_be_bytes::<32>()),
```
citizen-stig, line 110: I belive we can reducy memcpy here by using `Bytes::from`, (since bytes 1.x has From<Vec<u8>>):

```rust

    let be = U256::from(balance.0).to_be_bytes::<32>();
    Ok(PrecompileOutput {
        gas_used: BANK_BALANCE_GAS,
        bytes: Bytes::from(be.to_vec()),
    })
```

crates/module-system/module-implementations/sov-evm/src/precompiles/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+
+/// Error from a Sovereign precompile.
+#[derive(Debug, Clone, thiserror::Error, PartialEq, Eq)]
+pub enum PrecompileError {
+    /// The precompile ran out of EVM gas.
+    #[error("precompile out of gas")]
+    OutOfGas,
+    /// The precompile input was invalid.
+    #[error("invalid precompile input: {0}")]
+    InvalidInput(&'static str),
```
citizen-stig: If we also use `String` here, we can tell user more about the input, for example for bank we expect length to be 20 or 52, and we can tell it back, which length we received. 

We probably can assume user should be able to know the input size it passes, since it controls it, but by having 1 more allocation we can give user extra convinience. Considering that actual human or LLM can be behind layers and layers of Typescript and frameworks, knowing what size precompile actually received might be useful.

crates/module-system/module-implementations/sov-evm/src/precompiles/sequencing_timestamp.rs
```diff
@@ -0,0 +1,87 @@
...
+        .map(Ok)
+        .unwrap_or_else(|| {
+            chain_state
+                .get_oracle_time_nanos(env.state)
+                .map_err(|e| PrecompileError::State(e.to_string()))
+        })?;
+
+    Ok(PrecompileOutput {
+        gas_used: SEQUENCING_TIMESTAMP_GAS,
+        bytes: Bytes::copy_from_slice(&U256::from(nanos).to_be_bytes::<32>()),
```
citizen-stig, line 88: same not about memcpy and `Bytes::from`

citizen-stig (changes requested):
Neat! I had expected much larger PR!

I would like to iterate on error handling and some memory allocations

crates/module-system/module-implementations/sov-evm/src/evm/executor.rs
theodorebugnet: Yep, it's only used in a native only RPC.
Previously it was `pub` but wasn't really used anywhere else, which is why the compiler allowed it to be non-gated. I'm intentionally adding a breaking change by making it `pub (crate)` instead, because now it has all the internal Precompile types as part of the signature, and so as a result rustc flags it as unused in non-native compilation.

I can restore it as `pub` if we also make all the other types pub if we think there's a risk of anyone else using it, but I don't see why anyone would

crates/module-system/module-implementations/sov-evm/src/precompiles/traits.rs
theodorebugnet: Yep, that's provided by `revm` natively as `EthPrecompiles`, and `SovPrecompileProvider` constructs it unconditionally. `NoCustomPrecompiles` just means no extra sov-provided precompiles that we add on top

crates/module-system/module-implementations/sov-evm/src/precompiles/mod.rs
theodorebugnet: This is tricky because `InterpreterResult` is a revm internal type and is very limited, since it's part of EVM execution as far as I can tell.

But I think there's a way to pass context through to RPC/user errors, let me look into this, thanks for flagging

crates/module-system/module-implementations/sov-evm/src/precompiles/bank_balance.rs
theodorebugnet: I don't quite see how that would help - when doing `to_vec()`, doesn't that allocate a new Vec and copy the data there anyway?
I tried to ask codex but it says there's no difference in allocations either. Let me know if I've missed something.

crates/module-system/module-implementations/sov-evm/src/precompiles/mod.rs
theodorebugnet: This is the usage inside `revm` itself: the trait is defined as `PrecompileProvider<CTX>`. What do you think, should we rename it according to our own style, or should we use the upstream style for the upstream trait?

crates/module-system/module-implementations/sov-evm/src/precompiles/mod.rs
theodorebugnet: Added extra context

crates/module-system/module-implementations/sov-evm/src/precompiles/bank_balance.rs
citizen-stig: Good point, I've thought that to_vec won't allocate =(

citizen-stig (approved):

PR #2905, opened 2026-05-22
Make NodeDiscovery resilient to missed PostgreSQL notifications

bkolad:
# Description
`NodeDiscovery` previously blocked on `PgListener::recv()` and only re-polled cluster info when a PostgreSQL `NOTIFY` arrived. If a notification was dropped, e.g., the listener connection was severed while membership changed, the change was never observed, and a listener error would terminate the discovery task entirely.

This PR makes discovery resilient:
1. `NodeDiscovery::connect` now takes a `poll_interval`. The discovery loop waits
    *up to* that interval for a notification via `timeout(poll_interval, try_recv())`
    and then re-polls regardless, bounding how stale cluster info can get.
2. A reconnect (`try_recv` returning `None`) triggers an immediate re-poll, since
    notifications sent before `LISTEN` was restored may have been lost.
3. Listener errors are now logged + metered and retried instead of killing the
    task, so discovery survives database connection loss.
4. `ClusterInfoService` passes a `DEFAULT_POLL_INTERVAL` of 3s.

### DIFF
--- a/crates/utils/sov-proxy-utils/src/node_discovery.rs
+++ b/crates/utils/sov-proxy-utils/src/node_discovery.rs
@@ -73,6 +73,8 @@ impl NodeDiscoveryTask {
 
 #[derive(Debug, thiserror::Error)]
 enum HandleClusterUpdateError {
+    #[error("Failed to connect to the cluster database")]
+    Connect(#[source] anyhow::Error),
     #[error("Failed to fetch cluster info")]
     GetClusterInfo(#[source] anyhow::Error),
     #[error("Failed to notify on cluster update")]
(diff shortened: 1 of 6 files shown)


crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -216,6 +222,42 @@ impl NodeDiscovery {
         NodeDiscoveryTask { receiver, handle }
     }
 
+    /// Waits once for the next cluster change signal before returning to the
+    /// caller's loop, which always re-polls cluster info.
+    async fn recv_notification(&mut self) {
```
bkolad: This is the most important change.

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -216,6 +222,42 @@ impl NodeDiscovery {
...
+            // so re-poll cluster info immediately instead of waiting for the
+            // next poll interval.
+            Ok(Ok(None)) => {
+                tracing::debug!(
+                    "Cluster notification listener reconnected after connection loss, re-polling"
+                );
+            }
+            // No notification arrived within `poll_interval`. This is expected
+            // when the cluster is quiet; we re-poll anyway to bound staleness.
+            Err(_) => {
```
bkolad, line 276: We will poll handle_cluster_update even when there are no DB notifications, but handle_cluster_update will trigger a proxy reload only if the cluster has changed.

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -110,6 +136,22 @@ impl NodeDiscovery {
...
         connection_string: &str,
         max_age: Duration,
+        poll_interval: Duration,
+        notifier: Option<Box<dyn ClusterUpdateNotifier>>,
+    ) -> Result<Self> {
+        Self::try_connect(connection_string, max_age, poll_interval, notifier)
+            .await
+            .map_err(|error| {
+                let error = HandleClusterUpdateError::Connect(error);
+                error.report();
```
bkolad, line 146: Now the connection errors are visible in metrics

ross-weir (approved):

PR #2907, opened 2026-05-22
Add cache control to ledger API routes

### LINKED ISSUE #2513: Ledger REST API Cache friendliness improvements
LedgerDb effectively have 2 types endpoints:

* "live" which should never be cached.
  * `/ledger/slots/latest`
  * `/aggregated-proofs/latest/`
* historical, that can be cached forever (for instant finality, or after finalization):
  * `/ledger/slots/{slotId}`
  * `/ledger/txs/{txId}`
  * etc..

Currently no Cache-Control headers are set, leaving caching behaviour undefined for clients and intermediaries (CDNs, reverse proxies).

### Historical endpoints
```
Cache-Control: public, max-age=31536000, immutable
```

* `public` — cacheable by CDNs and shared proxies, not just the browser
* `max-age=31536000` — 1 year in seconds (practical "forever")
* `immutable` — tells clients: don't bother revalidating, this URL's content will never change. Only valid for content-addressed URLs (by hash). Do not use on height-based routes.

### Live endpoints
```
Cache-Control: no-store
```

`no-store` — strongest option; response must not be written to any cache at any layer. Prefer this over no-cache (which still stores but revalidates) since stale ledger state is actively harmful.

samsamtrum:
## Summary
- add `Cache-Control: no-store` to live ledger responses (`/ledger/slots/latest` and `/ledger/aggregated-proofs/latest`)
- add long-lived public cache headers to identifier-based ledger resources
- keep height-based finalized routes uncached because their result can change as finality advances

## Why
This improves cache behavior for clients, CDNs, and reverse proxies while avoiding stale data on live endpoints.

Fixes #2513

## Testing
- `cargo fmt --check --package sov-ledger-apis`
- `git diff --check`
- `cargo check -p sov-ledger-apis`

### DIFF
--- a/crates/full-node/sov-ledger-apis/src/lib.rs
+++ b/crates/full-node/sov-ledger-apis/src/lib.rs
@@ -3,7 +3,7 @@ use std::marker::PhantomData;
 use std::ops::Range;
 
 use axum::extract::{Request, State, WebSocketUpgrade};
-use axum::http::StatusCode;
+use axum::http::{header, HeaderValue, StatusCode};
 use axum::middleware::Next;
 use axum::response::{IntoResponse, Response};
 use axum::routing::get;
@@ -34,6 +34,7 @@ use sov_rollup_interface::node::ledger_api::{
 };
 use sov_rollup_interface::stf::TxReceiptContents;
 use tokio::sync::watch;
+use tower_http::set_header::SetResponseHeaderLayer;
 
 type PathMap = Path<HashMap<String, NumberOrHash>>;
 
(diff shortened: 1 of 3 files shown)


crates/full-node/sov-ledger-apis/src/lib.rs
```diff
@@ -146,34 +155,42 @@ where
...
-                Self::router_slot(state.clone()).route_layer(middleware::from_fn_with_state(
-                    state.clone(),
-                    Self::resolve_slot_id,
-                )),
+                Self::router_slot(state.clone())
+                    .route_layer(middleware::from_fn_with_state(
+                        state.clone(),
+                        Self::resolve_slot_id,
+                    ))
+                    .layer(cache_forever.clone()),
```
citizen-stig: This will poison cache for chain with non instant finality

samsamtrum:
recheck

samsamtrum:

samsamtrum:
Thanks for the review. I removed the long-lived cache headers from the slot/batch/tx/event identifier routes so they cannot poison caches on chains with non-instant finality. The PR now only adds `Cache-Control: no-store` to the live endpoints (`/ledger/slots/latest` and `/ledger/aggregated-proofs/latest`).

Re-ran locally:
- `cargo fmt --check --package sov-ledger-apis`
- `git diff --check`
- `cargo check -p sov-ledger-apis`

PR #2910, opened 2026-05-22
Unify `get_root_hash` with `get_value` semantically

citizen-stig:
# Description

Currently `Storage::get_root_hash` semantically deviates from other data fetching methods, when it return error on wrong version (while also bubbling up storage errors)

Other methods, such as `get_value` return Option and storage errors cause panics. This PR unifies methods to simplify reasoning. It also clarify version semantics.

---

## Linked Issues

Fixes https://github.com/Sovereign-Labs/sovereign-sdk-wip/issues/3030 

## Testing
Existing tests are passing

## Docs
Doc strings are updated


crates/module-system/sov-state/src/nomt/prover_storage.rs
```diff
@@ -905,18 +904,18 @@ where
             );
         }
 
-        Ok(storage_root_historical)
+        Some(storage_root_historical)
     }
 
-    fn get_root_hash_unbound(&self, version: SlotNumber) -> anyhow::Result<Self::Root> {
+    fn get_root_hash_unbound(&self, version: SlotNumber) -> Option<Self::Root> {
```
bkolad, line 875: We have this comment:
```
storage.rs:411
Returns `Some(root)` iff a root has been committed for that version, else `None`.
```
But this is not true here. If the last commited version =10 and we ask `get_root_hash_unbound(11)`
`get_serialized_root_hash` will still return the root hash for ver=10. 
This matches the old behavior, but we should update the comment.

bkolad (approved):

PR #2912, opened 2026-05-23
Add `prover-incentives` rejection and admin upgrade observability

bkolad:
# Description

Adds observability for `ProverIncentives::process_proof` outcomes:
  - Logs proof rejections, prover slashing, and prover penalties.
  - Emits `sov_prover_incentives_rejected_proof` metrics with rejection reason details.
  - Emits `sov_prover_incentives_admin_upgrade` metrics for successful and failed admin upgrade attempts.
  - Uses an enum for admin upgrade metric reasons.
  - Centralizes native/non-native metric handling in metrics.rs, with no-op helpers for non-native builds.

## Linked Issues
- Related to #2551

image

### DIFF
--- a/crates/module-system/module-implementations/sov-prover-incentives/src/capabilities.rs
+++ b/crates/module-system/module-implementations/sov-prover-incentives/src/capabilities.rs
@@ -279,6 +279,7 @@ impl<S: Spec> ProverIncentives<S> {
             prover_address,
             &ctx,
             is_admin,
+            execution_context,
             state,
         )?;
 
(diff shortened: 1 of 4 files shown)


crates/module-system/module-implementations/sov-prover-incentives/src/capabilities.rs
```diff
@@ -127,6 +127,51 @@ impl<S: Spec> ProverIncentives<S> {
...
+            }
+            ProcessProofError::ProverPenalizedNoRevert(reason) => {
+                tracing::warn!(
+                    %prover_address,
+                    %reason,
+                    "Prover penalized while processing proof"
+                );
+            }
+            _ => {
+                tracing::debug!(
```
ross-weir, line 156: The other variants seem like they're also worth logging at a `warn` level, unless they're already captured

ross-weir (approved):

crates/module-system/module-implementations/sov-prover-incentives/src/capabilities.rs
bkolad: The idea was to issue warnings only for penalization or slashing events—basically, whenever funds are involved.

PR #2917, opened 2026-05-26
Attempt to fix flaky `test_gas_limit_update`

citizen-stig:
# The fix

1. **`send_raw_tx_to_sequencer → send_raw_tx_to_sequencer_with_retry`** — Eliminates: the panic Tx N failed with error: ... 503 Service Unavailable WaitingOnBlobSender. This fires when the sequencer's in-flight batch-blob counter saturates at TEST_MAX_CONCURRENT_BATCH_BLOBS = 16  because the test creates batches via `force_close_batch` faster than the blob-sender drains them to "Finalized + Accepted". The retry helper transparently waits for the cap to drop before letting the tx submission succeed.
2. **Replace the fixed 10-block post-loop with a drain loop** that produces DA blocks until every sequencer-accepted tx has appeared in a slot (bounded at 60 blocks for safety). Under load, force_close_batch + produce_block sometimes produces empty DA blocks while the just-closed batch is still in flight; the tx lands later. The drain ensures ledger and sequencer counts match
3. Observe base_fee_per_gas inside the drain loop in addition to the main loop.  
   **Eliminates**: the panic Max observed base fee per gas after update must be greater than The base fee can only rise above 7 once a rollup block contains a post-gas-limit-change tx with gas_used > gas_target (33k vs 25k after the override kicks in). When batches arrive late, the main loop's observations land on rollup blocks with gas_used = 0, so the base fee sits at its empirical floor of 7 (the EIP-1559 decrease delta truncates to 0 there and stops moving  see crates/module-system/module-implementations/sov-chain-state/src/gas.rs:188-195). Only when the late-arriving batches finally land — during the drain — does the gas_used push the base fee up, and observing there catches that rise

No production code is changed; the underlying race in force_close_batch / tenderly_produce_blocks is still there, but the test is now resilient to it.

## What Claude learned along the way

It went down two wrong paths before landing here — both worth flagging:
- A wait_for_next_published_batch_blob helper (consuming events from subscribe_to_blobs_from_blob_sender) felt elegant but actually made the test worse (~95% failure rate). Setup-phase events buffered in the WS subscription throw off the 1:1 iteration ↔ Published-event matching.
- A "produce DA blocks until our tx lands by hash" loop pushed the test into a regime where many empty blocks separated tx blocks, killing the EIP-1559 rise. The base-fee dynamics in this test are surprisingly sensitive to tx density.

The retry-and-drain approach succeeds because it keeps the original force_close + 1 DA block per iter cadence intact and only patches the failure modes externally.

---

## Testing
Updates only test code

## Docs
No update

### DIFF
--- a/crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
+++ b/crates/full-node/sov-sequencer/tests/integration/preferred_end_to_end.rs
@@ -2431,10 +2431,12 @@ async fn test_gas_limit_update() {
 
     // 4. Main loop Submit a tx and save the receipt. Then, force close the batch and produce a block.
     for i in 0..20_u64 {
-        // 4.1 Submit a tx and save the receipt.
+        // 4.1 Submit a tx and save the receipt. Use the retrying client so that transient
+        //     `WaitingOnBlobSender` 503s from blob-sender backpressure are retried with backoff
+        //     instead of failing the test outright (one historical Mode-C failure mode).
         let receipt = test_rollup
             .api_client()
-            .send_raw_tx_to_sequencer(&tx_set_value_nonce::<TestRuntime<TestSpec>>(
+            .send_raw_tx_to_sequencer_with_retry(&tx_set_value_nonce::<TestRuntime<TestSpec>>(
                 &admin.private_key,
                 nonce,
                 i,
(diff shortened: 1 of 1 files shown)


bkolad (approved):

citizen-stig:
Merging this to temporarily relief CI, but this feature is going away in hardfork 1. 

Not for the rebase: this test should be removed in `theodore/multisig-upgrade`

PR #2918, opened 2026-05-26
Preferred sequencer Rate Limiter: add CIDR support

citizen-stig:
# Description

Adds CIDR / subnet support to the preferred sequencer's IP rate limiter. Buckets that were keyed by an exact `IpAddr` are now keyed by `ipnet::IpNet`, so operators can rate-limit whole networks, not just single IPs.

### 1. What & how
- **Config** — `ip_custom_limits` now accepts CIDR (`10.0.0.0/8`) as well as bare IPs. A bare IP becomes a host network (`/32` / `/128`), so **existing configs keep working** unchanged; a custom deserializer parses both forms and canonicalizes each key to its network address.
- **Resolution** — an incoming IP is charged to the **most-specific (longest-prefix) configured subnet** that contains it, falling back to its own host network on the default limits when nothing matches. The distinct configured prefix lengths are precomputed per address family at startup (`SubnetPrefixes`) and probed longest-first.
- **Overlap policy** — overlapping subnets are rejected at startup (fail-fast), so every IP maps to exactly one bucket.

### 2. Bugs fixed along the way
- **Wrong custom-limit budgets** — `to_limiter_config_map` was called with `max_requests_per_second` and the batch execution-time limit **swapped**, so every custom address/IP limit got the wrong execution-time budget and refill rate. Fixed by making the time parameter a `Duration` (so the two can't be transposed) and correcting the call.
- **Request-count budget 1000× too loose** — `req_counter` was a per-second rate × *milliseconds*, so the per-request limit was 1000× the true requests-per-batch (and 1000× looser than the byte/exec-time/gas dimensions, so it never bound). Fixed to requests-per-batch, with a floor so a non-zero rate can't round down to a zero (reject-all) budget. **Breaking (behavior).**
- **IPv4-mapped IPv6 bypass** — `::ffff:a.b.c.d` parsed as IPv6 and skipped IPv4 subnet rules; incoming IPs are now canonicalized (`to_canonical()`) before resolution.

### 3. Expected performance impact
Negligible on the admission hot path. The common case (**no IP rules configured**) resolves straight to the host network with no extra map lookups — identical to before. With subnets configured, admission does at most one HashMap probe per distinct configured prefix length (small, bounded), short-circuiting on the first match. Overlap validation is a one-time O(n²) check over configured networks at startup (n = number of IP rules, typically tiny). The richer error string allocates only on the rejection path, only when rendered, and only for subnet (not host) rejections.

## Testing
rate_limiter unit tests (longest-prefix resolution, overlap-rejection, IPv4-mapped canonicalization), config (de)serialization + snapshot tests, and the rate-limit integration tests (IPv4/IPv6 shared-subnet buckets, per-IP limits) all pass.

## Docs
The new `ip_custom_limits` behavior is documented on the `SovRateLimiterConfig` field doc-comments, and the rollup config JSON schema is regenerated.


crates/module-system/module-schemas/rollup-config.json
```diff
@@ -1154,15 +1154,15 @@
...
           "type": "array",
           "items": {
             "type": "array",
             "maxItems": 2,
             "minItems": 2,
             "prefixItems": [
               {
-                "type": "string",
-                "format": "ip"
+                "type": "string"
```
citizen-stig: investigate if we can add some hint to schemars about IP or CIDR

citizen-stig:
It is possible to write more efficient filter for incoming IP addresses, but this approach re-uses existing structs and less invasive

preston-evans98 (approved):

preston-evans98:
Codex found one minor issue: 

```
max_requests_per_second = 0 now permits one request when resources_per_bucket > 0. The new floor at crates/full-node/sov-sequencer/src/preferred/rate_limiter/
mod.rs:180 only checks resources_per_bucket, not whether the configured request rate is nonzero. Since allow checks usage before the current request is added, a computed
req_counter of 0 is raised to 1 and the first request passes. Either gate this on max_requests_per_second > 0 too, or explicitly reject max_requests_per_second == 0 at config
time.
```

PR #2924, opened 2026-05-28
Preferred Sequencer: deregister on graceful shutdown

citizen-stig:
# Description

This should minimize failed requests during regular deployment, because node-discovery service will immediately exclude stopping service from nginx

----

## Testing
Add unit test

## Docs
No updates

### DIFF
--- a/crates/utils/sov-proxy-utils/src/node_discovery.rs
+++ b/crates/utils/sov-proxy-utils/src/node_discovery.rs
@@ -230,6 +230,9 @@ impl NodeDiscovery {
             }
         }
 
+        // A `leader_id` with no matching node row is expected, not an error: a leader
+        // that gracefully shuts down deletes its `nodes` row while `sequencer_leader`
+        // still points at it. Callers treat `leader: None` as "no routable leader yet".
         Ok(ClusterInfo { leader, followers })
     }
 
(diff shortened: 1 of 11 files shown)


crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -230,12 +230,9 @@ impl NodeDiscovery {
             }
         }
 
-        if let Some(leader_id) = &leader_id {
-            if leader.is_none() {
-                anyhow::bail!("Leader is missing from the Nodes table. leader_id: {leader_id}, followers: {followers:?}");
-            }
-        }
-
+        // A `leader_id` with no matching node row is expected, not an error: a leader
```
citizen-stig, line 233: How is this acceptable and not a bug?

it is still an error from node discovery point of view, it needs to try again

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -435,10 +432,10 @@ mod tests {
             ],
         );
 
-        let err = result.unwrap_err();
-        assert!(err
-            .to_string()
-            .contains("Leader is missing from the Nodes table."));
```
citizen-stig, line 441: same about the issue

PR #2931, opened 2026-05-29
Add state related gas constant calibration

ross-weir:
# Description

Adds benchmarks that target state access/writes. Uses a differential approach like borsh to remove guest start-up noise from the prover gas value.

We're testing at a low level (NOMT funcs) to capture the remaining actual work that is done in ZK, serialization/hashing that happens as part of these operations is already capatured by other constants - so we don't double charge.

AI summary on using NOMT funcs:

```
  - What we're pricing. bias_to_charge_for_access is the cost of one Merkle path verify. verify_multi_proof is that work, and nothing else.
  - Higher levels double-count. compute_state_update bundles the verify with per-access key/leaf hashing — and that hashing is already charged separately via the hash constants in charge_read/charge_write. Measuring the bundle and attributing it to access_bias would silently double-charge users.
  - Depth control. We price at worst-case depth D = 64. A real NOMT trie produces depths ≈ log₂(N keys) and can't reach 64 without ~2⁶⁴ keys. Hand-built single-path proofs give exact depth sweep [1..256] and the clean R² = 0.99999 fit it depends on.
  - Single-path verify is data-independent. For one path it's just depth × hash_internal with no branching on sibling content. So a synthetic proof and a real-trie proof at the same depth cost identically by construction — there's no fidelity gap.
  - Matches the charging model. We deliberately charge each access as a disjoint worst-case path with no credit for batch path-sharing. Single-path verify_multi_proof is that model.
```

Bench output:

```
========== SUMMARY ==========
Raw fitted values (prover gas), cost = bias + per_sibling * depth:
  read:  bias = 3750.38, per_sibling = 2535.7829
  write: bias = 6582.74, per_sibling = 2634.7088
Cost at TARGET_DEPTH = 64:
  access (verify)        = 166040.48
  update (verify_update) = 175204.11

Suggested constants.toml values (full value charged to a single dimension, rounded ≥1):
  GAS_TO_CHARGE_PER_STORAGE_ACCESS = [166040, 0]
  BIAS_STORAGE_UPDATE              = [175204, 0]

Not proof-driven (NOMT proof cost is depth-driven, values hashed to 32 bytes; value
hashing already billed via the hash constants). Leave near-zero unless a value-size
materialization sweep says otherwise:
  GAS_TO_CHARGE_PER_READ                = [1, 0]
  GAS_TO_CHARGE_PER_BYTE_READ           = [1, 0]
  GAS_TO_CHARGE_PER_BYTE_STORAGE_UPDATE = [1, 0]
```

Here's some AI generated explanations:

Why are some values [1,0]?

```
  GAS_TO_CHARGE_PER_READ — fixed per-read bookkeeping beyond what access already covers. In the zk path that "extra" is essentially a NodeLeaf getting pulled from the witness (~40 bytes) and a small amount of control flow. The big costs of a read — proof verification, key hash, value hash — are priced by other constants (access bias, hash
  constants). What's left for this constant to pay for is genuinely tiny → [1, 0].

  GAS_TO_CHARGE_PER_BYTE_READ — pays for materializing the value byte-by-byte. In the zk path that's exactly witness.get_hint::<SlotValue>(), which is a borsh Vec<u8> decode. The borsh bench measured that at ~1.02 gas/byte (the decode-vec slope). At ~1 gas/byte, this only matches the flat 166k access cost at a value size of ~166 KB. Typical state
   values are tens to low hundreds of bytes, so this constant contributes <0.2% of the read cost in practice.

  We keep it at [1, 0] rather than [0, 0] because it's the only term that scales with value size — drop it to zero and reading a 100 MB blob costs the same as reading a 1-byte value (real witness-deserialization DoS). At 1 gas/byte, that 100 MB read costs 100M gas, which is the right shape.

  GAS_TO_CHARGE_PER_BYTE_STORAGE_UPDATE — pays for per-byte write work in the guest. The write path is cache.set(key, value) — a HashMap insert where the Vec gets moved in. There's no per-byte codec work in the guest at write time. The value's hash is computed later at slot end (charged via the hash constants). So per-byte non-hash write work ≈
  memmove ≈ ~0 gas/byte.

  Same DoS reasoning as above: [1, 0] rather than [0, 0] so a giant write isn't free, even though the actual per-byte work is essentially nothing.
```

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/bench/sp1-microbenches/guest-storage/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+name = "sov-microbench-guest-storage"
+version = "0.3.0"
+license = "MIT OR Apache-2.0"
+edition = "2021"
+resolver = "2"
+
+[workspace]
+
+[dependencies]
+sp1-zkvm = { version = "6.2.1" }
```
citizen-stig: we just bumped it 6.2.2.

crates/bench/sp1-microbenches/guest-storage/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+// Mode flags must match cmd/storage.rs.
+const MODE_READ: u8 = 0;
+const MODE_WRITE: u8 = 1;
+
+pub fn main() {
+    let mode: u8 = sp1_zkvm::io::read();
+    let depth: u32 = sp1_zkvm::io::read();
+    let iterations: u32 = sp1_zkvm::io::read();
+
+    let (proof, root, key_path) = build_single_path_proof(depth as usize);
```
citizen-stig, line 32: why do we build proof, root and keypath inside zk-vm and not passing it as param?

crates/bench/sp1-microbenches/guest-storage/Cargo.toml
ross-weir: done

crates/bench/sp1-microbenches/guest-storage/src/main.rs
ross-weir: No reason in particular

citizen-stig: Does it affect the estimation, doesn't it?

ross-weir: We're using a differential of 2 runs to attempt to eliminate as much unrelated work like this as possible, even if it was passed as a param we'd be affected by deserialization costs (and other work)

bkolad:
I think these constants are way too high. For example, in #2857 we had:

  DEFAULT_FIXED_GAS_TO_CHARGE_PER_SIGNATURE_VERIFICATION[1] ≈ 74465
  DEFAULT_GAS_TO_CHARGE_PER_BYTE_SIGNATURE_VERIFICATION[1]  ≈ 42

  With the values from this PR, state access seems more expensive than signature verification for medium-sized data. I would expect state access to be orders of magnitude cheaper than signature verification.

  I think the problem is this assumption:

  > We deliberately charge each access as a disjoint worst-case path with no credit for batch path-sharing. Single-path `verify_multi_proof` is that model.

  But in production we do the following:

  1. During execution, each set lands in the in-memory cache log. `charge_write` charges gas at this point, but no NOMT proof verification or trie-root update happens here.

  2. At checkpoint/slot materialization, `StateCheckpoint::materialize_update` freezes the whole cache log into `OrderedReadsAndWrites` and calls `compute_state_update` once.

  3. Inside `compute_state_update_namespace`, for the whole batch/namespace:
     - one `verify_multi_proof` runs over a single `MultiProof` covering all accessed keys,
     - each read is checked with `confirm_value` / `confirm_nonexistence` against that already verified proof,
     - one `verify_multi_proof_update` applies the sorted list of all writes and produces the new root.

  So the proof/root-update cost is amortized across the batch. The current benchmark measures repeated single-key proof verification and single-op proof updates.
  
 I think we should reuse the prod logic and bench the following:
  1. Build a storage microbench that feeds the guest a real batched `OrderedReadsAndWrites` plus one matching MultiProof.
  2. Bench`NomtVerifierStorage::compute_state_update_namespace`

ross-weir:
> I think the problem is this assumption:
> 
> > We deliberately charge each access as a disjoint worst-case path with no credit for batch path-sharing. Single-path `verify_multi_proof` is that model.
> 
> But in production we do the following:
> 
> 1. During execution, each set lands in the in-memory cache log. `charge_write` charges gas at this point, but no NOMT proof verification or trie-root update happens here.
> 2. At checkpoint/slot materialization, `StateCheckpoint::materialize_update` freezes the whole cache log into `OrderedReadsAndWrites` and calls `compute_state_update` once.
> 3. Inside `compute_state_update_namespace`, for the whole batch/namespace:
>    
>    * one `verify_multi_proof` runs over a single `MultiProof` covering all accessed keys,
>    * each read is checked with `confirm_value` / `confirm_nonexistence` against that already verified proof,
>    * one `verify_multi_proof_update` applies the sorted list of all writes and produces the new root.
> 
> So the proof/root-update cost is amortized across the batch. The current benchmark measures repeated single-key proof verification and single-op proof updates.

The gas is charged at execution time whereas amortization doesn't take effect until the end of the slot so we can't factor that into the equation IMO. There's no way to know ahead of time what impact it will have (or if it will even have any impact at all).

> I think we should reuse the prod logic and bench the following:
> 
> Build a storage microbench that feeds the guest a real batched OrderedReadsAndWrites plus one matching MultiProof.
> BenchNomtVerifierStorage::compute_state_update_namespace

- We will double charge for hashing, the current benches don't
- We will still have to come up with some value `K` for the batch size and `N` for the trie that is reasonable. Essentially what we're doing now but hidden with multiple levers instead of 1

> I would expect state access to be orders of magnitude cheaper than signature verification.

I'm not sure it will be orders of magnitude in ZK, but I guess it depends on the state size. In any case we can tweak the current price with `TARGET_DEPTH`

> The current benchmark measures repeated single-key proof verification and single-op proof updates.

I still think this is the best route given the above constraints, the goal is to charge the upper bounds since we can't charge precisely in this case

bkolad (approved):

PR #2941, opened 2026-06-03
 Filter advertised proxy followers by sequencer readiness

### LINKED ISSUE #2930: Prevent Proxy from Discovering Unsynced Replicas
By design, nodes are able to serve limited rpc queries during resync. However, in the high availability setup we do not wish to route requests to an unsynced node. We should prevent nodes from advertising themselves to the proxy during initial resync.

bkolad:
# Description

Adds a readiness probe to `sov-proxy-utils` node discovery so only followers returning success from `/sequencer/ready` are included in the advertised cluster view. 

Changes:
1. Probe follower readiness on every discovery poll.
2. Track raw vs advertised cluster changes separately in metrics.
3. Add integration coverage for excluding not-ready followers and re-advertising them once ready.

 The changes are backward compatible with the existing Grafana dashboards. New panels will be added once this PR is merged.

## Linked Issues
- Fixes #2930


crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -16,6 +16,9 @@ use tokio::task::JoinHandle;
 /// `poll_interval * LIVENESS_POLL_MULTIPLIER` even if the cluster is unchanged,
 /// so the absence of samples can be alerted on if the polling task dies silently.
 const LIVENESS_POLL_MULTIPLIER: u32 = 5;
+const READY_ENDPOINT_PATH: &str = "/sequencer/ready";
+const READY_REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
```
citizen-stig: In production proxy and rollup node in a single region (maybe cross DC) call. 

Non responding node can delay nginx update by 5 seconds. 

I think we can lower those.

crates/utils/sov-proxy-utils/src/node_discovery.rs
```diff
@@ -316,57 +438,121 @@ impl NodeDiscovery {
...
-            if leader_changed {
-                tracing::info!(
-                    old_leader = ?self.prev_leader_id,
-                    new_leader = ?leader_id,
-                    "Leader changed"
-                );
-            }
-
-            // Notify watchers that the cluster was updated.
+        if advertised_membership_change.has_changed() {
```
citizen-stig: From codex:

>   - [P2] Force an initial empty advertised update — /Users/nikolai/workspace/sovereign/sovereign-sdk-public/crates/utils/sov-proxy-utils/src/node_discovery.rs:465-465
    When the first poll sees only followers that are filtered out by readiness, with no advertised leader and no ready followers, advertised_membership equals the default prev_advertised_membership, so this branch is skipped. A notifier
    that materializes the advertised view is never called to clear any previous contents, so after a restart or stale output file those not-ready followers can remain advertised until a later advertised membership change.

citizen-stig (approved):

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: Changed to 3s

crates/utils/sov-proxy-utils/src/node_discovery.rs
bkolad: Fixed

Issue #2286, opened 2026-01-01
Exponential backpressure on sequencer


citizen-stig:
TBD

PR #2942, opened 2026-06-04
Add optional config to enable smooth load shedding

preston-evans98:
# Description

This PR adds a PI controller (a PID controller minus the Derivative term) to control the rate of batch growth across two dimensions - execution time and batch size. This changes the behavior of the sequencer from spiky ("accept all txs until the batch is full, then reject all txs") to smooth (reject X% of txs, updating X every 10ms based on current load and batch size). 

I also add a load shedding switch based on the node's sync distance.

All new functionality is gated behind a new config option: 
```
[sequencer.preferred]
use_pi_rate_limiter = true # defaults to false if omitted
```

## Linked Issues
- Related to #2286

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
@@ -130,17 +130,17 @@ where
 ///  - How much longer we expect the current batch to be open (based on estimated block times)
 ///
 /// Based on that data, we set probabilities for accepting or rejecting new transactions. For example,
-/// suppose that we 3 seconds in to a 6 second block time, and we've accepted 4 MB of our 6MB limit. THen the probabilty of accepting a new
+/// suppose that we 3 seconds in to a 6 second block time, and we've accepted 4 MB of our 6MB limit. Then the probability of accepting a new
 /// tx will drop to keep the batch size under control. Note that we compute probabilities for both execution time and batch size,
 /// and then we take the max rejection probability across those two dimensions.
 pub struct PIController {
     pub(crate) batch_start_time: std::time::Instant, // The time when the current batch was opened
     pub(crate) approximate_block_time: std::time::Duration,
-    // The bias term for the the batch size limiter (this lets us correct if we're repeatedly over or undershooting the target batch size)
+    // The bias term for the batch size limiter (this lets us correct if we're repeatedly over or undershooting the target batch size)
     // Expressed in bytes per second. (I.e. if the `P` term of our controller says to accept 1000 bytes per second, and this bias is 40, then we will try to accept 1040 bytes per second.)
     // Can be negative
     pub(crate) size_limit_bias: f64,
-    // The bias term for the the execution time limiter (this lets us correct if we're repeatedly over or undershooting the target execution time)
+    // The bias term for the execution time limiter (this lets us correct if we're repeatedly over or undershooting the target execution time)
     pub(crate) execution_time_limit_bias: f64,
     // The average time to execute a tx in microseconds. Computed as a EWMA over all txs since startup.
     pub(crate) estimated_tx_execution_time_micros: f64,
@@ -148,7 +148,7 @@ pub struct PIController {
     // The last time we ticked the PI controller.
     pub(crate) last_tick_time: std::time::Instant,
     pub(crate) bytes_offered_since_last_tick: u64, // How many bytes worth of tx data we would have accepted given 100% acceptance rate
-    pub(crate) bytes_offered_per_second_ewma: f64, // The weighted average of bytes offered per second
+    pub(crate) bytes_offered_per_second_ewma: f64, // The weighted average of bytes offered per second. EWMA = Exponentially Weighted Moving Average
     pub(crate) current_tx_accept_rate_bytes_per_second: f64,
     pub(crate) execution_time_offered_since_last_tick: f64,
     pub(crate) execution_time_offered_per_second_ewma: f64,
@@ -161,6 +161,7 @@ pub struct PIController {
     pub(crate) load_shed_rejection_debt: f64,
 }
 
+#[allow(clippy::float_arithmetic)]
 impl PIController {
     pub(crate) fn new(
         approximate_block_time: std::time::Duration,
@@ -837,6 +838,7 @@ where
 
     // Implements the I part of a PID controller for the batch size limit; we track our error rate over each batch and tune the bias
     // (does our controller over or under shoot the ideal rate?)
+    #[allow(clippy::float_arithmetic)]
     fn update_size_limit_bias_on_batch_close(&mut self) {
         let (target_size, target_rate_bytes_per_sec) = self.get_target_batch_size_and_growth_rate();
         let target_size = target_size as f64;
@@ -856,6 +858,8 @@ where
 
     // Implements the I part of a PID controller for the batch execution time limit. The I term is used to correct for systematic errors in our execution time limit.
     // It grows when we accept too few transactions over the batch lifetime, and shrinks when we accept too many.
+    // Float arithmetic is allowed because of the offchain nature of rate limiting.
+    #[allow(clippy::float_arithmetic)]
     fn update_execution_time_limit_bias_on_batch_close(&mut self) {
         // How many microseconds of execution time we'd like to spend on this batch.
         let (target_execution_time_micros, target_rate_micros_per_second) =
@@ -905,6 +909,7 @@ where
 
     // Returns the target batch size in bytes and the growth rate for the batch size (in bytes per second).
     // Returns a tuple (target, rate)
+    #[allow(clippy::float_arithmetic)]
     pub(crate) fn get_target_batch_size_and_growth_rate(&self) -> (u64, f64) {
         let target_size = (self.batch_size_tracker.max_batch_size as u64)
             .checked_div(20)
@@ -917,6 +922,7 @@ where
 
     // Returns the target execution time in microseconds and the growth rate for the execution time (in micros per second).
     // Returns a tuple (target, rate)
+    #[allow(clippy::float_arithmetic)]
     pub(crate) fn get_target_execution_time_and_growth_rate(&self) -> (u64, f64) {
         let target_execution_time_micros = self
             .batch_execution_time_limit_micros
@@ -928,6 +934,7 @@ where
         (target_execution_time_micros, target_rate)
     }
 
+    #[allow(clippy::float_arithmetic)]
     fn reset_current_accept_rates_on_batch_close(&mut self) {
         // Update the bytes per second target rate with the latest bias from the PI controller.
         let (_, target_rate) = self.get_target_batch_size_and_growth_rate();
@@ -941,6 +948,7 @@ where
             (target_rate_micros_per_second + self.pi_controller.execution_time_limit_bias).max(0.0);
     }
 
+    #[allow(clippy::float_arithmetic)]
     pub(crate) fn update_estimated_tx_execution_time_micros(&mut self, execution_time_micros: u64) {
         if execution_time_micros == 0 {
             return;
@@ -958,6 +966,7 @@ where
                 + execution_time_micros * 0.25;
     }
 
+    #[allow(clippy::float_arithmetic)]
     pub(crate) fn should_accept_load_shed_tx(&mut self, accept_probability: f64) -> bool {
         let accept_probability = accept_probability.clamp(0.0, 1.0);
         // If the accept probability is 1.0, we accept the tx and clear our rejection debt.
--- a/crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
@@ -57,6 +57,8 @@ const METRICS_BATCH_SIZE: usize = 32;
 /// The constant for the I part of the PID controller for the batch size limit.
 /// The larger this constant, the faster the bias will evolve, but the more likely we are to oscillate.
 const BATCH_SIZE_LIMIT_BIAS_K_I: f64 = 0.1;
+const EXECUTION_TIME_LIMIT_BIAS_POSITIVE_ERROR_K_I: f64 = 0.02; // Positive errors are slow to accumulate, because over-accepting txs early is not easily fixable
+const EXECUTION_TIME_LIMIT_BIAS_NEGATIVE_ERROR_K_I: f64 = 0.2; // Negative errors accumulate faster, because rejecting txs too aggressively is easy to fix by accepting more later.
 
 #[derive(Debug)]
 pub(crate) enum DoNewTxError<S: Spec> {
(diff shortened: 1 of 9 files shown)


citizen-stig:
I believe this could be related to https://github.com/Sovereign-Labs/sovereign-sdk/issues/2286

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -113,6 +116,75 @@ where
...
+/// switching suddenly from "accept all" to "accept none".
+///
+/// At most every 10 millis we "tick" the PI controller, updating the state with the latest data about...
+///  - The rate at which new transactions are being offered
+///  - The average time/bytes to accept a tx
+///  - The remaining available size/time limits.
+///  - How much longer we expect the current batch to be open (based on estimated block times)
+///
+/// Based on that data, we set probabilities for accepting or rejecting new transactions. For example,
+/// suppose that we 3 seconds in to a 6 second block time, and we've accepted 4 MB of our 6MB limit. THen the probabilty of accepting a new
```
citizen-stig:
```suggestion
/// suppose that we 3 seconds in to a 6 second block time, and we've accepted 4 MB of our 6MB limit. Then the probability of accepting a new
```

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -113,6 +116,75 @@ where
...
+    pub(crate) approximate_block_time: std::time::Duration,
+    // The bias term for the the batch size limiter (this lets us correct if we're repeatedly over or undershooting the target batch size)
+    // Expressed in bytes per second. (I.e. if the `P` term of our controller says to accept 1000 bytes per second, and this bias is 40, then we will try to accept 1040 bytes per second.)
+    // Can be negative
+    pub(crate) size_limit_bias: f64,
+    // The bias term for the the execution time limiter (this lets us correct if we're repeatedly over or undershooting the target execution time)
+    pub(crate) execution_time_limit_bias: f64,
+    // The average time to execute a tx in microseconds. Computed as a EWMA over all txs since startup.
+    pub(crate) estimated_tx_execution_time_micros: f64,
+    // pub(crate) has_estimated_tx_execution_time: bool, // TODO: Remove
```
citizen-stig: remove

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -113,6 +116,75 @@ where
...
+    pub(crate) batch_start_time: std::time::Instant, // The time when the current batch was opened
+    pub(crate) approximate_block_time: std::time::Duration,
+    // The bias term for the the batch size limiter (this lets us correct if we're repeatedly over or undershooting the target batch size)
+    // Expressed in bytes per second. (I.e. if the `P` term of our controller says to accept 1000 bytes per second, and this bias is 40, then we will try to accept 1040 bytes per second.)
+    // Can be negative
+    pub(crate) size_limit_bias: f64,
+    // The bias term for the the execution time limiter (this lets us correct if we're repeatedly over or undershooting the target execution time)
+    pub(crate) execution_time_limit_bias: f64,
+    // The average time to execute a tx in microseconds. Computed as a EWMA over all txs since startup.
+    pub(crate) estimated_tx_execution_time_micros: f64,
```
citizen-stig, line 148: is it in f64 for easier multiplication? I've noticed that when duration is stored as number, units sometime can be lot between passing variable to functions. So maybe storing std::time::Duration here and converting to micros where it needed can add some type safety.

Optional

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -761,6 +835,162 @@ where
...
 
+    // Implements the I part of a PID controller for the batch size limit; we track our error rate over each batch and tune the bias
+    // (does our controller over or under shoot the ideal rate?)
+    fn update_size_limit_bias_on_batch_close(&mut self) {
+        let (target_size, target_rate_bytes_per_sec) = self.get_target_batch_size_and_growth_rate();
+        let target_size = target_size as f64;
+
+        // How far we undershot our target batch size, expressed as a fraction of the target size.
+        let error_fraction =
+            (target_size - self.batch_size_tracker.current_batch_size as f64) / target_size;
```
citizen-stig, line 854: can it underflow?

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -761,6 +835,162 @@ where
...
+            return;
+        }
+        // Otherwise, use an EWMA to estimate the tx execution time.
+        self.pi_controller.estimated_tx_execution_time_micros =
+            self.pi_controller.estimated_tx_execution_time_micros * 0.75
+                + execution_time_micros * 0.25;
+    }
+
+    pub(crate) fn should_accept_load_shed_tx(&mut self, accept_probability: f64) -> bool {
+        let accept_probability = accept_probability.clamp(0.0, 1.0);
```
citizen-stig: stupid question, but why do we clamp here, if we still manually check boundaries later?

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -878,6 +879,213 @@ where
         inner.trigger_batch_production().await;
     }
 
+    // Takes the current rate limits and returns the next byte and execution-time rates, along with the offered rates.
+    // This implements the P and I parts of a PID controller; this function is responsible for both updating the P term (every tick interval) and combining
+    // it with the I term to get the current rate. (The I term is updated on each batch close.)
+    fn tick_rate_limiter(&mut self) -> RateLimitTick {
+        let minimum_tick_size = Duration::from_millis(10);
```
citizen-stig: nit: can be const

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -761,6 +835,162 @@ where
...
+    fn update_size_limit_bias_on_batch_close(&mut self) {
+        let (target_size, target_rate_bytes_per_sec) = self.get_target_batch_size_and_growth_rate();
+        let target_size = target_size as f64;
+
+        // How far we undershot our target batch size, expressed as a fraction of the target size.
+        let error_fraction =
+            (target_size - self.batch_size_tracker.current_batch_size as f64) / target_size;
+        // The next bias is old bias + (error fraction * target rate) * the "learning rate" constant "K"
+        let next_bias = self.pi_controller.size_limit_bias
+            + BATCH_SIZE_LIMIT_BIAS_K_I * error_fraction * target_rate_bytes_per_sec;
```
citizen-stig, line 857: can this overflow on big error cause by spike in tx rate?

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -935,6 +1148,15 @@ where
             .allow(ip_and_credential.ip_addr, ip_and_credential.address)
             .map_err(|err| AcceptTxError::RateLimiter(err))?;
 
+        // Shape probabilistic load shedding into a smooth accept stream. Independent random draws
+        // create accepted-transaction clusters, which is especially noisy when accepted txs are
+        // slow and rejected txs are cheap.
+        if !inner.should_accept_load_shed_tx(load_based_accept_probability) {
+            return Err(AcceptTxError::SequencerOverloaded503(
```
citizen-stig, line 1158: Found by codex:

> this early return drops the token without calling rate_limiter.update.

citizen-stig: > With a configured rate limiter and use_pi_rate_limiter=true, shed requests do not consume even the 1-request budget, so a client can flood load-shed 503s without tripping per-IP/address request limits; update the limiter with the cheap request usage before returning.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -761,6 +836,170 @@ where
...
+
+        // We use different learning rates for positive and negative errors.
+        // Positive errors (we've been accepting too few transactions) should accumulate slowly because we don't want to
+        // over-update and run out of space in the *next* batch. Running out of space is very bad - it causes complete downtime until the batch closes.
+        // Negative errors (we've been accepting too many transactions) can react faster to an
+        // over-full batch, because rejecting a bit too aggressively only causes the loss of low value txs.
+        let k_i = if error_fraction.is_sign_negative() {
+            0.2
+        } else {
+            0.02
```
citizen-stig: Minor, from claude:

> Magic numbers 0.2/0.02, −0.5/+0.1 inline while size K_I is a named const — name them.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -761,6 +836,170 @@ where
...
+            self.pi_controller.estimated_tx_execution_time_micros * 0.75
+                + execution_time_micros * 0.25;
+    }
+
+    #[allow(clippy::float_arithmetic)]
+    pub(crate) fn should_accept_load_shed_tx(&mut self, accept_probability: f64) -> bool {
+        let accept_probability = accept_probability.clamp(0.0, 1.0);
+        // If the accept probability is 1.0, we accept the tx and clear our rejection debt.
+        if accept_probability >= 1.0 {
+            self.pi_controller.load_shed_rejection_debt = 0.0;
```
citizen-stig: From claude:

> Debt reset on p≥1.0 defeats shedding in mixed-probability streams. Any p=1.0 tx wipes accumulated debt. Alternating p=0.6/p=1.0 → intended 40% shed of low-pri txs,actual 0%. Regimes: (a) Runtime::accept_tx_probability priority boosting (added in this PR), (b) load hovering at capacity (p flips 1.0/<1.0 across 10ms ticks). Degenerates to the bang-bang behavior the PR eliminates. Fix: remove the reset (debt already bounded [0,1)); optionally clear on batch close instead.

Seems legit, but I am not 100% that mixed probabilities is valid case

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -113,6 +116,76 @@ where
...
+    // we deterministically reject the next tx and reset the debt to 0.
+    pub(crate) load_shed_rejection_debt: f64,
+}
+
+#[allow(clippy::float_arithmetic)]
+impl PIController {
+    pub(crate) fn new(
+        approximate_block_time: std::time::Duration,
+        max_batch_size: usize,
+        batch_execution_time_limit_micros: u64,
```
citizen-stig, line 170: From claude:

> NaN robustness: NaN p → both guards false → debt = NaN permanently → accept-all silently. Only realistic NaN source: approximate_block_time == 0 (trait doesn't forbid; Mock=200ms, Celestia=6s today) → div-by-zero in PIController::new → inf rates → inf - inf = NaN in slew_limited_rate → f64::clamp panics on NaN bounds. Fix: assert non-zero block time in PIController::new.

I think we can eliminate bunch of this if we wrap it into NonZero<u64>

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
```diff
@@ -761,6 +836,170 @@ where
...
+            self.pi_controller.load_shed_rejection_debt -= 1.0;
+            false
+        } else {
+            true
+        }
+    }
+
+    // Updates the PI controller based on the current batch size and execution time.
+    fn update_pi_controller_on_batch_close(&mut self) {
+        self.update_size_limit_bias_on_batch_close();
```
citizen-stig, line 1005: From Devin:

> crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs:R988-992

**load_shed_rejection_debt is never reset across batch boundaries**

The `load_shed_rejection_debt` field in `PIController` (`inner.rs:161`) accumulates across batches — `update_pi_controller_on_batch_close` (`inner.rs:988-992`) calls `update_size_limit_bias_on_batch_close`, `update_execution_time_limit_bias_on_batch_close`, and `reset_current_accept_rates_on_batch_close`, but never resets `load_shed_rejection_debt`. This means if the sequencer accumulates rejection debt at the end of batch N (e.g. debt = 0.9), it will carry that into batch N+1 and may immediately reject the first transaction of the new batch even though load conditions have changed. This may be intentional for smoothness, but could cause surprising behavior at batch boundaries.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
preston-evans98: This is an f64 because it's a moving average, and Duration doesn't allow the math we need for that.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
preston-evans98: Nope, this is an f64 so there's no underflow. And batch sizes are capped well within range, so there's no worry of significant precision loss

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
preston-evans98: Good callout - I think that was an overly-cautious codex addition. Removing!

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
preston-evans98: No, the error fraction can't be larger than 1 in absolute value.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
preston-evans98: I think this is the correct semantic. If you get load shed before we spend any resources (not your fault) that shouldn't count against your request budget

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
preston-evans98: This is intentional for smoothness - although we can adjust later if we don't like the behavior in practice

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
preston-evans98: I added an assertion to protect against this.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
preston-evans98: Hmm, this is a good point. I don't think mixed probabilities are impossible. If we treat `1.0` as "temporarily exempt from load shedding", then modifying the rate does seem wrong - and of course the rate will only kick in once txs stop having a 100% acceptance probability. I'll get rid of this.

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs
preston-evans98: Fixed

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
preston-evans98: done

citizen-stig (approved):

PR #2944, opened 2026-06-04
Enrich hyperlane metrics

ross-weir:
# Description

Makes the following fields available for hyperlane rate limit metrics:

- Human readable route name
- Token values in decimal scaled form rather than raw

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/module-system/hyperlane/src/warp/execution_config.rs
+++ b/crates/module-system/hyperlane/src/warp/execution_config.rs
@@ -13,18 +13,28 @@ pub static WARP_EXECUTION_CONFIG: OnceLock<WarpExecutionConfig> = OnceLock::new(
 /// metrics emitted during block processing.
 #[derive(Clone, Default, Debug, serde::Serialize, serde::Deserialize, Eq, PartialEq)]
 pub struct WarpExecutionConfig {
-    /// List of warp route IDs to monitor for metrics.
+    /// Warp routes to monitor for rate limiter metrics.
     ///
     /// Metrics will be emitted for the rate limiters of these routes at the end
-    /// of each block.
+    /// of each block, labelled with the operator-supplied name.
     #[serde(default)]
-    pub monitored_route_ids: Vec<WarpRouteId>,
+    pub monitored_routes: Vec<MonitoredRoute>,
+}
+
+/// A warp route to emit rate limiter metrics for, paired with a human-readable name.
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize, Eq, PartialEq)]
+pub struct MonitoredRoute {
+    /// The id of the route to monitor.
+    pub id: WarpRouteId,
+    /// Human-readable name used to label the route's metrics.
+    pub name: String,
 }
 
 impl<S: Spec> ExecutionInit for Warp<S> {
     type Config = WarpExecutionConfig;
 
     fn init(config: &Self::Config) -> Result<(), Box<dyn std::error::Error>> {
+        tracing::debug!(?config, "Initializing Warp execution config");
         WARP_EXECUTION_CONFIG
             .set(config.clone())
             .map_err(|_| "Warp execution config already initialized")?;
@@ -33,14 +43,14 @@ impl<S: Spec> ExecutionInit for Warp<S> {
 }
 
 impl<S: Spec> Warp<S> {
-    /// Returns the list of warp route IDs that should be monitored for metrics.
+    /// Returns the warp routes that should be monitored for metrics.
     ///
     /// This reads from the global execution configuration set at startup.
-    /// If no configuration is set, returns an empty vector.
-    pub(super) fn get_monitored_route_ids(&self) -> &[WarpRouteId] {
+    /// If no configuration is set, returns an empty slice.
+    pub(super) fn get_monitored_routes(&self) -> &[MonitoredRoute] {
         WARP_EXECUTION_CONFIG
             .get()
-            .map(|config| config.monitored_route_ids.as_slice())
+            .map(|config| config.monitored_routes.as_slice())
             .unwrap_or_default()
     }
 }
--- a/crates/module-system/hyperlane/src/warp/execution_config.rs
+++ b/crates/module-system/hyperlane/src/warp/execution_config.rs
@@ -12,6 +12,7 @@ pub static WARP_EXECUTION_CONFIG: OnceLock<WarpExecutionConfig> = OnceLock::new(
 /// This configuration specifies which warp routes should have their rate limiter
 /// metrics emitted during block processing.
 #[derive(Clone, Default, Debug, serde::Serialize, serde::Deserialize, Eq, PartialEq)]
+#[serde(deny_unknown_fields)]
 pub struct WarpExecutionConfig {
     /// Warp routes to monitor for rate limiter metrics.
     ///
--- a/crates/module-system/hyperlane/src/warp/hooks.rs
+++ b/crates/module-system/hyperlane/src/warp/hooks.rs
@@ -12,12 +12,14 @@ impl<S: Spec> BlockHooks for Warp<S> {
 
 impl<S: Spec> Warp<S> {
     fn emit_rate_limiter_metrics(&self, state: &mut StateCheckpoint<S>) {
-        let route_ids = self.get_monitored_route_ids();
+        let monitored_routes = self.get_monitored_routes();
         let visible_slot = state.current_visible_slot_number();
 
-        for route_id in route_ids {
-            if let Ok(Some(route)) = self.warp_routes.get(route_id, state) {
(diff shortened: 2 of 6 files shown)


crates/module-system/hyperlane/src/warp/metrics.rs
```diff
@@ -47,20 +51,88 @@ impl sov_metrics::Metric for RateLimiterCapacityMetrics {
...
+            "{},direction={},route_id={},remote_domain={},route_name={}",
             self.measurement_name(),
             self.direction,
             self.route_id,
-            self.remote_domain
+            self.remote_domain,
+            escape_tag(&self.route_name),
         )?;
 
-        // fields
```
citizen-stig, line 57: nit: I think we still can keep this comment

crates/module-system/hyperlane/src/warp/metrics.rs
```diff
@@ -47,20 +51,88 @@ impl sov_metrics::Metric for RateLimiterCapacityMetrics {
...
+#[allow(
+    clippy::float_arithmetic,
+    reason = "lossy scaling is fine for a metric gauge"
+)]
+fn scale(amount: Amount, decimals: u8) -> f64 {
+    amount.0 as f64 / 10f64.powi(i32::from(decimals))
+}
+
+/// escapes a string for use as an InfluxDB line-protocol tag value.
+fn escape_tag(value: &str) -> String {
```
citizen-stig: I think there's already a function that does exactly this: `sov_metrics::safe_telegraf_string`, but it needs to be improved to add backslash support there.

crates/module-system/hyperlane/src/warp/execution_config.rs
```diff
@@ -13,18 +13,28 @@ pub static WARP_EXECUTION_CONFIG: OnceLock<WarpExecutionConfig> = OnceLock::new(
 /// metrics emitted during block processing.
 #[derive(Clone, Default, Debug, serde::Serialize, serde::Deserialize, Eq, PartialEq)]
```
citizen-stig, line 14: Can we add `#[serde(deny_unknown_fields)]` so users won't forget to update?

crates/module-system/hyperlane/src/warp/types.rs
```diff
@@ -180,6 +180,16 @@ impl StoredTokenKind {
...
         })?))
     }
+
+    /// The number of decimals of this route's local token.
+    pub fn local_decimals(&self) -> u8 {
+        match self {
+            StoredTokenKind::Synthetic { local_decimals, .. } => *local_decimals,
+            // Bank derives every token id with its decimals encoded in the final byte.
+            StoredTokenKind::Collateral { token } => token.as_bytes()[31],
+            StoredTokenKind::Native => sov_bank::config_gas_token_id().as_bytes()[31],
```
citizen-stig, line 190: This is from codex, but I think it is false positive, since we alwways require for gas token to have decimals to be in the last byte:

>   - [P2] Avoid inferring bank token decimals from IDs — /Users/nikolai/workspace/sovereign/sovereign-sdk-public/crates/module-system/hyperlane/src/warp/types.rs:188-190 This assumes every bank token ID carries its decimal count, but genesis/native tokens can be configured with predetermined IDs. For example the default GAS_TOKEN_ID in constants.toml:93 decodes with byte 31 = 89 while the demo/integration bank configs set token_decimals to 6, so a monitored Native route will divide capacities by 1e89 instead of 1e6 and report near-zero gauges. Please source the decimals from configuration/state or keep raw units when the ID  does not encode them.

citizen-stig: Surprisingly Claude also got concerned about it:

```
  🔴 Critical: local_decimals() returns garbage for Native and genesis-token Collateral routes

  types.rs:185-191 assumes "Bank derives every token id with its decimals encoded in the final byte." That invariant only holds for runtime-created tokens (sov-bank/src/utils.rs:25 — hash[31] = token_decimals). It does not hold for
  genesis tokens: TokenConfig.token_id is a "Predetermined ID … allowed only for genesis tokens" (sov-bank/src/genesis.rs:40-41), and init_module writes it with no validation of the last byte.

  The gas token is always a genesis token, and I verified the concrete failure in this repo:

  - constants.toml:93 → GAS_TOKEN_ID = token_1nyl0e... — last byte decodes to 89 (decoder validated against genesis.rs:236-239, where the token derived with Some(6) decimals decodes to 6)
  - The demo genesis configs set the gas token to 6 decimals (examples/test-data/genesis/demo/mock/bank.json)

  So for any Native route in this repo's default configuration, scale() divides by 10⁸⁹ instead of 10⁶, and all three gauges flatline at ~1e-80 ≈ 0. Same failure applies to Collateral routes whose token was created at genesis with a
  non-derived ID. Only Synthetic (which stores local_decimals in state) is reliable.

  Fix: Bank's Token state stores no decimals field, so there is no on-chain source of truth to fall back on. The simplest robust fix is to put decimals: u8 in MonitoredRoute — the operator is already supplying name per route, and this is
  an offchain-only, non-deterministic config (ExecutionInit docs explicitly permit per-instance config). That also deletes local_decimals() and its fragile invariant entirely.

  If you keep local_decimals(): scope it pub(super) like get_monitored_routes (it's only called from hooks.rs, and a pub method encoding a sometimes-false invariant on a widely-used type will get misused), and document the genesis-token
  caveat.
```

crates/module-system/hyperlane/src/warp/hooks.rs
```diff
@@ -12,12 +12,14 @@ impl<S: Spec> BlockHooks for Warp<S> {
...
     fn emit_rate_limiter_metrics(&self, state: &mut StateCheckpoint<S>) {
-        let route_ids = self.get_monitored_route_ids();
+        let monitored_routes = self.get_monitored_routes();
         let visible_slot = state.current_visible_slot_number();
 
-        for route_id in route_ids {
-            if let Ok(Some(route)) = self.warp_routes.get(route_id, state) {
-                let route_id = *route_id;
+        for monitored_route in monitored_routes {
+            if let Ok(Some(route)) = self.warp_routes.get(&monitored_route.id, state) {
```
citizen-stig: I think we follow hyperlane engineering guideline here.

So if warp routes return Ok(None) (misconfigured Id) or even Err, we silenltly skip it. 

Could we add info or debug log on the else branch? Or probably use match and have different lines on Ok(None) and err

crates/module-system/hyperlane/src/warp/execution_config.rs
```diff
@@ -13,18 +13,28 @@ pub static WARP_EXECUTION_CONFIG: OnceLock<WarpExecutionConfig> = OnceLock::new(
...
 pub struct WarpExecutionConfig {
-    /// List of warp route IDs to monitor for metrics.
+    /// Warp routes to monitor for rate limiter metrics.
     ///
     /// Metrics will be emitted for the rate limiters of these routes at the end
-    /// of each block.
+    /// of each block, labelled with the operator-supplied name.
     #[serde(default)]
-    pub monitored_route_ids: Vec<WarpRouteId>,
```
citizen-stig, line 21: If this is used only by Bullet and they aware about the change, then it is ok to skip changelog. Otherwise it is worth mentioning change in the module config setup.

crates/module-system/hyperlane/src/warp/types.rs
ross-weir: yeah I think this is legit, I've ran into it before a while ago but can't remember the outcome

I'll raise a post at standup

crates/module-system/hyperlane/src/warp/metrics.rs
ross-weir: readded

crates/module-system/hyperlane/src/warp/metrics.rs
ross-weir: nice, didnt know about this one

Looking at influx docs it seems backslash isn't need escaping for tag values & we handle escaping backslashes in field values already with `write_escaped_field_value`

crates/module-system/hyperlane/src/warp/execution_config.rs
ross-weir: done

crates/module-system/hyperlane/src/warp/hooks.rs
ross-weir: added a log

crates/module-system/hyperlane/src/warp/execution_config.rs
ross-weir: yeah only bullet and they're aware

citizen-stig (approved):
Approved, I hope we find a robust solution to `local_decimals`

PR #2945, opened 2026-06-04
Sanity check: preferred sequencer refuses startup if blobsender DB has a higher sequencer number than sequencer DB

theodorebugnet:
# Description

Iterates over PrefferedBlobSender blobs on startup, gets the max sequence number and ensures that it's not higher than the highest sequence number from the preferred sequencer DB. If it is, that means the sequencer doesn't know about some blobs that will be sent to DA, which is an inconsistent state and the sequencer cannot operate.

This prevents a footgun in case the preferred sequencer DB is wiped. Now, the node will immediately shut down on startup, with an informative error message to users ("if you wiped the sequencer DB, then delete the blobsender DB as well").

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/initialization.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/initialization.rs
@@ -112,6 +112,20 @@ where
         )
         .await?;
 
+        if let Some(blob_sender_sequence_number) =
+            blob_sender.highest_sequence_number_to_send_after_restart()?
+        {
+            if blob_sender_sequence_number >= next_sequence_number {
+                let _ = shutdown_sender.send(());
+                let preferred_db_highest_sequence_number = next_sequence_number
+                    .checked_sub(1)
+                    .map_or_else(|| "none".to_owned(), |seq| seq.to_string());
+                anyhow::bail!(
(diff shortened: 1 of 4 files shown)


crates/full-node/sov-sequencer/src/preferred/initialization.rs
```diff
@@ -111,6 +111,20 @@ where
...
+        if let Some(blob_sender_sequence_number) =
+            blob_sender.highest_sequence_number_to_send_after_restart()?
+        {
+            if blob_sender_sequence_number >= next_sequence_number {
+                let _ = shutdown_sender.send(());
+                let preferred_db_highest_sequence_number = next_sequence_number
+                    .checked_sub(1)
+                    .map_or_else(|| "none".to_owned(), |seq| seq.to_string());
+                anyhow::bail!(
+                    "Node state is inconsistent, aborting startup. The BlobSender DB contains a preferred blob with sequence number {blob_sender_sequence_number}, but the preferred sequencer DB's highest known sequence number is {preferred_db_highest_sequence_number}. This could mean the preferred sequencer DB was wiped; this is not supported, but to proceed, the BlobSender DB must also be deleted in the rollup state directory. Otherwise, this as a bug, please report it."
```
citizen-stig: I think it might be worth stating explicitly that it is bad that blob sender has **higher** sequencer number

citizen-stig (approved):

crates/full-node/sov-sequencer/src/preferred/initialization.rs
theodorebugnet: Great point

PR #2946, opened 2026-06-05
Mutable sequencing data

theodorebugnet:
# Description

Adds machinery to enable `sequencing_data` to be mutated post-execution by the sequencer, and uses it to implement a proof-of-concept for oracle data mutation in an integration test which serves as an example of how this can actually be used by e.g. Relay.

The mutation is implemented in a capability trait method that gives a lot of freedom to rollup developers. This can let them break their own consensus, but the pitfalls are documented in a doc comment. I don't think there's a convenient way to statically prevent footguns here, so it's left as a well-documented but dangerous method. For our own usecase this is not a problem.

**Note to reviewers**: 700 lines are in the PoC integration test, of which a huge chunk is boilerplate overriting `HasCapabilities`. The ergonomics of this could be improved in a future PR.

**Future improvements**
A couple of aspects could be made more complex in the future if needed:
* Size tracking: the oracle data could be pretty large. Right now this is recorded in the batch size tracker pessimistically, and then updated after pruning. If this becomes a concern, the actual implementation could specify a max final size (i.e.: max oracle reads in a transaction) that's smaller than the total oracle snapshot, use that for the initial estimate, and fail transactions that would read more items than this limit.
* Intelligent (de)serialization: a naive oracle implementation requires serializing and deserializing the entire `sequencing_data` snapshot at every transaction and every precompile call. For a key-value map of oracle values, a custom serialization method could be implemented that does zero-copy accesses and can extract values for specific keys, keeping the entire oracle snapshot as serialized `Bytes` and avoiding the roundtrip full map ser/de.

I'll convert these to issues if we proceed with this PR and merge it into the SDK

## Linked Issues
- Related to # (issue) 

## Testing
This
## Docs

### DIFF
--- a/crates/full-node/sov-ethereum/Cargo.toml
+++ b/crates/full-node/sov-ethereum/Cargo.toml
@@ -49,10 +49,10 @@ proptest = { workspace = true }
 strum = { workspace = true }
 test-strategy = { workspace = true }
 
-# Enable the `local` feature on this crate for the integration test target so the
-# eth_accounts / eth_sendTransaction handlers are registered and the eth_signer field
-# of EthRpcConfig exists.
-sov-ethereum = { path = ".", features = ["local"] }
+# Enable test-only cfgs on this crate for the integration test target.
+# `local` registers eth_accounts / eth_sendTransaction handlers; `native` lets tests
+# exercise native-only paths explicitly.
+sov-ethereum = { path = ".", features = ["local", "native"] }
 
 sov-mock-da = { workspace = true, features = ["native"] }
 sov-rollup-interface = { workspace = true }
@@ -63,6 +63,7 @@ sov-state = { workspace = true }
 sov-kernels = { workspace = true, features = ["native"] }
 sov-accounts = { workspace = true }
 sov-paymaster = { workspace = true, features = ["native"] }
+sov-blob-storage = { workspace = true }
 sov-test-utils = { workspace = true }
 
 alloy = { workspace = true, features = ["signers", "transport-http"] }
@@ -85,3 +86,4 @@ default = []
 local = [
 	"sov-ethereum/local"
 ]
+native = []
--- a/crates/full-node/sov-ethereum/Cargo.toml
+++ b/crates/full-node/sov-ethereum/Cargo.toml
@@ -86,4 +86,20 @@ default = []
 local = [
 	"sov-ethereum/local"
 ]
-native = []
+native = [
+	"sov-accounts/native",
+	"sov-address/native",
+	"sov-bank/native",
+	"sov-blob-storage/native",
+	"sov-ethereum/native",
+	"sov-evm/native",
+	"sov-kernels/native",
+	"sov-metrics/native",
+	"sov-mock-da/native",
+	"sov-modules-api/native",
+	"sov-modules-stf-blueprint/native",
+	"sov-paymaster/native",
+	"sov-rollup-interface/native",
+	"sov-state/native",
+	"sov-uniqueness/native"
+]
--- a/crates/full-node/sov-ethereum/tests/integration/evm/evm_sequencing_data_pruning.rs
+++ b/crates/full-node/sov-ethereum/tests/integration/evm/evm_sequencing_data_pruning.rs
@@ -1,0 +1,706 @@
+use std::collections::BTreeMap;
+use std::convert::Infallible;
+use std::marker::PhantomData;
+use std::str::FromStr;
+use std::sync::{Arc, LazyLock, Mutex};
+use std::time::Duration;
+
+use alloy_primitives::{Address, Bytes, TxHash, U256};
+use borsh::{BorshDeserialize, BorshSerialize};
+use sov_address::{EthereumAddress, FromVmAddress, MultiAddress};
+use sov_blob_storage::PreferredBatchData;
+use sov_evm::precompiles::{
+    EvmPrecompile, EvmPrecompileEnv, PrecompileError, PrecompileOutput, PrecompileResult,
+};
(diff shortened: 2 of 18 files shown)


crates/full-node/sov-ethereum/Cargo.toml
```diff
@@ -85,3 +86,20 @@ default = []
 local = [
 	"sov-ethereum/local"
 ]
+native = [
```
theodorebugnet, line 89: This is necessary because we add a `"native"`-gated section to the test precompile, for the PoC pruning precompile in the integration test. These dependencies are already brought in as `"native"` because sov-ethereum is itself native-only, but zepter still complains.

We could simplify this away; the test precompile would technically not be correct, but because we don't integrate ZK-proving into the test that doesn't matter. I'm not sure if adding this noise to Cargo.toml is worthwhile, but I've left it for now just so the PoC implementation is as realistic as possible and can serve as a template for building the real oracle integration.

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
```diff
@@ -658,20 +678,35 @@ impl<S: Spec, Rt: Runtime<S>> RollupBlockExecutor<S, Rt> {
...
+            return;
+        };
+        let Ok(decoded) = <Rt as HasCapabilities<S>>::SequencingData::try_from_slice(data) else {
+            tracing::warn!("Failed to deserialize sequencing data while finalizing transaction");
+            return;
+        };
+
+        let mut runtime = Rt::default();
+        let mut handler = runtime.sequencing_data_handler();
+        let finalized = handler.finalize_sequencing_data(decoded, scratchpad);
```
bkolad, line 706: Should we reject/assert that finalized data is smaller that the `sequencing_data`?

crates/full-node/sov-ethereum/tests/integration/evm/evm_sequencing_data_pruning.rs
```diff
@@ -0,0 +1,706 @@
...
+        let value = sequencing_data
+            .0
+            .get(&key)
+            .ok_or_else(|| PrecompileError::InvalidInput(format!("missing oracle key {key}")))?;
+
+        #[cfg(feature = "native")]
+        env.sov_context
+            .expect("sov context was checked above")
+            .sequencing_scratchpad()
+            .set(
```
bkolad: My understanding is the following:
The scratchpad should hold a serialized `OracleSequencingData`, instead of  a single key? And we should keep appending all the read keys to OracleSequencingData instead of overrding them with `set` method.
Is it correct? In that case let's update the test.

bkolad (approved):

crates/module-system/sov-modules-api/src/runtime/capabilities/sequencing_data.rs
```diff
@@ -21,6 +21,34 @@ pub trait SequencingDataHandler<S: Spec> {
     /// Generates the sequencing data for a new transaction.
     #[cfg(feature = "native")]
     fn create_sequencing_data(&self) -> Self::SequencingData;
+
+    /// Finalizes sequencing data after native transaction execution.
+    ///
+    /// WARNING: this allows the sequencer to modify concensus data. If used carelessly, this can
```
bkolad, line 27: If we narrow `finalize_sequencing_data` functinality so that it is only responsible for shrinking data, then we could record all the keys automatically. The framework would accumulate the access set and prune to exactly that set. This would limit the functionality, but it would also remove the footgun. Is that acceptable?

crates/full-node/sov-sequencer/src/preferred/block_executor.rs
theodorebugnet: We could but the main question is, what happens if it is larger?
* Do we reject the transaction? But we just executed it all the way, so seems like a big waste. It's a bug in the sequencer so we can't just charge the user gas for it.
* Do we error/panic and shutdown? That might make it easier to debug the sequencer, but also risks bringing down the rollup and causing downtime on a bug that's usually not fatal.

If you think either of these (or something else) is a good way to handle the condition, then I'm happy to add this check. But for now, since we will be the only ones implementing it, I just went with putting responsibility on the dev to make sure it's done correctly

crates/module-system/sov-modules-api/src/runtime/capabilities/sequencing_data.rs
theodorebugnet: We could, but that would require hardcoding the `SequencingData` type to something that can record accesses and get pruned - basically restricting it to some sort of map. Right now it's fully generic so we can't do that.

For example existing HDTimestamp implementations would need to be reworked into a pseudo-"map" with a single key and value, and access tracking logic as well.

Hmm, if we do require this in a trait, it would make pruning safer. Let me think about this

crates/full-node/sov-ethereum/tests/integration/evm/evm_sequencing_data_pruning.rs
theodorebugnet: Good catch actually, if the precompile is called multiple times it should append to the scratchpad, not `set` it, thanks Blazej!

crates/module-system/sov-modules-api/src/runtime/capabilities/sequencing_data.rs
theodorebugnet: Thinking about this further, this will break the existing `HDTimestamp` format. So I think we can implement this in the hard fork if we need to, but not in this PR.

bkolad (approved):

PR #2949, opened 2026-06-05
Add cheap user state root sequencer-node consistency check, even when expensive checks are disabled

theodorebugnet:
# Description

Adds an unconditional check of the sequencer user root against the node user root in `process_new_storage()`.

This does require a RocksDB lookup for the node root. We could avoid this by adding the node state root to the `StateUpdateInfo`, since I think the node has it available in-memory and just doesn't send this to the sequencer; I don't know whether the extra complexity in changing the type signature here is worth the tradeoff, so I've left it simple for now.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
@@ -33,9 +33,9 @@ use sov_blob_sender::{new_blob_id, BlobInternalId};
 use sov_blob_storage::SequenceNumber;
 use sov_modules_api::capabilities::{RollupHeight, SequencingDataHandler};
 use sov_modules_api::state::{ApiStateAccessor, ConcurrentStateCheckpoint};
-use sov_modules_api::{FullyBakedTx, Runtime, Spec, StateCheckpoint, VersionReader};
+use sov_modules_api::{FullyBakedTx, HexString, Runtime, Spec, StateCheckpoint, VersionReader};
 use sov_rollup_full_node_interface::StateUpdateInfo;
-use sov_state::Storage;
+use sov_state::{NativeStorage, ProvableNamespace, StateRoot, Storage};
 use std::collections::BTreeMap;
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
```diff
@@ -685,6 +687,51 @@ where
...
+    ) {
+        let Some(sequencer_root) = inner.executor.state_roots.get(&new_rollup_height) else {
+            debug!(
+                %new_rollup_height,
+                "Skipping sequencer/node state root sanity check; no cached sequencer root is available"
+            );
+            return;
+        };
+
+        let Some(node_root) = info.storage.get_root_hash(info.slot_number) else {
```
citizen-stig, line 703: Do I understand correctly that user space root hash on `slot_number` should be always equal on the same `rollup_height`? I thought those are different numbers

citizen-stig (approved):

crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/sync_state.rs
preston-evans98: It looks to me like what this is doing is...
- Getting the sequencer state root from the latest rollup height `R` processed by the node
- Getting the node state from the latest slot number `S` processed by the node (i.e. the slot corresponding to `new_rollup_height`)

That's correct, as long as we compare only the user state roots.

Proof sketch below:
- Case 1: The *slot* S that the node just processed contained a rollup block. That block will have height `R`, so comparing the user roots `S` and `R` is safe
- Case 2: The *slot* S that the node just processed did *not* contain a rollup block. In that case, rollup block `R` appeared in some earlier slot `S-K`.  Because we `R` is the latest rollup height, we know that all of `(S-(K-1))`, `(S-(K-2))`, ..., `S` are empty. That means that the *user* space root from `S` will still match the user root from `R`.

citizen-stig: Yep, that seams correct. 

If I understand correctly, this check might fail if preferred sequencer failed behind defferred slots count and there are batches from standard sequencer.

PR #2960, opened 2026-06-10
Add rollup height websocket subscription

citizen-stig:
# Description

So it can be used by sov-rollup-manager for more precise detection of the height instead of polling

Adds shutdown receiver watch channel to API state.

## Follow up

Potential optimization can be done for waking up subscription only when rollup height changed, instead of on every accept tx

Current path:

- The WS stream waits on checkpoint_rx.changed() in:
  https://github.com/Sovereign-Labs/sovereign-sdk/blob/19b33b6a222e68a715acc9e3ce73fe6477a8d299/crates/module-system/module-implementations/sov-chain-state/src/query.rs#L82-L84
- Preferred sequencer sends checkpoint notifications for accepted tx changes even when rollup height is unchanged:
  https://github.com/Sovereign-Labs/sovereign-sdk/blob/26e6a0bbf2779488c3b0f65485b74da303f6b152/crates/full-node/sov-sequencer/src/preferred/side_effects.rs#L195-L200
- The subscription then borrows the checkpoint, sees the same height, and emits nothing.

So with N connected height subscribers and U accepted-tx checkpoint notifications/sec, you can get about N * U useless async wakeups/sec. For a few clients this is noise. For many dashboards/indexers, or a busy preferred sequencer, this becomes avoidable executor churn on the same node that is handling tx intake.

----

## Linked Issues
- Related to https://github.com/Sovereign-Labs/sovereign-sdk-wip/issues/2341

## Testing
New tests are added

## Docs
Docstrings are provided

### DIFF
--- a/crates/module-system/module-implementations/sov-chain-state/src/query.rs
+++ b/crates/module-system/module-implementations/sov-chain-state/src/query.rs
@@ -55,15 +55,6 @@ impl<S: Spec> ChainState<S> {
         use futures::StreamExt;
 
         ws.on_upgrade(move |socket| async move {
-            // No shutdown receiver is plumbed through to module custom REST APIs, so we
-            // keep a local sender alive for the lifetime of the connection: the helper's
-            // shutdown arm then never fires, and the task instead terminates on client
-            // disconnect, ping timeout, or stream end. Stream end covers node shutdown:
-            // dropping the checkpoint sender makes `changed()` return `Err`, ending the
-            // stream.
-            let (_shutdown_tx, shutdown_rx) =
-                sov_modules_api::prelude::tokio::sync::watch::channel(());
-
             let stream = futures::stream::unfold(
                 (
                     state.checkpoint_receiver(),
@@ -105,7 +96,7 @@ impl<S: Spec> ChainState<S> {
             sov_modules_api::rest::utils::serve_generic_ws_subscription(
                 socket,
                 stream,
-                shutdown_rx,
+                state.shutdown_receiver(),
             )
             .await;
         })
--- a/crates/module-system/module-implementations/sov-chain-state/src/query.rs
+++ b/crates/module-system/module-implementations/sov-chain-state/src/query.rs
@@ -43,9 +43,10 @@ impl<S: Spec> ChainState<S> {
     }
 
     /// WebSocket handler that streams the current [`RollupHeight`] to the client: the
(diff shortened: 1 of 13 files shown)


crates/module-system/module-implementations/sov-chain-state/src/query.rs
```diff
@@ -41,6 +41,86 @@ impl<S: Spec> ChainState<S> {
...
+                        loop {
+                            // Skip the wait on the first iteration so the client receives
+                            // the current height immediately on connection.
+                            if started && checkpoint_rx.changed().await.is_err() {
+                                return None;
+                            }
+
+                            let mut accessor = match state.build_api_state_accessor(None) {
+                                Ok(accessor) => accessor,
+                                Err(_) => return None,
```
citizen-stig: log it here?

crates/module-system/module-implementations/sov-chain-state/src/query.rs
```diff
@@ -41,6 +41,78 @@ impl<S: Spec> ChainState<S> {
...
+    /// checkpoint updates that don't advance the height are ignored, and a height jump
+    /// is streamed one value at a time.
+    #[cfg(feature = "native")]
+    async fn subscribe_rollup_height(
+        state: ApiState<S, Self>,
+        ws: axum::extract::ws::WebSocketUpgrade,
+    ) -> axum::response::Response {
+        use futures::StreamExt;
+
+        ws.on_upgrade(move |socket| async move {
```
bkolad, line 57: If a client keeps that websocket open and continues responding to ping/pong, the rollup may not exit cleanly because the HTTP server’s graceful shutdown waits for active upgraded websocket tasks to finish. Since this handler passes a local shutdown
  receiver that never fires, that websocket task has no direct signal to stop.

  It may still exit if the client disconnects or if the checkpoint sender is dropped in the right order. The bug is that clean rollup shutdown is no longer guaranteed while that subscription is open.

bkolad (approved):

crates/module-system/module-implementations/sov-chain-state/src/query.rs
citizen-stig: I've wired in proper shutdown receiver into API state

citizen-stig: But probably demo-rollup deserves a test for that

PR #2971, opened 2026-06-12
Support in-place nomt bucket growth

theodorebugnet:
# Description

Updates to the upstream NOMT commit that added support to `grow_hashtable()`. Adds rollup support to grow the on-disk NOMT DB in-place if the config is changed.

Since this is non-reversible, gate the operation behind an explicit CLI flag to avoid accidental config changes from blowing up disk usage.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -213,10 +213,8 @@ reltester = { git = "https://github.com/sovereign-labs/reltester.git", rev = "d6
 rockbound = { git = "https://github.com/sovereign-labs/rockbound", rev = "0aa43d4e581e047ba8d8e7a662ec6ade32b2cae3" }
 nmt-rs = { version = "0.2.5", features = ["serde", "borsh"] }
 # External dependencies
-# nomt-core = { version = "1.0.4", default-features = false, features = ["borsh", "serde"] }
-# nomt = { version = "1.0.4", default-features = false, features = ["borsh", "serde"] }
-nomt-core = { git = "https://github.com/thrumdev/nomt.git", rev = "cf1672b2bb9e89e07f6d939f6b7e404417c3120a", features = ["borsh", "serde"] }
-nomt = { git = "https://github.com/thrumdev/nomt.git", rev = "cf1672b2bb9e89e07f6d939f6b7e404417c3120a", features = ["borsh", "serde"] }
+nomt-core = { version = "1.0.5", default-features = false, features = ["borsh", "serde"] }
+nomt = { version = "1.0.5", default-features = false, features = ["borsh", "serde"] }
 concread = { version = "0.5.7" }
 
 anyhow = { version = "1.0.95" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -213,8 +213,10 @@ reltester = { git = "https://github.com/sovereign-labs/reltester.git", rev = "d6
 rockbound = { git = "https://github.com/sovereign-labs/rockbound", rev = "0aa43d4e581e047ba8d8e7a662ec6ade32b2cae3" }
 nmt-rs = { version = "0.2.5", features = ["serde", "borsh"] }
 # External dependencies
-nomt-core = { version = "1.0.4", default-features = false, features = ["borsh", "serde"] }
-nomt = { version = "1.0.4", default-features = false, features = ["borsh", "serde"] }
+# nomt-core = { version = "1.0.4", default-features = false, features = ["borsh", "serde"] }
+# nomt = { version = "1.0.4", default-features = false, features = ["borsh", "serde"] }
+nomt-core = { git = "https://github.com/Sovereign-Labs/nomt", rev = "c85fb01f526b9eee742bafbdc89c4a58313b249b", features = ["borsh", "serde"] }
+nomt = { git = "https://github.com/Sovereign-Labs/nomt", rev = "c85fb01f526b9eee742bafbdc89c4a58313b249b", features = ["borsh", "serde"] }
 concread = { version = "0.5.7" }
 
 anyhow = { version = "1.0.95" }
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -208,8 +208,9 @@ reltester = { git = "https://github.com/sovereign-labs/reltester.git", rev = "d6
 rockbound = { git = "https://github.com/sovereign-labs/rockbound", rev = "d8b270d399d0d355a19fbeedf9af009baee8ae66" }
 nmt-rs = { version = "0.2.5", features = ["serde", "borsh"] }
 # External dependencies
-nomt-core = { version = "1.0.4", default-features = false, features = ["borsh", "serde"] }
-nomt = { version = "1.0.4", default-features = false, features = ["borsh", "serde"] }
+# TODO: switch back to the crates-io release once one containing `nomt::grow_hashtable` is published.
+nomt-core = { git = "https://github.com/thrumdev/nomt.git", rev = "cf1672b2bb9e89e07f6d939f6b7e404417c3120a", default-features = false, features = ["borsh", "serde"] }
```
citizen-stig: Could we ask Thumdev to release new version and just keep using version. 

Less git-dependencies, closer we to crates.io. Currently only rockbound is git revision and it is under our control. So if we urgently need do crates.io release, we can.

examples/demo-rollup/src/main.rs
```diff
@@ -387,10 +404,11 @@ async fn new_rollup_with_sp1_mock_da(
     )
     .await?;
 
-    let rollup_config: RollupConfig<MultiAddressEvmSolana, StorableMockDaService> =
+    let mut rollup_config: RollupConfig<MultiAddressEvmSolana, StorableMockDaService> =
         from_toml_path(rollup_config_path).with_context(|| {
             format!("Failed to read rollup configuration from {rollup_config_path}")
         })?;
+    rollup_config.storage.allow_nomt_grow = allow_nomt_grow;
```
citizen-stig: Why we do this and not populate value in the config?

examples/demo-rollup/src/main.rs
```diff
@@ -336,6 +350,7 @@ async fn new_rollup_with_mock_da(
     stop_at_rollup_height: Option<RollupHeight>,
     override_code_commitments: bool,
     start_fresh_outer_proof_on_resync: bool,
+    allow_nomt_grow: bool,
```
citizen-stig: same about why value is not coming via config?

crates/full-node/sov-db/src/metrics/nomt.rs
```diff
@@ -43,7 +44,7 @@ impl NomtDbMetric {
             tracing::warn!(
                 %db,
                 rate = hash_table_utilization.occupancy_rate(),
-                "Occupancy rate for NOMT hashtable is too high. Please update buckets size and resync database");
+                "Occupancy rate for NOMT hashtable is too high. Please increase the configured bucket count and restart the node with `--allow-nomt-grow`");
```
citizen-stig: is this cmd param for the rollup? So existing rollup need to add support for that? Why not keep it config.toml with default false? Then operator only has to change value in config. And not touching systemd unit, docker exec. Also it will be more sov-rollup-manager friendly.

crates/full-node/sov-db/src/config.rs
```diff
@@ -133,6 +133,17 @@ pub struct RollupDbConfig {
     /// More details at [`Options::page_cache_upper_levels`]
     pub kernel_page_cache_upper_levels: Option<usize>,
 
+    /// Allow growing the NOMT hashtables of an existing database when the
+    /// configured bucket count is larger than the on-disk one.
+    ///
+    /// Growing is a one-way operation that permanently increases disk usage
+    /// (shrinking is not supported), so it must be explicitly acknowledged at
```
citizen-stig: Why should it be explicitly acknowledged? It won't happen if buckets are unchanged. If buckets are changed, then without this option it is going to be an error at startup (current dev behaviour).

Also let's take step back and think about UX. Users want their rollup to keep functioning. So if they see warning, and this value is already set to true, they only have to increase buckets number (considering they have enough disk, which is other concern). Maybe default = true can be even more prefferable from users point of view. 

Rollup as a whole has a already plenty of params operator has to keep track of. Adding one more does not make life easier. And nothing prevents operator from keeping the flag in systemd unit / kubernetes template after it has applied an increase.

theodorebugnet: Hmm, this is a good point. I was thinking that because the operating is destructive, explicit acknowledgement sounds better. I want to avoid the situation where someone mis-copies the wrong config, and suddenly his nomt db takes up 500GB of extra space on disk for example.

With this view, being able to set it to default=true in the config defeats the purpose. It's meant to be an exceptional parameter you only add for special cases.

You're right that the rollup already has a lot of config options though, and this adds more friction to rollup operation. Maybe the right tradeoff is to just silently grow if the config changes, and the "surprise" accidental changes will almost never happen realistically. I'm not sure though, because intentionally growing NOMT is likely going to be a very rare operation, so I'm not sure how frequent accidental changes might be relative to intentional changes.

Then again most rollup operators will probably not change their config often so perhaps I am overthinking this entirely. I think I'll bring it up on daily

theodorebugnet: Removed the param as per sync discussion

crates/full-node/sov-db/src/metrics/nomt.rs
```diff
@@ -43,7 +43,7 @@ impl NomtDbMetric {
             tracing::warn!(
                 %db,
                 rate = hash_table_utilization.occupancy_rate(),
-                "Occupancy rate for NOMT hashtable is too high. Please update buckets size and resync database");
+                "Occupancy rate for NOMT hashtable is too high. Please increase the configured bucket count and restart the node");
```
citizen-stig: I think it is important to add small node that buckets resizing can take some time.

citizen-stig:
LGTM, the git revision is the only minor blocker

citizen-stig (approved):

PR #2976, opened 2026-06-15
Add gas decimals constant

ross-weir:
# Description

Adds a new constant `GAS_TOKEN_DECIMALS` that is used to provide the amount of decimals used for the gas token. This is needed because the current gas token ids in genesis weren't created with the decimal in the last byte of the token id.

Updates universal wallet to handle this and also adds a new `/tokens/{id}/metadata` endpoint that returns the token metadata including the correct decimal count 

- [ ] TODO: Think I'll need to regen test data

--- 

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


examples/demo-rollup/demo-rollup-schema.json
```diff
@@ -5810,7 +5880,7 @@
           "name": "read_and_set_heavy_state"
         },
         {
-          "name": "run_cpu_heavy_operation"
+          "name": "run_c_p_u_heavy_operation"
```
ross-weir, line 5883: This is because we recently moved over to using serde for case conversion to be consistent, this is how serde formats it

citizen-stig (approved):
LGTM. I am not that familiar with wallet macros, especially crates/universal-wallet/macro-helpers/src/fixed_point_ints.rs changes

crates/universal-wallet/schema/src/ty/mod.rs
```diff
@@ -175,6 +175,17 @@ impl<L: LinkingScheme> Ty<L> {
...
+            Ty::Integer(
+                _,
+                IntegerDisplay::FixedPoint(FixedPointDisplay::FromSiblingFieldWithOverride {
+                    field_index,
+                    ..
+                }),
+            ) => {
+                // We need all 32 bytes of the sibling field to compare against `override_match`.
+                // This also covers the fallback `byte_offset`, which is always within 0..32.
+                (0..32).map(|i| (*field_index, i)).collect()
```
theodorebugnet, line 188: I don't think there's anything limiting the sibling field to being 32 bytes only, is there?
Not a big deal functionality-wise since we don't expect this to be widely used outside of our own special case override, but might be worth clarifying the comments a bit to
a) Not document it publicly (i.e. in the error message) and
b) Call out the special-casing here, that it intentionally only supports 32 byte values because that's how we use it and we don't expect it to be used anywhere else

crates/universal-wallet/macro-helpers/src/fixed_point_ints.rs
```diff
@@ -59,39 +93,41 @@ impl FromMeta for FixedPointDisplay {
             NestedMeta::Meta(Meta::List(list)) => {
                 match list.path.get_ident().map(Ident::to_string) {
                     Some(s) if s == "from_field" => {
+                        const USAGE: &str = "Field references for fixed points must provide a field index: `from_field(1)`, optionally a 0-indexed byte offset: `from_field(5, offset=31)`, and optionally an override pair: `from_field(1, offset=31, override_eq=<expr>, override_decimals=<expr>)`";
```
theodorebugnet: Same here w.r.t. hiding it from the error message to make it undocumented

theodorebugnet (approved):
LGTM!

crates/universal-wallet/schema/src/ty/mod.rs
ross-weir: done

crates/universal-wallet/macro-helpers/src/fixed_point_ints.rs
ross-weir: done

PR #2981, opened 2026-06-16
Do not mark skipped blob during shutdown as accepted

citizen-stig:
# Description

Previously blob sender returns `Ok(())` for blob that is not send if it happens during shutdown. Which is semantically wrong, because it never attempted to submit this blob.

This also resulted in following changes:

- `crates/full-node/sov-blob-sender/src/lib.rs:50`: added typed `BlobSenderShutdownError`; blob sender still rejects post-shutdown blobs, but callers can now distinguish graceful shutdown from real publish failures.
- `crates/full-node/sov-sequencer/src/preferred/preferred_blob_sender.rs:128`: preferred blob sender swallows only that typed shutdown error, and only after shutdown is observed.
- `crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs:335`: aggregator now exits on shutdown while waiting for metadata, checking sender status, proving, or publishing.
- `crates/full-node/sov-sequencer/src/preferred/sync_sequencer_state/inner.rs:671`: preferred sequencer skips new batch start and proof publication once shutdown is observed.
 
---

## Linked Issues
- Related to # (issue) 

## Testing
New test is added

## Docs
No updates

---

<a href="https://app.devin.ai/review/sovereign-labs/sovereign-sdk/pull/2981" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/preferred_blob_sender.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/preferred_blob_sender.rs
@@ -1,4 +1,4 @@
-use sov_blob_sender::BlobExecutionStatus;
+use sov_blob_sender::{BlobExecutionStatus, BlobSenderShutdownError};
 use sov_blob_sender::{BlobInternalId, BlobSender, BlobToSend};
 use sov_blob_storage::{PreferredBatchData, PreferredProofData};
 use sov_db::ledger_db::LedgerDb;
@@ -25,6 +25,7 @@ use crate::{common::TxStatusBlobSenderHooks, TxStatusManager};
 /// Wrapper around [`BlobSender`] with preferred blob -specific logic.
 pub struct PreferredBlobSender<Da: DaService> {
     inner: Option<BlobSender<Da, TxStatusBlobSenderHooks<Da::Spec>, LedgerDb>>,
+    shutdown_receiver: watch::Receiver<()>,
     nb_of_concurrent_batch_blob_submissions: Arc<AtomicUsize>,
     nb_of_concurrent_proof_blob_submissions: Arc<AtomicUsize>,
 }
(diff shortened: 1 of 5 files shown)


crates/full-node/sov-sequencer/src/preferred/preferred_blob_sender.rs
```diff
@@ -114,11 +119,26 @@ impl<Da: DaService> PreferredBlobSender<Da> {
...
         let data = batch_bytes(batch)?;
 
-        inner.publish_batch_blob(data, blob_id).await?;
+        let result = inner.publish_batch_blob(data, blob_id).await;
+        self.handle_blob_publish_result(result)?;
 
         Ok(())
     }
 
+    fn handle_blob_publish_result(&self, result: anyhow::Result<()>) -> anyhow::Result<()> {
```
bkolad, line 128: What advantage does this approach provide? Previously, publish_batch_blob logged an info message and returned Ok during shutdown. The end result appears to be the same now, but with additional plumbing code.

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
```diff
@@ -330,18 +332,37 @@ impl<Ps: ProverService> AggregatorTask<Ps>
...
+        mut self,
+        shutdown_receiver: tokio::sync::watch::Receiver<()>,
+    ) -> anyhow::Result<()> {
         loop {
-            let (metadata, window_size) = match self.metadata_rx.recv().await {
-                // Intake side dropped its sender (clean shutdown or intake error).
-                None => {
-                    tracing::debug!("Intake task closed metadata channel; aggregator exiting");
+            let (metadata, window_size) =
+                match future_or_shutdown(self.metadata_rx.recv(), &shutdown_receiver).await {
```
bkolad, line 341: What is the advantage of calling future_or_shutdown inside this method at every await point, instead of the previous approach where we made a single call to run_task_until_shutdown?

crates/full-node/sov-sequencer/src/preferred/preferred_blob_sender.rs
citizen-stig: The problem was that blob is marked as accepted, but it never was.

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
citizen-stig: That's good point, could be simplified.

crates/full-node/sov-sequencer/src/preferred/preferred_blob_sender.rs
citizen-stig: >  So a caller got "success" for a blob that was neither sent to DA nor durably stored.

PR #2982, opened 2026-06-16
Celestia Compression PR-2.1: Consensus breaking split in blob storage

citizen-stig:
# Description

**PR-A is a small, consensus-breaking prerequisite for PR3 (Celestia streaming/chunked compression).** It teaches the standard blob-storage consumer to distinguish *"the sender posted authenticated bytes that don't decode"* (→ **slash**) from *"the prover withheld bytes"* (→ **fail closed**), and — critically — to make that decision **before** trusting Borsh. No compression code lands here; this only prepares the trust boundary so PR3 can plug in.

## What changed

**1. `rollup-interface` — new trait method (`da.rs`)**
- `BlobReaderTrait::logical_decode_failed() -> bool`, defaulting to `false`.
- Signals that a blob's fully-provided, *authenticated* DA bytes do not canonically decode into their claimed logical payload (e.g. a compressed envelope that fails to decompress to the declared length). Derived deterministically from authenticated bytes — never from a serialized claim — so native and zk execution agree. Adapters with no decode layer (logical == physical bytes) keep the `false` default, so **behavior is unchanged for them at the trait level.**

**2. `sov-blob-storage` — pre-Borsh decision gate (`capabilities.rs`)**
- `deserialize_or_try_slash_sender` now forces the full read/decode, then routes through a new pure function `blob_deserialization_gate(logical_decode_failed, verified_len, total_len)` returning one of three outcomes, evaluated **before** `try_from_slice`:
  - `SlashUndecodable` — authenticated bytes present but undecodable (sender's fault) → slash, return `None`.
  - `WithheldFailClosed` — fewer logical bytes than claimed (prover's fault) → panic / fail closed.
  - `Deserialize` — complete, decodable payload → attempt Borsh (a Borsh `Err` on a complete payload still slashes).
- Branch **order is security-critical**: the decode-failure check takes precedence over the length check, because a non-canonical envelope can decode to *exactly* the claimed length yet still be invalid.

## Why it's consensus-breaking (the core review point)

Previously the completeness assert lived **only inside the Borsh-`Err` branch**:

```rust
match B::try_from_slice(...) {
    Ok(batch) => Some(batch),               // accepted with NO completeness check
    Err(e) => { assert_eq!(verified_len, total_len, ...); /* slash */ }
}
```

A malicious prover could supply a **short prefix that itself deserializes** into a valid `B` and have it accepted, bypassing completeness. PR-A moves the completeness check ahead of Borsh and runs it **unconditionally**, closing that hole. For legacy (non-compressed) blobs this changes one path: a short-but-valid prefix now fails closed instead of being accepted — hence the consensus break.

**Reviewer focus:** the branch ordering in `blob_deserialization_gate` and the "check completeness before Borsh" invariant — that's the consensus-relevant behavior change. The rest is plumbing for PR3.

## Linked Issues
- Related to #2990 

## Testing
New unit tests on the extracted pure gate (`deser_gate_*`) lock in all three outcomes and the precedence rule: decode-failure slashes even at full length; an incomplete read fails closed regardless of whether the prefix would deserialize; complete payloads (including empty) proceed.

## Docs
`docs/design/compression_actual_plan.md` updated to v10 (design/rationale for the chunked envelope and PR-A's role). Doc-only.


crates/module-system/module-implementations/sov-blob-storage/src/capabilities.rs
```diff
@@ -1056,11 +1056,53 @@ impl<S: Spec> BlobStorage<S> {
                 state,
             ).expect("Failed to remove funds for deserialization even though the sender has enough balance. This should never happen.");
         }
-        match B::try_from_slice(data_for_deserialization(blob)) {
+        // Force the full read/decode before inspecting decode status or completeness.
+        // In native mode `data_for_deserialization` is `full_data()`, which drives the
+        // blob to its end (for a compressed envelope: decoding to logical completion /
+        // decode failure / physical EOF); in the guest it is the already-authenticated
+        // `verified_data()`.
+        let _ = data_for_deserialization(blob);
```
citizen-stig: I don't like workarounds those. We probably should move everything including results inside `data_for_deserialization` and do logic there.

PR #2989, opened 2026-06-18
Celestia Adapter: Add compression support

### LINKED ISSUE #2990: Celestia: Add support for compressing blobs
## Requirements

Celestia adapter should be able to compress blobs before submission to save DA cost.

Adapter should remain compatible with "old" uncompressed blobs

Compression should be streamed/chunked, so adapter retains capability of verifying and proving only shares being touched and not the whole blob.

Feature should be self-contained inside celestia adapter and not affecting the rest of the STF. - **Transparent for `CelestiaService` callers** — the public submit/read API takes/returns raw bytes; compression is internal.

## Upgrade procedure

1. Update rollup to use celestia adapter that supports compression, but keep compression off
2. Enable compression (`da.compress_on_submit = "lz4"`) in `rollup_config.toml` after all nodes have been updated

This change does not have any constant or activation height

## High Level Design

Note. We use `.advance` as example to focus on more complicated partial reads. `.full_data` is simplified version of `.advance(total_len)` so it is simpler in terms that only single `.advance` is called. 

1. Magic bytes for blobs with compression added in the beginning. Read path check them or fallsback to legacy raw bytes
2. BlobWithSender gains `envelope_state: OnceLock<EnvelopeState>` which hold decoded values. In native this Envelope is build during `.advance`. On sequential advance previous is state is taken to save CPU cycles. 
  - Guest builds `EnvelopeState` once from shares data that it verified as usually, so decompressed data is verifed
  - Honest prover re-uses same logic for decompress, but populates it only with what is needed.

Currenlty we've only added LZ4 compression support on the assumption that its pure rust implementation will be cheaper for ZK, but still deliver significant compression benefits.

`zstd` can be added later for operator mode rollups

## Implementation Progress

- [x] PR 1: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2977
- [x] PR 2: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2978
- [x] PR 2.1: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2982
- [x] PR 3: https://github.com/Sovereign-Labs/sovereign-sdk/pull/2989 (cummulative of all 4)

PRs are stacked on top of each other and the first one open against `nikolai/celestia-compression-feature` branch, so there will be another short round of review after all PRs are merged.

## Accepted Risks And Tradeoffs

- No read-side activation gate means upgraded code interprets magic-prefixed historical blobs as envelopes. The 128-bit magic is chosen so a standard Borsh batch/proof payload that is also a *valid* envelope is negligible; a magic-prefixed *invalid* frame is malformed-as-raw, i.e. identical bytes to legacy. Accepted collision risk.
- Runtime config cannot make read semantics safe for mixed old/new networks; it only prevents this node from emitting envelopes. Coordinated rollout remains an operational requirement.
- Chunked compression **preserves** partial-read savings and the verifier DoS bound (work scales with compressed bytes consumed, chunk-granular). Share-aligned chunks add small framing overhead (~0.8% on a large batch); accepted for the tightest granularity.
- Decompression cost is bounded per-chunk (caps enforced before allocation) and by `MAX_LOGICAL_BLOB_LEN`. Compression economics are left to metrics; default emission is off.
- Proof blobs are not compressed. Can be enabled in follow up

## Other notes

* Abstraction is leaky. In order to preserve rollup security, new method added to BlobReaderTrait and let the rest of the system know that DaService allowed to modify bytes from the DA

citizen-stig:
# Description

Wires transparent LZ4 compression for Celestia **batch** blobs, self-contained in
`crates/adapters/celestia` apart from one shared-trait signal (justified below). Blobs are
posted as a chunked, prefix-decodable envelope so the ZK verifier's work still scales with
bytes consumed (no full-blob-decode DoS); legacy raw blobs are unchanged; emission is
config-gated and defaults **off**. Reads decode the authenticated compressed bytes into the
logical payload via a single decoder shared by native and guest; native `advance` decodes
incrementally (O(N), not O(N²)).

## Why `rollup_decode_failed` lives on `BlobReaderTrait` (not celestia-only)

The one part that isn't self-contained in the adapter is the
`BlobReaderTrait::logical_decode_failed` signal that `sov-blob-storage` consults. It is
necessary; here is the precise reason.

**Two different failures produce the identical observable.** After the forced read, the gate
(`sov-blob-storage/src/capabilities.rs:1307`) sees only `verified_len` (logical bytes decoded)
and `total_len` (authenticated `logical_len`). Both yield `verified_len < total_len`:

- **Corrupt envelope** (permissionless sender posts bad chunks): the full physical blob is
  present and authenticated, but decodes short → the sender is at fault → slash.
- **Prover withholding**: the prover hands over a witness whose accumulator is a short prefix
  of the blob, so the decode is short → the prover is cheating → fail closed (panic).
  Slashing here would punish the honest sequencer for the prover's misbehavior, and the
  panic is what stops a state transition computed over a truncated blob.

**A short accumulator is a real, accepted witness** — which is why withholding is a genuine
threat. The verifier proves shares for the consumed bytes
(`shares_needed_for_bytes_with_signer(compressed_verified_data().len(), …)`,
`celestia/src/verifier/mod.rs:355`) and only checks `compressed_total_len() == sequence_length`
(`:451`), where `compressed_total_len() = accumulator + remaining`. So a prover can submit a
short accumulator (the rest sits unread in `remaining`) and the verifier accepts it. Partial
reads are the DoS optimization; we can't force full reads.

**So the two alternatives one might reach for both break:**
- Let the corrupt envelope flow to the completeness check → it panics → any permissionless
  poster halts the chain.
- Make `verified < total` slash unconditionally (no panic) → a malicious prover truncates any
  blob and slashes honest sequencers.

The only escape is telling the generic gate which case it is — and the distinguishing fact
("are all the authenticated DA bytes present?", i.e. `compressed_verified == compressed_total`)
is the adapter-internal compressed-vs-logical split. Generic blob-storage sees only logical
lengths, so it cannot compute it. `logical_decode_failed` is therefore a **defaulted (`false`)**
method on `BlobReaderTrait` — the validity companion to `verified_data`'s existing contract
that an adapter may post a different physical representation — so identity adapters are
byte-for-byte unaffected and only transforming adapters (celestia compression) override it.

---

## Linked Issues
- Related to #2990

## Testing
`SKIP_GUEST_BUILD=1 cargo nextest run -p sov-celestia-adapter --features native` — full adapter
suite (99 tests) green, covering: legacy raw blobs unchanged; partial-read matrix;
incremental-`advance` == wholesale (byte-for-byte); corrupt-chunk → slash signal;
malformed-header → read-as-raw; and a live docker `test_submit_compressed_batch_round_trips`
e2e. `cargo check -p sov-celestia-adapter --no-default-features` (guest build). clippy + fmt
clean.

## Docs
Rationale lives on `BlobReaderTrait::logical_decode_failed` (rollup-interface) and in the
design docs under `docs/design/`.


crates/adapters/celestia/src/da_service/mod.rs
```diff
@@ -109,16 +116,42 @@ impl CelestiaService {
...
         };
-        let blob = JsonBlob::new(namespace, blob.to_vec(), Some(signer.0))
-            .expect("Bug in CelestiaAdapter");
+
+        // Compress batch blobs when configured; proofs always post verbatim. This is
+        // emission only — read/verify semantics never depend on it.
+        let posted = if matches!(ns, RollupNamespace::Batch) {
+            let posted = crate::envelope::encode_for_submission(
+                blob,
+                matches!(self.compression, crate::config::CompressOnSubmit::Lz4),
```
theodorebugnet: Small nit: we could add a method `CompressOnSubmit::is_enabled()` with exhaustive pattern matching, that way if new variants are added later we won't risk forgetting to expand this `matches!`

crates/adapters/celestia/src/types/mod.rs
```diff
@@ -97,15 +99,113 @@ impl From<[u8; 32]> for TmHash {
...
+
+/// Celestia-private accessors distinguishing the two byte streams a blob represents:
+///
+/// * **DA-physical ("compressed")**: the payload bytes as actually posted to Celestia.
+///   Share-occupancy math (inclusion proofs, namespace continuity) is defined over
+///   these bytes and only these bytes.
+/// * **Logical**: the payload bytes exposed to the rollup via [`BlobReaderTrait`].
+///
+/// Today the two streams are identical. Once blobs can be posted in a compressed
+/// envelope, they diverge; proof generation and verification must keep using the
```
theodorebugnet: Nit: comment from an intermediate PR?

crates/adapters/celestia/src/types/mod.rs
```diff
@@ -97,15 +99,113 @@ impl From<[u8; 32]> for TmHash {
...
 }
 
+/// Celestia-private accessors distinguishing the two byte streams a blob represents:
+///
+/// * **DA-physical ("compressed")**: the payload bytes as actually posted to Celestia.
+///   Share-occupancy math (inclusion proofs, namespace continuity) is defined over
+///   these bytes and only these bytes.
+/// * **Logical**: the payload bytes exposed to the rollup via [`BlobReaderTrait`].
+///
+/// Today the two streams are identical. Once blobs can be posted in a compressed
```
theodorebugnet: Same nit (outdated comment)

crates/adapters/celestia/src/envelope.rs
```diff
@@ -0,0 +1,756 @@
...
+//!   chunk_payload     [chunk_encoded_len]
+//! ```
+//!
+//! Chunks are decodable on their own, so reading the first N logical bytes only
+//! authenticates and decodes the chunks covering them: verifier/guest work scales
+//! with bytes consumed, not blob size.
+
+/// Envelope magic prefix. 128 bits, chosen so a standard Borsh batch/proof
+/// payload beginning with these exact bytes would already be malformed in
+/// practice.
```
theodorebugnet, line 37: Could we add a comment explaining why a legacy/uncompressed blob starting with these bytes would be malformed?

crates/rollup-interface/src/state_machine/da.rs
```diff
@@ -164,11 +164,43 @@ pub trait BlobReaderTrait: Serialize + DeserializeOwned + Send + Sync + 'static
...
+    /// The returned bytes are the *logical* payload as submitted by the original caller.
+    /// Adapters are free to post a different physical representation to the DA layer
+    /// (e.g. a compressed envelope), as long as the logical bytes exposed here are
+    /// authenticated by the adapter's verifier.
     fn verified_data(&self) -> &[u8];
 
     /// Returns the total number of bytes in the blob. Note that this may be unequal to `verified_data.len()`.
+    ///
+    /// Like [`Self::verified_data`], this is the length of the *logical* payload, which may
+    /// differ from the number of bytes the blob occupies on the DA layer.
```
theodorebugnet, line 177: Might be worth a comment clarifying that implementers should guarantee this comes from proven DA bytes and isn't host-forgeable, if I'm understanding the length check logic correctly

crates/adapters/celestia/src/config.rs
```diff
@@ -276,6 +309,15 @@ pub(crate) const fn default_safe_lead_time_ms() -> u64 {
     500
 }
 
+/// Default chunk size: one continuation sparse-share payload (share-aligned).
+/// Decoupled from the per-chunk cap [`crate::envelope::MAX_LOGICAL_CHUNK_LEN`] so the
+/// conservative default does not move when the cap is raised.
+pub(crate) const fn default_compression_chunk_size() -> usize {
+    // = CONTINUATION_SPARSE_SHARE_CONTENT_SIZE. A literal (not `MAX_LOGICAL_CHUNK_LEN`)
+    // keeps the default pinned to one share when the cap is raised.
+    482
```
theodorebugnet: Should we just use `CONTINUATION_SPARSE_SHARE_CONTENT_SIZE` here? If Celestia ever changes its share size (however unlikely), it might make sense to also have the default chunk size follow.

crates/adapters/celestia/src/envelope.rs
```diff
@@ -0,0 +1,756 @@
...
+        let cut = frame.len() - 7;
+        let d = envelope(&frame[..cut]);
+        assert!(
+            d.clean,
+            "a truncated trailing chunk is incomplete, not corrupt"
+        );
+        assert!(d.logical.len() < logical.len());
+        assert_eq!(
+            &logical[..d.logical.len()],
+            &d.logical[..],
```
theodorebugnet: Minor nit: I think we try (not very successfully) to enforce the order of `assert_eq!(actual, expected)`, it looks reversed here

crates/adapters/celestia/src/types/mod.rs
```diff
@@ -384,6 +618,301 @@ pub mod tests {
...
+        let _ = b.full_data();
+
+        assert!(
+            a.logical_decode_failed(),
+            "incremental advance must surface a mid-stream corrupt chunk"
+        );
+        assert_eq!(a.logical_decode_failed(), b.logical_decode_failed());
+        // Both expose the same clean prefix decoded before the corrupt chunk.
+        assert_eq!(a.verified_data(), b.verified_data());
+        assert_eq!(&logical[..a.verified_data().len()], a.verified_data());
```
theodorebugnet: Minor nit: `assert!(actual, expected)` reversed

crates/adapters/celestia/src/types/mod.rs
```diff
@@ -384,6 +618,301 @@ pub mod tests {
...
+        let mut a = blob_with_posted_payload(&posted);
+        let mut b = blob_with_posted_payload(&posted);
+
+        // An intermediate small advance yields a chunk-aligned prefix of the logical.
+        let prefix = a.advance(700).to_vec();
+        assert!(
+            !prefix.is_empty() && prefix.len() < logical.len(),
+            "a chunk-aligned logical prefix"
+        );
+        assert_eq!(&logical[..prefix.len()], prefix.as_slice());
```
theodorebugnet: Minor nit: `assert!(actual, expected)` reversed

crates/adapters/celestia/src/types/mod.rs
```diff
@@ -384,6 +618,301 @@ pub mod tests {
...
+        );
+
+        let mut blob = blob_with_posted_payload(&posted);
+        // Reading one logical byte pulls whole chunks until >= 1 byte is covered.
+        let prefix = blob.advance(1).to_vec();
+        assert!(
+            !prefix.is_empty() && prefix.len() < logical.len(),
+            "a chunk-aligned logical prefix"
+        );
+        assert_eq!(
```
theodorebugnet, line 822: Minor nit: `assert!(actual, expected)` reversed

theodorebugnet (approved):
LGTM, nice!

Just to check my understanding: is it safe/intended for both compressed and uncompressed blobs to coexist? And is changing the default compressed chunk len safe? For example imagine a replica sequencer with a different config from the master, and then failover ocurs so the new leader starts submitting blobs with different settings. It looks safe to me, I just want to double-check my understanding.

citizen-stig:
> LGTM, nice!
> 
> Just to check my understanding: is it safe/intended for both compressed and uncompressed blobs to coexist?

@theodorebugnet, yes. if we upgrade adapter, but never turn on compression, old version of adapter still can submit blobs and they will be processed.

> And is changing the default compressed chunk len safe? For example imagine a replica sequencer with a different config from the master, and then failover ocurs so the new leader starts submitting blobs with different settings. It looks safe to me, I just want to double-check my understanding.

Yes, chunk size is encoded into the blob(each chunk has its metadat) So each blob can be decoded on its own, just by parsing the header. 

And changing chunk size is safe, within limits imposed by MAX_ALLOWED_CHUNK size which gates zk-proving whole 8MB compressed blob from unregistered sender.

> The only chunking number every node must agree on is the compile-time constant MAX_LOGICAL_CHUNK_LEN = 1446

The only thing I will add later today, is exposing `MAX_LOGICAL_CHUNK_LEN` as constants.toml, so operator only mode rollups can max out compression

crates/adapters/celestia/src/envelope.rs
citizen-stig: Fixed in 83ded7473510a8df487a63c9ff0298a63b869c24

crates/adapters/celestia/src/types/mod.rs
citizen-stig: Fixed in 83ded7473510a8df487a63c9ff0298a63b869c24

crates/adapters/celestia/src/types/mod.rs
citizen-stig: Fixed in 83ded7473510a8df487a63c9ff0298a63b869c24

crates/adapters/celestia/src/da_service/mod.rs
citizen-stig: Good point! Done in 48b297e8159af5015de2088de15294bfe2bc5078

crates/adapters/celestia/src/types/mod.rs
citizen-stig: Fixed in d6a7fe5da1d125a37c8a60d35c531786966957b9

crates/adapters/celestia/src/types/mod.rs
citizen-stig: Fixed in d6a7fe5da1d125a37c8a60d35c531786966957b9

crates/adapters/celestia/src/envelope.rs
citizen-stig: Fixed in d6a7fe5da1d125a37c8a60d35c531786966957b9

crates/rollup-interface/src/state_machine/da.rs
citizen-stig: Fixed in d6a7fe5da1d125a37c8a60d35c531786966957b9

PR #3002, opened 2026-06-22
Introduce secondary shutdown controller for post-run background tasks

bkolad:
# Description

  This PR hides secondary shutdown signaling behind `SecondaryShutdownController` and threads it through DA services, proof workflows, metrics, and rollup blueprint setup.

  Changes:
  - Adds `SecondaryShutdownController` in sov-rollup-interface.
  - Updates Celestia and mock DA background tasks to use the controller.
  - Updates OP/ZK/operator proof workflows to use secondary shutdown through the controller.
  - Updates metrics publisher initialization to use the shared shutdown controller.
  - Propagates sov-rollup-interface/native through sov-metrics.
  - Adjusts tests and demo rollups for the new shutdown API.

### DIFF
--- a/crates/full-node/sov-metrics/Cargo.toml
+++ b/crates/full-node/sov-metrics/Cargo.toml
@@ -24,7 +24,7 @@ derive_more = { workspace = true, features = ["deref_mut", "deref"], optional = 
 derive-new = { workspace = true }
 tokio-metrics = { workspace = true, optional = true }
 hex = { workspace = true }
-sov-rollup-interface = { workspace = true, features = ["native"], optional = true }
+sov-rollup-interface = { workspace = true, optional = true }
 
 anyhow = { workspace = true }
 risc0-zkvm = { workspace = true, default-features = false, features = [
--- a/crates/full-node/sov-metrics/Cargo.toml
+++ b/crates/full-node/sov-metrics/Cargo.toml
@@ -24,6 +24,7 @@ derive_more = { workspace = true, features = ["deref_mut", "deref"], optional = 
 derive-new = { workspace = true }
 tokio-metrics = { workspace = true, optional = true }
 hex = { workspace = true }
+sov-rollup-interface = { workspace = true, features = ["native"], optional = true }
 
 anyhow = { workspace = true }
 risc0-zkvm = { workspace = true, default-features = false, features = [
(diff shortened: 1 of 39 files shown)


crates/rollup-interface/src/node/secondary_shutdown.rs
```diff
@@ -0,0 +1,49 @@
...
+impl SecondaryShutdownController {
+    /// Creates a new secondary shutdown controller.
+    pub fn new() -> Self {
+        let (sender, mut receiver) = watch::channel(());
+        receiver.mark_unchanged();
+        Self { sender, receiver }
+    }
+
+    /// Waits until a secondary shutdown notification is sent.
+    pub async fn changed(&self) -> Result<(), watch::error::RecvError> {
```
ross-weir: IMO something like `wait_for_shutdown` would be more explanatory and match the other method names

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
```diff
@@ -71,7 +71,9 @@ where
                 .with_max_times(BACKOFF_POLICY_MAX_NUM_RETRIES),
             stf_info_receiver,
             da_sync_state,
-            shutdown_receiver,
+            secondary_shutdown_controller: SecondaryShutdownController::clone(
```
ross-weir: nit: `secondary_shutdown_controller.clone` is probably more idomatic

ross-weir (approved):

crates/rollup-interface/src/node/secondary_shutdown.rs
bkolad: fixed

examples/demo-rollup/src/da_tester/demo_celestia_tester.rs
```diff
@@ -37,19 +38,18 @@ async fn main() -> anyhow::Result<()> {
...
             rollup_proof_namespace: ROLLUP_PROOF_NAMESPACE,
         },
-        shutdown_rx,
+        &secondary_shutdown_controller,
     )
     .await;
 
     sov_celestia_adapter::checker::check_da_service(&da_service, args.rounds).await?;
-    shutdown_tx.send(())?;
+    let _ = secondary_shutdown_controller.shutdown();
```
citizen-stig: Why error is silently ignored? Should we at least warn? Our shutdown logic often tends to be complicated, so skipping errors could hide issues.

bkolad: There is no errow anymore (the receiver is never dropped before the sender).
Fixed it in the latest PR to `secondary_shutdown_controller.shutdown()`;

crates/full-node/sov-stf-runner/src/processes/zk_manager/mod.rs
bkolad: fixed

crates/full-node/sov-metrics/Cargo.toml
```diff
@@ -24,6 +24,7 @@ derive_more = { workspace = true, features = ["deref_mut", "deref"], optional =
 derive-new = { workspace = true }
 tokio-metrics = { workspace = true, optional = true }
 hex = { workspace = true }
+sov-rollup-interface = { workspace = true, features = ["native"], optional = true }
```
citizen-stig: I understand where it is coming from, we have `native` here, but also native feature on sov-metrics, which also enabled native if sov-rollup-interface is present. 

I think it might be a little bit simpler, if this definition didn't have explicit native and it is only enable via sov-metrics native.

citizen-stig (approved):

crates/full-node/sov-metrics/Cargo.toml
bkolad: fixed

PR #3009, opened 2026-06-23
Celestia: remove unnecessary `Blob(Vec<Shares>)` from ZK Witness

citizen-stig:
# Description

The ZK witness serialized each Celestia blob as its full live reader, which carried the blob's entire `Vec<Share>` — including shares the rollup never reads. Those unread shares bloat the witness and let a sequencer inflate proving cost without changing what the rollup actually executes.

This PR serializes only the **authenticated DA projection** the guest needs, and drops the shares:

- `BlobWithSender` now (de)serializes through a pruned wire form (`BlobWithSenderWire` /  borrowed `BlobWithSenderWireRef`) carrying just the verified prefix (`accumulator`), the total DA length, and metadata (namespace range, sender, hash).
- On deserialize the guest rebuilds a **share-less placeholder** (`BlobIterator::verified_placeholder`) and reads through the `accumulator`; it never touches shares. Authentication is unchanged - it runs against the separate inclusion proof, which pins `total_len` to the first share's `sequence_length` and checks the prefix byte-for-byte.
- Deserialization validates the witness: rejects `total_len > u32::MAX` (32-bit guest  truncation) and `accumulator.len() > total_len` (`remaining()` underflow).
- `BlobWithSender` no longer implements `PartialEq` — after pruning, structural equality is meaningless; tests compare the projection via an explicit `test_helper::compare_equality`.

Depends on #2989.

### Examples

**Normal blob — N txs across 10 shares.** The host reads the whole payload, so the witness carries the full rollup-payload prefix + `total_len`, but not the 10 framed/padded `Share`s (≈ half the per-blob bytes here). The guest reads the same txs from the `accumulator`; the inclusion proof authenticates them.

**Recovery blob with 30 shares of spam appended.** A prover can no longer pad the witness with the unread spam — the wire form has no share field, and `accumulator` is capped at the authenticated `total_len`. If the spam is a non-canonical tail past the declared payload, the rollup flags it (`rollup_decode_failed` → slash) without reading it, so it never reaches the `accumulator` or the witness.

----

## Testing
New unit tests in the celestia adapter: the serde round-trip drops and recomputes the envelope cache; a pruned witness round-trips the read prefix and is smaller than its payload;
deserialization rejects a forged `total_len` / over-long `accumulator`; and the existing block / relevant-blobs serialization tests now compare blobs via `compare_equality` / `compare_blobs` (since `BlobWithSender` no longer derives equality).

## Docs
The pruned wire form and the native-vs-guest split are documented inline on `BlobWithSenderWire` and `BlobWithSender`.


crates/adapters/celestia/src/types/mod.rs
```diff
@@ -129,19 +130,111 @@ pub struct BlobWithSender {
...
+    /// the DA block's share count (orders of magnitude below `u32::MAX`) and is consumed only
+    /// when building proofs on the host (`verifier::proofs`); the 32-bit guest verifier never
+    /// reads it, so it carries no truncation risk.
+    range_in_namespace: Range<usize>,
+    sender: CelestiaAddress,
+    hash: HexHash,
+}
+
+#[derive(Serialize)]
+struct BlobWithSenderWireRef<'a> {
```
bkolad: NIT: This is mostly duplication of `BlobWithSenderWire`, what about using a single struct:
```
  #[derive(Debug, Serialize, Deserialize)]
  struct BlobWithSenderWire<'a> {
      #[serde(borrow)]
      accumulator: Cow<'a, [u8]>,
      total_len: u64,
      range_in_namespace: Cow<'a, Range<usize>>,
      sender: Cow<'a, CelestiaAddress>,
      hash: Cow<'a, HexHash>,
  }
```
 This way, reordering, adding, or removing a field only has to be done once, instead of being kept in sync across both structs.

bkolad (approved):

PR #3017, opened 2026-06-25
 Extract shutdown utilities into sov-shutdown

bkolad:
# Description

 -  Adds a new `sov-shutdown` crate for shutdown controllers and helper utilities.
  - Moves shutdown controller APIs out of `sov-rollup-interface` and updates in-repo call sites.


crates/utils/sov-shutdown/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+    (
+        $name:expr,
+        $recv:expr,
+        $shutdown:expr,
+        $var:ident => $body:block
+    ) => {
+        let name = $name;
+        loop {
+            tokio::select! {
+                _ = $shutdown.wait_for_shutdown() => {
```
citizen-stig, line 80: can we add #![biased]  so shutdown has priority?

crates/utils/sov-shutdown/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+    shutdown: &watch::Receiver<()>,
+) -> FutureOrShutdownOutput<T::Output>
+where
+    T: Future,
+{
+    let mut shutdown = shutdown.clone();
+
+    select! {
+        res = inner => FutureOrShutdownOutput::Output(res),
+        _ = shutdown.changed() => FutureOrShutdownOutput::Shutdown,
```
citizen-stig, line 31: Can we add #[bias] here too?

crates/rollup-interface/src/node/mod.rs
```diff
@@ -7,91 +7,5 @@
 pub mod da;
 mod da_sync_state;
 pub mod ledger_api;
-mod shutdown_controller;
```
citizen-stig, line 10: Slightly related: does rollup-interface still needs tokio dep? With all features?

citizen-stig (approved):

crates/utils/sov-shutdown/src/lib.rs
bkolad: I would do it in a separate PR (this already modifies 90 files)

crates/utils/sov-shutdown/src/lib.rs
bkolad: same

crates/rollup-interface/src/node/mod.rs
bkolad: We use tokio types in DaService (will check the features in the next PR)

PR #3018, opened 2026-06-25
Split demo rollup into 2 features, by DA

citizen-stig:
# Description

Previously, `demo-rollup` has been compiling all DAs and ZKVMs all the time, and deciding what to run during the runtime. This adds a lot of time to compilation process and slows down CI.

In reality following combinations are used mostly

* `mock_da` with `mock_zkvm` or `sp1` in unit and integration tests.
* `celestia_da` with `risc0` in README_CELESTIA.md

This PR does this split, in attempt to reduce compilation time, but preserve all tests.

So now `mock_da` feature is default and even chosen for `--all-features`, which also allows choosing ZKVM at the runtime. This feature excludes 2 big subtrees: `sov-celestia-adapter` and `sov-risc0-adapter`.

It removes `celestia` + `sp1` ZKVM guest combination, which has never been tested.

---

## Testing
All existing test are passing

## Docs
No updates


CHANGELOG.md
```diff
@@ -1,6 +1,5 @@
...
-  * **Breaking Change** `CelestiaConfig` and `GrpcEndpointConfig` now set `#[serde(deny_unknown_fields)]`: unknown or misspelled keys in the `[da]` config table fail at startup instead of being silently ignored. Existing field aliases (e.g. `celestia_rpc_address`) are unaffected.
+# 2026-06-25
+- #XXXX demo-rollup: split the example into two DA-layer Cargo features — `mock_da` (default; mock + SP1 zkVMs) and `celestia_da` (Risc0) — to cut compile time. The DA layer is selected at compile time while the zkVM stays a runtime `--zk-vm` choice within `mock_da`, so a build only compiles the selected DA's zkVM guests: a default `mock_da` build no longer pulls the Celestia/Risc0 adapters, and a `celestia_da` build skips the SP1 guests. When both DA features are enabled (`--all-features`), mock DA is selected by default. Also drops the unused SP1 Celestia guest build. Example crate only (`publish = false`) — no SDK API, state, or protocol change.
```
citizen-stig:
```suggestion
- #3018 demo-rollup: split the example into two DA-layer Cargo features — `mock_da` (default; mock + SP1 zkVMs) and `celestia_da` (Risc0) — to cut compile time. The DA layer is selected at compile time while the zkVM stays a runtime `--zk-vm` choice within `mock_da`, so a build only compiles the selected DA's zkVM guests: a default `mock_da` build no longer pulls the Celestia/Risc0 adapters, and a `celestia_da` build skips the SP1 guests. When both DA features are enabled (`--all-features`), mock DA is selected by default. Also drops the unused SP1 Celestia guest build. Example crate only (`publish = false`) — no SDK API, state, or protocol change.
```

theodorebugnet (approved):
Code LGTM, but I wonder if removing celestia+SP1 is the right approach considering that's currently our most production ready ZK rollup configuration.
Though I suppose if we want to switch the risc0 tests to sp1, this can be done as a follow-up, with the split already in place

citizen-stig:
> Code LGTM, but I wonder if removing celestia+SP1 is the right approach considering that's currently our most production ready ZK rollup configuration. Though I suppose if we want to switch the risc0 tests to sp1, this can be done as a follow-up, with the split already in place

Yep, we can change how demo celestia is tested.

PR #3020, opened 2026-06-26
Move rollup HTTP/RPC shutdown into the STF runner

bkolad:
# Description
This PR introduces the following changes

  - Remove `NodeEndpoints::background_handles` (it was never used, but adds additional cleanup logic)
  - Keep HTTP/RPC server task ownership inside `StateTransitionRunner`
  - Add explicit rollup and runner shutdown helpers to signal and await background tasks
  - Update endpoint construction sites after the `NodeEndpoints` shape change

### DIFF
--- a/crates/module-system/sov-modules-api/src/runtime/mod.rs
+++ b/crates/module-system/sov-modules-api/src/runtime/mod.rs
@@ -237,10 +237,6 @@ pub struct NodeEndpoints {
     pub axum_router: axum::Router<()>,
     /// A [`jsonrpsee::RpcModule`] for the runtime's JSON-RPC server.
     pub jsonrpsee_module: jsonrpsee::RpcModule<()>,
-    /// A list of optional background tasks that have been spawned for the endpoints' purposes.
-    ///
-    /// These will be joined upon node shutdown.
-    pub background_handles: Vec<tokio::task::JoinHandle<anyhow::Result<()>>>,
 }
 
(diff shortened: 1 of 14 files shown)


crates/module-system/sov-modules-api/src/runtime/mod.rs
```diff
@@ -237,10 +237,6 @@ pub struct NodeEndpoints {
     pub axum_router: axum::Router<()>,
     /// A [`jsonrpsee::RpcModule`] for the runtime's JSON-RPC server.
     pub jsonrpsee_module: jsonrpsee::RpcModule<()>,
-    /// A list of optional background tasks that have been spawned for the endpoints' purposes.
-    ///
-    /// These will be joined upon node shutdown.
-    pub background_handles: Vec<tokio::task::JoinHandle<anyhow::Result<()>>>,
```
bkolad, line 243: It was never used so I removed it in order to simplify shutdown logic.

citizen-stig: Could it be used by external clients?

citizen-stig (approved):

crates/module-system/sov-modules-api/src/runtime/mod.rs
bkolad: I am not aware of anyone using it.

PR #3032, opened 2026-07-08
Limit generation/windowed-nonce growth

theodorebugnet:
# Description

Add a sane limit to how much a generation or windowed nonce can grow in a single transaction, to prevent accidental major changes to account state.

The limit is 2x the current value, with exceptions for low values (at least the minimum required for pruning in generations, and at least PAST_WINDOWS for windowed nonces) to avoid excessive friction when the current value is small.

Also updates the module README which was out of date.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


bkolad (approved):
The PR LGTM. One question,  generation uniqueness now rejects jumps above `max(latest * 2, latest + PAST_TRANSACTION_GENERATIONS + 1) ` but if last is let's say 1000_0000 so the gap can be 1000_0000 and it grows as latest grows is that ok?

theodorebugnet:
Yep, the main idea is that you can't do 1000x e.g. when accidentally submitting ms time instead of s time. That's been Zeta's main issue

PR #3034, opened 2026-07-09
Sequencing data refactor

theodorebugnet:
# Description

Refactors the mutable sequencing data to be easier and safer to use, and require a lot less boilerplate from rollups.
* `HasSequencingData` is now a trait independent from HasCapabilities, which avoids having to re-impl all capability traits when using custom sequencing data.
* `HDTimestamp` now has first-class support, and the ChainState oracle is updated through the ChainState trait, separate from SequencingData handling.
* Sequencing data is now gated on `SequencerType::Preferred` during `Context` construction. This removes the need to gate every usage of the data on checking if it comes from the preferred sequencer.
* Sequencing data as a whole now provides key-based access, keeping the raw byte representation opaque. Access recording is implemented directly in the SDK; rollup developers now only need to implement the `get()` and `prune_to_used_keys()` methods for their type. (Those are intentionally left free-form to allow for arbitrary serialization and data handling; forcing a BTreeMap representation would force deserialization of the full data on every tx which might be inefficient.)

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/module-system/module-implementations/sov-evm/src/precompiles/sequencing_timestamp.rs
```diff
@@ -71,17 +77,9 @@ fn sequencing_timestamp_precompile<S: Spec, ST: TxState<S>>(
...
-        .map(|timestamp| timestamp.as_nanos())
-        .map(Ok)
-        .unwrap_or_else(|| {
-            chain_state
-                .get_oracle_time_nanos(env.state)
-                .map_err(|e| PrecompileError::State(e.to_string()))
-        })?;
+    let nanos = chain_state
+        .get_oracle_time_nanos(env.state)
+        .map_err(|e| PrecompileError::State(e.to_string()))?;
```
bkolad, line 82: Now this path performs a metered state read, which charges fees. That makes this a hard fork, so we need to handle it.

theodorebugnet: The PR is into the hard fork - but also the sequencing_timestamp precompile is mostly example code, nobody uses it in production

bkolad (approved):

PR #3035, opened 2026-07-09
Replace chain_id with chain_hash_fragment in TxDetails, to aid in deterministic signature checking

theodorebugnet:
# Description

Request from Bullet. `chain_id` is not needed for security - it's already hashed as part of the chain hash - and seems to be rarely, if ever, checked by users. So remove it in favor of adding a chain hash fragment which allows eliminating the trial-and-error loop in verifying transactions when multiple chain hash overrides are active.

Also adds consistency verification to V1 signing flows to assert that all signatures are done over the same chain has, helping avoid footguns if users sign a multisig around the exact time a rollup upgrades its chain hash.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/full-node/sov-api-spec/openapi-v3.yaml
```diff
@@ -1000,53 +1000,6 @@ components:
       items:
         type: integer
         format: uint64
-    PartialTransaction:
```
theodorebugnet, line 1003: These were unused anywhere in the spec and stale

crates/universal-wallet/schema/src/schema/mod.rs
```diff
@@ -32,6 +32,25 @@ use crate::visitors::eip712::{Context as Eip712Context, Eip712Error, Eip712Visit
 #[cfg(feature = "eip712")]
 use alloy_dyn_abi::{Eip712Types, Error as AlloyEip712Error, PropertyDef, TypedData};
 
+/// Returns the 64-bit fragment used to identify a full chain hash in transaction
+/// details.
+///
+/// The fragment is the first eight bytes of the chain hash. Interpreting it as
+/// little-endian means Borsh serializes the `u64` back to those same eight bytes.
+#[must_use]
+pub const fn chain_hash_fragment(chain_hash: &[u8; 32]) -> u64 {
```
theodorebugnet, line 41: The reason it's in the universal-wallet crate is that it's a dependency of the `web3` crate, and generally going to be needed by all clients (at least all rust clients), so it's added as a helper here to avoid requiring duplication of the logic.

bkolad (approved):

PR #3042, opened 2026-07-21
Fix performance in list_events, preventing a tokio hang and allocation explosion.

theodorebugnet:
# Description

* Fix bug that would attempt to query unbounded events if they were not in the cache - bound all reads by the actual range param provided
* Filter events from the transaction cache by the range before cloning, not after
* Drop the cache lock after it's been read, instead of continuing to hold the lock across the DB queries
* Handle missing ranges gracefully, including future/nonexistent ranges

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
+++ b/crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
@@ -166,25 +166,29 @@ impl<S: Spec, Rt: Runtime<S>> TransactionCache<S, Rt> {
         &self,
         event_numbers: std::ops::Range<u64>,
     ) -> anyhow::Result<Vec<RuntimeEventResponse<Rt::RuntimeEvent>>> {
-        let transaction_cache = self.inner.read().await;
-        let Some(last_needed_event) = event_numbers.end.checked_sub(1) else {
+        if event_numbers.start >= event_numbers.end {
             return Ok(vec![]);
+        }
+        let cached_events = {
+            let transaction_cache = self.inner.read().await;
+            let mut relevant_txs = transaction_cache
+                .event_numbers_index
+                .range(event_numbers.clone())
+                .map(|(_, tx_number)| *tx_number);
+            match relevant_txs.next() {
+                Some(first_needed_tx) => {
+                    let last_needed_tx = relevant_txs.next_back().unwrap_or(first_needed_tx);
+                    transaction_cache
+                        .cache
+                        .range(first_needed_tx..=last_needed_tx)
+                        .flat_map(|(_, tx)| tx.confirmation.events.iter())
+                        .filter(|event| event_numbers.contains(&event.number))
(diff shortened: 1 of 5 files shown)


crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
```diff
@@ -166,25 +166,29 @@ impl<S: Spec, Rt: Runtime<S>> TransactionCache<S, Rt> {
         &self,
         event_numbers: std::ops::Range<u64>,
     ) -> anyhow::Result<Vec<RuntimeEventResponse<Rt::RuntimeEvent>>> {
-        let transaction_cache = self.inner.read().await;
-        let Some(last_needed_event) = event_numbers.end.checked_sub(1) else {
+        if event_numbers.start >= event_numbers.end {
```
citizen-stig: non blocking: why not error though?

crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
```diff
@@ -166,25 +166,29 @@ impl<S: Spec, Rt: Runtime<S>> TransactionCache<S, Rt> {
...
+            match relevant_txs.next() {
+                Some(first_needed_tx) => {
+                    let last_needed_tx = relevant_txs.next_back().unwrap_or(first_needed_tx);
+                    transaction_cache
+                        .cache
+                        .range(first_needed_tx..=last_needed_tx)
+                        .flat_map(|(_, tx)| tx.confirmation.events.iter())
+                        .filter(|event| event_numbers.contains(&event.number))
+                        .cloned()
+                        .collect::<Vec<_>>()
```
citizen-stig, line 191: do we truly need a vector here or is it enough with iterator?

citizen-stig (approved):

crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
theodorebugnet: Good idea actually. I think it follows the old logic, but in this case it might be better to send an error to the clients since it might be a mistaken input

crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
theodorebugnet: This lets us clone the events and drop the `inner` lock before querying the ledger DB, otherwise the iterator would be lazy and would need to retain the references as far as I can tell.

crates/full-node/sov-sequencer/src/preferred/transaction_subscriptions.rs
citizen-stig: follow up issue is fine. Also we want to be consistent on this types of cases

PR #3050, opened 2026-08-13
fix(node-client): don't treat HTTP errors as nonce zero

questfever:
# Description

Fix `NodeClient::get_nonce_for_public_key` so it only returns nonce `0` when the nonce state item is actually missing.

Previously, the method called `error_for_status()` and converted any HTTP error into `0`. This meant responses such as 429, 500, or 503 were treated the same as a missing nonce. Callers such as the CLI `SubmitBatch` workflow could then sign a batch starting from nonce `0` instead of surfacing the node error.

This changes the status handling to:

- return `0` for `404 Not Found`
- return an error for other non-success HTTP statuses
- continue parsing the nonce response for successful statuses

This matches the existing pattern used by `sequencer_rollup_address`, where `404` is treated as absence and other non-success responses are errors.

## Linked Issues
- Related to # (issue) 

## Testing

- `SKIP_GUEST_BUILD=1 cargo test -p sov-node-client`
- `cargo fmt --check`
- `git diff --check`

## Docs

### DIFF
--- a/crates/utils/sov-node-client/src/lib.rs
+++ b/crates/utils/sov-node-client/src/lib.rs
@@ -117,9 +117,15 @@ impl NodeClient {
         );
         let response = self.http_client.get(&nonce_url).send().await?;
 
-        let nonce = match response.error_for_status() {
-            Ok(res) => res.json::<NonceResponse>().await?.value,
-            Err(_) => 0,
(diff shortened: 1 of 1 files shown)


questfever:

questfever:
@bkolad @citizen-stig Hi, Could you please review this PR at your convenience? Thank you very much.

